PR #11, opened 2017-06-08
fix deprecated warning

anhtuank7c:


anhtuank7c:
There are many files contain this warning :|

PR #15, opened 2017-06-29
add ability to style tabBar

mbardauskas:


ethanshar:
I will probably change containerStyle to plain "style"

PR #33, opened 2017-10-05
Tweak ActionSheet style to more closely match native component on Android

overengineered:
ActionSheet before:
current-uilib-sheet

ActionSheet after:
fixed-uilib-sheet

Native component on Android (Google Photos app):
native-sheet-2

Ideally, ActionSheet action items would have TouchableHighlight (as pictured in native component screenshot), not sure how to add it though.

### DIFF
--- a/src/components/actionSheet/index.js
+++ b/src/components/actionSheet/index.js
@@ -66,7 +66,7 @@ export default class ActionSheet extends BaseComponent {
         duration={600}
         easing="ease-out-quint"
       >
-        <View paddingB-8 bg-white>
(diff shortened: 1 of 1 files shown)


ethanshar:
Hey @overengineered, 
Thanks, I'll check it out. 
Regarding the highlight behavior. I added to our [TouchableOpacity](https://github.com/wix/react-native-ui-lib/blob/master/src/components/touchableOpacity/index.js) an activeBackgroundColor prop to allow set this behavior. I will add it also.

overengineered:
Already fixed

PR #36, opened 2017-10-19
Toast component: added blurring of underlying content

trofima:


src/components/toast/index.js
```diff
@@ -172,7 +173,7 @@ export default class Toast extends BaseComponent {
   }
 
   render() {
-    const {backgroundColor, actions, allowDismiss } = this.getThemeProps();
+    const {backgroundColor, actions, allowDismiss, blur} = this.getThemeProps();
```
ethanshar: blur is missing the component propTypes. 
also to prefer maybe calling it blurOptions to be more explicit

src/components/toast/index.js
```diff
@@ -252,5 +254,12 @@ function createStyles() {
       height: 12,
       tintColor: Colors.white,
     },
+    blur: {
```
ethanshar: quicker/cleaner to use `StyleSheet.absoluteFillObject`
https://facebook.github.io/react-native/docs/stylesheet.html#absolutefillobject

src/components/toast/index.js
```diff
@@ -195,6 +196,7 @@ export default class Toast extends BaseComponent {
         ]}
         {...animationConfig}
       >
+        {blur && <BlurView style={this.styles.blur} blurType={blur.type || 'light'} blurAmount={blur.amount || 3}/>}
```
ethanshar: maybe create helper function to get blurOptions (handling defaults) and spread it here

Issue #48, opened 2017-12-08
Ellipse Button


VanMen:
A sample code that I use in my code somehow ends up with complete ellipse for the button like below:

![image](https://user-images.githubusercontent.com/11056366/33789679-ec2442da-dcac-11e7-8777-d8e7a8d009ba.png)

Is there any parameter in button tag that can be use to make a better button shape ?

Windows 7
Android Emulator SDK 16

ethanshar:
Hey @VanMen 
Sorry for the late response, can you tell me which Android device are you running on here? (device model and everything..)
Thanks

VanMen:
Hey @ethanshar,

I'm using AVD (Android Virtual Device) from Android Studio 2.3.3
Screen size: 5.2 (1080 x 1920)
API version: 16
Target Android 4.1 (Google APIs)
CPU: x86
Emulated Performance, Graphics: Software - GLES 2.0


Thanks

inbalwix:
Seems this issue is related to API 4.1 (Google APIs).
You can set the Button's borderRadius prop to change the shape of the button (value of 26 will generate a shape similar to the default button shape in other Android versions).

ethanshar:
resolved

PR #56, opened 2018-01-07
Docs fix2

inbalwix:
@modifiers 'text' to 'Typography'; Issue #48 fix (button borderRadius in Android API 16)

### DIFF
--- a/src/components/button/index.js
+++ b/src/components/button/index.js
@@ -128,7 +128,7 @@ export default class Button extends BaseComponent {
 
   // This method will be called more than once in case of layout change!
   getComponentDimensions(event) {
-    if (Platform.OS === 'android' && Platform.Version === 16) {
+    if (Platform.OS === 'android' && Platform.Version <= 17) {
       const height = event.nativeEvent.layout.height;
       this.setState({borderRadius: height / 2});
     }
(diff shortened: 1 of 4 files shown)


src/components/button/index.js
```diff
@@ -127,6 +126,14 @@ export default class Button extends BaseComponent {
     }
   }
 
+  // This method will be called more than once in case of layout change!
+  getComponentDimensions(event) {
+    if (Platform.OS === 'android' && Platform.Version <= 17) {
```
ethanshar: Please use Constants.isAndroid for checking platform.

src/components/button/index.js
```diff
@@ -127,6 +126,14 @@ export default class Button extends BaseComponent {
...
+      const height = event.nativeEvent.layout.height;
```
ethanshar: Use destructor for extracting the height as follow
const {height} = event.nativeEvent.layout;

src/components/button/index.js
```diff
@@ -268,11 +275,14 @@ export default class Button extends BaseComponent {
   }
 
   getBorderRadiusStyle() {
-    const {link, borderRadius, fullWidth} = this.props;
-    if (link || fullWidth) {
+    const {link, fullWidth, borderRadius: borderRadiusFromProps} = this.props;
+    if (link || fullWidth || borderRadiusFromProps === 0) {
```
ethanshar: why are you checking here if borderRadiusFromProps === 0 ?

PR #66, opened 2018-01-17
more consistent z-order and android zIndex bug fix

trofima:


ethanshar:
@trofima can I close that, If I remember correctly it's irrelevant. right?

PR #72, opened 2018-01-31
Fix for Badge stretch when label value change number of chars

inbalwix:
fix badge

### DIFF
--- a/demo/src/screens/componentScreens/BadgesScreen.js
+++ b/demo/src/screens/componentScreens/BadgesScreen.js
@@ -10,7 +10,7 @@ export default class BadgesScreen extends Component {
     super(props);
 
     this.state = {
-      value: 1,
+      value: 10,
     };
   }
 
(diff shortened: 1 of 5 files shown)


src/components/badge/index.js
```diff
@@ -47,6 +47,7 @@ export default class Badge extends BaseComponent {
   }
 
   render() {
+    this.generateStyles();
```
ethanshar: I think we should avoid using generateStyles in render (and in general) since it creates the whole StyleSheet object and it might be heavy to do in each render.
 
It would be better to move the functionality based on the label prop to other functions that returns the badge style parts.  similar to what we do in the Button component (getContainerSizeStyle, getOutlineStyle, etc..)
This way we can also add unit tests for this functionalities. 

Generally I would like to slowly avoid using the generateStyle and deprecate it eventually.

demo/src/screens/componentScreens/BadgesScreen.js
```diff
@@ -1,25 +1,68 @@
 import React, {Component} from 'react';
 import {ScrollView, StyleSheet} from 'react-native';
-import {View, Badge, Colors} from 'react-native-ui-lib';//eslint-disable-line
+import {Colors, View, Badge, Button, Text} from 'react-native-ui-lib';//eslint-disable-line
 
 const BadgesSpace = 30;
 
 export default class BadgesScreen extends Component {
 
+  constructor(props) {
```
ethanshar: you can avoid adding a constructor if you just need to initialize the state. 
you can just add a state member in the Class, it's cleaner.. 

```
Class BadgeScreen extends Component {
  state = {
    value: 1
  }
  ....
}
```

PR #93, opened 2018-02-18
Text input fixed title

inbalwix:
WOA-9160 adding fixed title option (will be rendered only when floatingPlaceholder = false)

### DIFF
--- a/src/components/inputs/TextInput.js
+++ b/src/components/inputs/TextInput.js
@@ -77,10 +77,6 @@ export default class TextInput extends BaseInput {
      */
     titleColor: PropTypes.string,
     /**
-     * Limits the maximum number of characters that can be entered
-     */
-    maxLength: PropTypes.number,
-    /**
      * should the input display a character counter (only when passing 'maxLength')
      */
     showCharacterCounter: PropTypes.bool,
@@ -120,6 +116,8 @@ export default class TextInput extends BaseInput {
       ),
       showExpandableModal: false,
     };
+
+    this.generatePropsWarnings(props);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -131,8 +129,14 @@ export default class TextInput extends BaseInput {
         this.updateFloatingPlaceholderState,
       );
     }
-    if (nextProps.maxLength === 0) {
-      console.warn('setting maxLength to zero will block typing in this input');
+  }
+
+  generatePropsWarnings(props) {
+    if (props.maxLength === 0) {
+      console.warn('Setting maxLength to zero will block typing in this input');
+    }
+    if (props.showCharacterCounter && !props.maxLength) {
+      console.warn("In order to use showCharacterCount please pass 'maxLength' prop");
     }
   }
 
--- a/src/components/inputs/TextInput.js
+++ b/src/components/inputs/TextInput.js
@@ -18,7 +18,6 @@ const DEFAULT_UNDERLINE_COLOR_BY_STATE = {
 const charCountColorDefault = Colors.dark40;
 const charCountColorLimit = Colors.red30;
 
-
 /**
  * @description: A wrapper for Text Input component with extra functionality like floating placeholder
  * @extends: TextInput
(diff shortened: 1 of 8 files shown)


src/components/inputs/TextInput.js
```diff
@@ -163,6 +175,10 @@ export default class TextInput extends BaseInput {
     return typography.lineHeight;
   }
 
+  capitalize(string) {
+    return string.charAt(0).toUpperCase() + string.slice(1);
+  }
+
```
ethanshar: I think it's better to use
https://lodash.com/docs#capitalize

src/components/inputs/TextInput.js
```diff
@@ -209,6 +225,25 @@ export default class TextInput extends BaseInput {
...
+    const {floatingPlaceholder, title, titleColor} = this.props;
+
+    if (!floatingPlaceholder && title) {
+      const marginBottom = Constants.isIOS ? 5 : 4;
+      const titleFontSize = 13;
+      const capitalizedTitle = this.capitalize(title);
+
+      return (
+        <Text
+          style={{top: 0, color: titleColor, marginBottom}}
```
ethanshar: inline style is less explanatory, we should use it just for changing styling (titleColor), the rest should 

instead of defining style variables, create a style object using StyleSheet.
it's cleaner and better for performance.

inbalwix: already changed

src/components/inputs/TextInput.js
inbalwix: Cool. Thanks!

src/components/inputs/TextInput.js
```diff
@@ -65,13 +65,33 @@ export default class TextInput extends BaseInput {
...
+     */
+    titleColor: PropTypes.string,
+    /**
+     * Limits the maximum number of characters that can be entered
+     */
+    maxLength: PropTypes.number,
+    /**
+     * should the input display a character counter (only when passing 'maxLength')
+     */
+    characterCounter: PropTypes.bool,
```
ethanshar: let's give it a name that identify it as boolean prop like showCharacterCounter

src/components/inputs/TextInput.js
```diff
@@ -209,6 +229,38 @@ export default class TextInput extends BaseInput {
...
+          {capitalizedTitle}
+        </Text>
+      );
+    }
+  }
+
+  renderCharCounter() {
+    const {maxLength, characterCounter} = this.props;
+    if (maxLength && characterCounter) {
+      const {value} = this.state || '';
```
ethanshar: This line is wrong, basically if the state is undefined you fallback to an empty string and then try to destruct "value" from it, theoretically it would throw and error if state was undefined.

src/components/inputs/TextInput.js
```diff
@@ -209,6 +229,38 @@ export default class TextInput extends BaseInput {
...
+      const counter = value !== undefined ? value.length : 0;
```
ethanshar: Use (lodash) _.size here for cleaner code, it will return 0 for undefined value otherwise will return the string length

src/components/inputs/TextInput.js
```diff
@@ -65,13 +65,33 @@ export default class TextInput extends BaseInput {
...
+     */
+    title: PropTypes.string,
+    /**
+     * The title's color
+     */
+    titleColor: PropTypes.string,
+    /**
+     * Limits the maximum number of characters that can be entered
+     */
+    maxLength: PropTypes.number,
```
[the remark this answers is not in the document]
inbalwix: I know, I thought of making it more visible since it is required if wanting to use the characterCounter

src/components/inputs/TextInput.js
inbalwix: you mean this line: const {value} = this.state || ''; ? 
I already changed it to: const {value} = this.state;

src/components/inputs/TextInput.js
inbalwix: ok

src/components/inputs/TextInput.js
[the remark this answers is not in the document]
ethanshar: so maybe add console warning if the user pass "showCharacterCount" and maxLength is undefined?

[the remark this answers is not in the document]
inbalwix: Done.

PR #95, opened 2018-02-19
Feature highlighter

inbalwix:
WOA-9193 bug fix


src/components/featureHighlight/index.js
```diff
@@ -253,7 +256,8 @@ const styles = StyleSheet.create({
     lineHeight: messageLineHeight,
   },
   touchableOverlay: {
-    flex: 1,
+    ...StyleSheet.absoluteFillObject,
+    backgroundColor: 'red',
```
ethanshar: let's remove the red background here

Issue #120, opened 2018-03-21
Modal enableModalBlur broken on Android


isaachinman:
The `Modal` component's `enableModalBlur` works great on iOS.

However, I get this critical breaking error on Android:

    Error: [ReactNativeBlur]: BlurView cannot contain any child views on Android. You should use "position: absolute" on the BlurView, and place other views in front of it.

This error occurs when I try to launch the modal in question. Previous to launching the modal, I see this yellow box warning:

    Warning: Failed prop type: The prop `viewRef` is marked as required in `BlurView` but its value is `undefined`.

Not entirely sure what's happening here, or if it's my fault, but it seems the `BlurView` setup with this `Modal` component is broken on Android.

Here's my code:

    <Modal
      visible={ui.modals.myModal.open}
      onRequestClose={this.handleClose}
      animationType='slide'
      transparent
      hardwareAccelerated
      enableModalBlur
    >
      <Container>
        <Header>
          <Left>
            <Button
              onPress={this.handleClose}
              transparent
            >
              <Text>Close icon</Text>
            </Button>
          </Left>
          <Body>
            <Text style={styles.headerTitle}>Choose something</Text>
          </Body>
          <Right />
        </Header>
        <Content contentContainerStyle={styles.content}>
          <Grid>
            <Row>
              <Col style={styles.imageContainer}>
                <Image
                  style={styles.image}
                  resizeMode='contain'
                  source={item.image}
                />
              </Col>
            </Row>
          </Grid>
        </Content>
      </Container>
    </Modal>

Inbal-Tish:
I added a fix for the crash in Android. Please note that we are supporting modal blueView only in iOS at the moment

isaachinman:
Okay, I wasn't aware the blurView is only for iOS. That's a shame!

PR #128, opened 2018-04-01
fix for blurView crash in Android

Inbal-Tish:
issue #120


demo/src/screens/PlaygroundScreen.js
```diff
@@ -1,38 +1,57 @@
 import React, {Component} from 'react';
```
ethanshar: can you remove the PlaygroundScreen file from this PR, the PlaygroundScreen should be private and should not be committed.

PR #135, opened 2018-04-05
Expo.io support for BlurView

LaszloDev:
Expo uses its own bundled version of BlurView. So we need to have a check in the Modal component wether its Expo or a regular React-Native app and change the import.
Would be nice to see more expo support.


ethanshar:
We rather not depend on expo, but we can provide a way to pass the BlurView element you wish.

Inbal-Tish:
@LaszloDev Hi. Please see our last update with the addition blurView prop to the Modal component.

PR #141, opened 2018-04-15
Android wheel picker and android wheel picker dialog

bogobogo:


lib/android/src/main/java/com/wix/reactnativeuilib/wheelpicker/ThemedNumberPicker.java
```diff
@@ -0,0 +1,25 @@
+package com.wix.reactnativeuilib.wheelpicker;
```
ethanshar: @bogobogo 
What is this file? What's ThemedNumberPicker for ?

src/nativeComponents/WheelPicker.js
```diff
@@ -0,0 +1,101 @@
...
+  constructor(props) {
+    super(props);
+    this.onValueChange = this.onValueChange.bind(this);  //eslint-disable-line
+  }
+
+  state = {
+    ...this.stateFromProps(this.props),
+  }
+
+  UNSAFE_componentWillReceiveProps(nextProps) {  //eslint-disable-line
```
ethanshar: What's that?

src/nativeComponents/WheelPicker.js
```diff
@@ -0,0 +1,101 @@
...
+  itemStyle: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.number,
+  ]),
+  children: PropTypes.any,
+};
+
+WheelPickerView.Item = class extends React.Component { // eslint-disable-line react/no-multi-comp
+  static propTypes = {
+    value: PropTypes.any, // string or integer basically
```
ethanshar: you can use here `PropTypes.oneOfType([PropTypes.string, PropTypes.number])`

src/nativeComponents/WheelPicker.js
bogobogo: Since wheel picker is basically PickerIOS but for android, I thought it would make sense to add this addition as a pull request to react-native. To make this easier I am using almost exactly the same logic that react-native uses in PickerIOS. Specifically, I think that this is named "UNSAFE_componentWillReceiveProp" because this lifecycle will probably be deprecated in the future, but I am not sure.

lib/android/src/main/java/com/wix/reactnativeuilib/wheelpicker/ThemedNumberPicker.java
bogobogo: NumberPicker, the original native android component, does not expose an api to change the style of the divider, and its text inputs. One way to solve this is to use reflection. The other is to create a "theme" of styles for the app, that "trickles down" (i don't really know how it works under the hood) to the the component it self. Since I don't want to create a theme for the app, i wrapped the component with a theme where I define the styles. These themes are than implemented only on this component

src/nativeComponents/WheelPicker.js
bogobogo: oops.. my bad. Pushing a fix

PR #142, opened 2018-04-23
WOA-10013 - FeatureHighlight is "jump" on first time it visible

Inbal-Tish:


src/components/featureHighlight/index.js
```diff
@@ -136,56 +145,82 @@ class FeatureHighlight extends BaseComponent {
     return findNodeHandle(target);
   }
 
+  animate(toValue) {
+    Animated.timing(                  // Animate over time
+      this.state.fadeAnim,            // The animated value to drive
+      {
+        toValue,                      // Animate to value
+        duration: toValue === 0 ? 0 : 100,   // Make it take a while
```
ethanshar: toValue is either 1 or 0, it already acts as a boolean. We can omit the `=== 0` (and change the order of condition)

src/components/featureHighlight/index.js
```diff
@@ -136,56 +145,82 @@ class FeatureHighlight extends BaseComponent {
...
           });
         }, 0);
       }
     } else {
       const frame = props.highlightFrame;
       if (frame) {
-        this.setState({
-          targetPosition: {left: frame.x, top: frame.y, width: frame.width, height: frame.height},
-        });
+        this.targetPosition = {left: frame.x, top: frame.y, width: frame.width, height: frame.height};
```
ethanshar: if we're setting `this.targetPosition` every time we're calling `this.setContentPosition()`, let's just move the assignment inside the function, it has the right name anyway.

src/components/featureHighlight/index.js
```diff
@@ -136,56 +145,82 @@ class FeatureHighlight extends BaseComponent {
...
-    const height = event.nativeEvent.layout.height;
-    this.setState({contentViewHeight: height});
+    this.didLayout = true;
+    this.contentHeight = event.nativeEvent.layout.height;
+    if (this.targetPosition !== undefined) {
+      this.setContentPosition();
+    }
+  }
+
+  onCustomPress = () => {
```
ethanshar: I would change this name also, "custom" is too general and we might have all kind of clicks in that screen.

PR #146, opened 2018-04-29
Eslint colors

Inbal-Tish:
Adding extra color to map; updating version


eslint-rules/lib/utils.js
```diff
@@ -9,6 +9,17 @@ const colorExceptions = [
   'transparent', 'rgba(0,0,0,0)',
 ];
 
+const extraColorsDict = {
```
ethanshar: I don't think the extra Colors should be located inside the linter utils. since it's an open source, anyone would want to use it differently. (not all people are going to use red30, dark70, etc..)

I think a better way to load extra colors should be similar to what we do here
https://github.com/wix/react-native-ui-lib/blob/master/.eslintrc.js#L1
so the colors are being imported privately

PR #169, opened 2018-06-07
this.seState undefined

chenzhenjia:


chenzhenjia:
@ethanshar When the Android version is 17, report this.seState undefined

PR #175, opened 2018-06-11
add right to left (rtl) to textinput (V2)

rghorbani:
Hi,
I can make more of your components rtl if it's ok with you.
As requested by @Inbal-Tish.


Inbal-Tish:
@rghorbani I have a question: why passing 'rtl' prop and not using RN's  I18nManager? RN is making some layout changes when isRTL. Why not only make adjustments to cover what RN isn't?

rghorbani:
@Inbal-Tish In some case like in rtl languages we need to input to take a name and we need the input be rtl in that case but not when taking email.

PR #177, opened 2018-06-14
Tab bar style

Inbal-Tish:
Adding scroll mode to TabBar; 
Adding support for gradient effect on scroll bar; 
Adding support for maxLines, width, divider to TabBarItem.

### DIFF
--- a/src/components/tabBar/TabBarItem.js
+++ b/src/components/tabBar/TabBarItem.js
@@ -26,7 +26,7 @@ export default class TabBarItem extends BaseComponent {
      */
     labelStyle: Text.propTypes.style,
     /**
-     * maximun number of lines the label can break 
+     * maximun number of lines the label can break
      */
     maxLines: PropTypes.number,
     /**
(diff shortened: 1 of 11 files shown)


src/components/tabBar/TabBarItem.js
```diff
@@ -37,36 +41,79 @@ export default class TabBarItem extends BaseComponent {
      * callback for when pressing a tab
      */
     onPress: PropTypes.func,
+    /**
+     * whether the tab will have a divider on its right
+     */
+    divider: PropTypes.bool,
```
ethanshar: let's use boolean names for boolean props like showDivider

src/components/tabBar/TabBarItem.js
```diff
@@ -37,36 +41,79 @@ export default class TabBarItem extends BaseComponent {
...
+    super(props);
+
+    this.state = {
+      fontStyle: this.getFontStyle(props),
+    };
+  }
+
+  // HACK: for indicator width in TabBar
+  getFontStyle(props) {
+    const userProp = props.selectedLabelStyle ? props.selectedLabelStyle.fontWeight : undefined;
```
ethanshar: this could break if the style has passed as number (when passing a StyleSheet object the style translate to a number). 
you can use flatten for this https://facebook.github.io/react-native/docs/stylesheet.html#flatten

PR #178, opened 2018-06-14
Create RadioButton and RadioGroup components

ethanshar:


src/components/radioButton/RadioButton.js
```diff
@@ -0,0 +1,94 @@
...
+  return StyleSheet.create({
+    container: {
+      borderWidth: 1,
+      borderColor: color,
+      width: size,
+      height: size,
+      borderRadius,
+      padding: 3,
+    },
+    checkmark: {
```
Inbal-Tish: I would change the name 'checkmark' to 'filler' or something similar

src/components/radioButton/RadioGroup.js
```diff
@@ -0,0 +1,56 @@
...
+
+  static displayName = 'RadioGroup';
+
+  static propTypes = {
+    /**
+     * The value of the selected radio button
+     */
+    value: PropTypes.string,
+    /**
+     * Invoked once the value has changed, by selecting once of the radio buttons in the group
```
Inbal-Tish: type: "... by selecting **one**  of the..." (instead of 'once')

PR #179, opened 2018-06-14
create Checkbox component

ethanshar:


src/components/checkbox/index.js
```diff
@@ -0,0 +1,90 @@
...
+      borderWidth: 1,
+      borderColor: color,
+      alignItems: 'center',
+      justifyContent: 'center',
+    },
+    containerSelected: {
+      backgroundColor: color,
+    },
+    selectedIcon: {
+      tintColor: Colors.white,
```
Inbal-Tish: Why not let the user pass the tint color as well if he already can pass backgroundColor?

PR #183, opened 2018-07-08
Adding no-hard-coded-font lint rule

Inbal-Tish:


src/style/typography.js
```diff
@@ -1,64 +1,75 @@
 import _ from 'lodash';
 import {Constants} from '../helpers';
 
+const WEIGHT_TYPES = {
+  SLIM: '100',
```
ethanshar: I don't it's called SLIM, I think it's ultra light, in our internal lib I called it ULIGHT

eslint-rules/tests/lib/rules/no-hard-coded-font.js
```diff
@@ -13,35 +13,41 @@ RuleTester.setDefaultConfig({
...
-//       code: 'const badUsage = <Text style={{fontSize: size}}/>;',
-//       errors: [
-//         {
-//           messageId: 'foo',
-//         },
-//       ],
-//     },
-//   ],
-// });
+ruleTester.run('no-hard-coded-font', rule, {
```
ethanshar: can you add a good scenario for when some spread the typography preset

Inbal-Tish: I added another valid case

PR #193, opened 2018-07-16
TextInput - char count fix

Inbal-Tish:


src/components/inputs/TextInput.js
```diff
@@ -460,7 +463,16 @@ export default class TextInput extends BaseInput {
     this.toggleExpandableModal(false);
   }
 
-  onChangeText(text) {
+  onKeyPress = (event) => {
+    this.lastKey = event.nativeEvent.key;
+  }
+
+  onChangeText = (text) => {
```
ethanshar: did u check the keypress thing on Android ?
I think it requires a special handling, see TagsInput component. 
It might work because we already added support to it.

PR #202, opened 2018-07-24
Text input - top errors option

Inbal-Tish:


src/components/inputs/TextInput.js
```diff
@@ -420,48 +464,39 @@ export default class TextInput extends BaseInput {
     );
   }
 
+  getTopPaddings() {
+    const {floatingPlaceholder} = this.props;
+    return floatingPlaceholder ? (this.shouldShowTopError() ? undefined : 25) : undefined;
```
ethanshar: let's avoid nested ternary condition. 
it's not readable 
also since we return undefined twice there's no need for it, the function by default will return undefined if there's no return value

src/components/inputs/TextInput.js
```diff
@@ -289,24 +331,27 @@ export default class TextInput extends BaseInput {
...
           {counter} / {maxLength}
         </Text>
       );
     }
   }
 
-  renderError() {
-    const {enableErrors, error} = this.props;
-    if (enableErrors) {
+  renderError(visible) {
```
ethanshar: why do we need to pass visible to `renderError` if the value comes from the props. 
shouldn't  we extract the value directly from the props inside the method to avoid unnecessary params ?

src/components/inputs/TextInput.js
```diff
@@ -98,6 +99,14 @@ export default class TextInput extends BaseInput {
      * should the input display a character counter (only when passing 'maxLength')
      */
     showCharacterCounter: PropTypes.bool,
+    /**
+     * should clear the input's placeholer when focused (instead of when typing)
+     */
+    clearOnFocus: PropTypes.bool,
```
ethanshar: confusing name

PR #209, opened 2018-08-09
containerStyle for LoaderScreen

NikoMontana:
Set the backgroundColor for this screen...

containerStyle prop has to have flex: 1 in it

### DIFF
--- a/src/screensComponents/loaderScreen/index.js
+++ b/src/screensComponents/loaderScreen/index.js
@@ -37,7 +37,7 @@ export default class LoaderScreen extends BaseComponent {
     /**
     * Styles for the container of this component
     */
-    containerStyle: View.propTypes.style
+    containerStyle: View.propTypes.style,
   };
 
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@NikoMontana Hi. I was wondering if you need to style the LoaderScreen container when it's not in 'layout' mode, why not just use activityIndicator strait from RN? what is the use for LoaderScreen if not flexed?

NikoMontana:
@Inbal-Tish I have set an absolute view inside a screen and by opening this screen with the loaderscreen overlay it is initially not covering the view. After a half of a second it applies the overlay to the absolute view (@thebottom)

Inbal-Tish:
@NikoMontana Sorry, I'm not sure I understand your use. Can you explain again or send some video/screenshot?

ethanshar:
I added support for `containerStyle` to LoaderScreen

PR #215, opened 2018-08-19
Card blur

Inbal-Tish:
cleaning component, resetting shadows


src/components/card/index.js
```diff
@@ -76,10 +75,17 @@ class Card extends BaseComponent {
...
 
   static defaultProps = {
-    // borderRadius: BorderRadiuses.br40,
     enableShadow: true,
   };
 
+  constructor(props) {
+    super(props);
+
+    if (props.containerStyle !== undefined) {
```
ethanshar: u wrote `ward` instead of `warn`

Issue #138, opened 2018-04-11
MaskedInput - renderMaskedText() cannot set state


NikoMontana:
First I bind the function: `renderPrice` in the constructor:

`this.renderPrice = this.renderPrice.bind(this);`

In the function I am trying to set the input to signed numbers only by simple iteration. After the iteration I am setting the new signed number with `setState()`. It looks like this:

 ```
renderPrice(value) {
    const hasValue = Boolean(value && value.length > 0);

    let newText = "";
    if (hasValue) {
      let numbers = "0123456789";

      for (var i = 0; i < value.length; i++) {
        if (numbers.indexOf(value[i]) > -1) {
          newText = newText + value[i];
        }
      }

      if (newText === "") newText = "15";

      if (newText.length > 5) newText = newText.substring(0, 5);
    }

    this.setState({ task_posting_budget: newText });

    return (
      <View row center>
        <Text text30 dark50>
          -
        </Text>
        <Text text30 dark10={hasValue} dark60={!hasValue}>
          {hasValue ? newText : "15"}
        </Text>
        <Text text80 dark60>
          $
        </Text>
      </View>
    );
  }
```

`                    <MaskedInput
                      renderMaskedText={this.renderPrice}
                      caretHidden
                      keyboardType={"numeric"}
                    />`

The problem is that by opening this component/screen the app doesnt react anymore, because setState is bein called to often. 

I will try to find a workaround for this and post a solution if possible.

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
> scheduleUpdateImpl
>     index.bundle?platform=android&dev=true&minify=false:5483:109
> scheduleUpdate
>     index.bundle?platform=android&dev=true&minify=false:5480:38
> enqueueSetState
>     index.bundle?platform=android&dev=true&minify=false:4100:185
> setState
>     index.bundle?platform=android&dev=true&minify=false:8063:35
> componentWillReceiveProps
>     index.bundle?platform=android&dev=true&minify=false:78597:20
> callComponentWillReceiveProps
>     index.bundle?platform=android&dev=true&minify=false:4173:47
> updateClassInstance
>     index.bundle?platform=android&dev=true&minify=false:4199:155
> updateClassComponent
>     index.bundle?platform=android&dev=true&minify=false:4290:296
> beginWork
>     index.bundle?platform=android&dev=true&minify=false:4403:48
[...]

PR #222, opened 2018-08-27
Fix Toast animated prop

CodyBrouwers:
Hi there, this is a fantastic UI library that we just started using for our app and we've all been extremely happy with so thank you very much for putting so much care and effort into it!

I just started using the Toast component and came across a situation where I don't want it to be animated at all, so I set `animated={false}` but this results in the following error:

Took a look into the source code and came up with this small solution. The issue it seems is that when the `animated` prop is false, registering the animations is skipped, but then the unregistered animations are still attempted to be used in the `Animated.View` props.

According to [this issue comment](https://github.com/oblador/react-native-animatable/issues/129#issuecomment-352246126) over at react-native-animatible, you can just pass `null` or `undefined` as the animation prop value to disable it, so I'm checking the Toast `animated` prop and doing that based on its value. This seems to work pretty well, but I don't have a lot of knowledge about animations so not sure if this is the optimal solution. Also I didn't see any tests for this component to confirm my fix didn't break anything else.

Please take a look and if it works for you great, if not then I can create an issue and let you tackle in this in another way!

Thanks again!

### DIFF
--- a/src/components/toast/index.js
+++ b/src/components/toast/index.js
@@ -170,7 +170,7 @@ export default class Toast extends BaseComponent {
       delay, 
       useNativeDriver, 
       onAnimationEnd: () => this.onAnimationEnd()
-    };    
+    };
   }
 
   getContentAnimation(shouldShow) {
--- a/src/components/toast/index.js
+++ b/src/components/toast/index.js
@@ -169,7 +169,7 @@ export default class Toast extends BaseComponent {
       duration, 
       delay, 
       useNativeDriver, 
-      onAnimationEnd: () => this.onAnimationEnd()
+      onAnimationEnd: () => this.onAnimationEnd(),
     };
   }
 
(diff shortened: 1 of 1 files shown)


src/components/toast/index.js
```diff
@@ -160,11 +160,17 @@ export default class Toast extends BaseComponent {
...
 
   getAnimation(shouldShow) {
-    const {position, useNativeDriver} = this.props;
+    const {position, useNativeDriver, animated} = this.props;
     const animationDescriptor = getAnimationDescriptor(position, this.state);
     const {animation, duration, delay} = shouldShow ? animationDescriptor.enter : animationDescriptor.exit;
 
-    return {animation, duration, delay, useNativeDriver, onAnimationEnd: () => this.onAnimationEnd()};
+    return {
+      animation: animated ? animation : null,
```
Inbal-Tish: I would move this condition to the 'animationConfig' state set in the constructor to avoid the call to getAnimation() altogether, as done on line #138.

PR #257, opened 2018-10-14
improve getColorTint util to support custom colors tints

ethanshar:


src/style/colors.js
```diff
@@ -60,10 +62,14 @@ class Colors {
...
   getColorTint(color, tintKey) {
+    const BASE_COLOR_LEVEL = 3;
+    const darkRatios = [0.13, 0.08];
+    const lightRatios = [0.27, 0.55, 0.72, 0.83, 0.9];
+
     const colorKey = _.findKey(this, (value, key) => this[key] === color);
 
     if (_.isUndefined(tintKey)) {
-      throw new Error('"Colors.getColorTint" must accept a color and tintKey params');
+      console.error('"Colors.getColorTint" must accept a color and tintKey params');
```
Inbal-Tish: If you don't return here you might use the undefined 'tintKey' later

PR #258, opened 2018-10-17
adding assets-deprecation rule

Inbal-Tish:


eslint-rules/lib/rules/assets-deprecation.js
```diff
@@ -0,0 +1,127 @@
...
+    const {source} = context.options[0];
+    const assetsName = 'Assets';
+    let shouldCheckDeprecation = false;
+
+    function checkAssetsImport(node) {
+      const importSource = node.source.value;
+      
+      if (source === importSource) {
+        const specifiers = node.specifiers;
+        if (specifiers) {
```
ethanshar: why not using lodash find `util`, it's cleaner

eslint-rules/lib/rules/assets-deprecation.js
```diff
@@ -0,0 +1,127 @@
...
+        }
+      }
+      pathString = '';
+      return false;
+    }
+
+    const {deprecations} = context.options[0];
+
+    function getDeprecatedObject(path) {
+      let jsonElement;
```
ethanshar: you can use `_.find` here too

eslint-rules/lib/rules/assets-deprecation.js
```diff
@@ -0,0 +1,127 @@
...
+    // Path
+    let pathString = '';
+
+    function deprecationCheck(node) {
+      if (node) {
+        const check = isAssetsObject(node);
+        if (check && pathString !== '') {
+          const deprecatedObject = getDeprecatedObject(pathString);
+          if (deprecatedObject) {
+            const path = deprecatedObject.path;
```
ethanshar: you can destruct the values here
`const {path, message, fix}  = deprecatedObject`

eslint-rules/lib/rules/assets-deprecation.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    function isAssetsObject(node) {
+      if (node) {
+        if (node.object) {
+          if (node.property && node.property.name) {
+            pathString = (pathString === '') ? `${node.property.name}` : `${node.property.name}.${pathString}`;
+            return isAssetsObject(node.object);
+          }
+        } else if (node.name && node.name === assetsName) {
```
ethanshar: isn't just enough to check if `node.name === assetsName` (if name is undefined it still needs to be equal to assetName)

PR #260, opened 2018-10-19
Picker useNativePicker

hand-dot:
When i change the property useNativePicker of Picker component I need
to change also the following properties: value, onChange, children.

And since useNativePicker is used only to change display mode,
i made changes that haven't any effect in other properties.

Do you think it is ok?

### DIFF
--- a/demo/src/screens/componentScreens/FormScreen.js
+++ b/demo/src/screens/componentScreens/FormScreen.js
@@ -32,6 +32,7 @@ export default class FormScreen extends Component {
       // language: {value: 'java', label: 'Java'},
       language: undefined,
       languages: [],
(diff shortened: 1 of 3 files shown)


hand-dot:
@ethanshar  Should I close this PR?

ethanshar:
hey @hand-dot, 
sorry for late response. 
The Picker component has become a bit big and complex, and currently, small changes in it affect us internally. 
My plan is to do a big refactoring, clean this component and provide better API. 
Till then I rather not change anything in it. 
I do hope to get to refactor it soon since it has other issues.

ethanshar:
I believe this issue has been resolved

PR #265, opened 2018-10-24
Adding swipe gesture (vertical) to Dialog component

Inbal-Tish:


src/components/dialog/index.js
```diff
@@ -71,10 +72,32 @@ class Dialog extends BaseComponent {
     };
   }
 
+  onSwipe(gestureName) {
+    const {SWIPE_UP, SWIPE_DOWN} = swipeDirections;
+    switch (gestureName) {
+      case SWIPE_UP:
+        if (this.props.top) {
```
ethanshar: I wouldn't use the top modifier to decide how and if to swipe, the modifiers are "nice to have" props and not mandatory. 
User can align his dialog using simple style. 
I would expose a dedicated prop for that behaviour (swipeToDismissDirection or something like that)

src/components/dialog/index.js
```diff
@@ -28,6 +35,10 @@ class Dialog extends BaseComponent {
      * dismiss callback for when clicking on the background
      */
     onDismiss: PropTypes.func,
+    /**
+     * the direction of the swipe to dismiss the dialog (default is 'down')
+     */
+    dismissSwipeDirection: PropTypes.oneOfType(SWIPE_DIRECTIONS),
```
ethanshar: I think `oneOfType` is not correct here, you need oneOf, something like that: 
`PropTypes.oneOf(_.values(SWIPE_DIRECTIONS))`

Inbal-Tish: What do you mean by not correct? it does work...

Inbal-Tish: It works even when sending the string value (i.e. 'up' instead of Dialog.swipeDirections.UP)

PR #266, opened 2018-10-24
Main screen update - injected navigationData option

Inbal-Tish:


demo/src/screens/MainScreen.js
```diff
@@ -233,7 +240,9 @@ export default class UiLibExplorerMenu extends Component {
 
   renderBreadcrumbs() {
     const {currentPage} = this.state;
-    const pages = Object.keys(navigationData);
+    const data = this.props.navigationData || navigationData;
```
ethanshar: we have this line more than once, 
`const data = this.props.navigationData || navigationData;`

let's unify it under a get function (getNavigationData)

PR #302, opened 2018-11-19
feat/Pimple badge

ArnonZ:
Added pimple badge and updated Badge demo screen

### DIFF
--- a/src/components/badge/index.js
+++ b/src/components/badge/index.js
@@ -63,7 +63,6 @@ export default class Badge extends BaseComponent {
 
   static defaultProps = {
     size: 'default',
-    pimple: false,
     label: '',
   };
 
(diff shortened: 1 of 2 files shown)


src/components/badge/index.js
```diff
@@ -58,6 +63,8 @@ export default class Badge extends BaseComponent {
 
   static defaultProps = {
     size: 'default',
+    pimple: false,
```
ethanshar: I would remove default values that are redundant
label by default is empty
and pimple by default is undefined (which is falsy)

ArnonZ: A badge without a label results in an undefined call to label.length which calculates how many chars you have inside the badge. this is the only reason I have to give the label an empty string as a default prop.
I imagined that this logic is intentional and didn't want to change it

ArnonZ: regarding the pimple obviously you are correct :P

PR #303, opened 2018-11-19
Feat/interactable drawer

Inbal-Tish:


src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,469 @@
...
+          {item.text}
+        </Animated.Text>}
+      </TouchableOpacity>
+    );
+  }
+  renderRightItems() {
+    const {height, rightItems} = this.props;
+
+    return (
+      <View style={{position: 'absolute', right: 0, height, flexDirection: 'row'}}>
```
ethanshar: Let's avoid inline style wherever we can

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,469 @@
...
+      const end = itemWidth - (size * BLEED);
+      const min = -(itemWidth * (i + 1));
+      const max = -(end + (interval * i));
+      // const range = [-(this.minItemWidth * (i + 1)), -(end + (interval * i))];
+      inputRanges.push([min, max]);
+    }
+    return inputRanges.reverse();
+  }
+
+  renderLeftItem() {
```
ethanshar: I think it will be cleaner to separate both `LeftItem` and `RightItem` into separate subcomponents.
It will make their code readable instead of having one big render method for each.

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,469 @@
...
+ * @extendslink: 
+ */
+export default class Drawer extends BaseComponent {
+  static displayName = 'Drawer';
+
+  static propTypes = {
+    /**
+     * The drawer's height
+     */
+    height: PropTypes.number.isRequired,
```
ethanshar: The `height` & `width` props, in my opinion, should be derived from the wrapped component. 
I know we talked about it already. But, I saw other libraries and usually, you don't send these props. 

I know we might need them for all kind of calculations, but, I think the internal behavior define in our guidelines  should be part of the internal (private) uilib code. 

This will make this component way more generic and simple.

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,469 @@
...
+    damping: 1 - 0.6,
+    tension: 300,
+    itemsTintColor: Colors.white,
+    itemsIconSize: DEFAULT_ICON_SIZE,
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.deltaX = new Animated.Value(0);
```
ethanshar: if we repeat this code in `componentWillReceiveProps` I would have turned it into a separate method.

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,469 @@
...
+  }
+  getMaxItemWidth() {
+    const {rightItems, width} = this.props;
+    return rightItems ? (width - MIN_LEFT_MARGIN) / rightItems.length : (width - MIN_LEFT_MARGIN);
+  }
+  getRightItemsTotalWidth(numberOfItems) {
+    const {rightItems} = this.props;
+    let total = 0;
+
+    if (rightItems && rightItems.length > 0) {
```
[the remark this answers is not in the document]
Inbal-Tish: Done. Thanks!

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,469 @@
...
+      const items = rightItems.reverse();
+      const size = numberOfItems && numberOfItems >= 0 ? numberOfItems : items.length;
+      
+      for (let i = 0; i < size; i++) {
+        total += this.getItemWidth(items[i]);
+      }
+    }
+    return total;
+  }
+  getItemWidth(item) {
```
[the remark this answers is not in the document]
Inbal-Tish: Not sure I understand...

src/interactableComponents/drawer/index.js
Inbal-Tish: Done.

src/interactableComponents/drawer/index.js
Inbal-Tish: Done.

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,469 @@
...
+    this.inputRanges = this.getInputRanges();
+    
+    this.state = {
+      inMotion: false,
+      position: 1,
+    };
+  }
+
+  componentWillReceiveProps(nextProps) { //eslint-disable-line
+    if (nextProps !== this.props) {
```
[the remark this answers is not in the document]
Inbal-Tish: Not sure I understand 'when relevant'...

src/interactableComponents/drawer/index.js
Inbal-Tish: Fixed!

src/interactableComponents/drawer/index.js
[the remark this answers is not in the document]
Inbal-Tish: Done

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,475 @@
...
+
+    this.state = {
+      inMotion: false,
+      position: 1,
+      width: Constants.screenWidth,
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.leftItemJSON !== JSON.stringify(nextProps.leftItem) ||
```
ethanshar: this comparison will always be against the initial value of the buttons. 

What if the value of buttons was initially X then changed to Y and then again to X, the second time it will compare it, it will compare X to X instead of Y to X.

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,475 @@
...
+    const inputRanges = this.getInputRanges();
+
+    return (
+      <TouchableOpacity
+        key={index}
+        style={[
+          this.styles.item, {
+            width: this.getItemWidth(item),
+            backgroundColor: item.background || ITEM_BG,
+            padding: this.state.itemPadding,
```
ethanshar: Destruct `state` & `props` at the beginning of the method like we usually do, it helps with readability.

src/interactableComponents/drawer/index.js
```diff
@@ -0,0 +1,475 @@
...
+     */
+    damping: PropTypes.number,
+    /**
+     * The drawer top layer's tention
+     */
+    tension: PropTypes.number,
+    /**
+     * The bottom layer's items to appear when opened from the right
+     */
+    rightItems: PropTypes.arrayOf(PropTypes.shape(ITEM_PROP_TYPES)),
```
ethanshar: I would put `rightItems` & `leftItem` props, first, since they're most important.

PR #313, opened 2018-12-10
Adding safeAreas to Dialog

Inbal-Tish:


src/components/dialog/index.js
```diff
@@ -131,7 +133,10 @@ class Dialog extends BaseComponent {
               style={this.styles.gestureContainer}
             >
               <TouchableWithoutFeedback>
-                {this.props.children}
+                <SafeAreaView style={{flex: 1}}>
```
ethanshar: why are u using here `SafeAreaView` and not our `View` with the `useSafeArea` prop?

src/helpers/Constants.js
```diff
@@ -17,6 +17,12 @@ export function getAndroidVersion() {
   return isAndroid ? parseInt(Platform.Version, 10) : undefined;
 }
 
+export function getSafeAreaInsets(mode) {
+  return (mode === 'landscape') ? 
+    {paddingLeft: 44, paddingRight: 44, paddingBottom: 24, paddingTop: 0} :
```
ethanshar: I would change the names here to something more general like top, left, right, bottom instead of padding. 
cause they can be used for all kind of scenarios.

Inbal-Tish: Ok.

src/components/dialog/index.js
Inbal-Tish: I just wanted to be more explicit. A View has no meaning here since it was added only for the purpose of adding the safe area.

PR #318, opened 2018-12-12
bumping lodash version

Inbal-Tish:


package.json
```diff
@@ -39,7 +39,7 @@
   "dependencies": {
     "babel-plugin-transform-inline-environment-variables": "^0.0.2",
     "hoist-non-react-statics": "^2.3.0",
-    "lodash": "^4.0.0",
+    "lodash": "^4.12.0",
```
ethanshar: Now that I look at it, this both will bring the save version 
^4.0.0 and ^4.12.0 will bring 4.x.x so I guess it's fine..

PR #323, opened 2018-12-13
Feat/animatable definitions

Inbal-Tish:
AnimationManager; Card animation + demo screen


src/style/animatableManager.js
```diff
@@ -0,0 +1,122 @@
...
+    to: {opacity: 1, translateY: 0},
+  }, 
+  cardEntrance: {
+    from: {opacity: 0, translateY: 40},
+    to: {opacity: 1, translateY: 0},
+  },
+};
+
+class AnimatableManager {
+  loadCustomDefinitions(costumDefinitions) {
```
ethanshar: The name is wrong, this method doesn't load but get definitions. 
Having said that, I actually do think should it should load the definitions, but should load only the custom definitions, while our definitions should be loaded in the manager constructor. 

I think it should work similar to our other presets. Where developers get uilib's definitions out of the box but still can load more if they want.

src/style/animatableManager.js
```diff
@@ -0,0 +1,122 @@
...
+    return {
+      animation: 'fadeIn',
+      duration: 300,
+      easing: 'ease-out-quint',
+      useNativeDriver: true,
+    };
+  }
+
+  // Components
+  getDialogAnimationProps(options) {
```
ethanshar: - I think the suffix of "AnimationProps" at the end of each method is redundant since it's clear we're talking about animaton presets, also it makes the names too long.

src/style/animatableManager.js
```diff
@@ -0,0 +1,122 @@
+import _ from 'lodash';
+
+
+/** Definitions */
+const definitions = {
+  basicListEntrance: {
```
ethanshar: the example screens shouldn't dictate our presets naming, basicList is just an example we have here. 
I would go with more generic names, maybe one that suits the component name, for example, `listItemEnrance`. 

In general, we need to think about naming standards here, because soon enough we'll have all kind of weird, conflicting names. 

maybe the following format: 
[component][entrance/exit/other action][slide/fade/other behaviour]

we can think of something else, as long we have a clear naming convention.

src/style/animatableManager.js
```diff
@@ -0,0 +1,122 @@
...
+  getDialogAnimationProps(options) {
+    return {
+      animation: 'slideInUp',
+      duration: 400,
+      useNativeDriver: true,
+      ...options,
+    };
+  }
+
+  getCardEntranceAnimationProps(index) {
```
ethanshar: This preset also comes from the example screens, if we do want to include it as part of presets it has to have more clear name, for instance, this one creates a random appearance (with a random delay), I would make it way more generic so that you can send a name of an animation, and it will return a preset with that name with the random delay behavior. 
also, currently there's a bug here, if the `index` was not sent, I think you'll get an error. 

In general, I think the methods should behave more as tools and less than just tons of presets

src/components/card/index.js
```diff
@@ -170,29 +161,32 @@ class Card extends BaseComponent {
...
-        activeOpacity={0.6}
-        testID={testID}
-        {...others}
-      >
-        {Constants.isIOS && enableBlur && <BlurView style={[this.styles.blurView, {borderRadius: brRadius}]} {...blurOptions}/>}
-        <View width={width} height={height} row={row} style={[this.styles.innerContainer, {borderRadius: brRadius}]}>
-          {this.renderChildren()}
-        </View>
-      </Container>
+      <Animatable.View {...animationProps}>
```
ethanshar: I thought we've decided on letting the developer wrap the components instead of us doing it for him.
If we do decide to keep it, I personally think we should condition this, otherwise, it creates an unnecessary hierarchy for our components (in most of the cases)

demo/src/demoApp.js
```diff
@@ -41,12 +41,26 @@ import {registerScreens} from './screens';
...
+//     from: {opacity: 0, translateY: 20},
+//     to: {opacity: 1, translateY: 0},
+//   }, 
+//   costumAnimation2: {
+//     from: {opacity: 0, translateY: 40},
+//     to: {opacity: 1, translateY: 0},
+//   },
+// };
+
+Animatable.initializeRegistryWithDefinitions(AnimatableManager.loadCustomDefinitions(/** costumDefinitions */));
```
ethanshar: The uilib animation presets should already be part of uilib when loaded, exactly like colors, typography and other presets.

demo/src/screens/animationScreens/CardAnimationsScreen.js
```diff
@@ -0,0 +1,100 @@
+import _ from 'lodash';
```
ethanshar: This example screen is great!

src/style/animatableManager.js
```diff
@@ -0,0 +1,122 @@
...
+
+  getActionSheetAnimationProps() {
+    return {
+      animation: 'slideInUp',
+      duration: 600,
+      easing: 'ease-out-quint',
+    };
+  }
+
+  getActionSheetContainerAnimationProps() {
```
ethanshar: Why create a wrapper function for a function that already exists here, why not just use it directly?

src/style/animatableManager.js
```diff
@@ -0,0 +1,122 @@
...
+  getListFadeInAnimationProps() {
+    return {
+      animation: 'fadeIn',
+      easing: 'ease-out-expo',
+      duration: 1000,
+      useNativeDriver: true,
+    };
+  }
+
+  getListLeftFadeInAnimationProps() {
```
ethanshar: This one could be unified with the other preset that gives random delays

src/style/animatableManager.js
Inbal-Tish: Done.

src/style/animatableManager.js
Inbal-Tish: Agreed.

demo/src/demoApp.js
Inbal-Tish: Done.

src/style/animatableManager.js
Inbal-Tish: Ok.

src/style/animatableManager.js
Inbal-Tish: Done.

src/components/card/index.js
Inbal-Tish: Condition added.

demo/src/screens/animationScreens/CardAnimationsScreen.js
Inbal-Tish: Tnx!

src/style/animatableManager.js
Inbal-Tish: Done.

src/style/animatableManager.js
Inbal-Tish: Done.

src/style/animatableManager.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  loadCustomDefinitions(customDefinitions) {
+    if (customDefinitions) {
+      Animatable.initializeRegistryWithDefinitions(customDefinitions);
+      this.updateDefinitions(customDefinitions);
+    }
+  }
+
+  loadSlideByHeightDefinitions(height, sufix) {
```
ethanshar: typo in suffix

src/style/animatableManager.js
```diff
@@ -0,0 +1,146 @@
...
+    definition[`slideInDown_${sufix}`] = {
+      from: {translateY: -height},
+      to: {translateY: 0},
+    };
+    definition[`slideOutUp_${sufix}`] = {
+      from: {translateY: 0},
+      to: {translateY: -height},
+    };
+    // relative
+    definition[`growUp_${sufix}`] = {
```
ethanshar: growUp/Down is not part of slide definitions, maybe we need to accept the animation type also as a param

src/style/animatableManager.js
```diff
@@ -0,0 +1,146 @@
...
+    this.updateDefinitions(definition);
+  }
+
+  updateDefinitions(newDefinitions) {
+    Object.assign(definitions, newDefinitions);
+    this.animations = getAnimations();
+  }
+
+  /** Presets */
+  getFadeIn(options) {
```
ethanshar: `getFadeIn`, `getFadeInRight` and `getSlideInUp` are simple presets, I don't think there's a need for them in the presets.

src/style/animatableManager.js
```diff
@@ -0,0 +1,146 @@
...
+    return {
+      animation: 'fadeInRight',
+      easing: 'ease-out-expo',
+      duration: 500,
+      useNativeDriver: true,
+      ...options,
+    };
+  }
+
+  getFadeInLeft(options) {
```
ethanshar: I would rename this preset to something that explains the delay behavior and also expose the delays arrays as a param to control.

src/style/animatableManager.js
```diff
@@ -0,0 +1,146 @@
...
+    return {
+      animation,
+      easing: 'ease-out-quint',
+      duration: 600,
+      delay: 10 + ((Number(index) % 12) * 100),
+      useNativeDriver: true,
+    };
+  }
+
+  getZoomInSlideDown(index = 0, options) {
```
ethanshar: the name is not clear enough, also the options are being passed both to the first item's animation and to the other items' animations. 
So if for instance, if I want to change the behavior only for the first item, it will affect the other items as well.

src/style/animatableManager.js
Inbal-Tish: Done.

src/style/animatableManager.js
Inbal-Tish: fixed the options. Not sure about the name...

src/style/animatableManager.js
Inbal-Tish: I thought one of the points is to get all the components' animatable presets in one spot...

src/style/animatableManager.js
Inbal-Tish: Or we can just call the "slideUp/Down" without the In/Out indication

demo/src/demoApp.js
```diff
@@ -48,7 +47,7 @@ import {registerScreens} from './screens';
 //   },
 // });
 
-// const costumDefinitions = {
+// const costumAnimationsDefinitions = {
```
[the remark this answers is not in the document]
Inbal-Tish: already fixed

src/style/animatableManager.js
```diff
@@ -0,0 +1,147 @@
+import _ from 'lodash';
+import * as Animatable from 'react-native-animatable';
+
+
+/** Animations Definitions */
+const definitions = {
+  indexEntrance: {
```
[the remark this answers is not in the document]
Inbal-Tish: done

src/style/animatableManager.js
```diff
@@ -0,0 +1,147 @@
...
+      animation: 'fadeInLeft',
+      easing: 'ease-out-expo',
+      duration: 600,
+      delay: _.sample(delays),
+      useNativeDriver: true,
+      ...options,
+    };
+  }
+
+  getEntranceByIndex(animation, index = 0) {
```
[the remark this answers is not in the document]
Inbal-Tish: done

src/style/animatableManager.js
```diff
@@ -0,0 +1,147 @@
...
+    to: {opacity: 1, translateY: 0},
+  },
+};
+
+class AnimatableManager {
+  constructor() {
+    this.loadCustomDefinitions(definitions);
+    this.animations = getAnimations();
+  }
+
```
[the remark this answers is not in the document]
Inbal-Tish: Done.

src/style/animatableManager.js
```diff
@@ -0,0 +1,155 @@
...
+    definition[`slideIn_${suffix}`] = {
+      from: {height: 0},
+      to: {height},
+    };
+    definition[`slideOut_${suffix}`] = {
+      from: {height},
+      to: {height: 0},
+    };
+
+    this.loadAnimationDefinitions(definition);
```
ethanshar: add return here

src/style/animatableManager.js
```diff
@@ -0,0 +1,155 @@
...
+    if (animationPresets) {
+      this.presets = Object.assign(presets, animationPresets);
+    }
+  }
+
+  // NOTE: to load globally send as a parameter to Animatable.initializeRegistryWithDefinitions() call
+  loadAnimationDefinitions(animationDefinitions) {
+    if (animationDefinitions) {
+      Animatable.initializeRegistryWithDefinitions(animationDefinitions); // Make available globally in uilib
+      const allDefinitions = Object.assign(definitions, animationDefinitions);
```
ethanshar: remove allDefinitions

PR #330, opened 2018-12-20
create a replacement component for TextInput, the TextField.

ethanshar:
TextField has some unit tests included

### DIFF
--- a/src/components/inputs/TextInput.js
+++ b/src/components/inputs/TextInput.js
@@ -129,6 +129,8 @@ export default class TextInput extends BaseInput {
 
   constructor(props) {
     super(props);
+    
+    console.warn('uilib TextInput component will be deprecated soon, please use TextField instead');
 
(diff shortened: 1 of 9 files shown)


src/components/inputs/TextInput.js
```diff
@@ -129,6 +129,8 @@ export default class TextInput extends BaseInput {
 
   constructor(props) {
     super(props);
+    
+    console.warn('uilib TextInput component will be deprecated soon, please use TextField instead');
```
Inbal-Tish: It's not really deprecated, no? We'll use it for the basic ui, so it will break only the ui, not the build.

ethanshar: Of course, currently, we have duplicates, from now on we'll only work on `TextField` for fixes and I'll create a breaking change PR for the change of `TextInput`

once we'll have few breaking PRs we'll merge them and update major version

PR #336, opened 2018-12-27
Feat/drawer

Inbal-Tish:
Adding spring to drag

### DIFF
--- a/src/interactableComponents/drawer/index.js
+++ b/src/interactableComponents/drawer/index.js
@@ -144,10 +144,10 @@ export default class Drawer extends BaseComponent {
     this.interactableElem.snapTo({index: 1});
   }
   generateStyles() {
-    this.styles = createStyles(this.props);
+    this.styles = createStyles(this.getThemeProps());
   }
   getAnimationConfig() {
-    const {animationConfig} = this.props;
+    const {animationConfig} = this.getThemeProps();
     return {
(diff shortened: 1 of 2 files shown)


src/interactableComponents/drawer/index.js
```diff
@@ -74,8 +74,8 @@ export default class Drawer extends BaseComponent {
   };
 
   static defaultProps = {
-    damping: 1 - 0.6,
-    tension: 300,
+    damping: 0.7,
```
ethanshar: I played with it, it's a bit too jumpy (: 
can you tone it down a bit?

src/interactableComponents/drawer/index.js
```diff
@@ -441,11 +447,12 @@ export default class Drawer extends BaseComponent {
     );
   }
   render() {
-    const {style, onPress, rightItems} = this.props;
+    const {style, onPress, rightItems} = this.getThemeProps();
     const Container = onPress ? TouchableOpacity : View;
+    const backgroundColor = rightItems ? rightItems[0].background || ITEM_BG : ITEM_BG;
```
ethanshar: with `_.get` you can do this line pretty easily
`_.get(rightItems, '[0].background', ITEM_BG)`, the last param is the default in case what you're looking for is undefined

PR #338, opened 2018-12-27
conditioning Animatable.View with animationsProps;

Inbal-Tish:
Adding deprecation warning for animationProps


nrutherford:
Won't this only work once you have removed the useNativeDriver prop from BaseComponent?

Inbal-Tish:
@nrutherford Good catch. Thanks!

PR #341, opened 2018-12-31
Feat/text measure

Inbal-Tish:


demo/src/screens/interactableComponentScreens/DrawerScreen.js
```diff
@@ -50,15 +61,45 @@ export default class DrawerScreen extends Component {
     }
   }
 
+  // Measure item text to calculate items' widths
+  async setRightItemsWidths() {
```
ethanshar: just for code cleanness, can we unified `setRightItemsWidths` & `setLeftItemWidth` to one method that knows to handle either one item or list of items. 
it just the code looks a bit repetitive

package.json
```diff
@@ -46,6 +46,7 @@
     "react-native-blur": "^3.1.1",
     "react-native-interactable": "1.0.0-alpha.1",
     "react-native-swipe-gestures": "^1.0.2",
+    "react-native-text-size": "^2.0.4",
```
ethanshar: we need to PR the engine to add this native as soon as possible, cause it takes time till they release versions and the native stuff are has to built there.

src/style/measureText.js
```diff
@@ -0,0 +1,23 @@
+import rnTextSize from 'react-native-text-size';
```
ethanshar: What about adding this functionality to `Typography` instead of having a new class. 
It'll be similar to `Colors` that has our presets and util functions altogether.

Inbal-Tish: But if you don't have the native linked won't you crash when importing typography in your screen?

demo/src/screens/interactableComponentScreens/DrawerScreen.js
Inbal-Tish: Done

src/style/measureText.js
ethanshar: we can require it inside the method anyway, using it will crash, but just importing it won't (I think) 
which I'm fine, we can wrap with try catch and throw nice error that says you need to link the module or something..

Inbal-Tish: require inside the method? you mean `const textMeasure = require('react-native-text-measure';`?

PR #343, opened 2019-01-02
Fix for deprecated components rule

Inbal-Tish:
To have a second alert on import (unless it's sub-component deprecation or prop deprecation)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,51 +1,59 @@
 // const validColors = require("./src/style/colorsPalette").colorsPalette;
 // const extraFixColorsMap = require("./src/style/colorsPalette").extraFixColorsMap;
 // const assetsDepJson = require('./eslint-rules/tests/assets_deprecation.json');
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -1,51 +1,59 @@
...
-    "no-unneeded-ternary": "off",
-    "object-curly-spacing": "off",
-    "react/forbid-prop-types": "off",
-    "react/jsx-filename-extension": "off",
-    "react/jsx-space-before-closing": "off",
-    "react/jsx-tag-spacing": "off",
-    "react/prefer-stateless-function": "off",
-    "react/require-default-props": "off",
+  'parser': 'babel-eslint',
+  'plugins': ['react-native', 'uilib'],
```
ethanshar: does it mean uilib rules will be applied on our src files?

Inbal-Tish: I hoped so. But it doesn't seem to work...

Inbal-Tish: Only the colors rule doesn't work because of the export vs. module.exports things

PR #344, opened 2019-01-02
adding modifiers support to TouchableOpacity

ethanshar:


src/components/touchableOpacity/index.js
```diff
@@ -67,15 +69,25 @@ export default class TouchableOpacity extends BaseComponent {
...
     return (
       <RNTouchableOpacity
         {...others}
         onPress={this.onPress}
         onPressIn={this.onPressIn}
         onPressOut={this.onPressOut}
-        style={[this.props.style, this.backgroundStyle]}
+        style={[
+          backgroundColor && {backgroundColor},
+          borderRadius && {borderRadius},
```
Inbal-Tish: Why do you do backgroundColor && {backgroundColor}?

ethanshar: I basically copied it from what I did in our `View` 
I guess the reason was not to assign something if it doesn't exist.

Inbal-Tish: oh. ok

Issue #113, opened 2018-03-08
does the Button's iconSource support the popular react-native-vector-icons's icon?


wilbyang:

ethanshar:
Currently we didn't try to support it, you're welcome to PR us with a solution.

skuzoluk:
Do you think it is a good idea to add a new prop to Button component named iconRenderer that takes function type like `iconStyle => <SomeIconComponent style={iconStyle} name="bla-bla-icon" />`
So we can check the new prop and if it exists we can just return the custom icon component, if not fallback to iconSource. Any other solution ?

RZsam:
I think it's a good idea to add `iconRenderer` to `<TextField>`  too

PR #346, opened 2019-01-03
Custom icon renderer for button component

skuzoluk:
This feature will let you to pass function to iconSource with iconStyle argument and return a custom icon component. Then we can use vector icon or some other components as in the example below:

`iconSource={iconStyle => <Icon active name="ios-add" size={24} color=iconStyle[0].tintColor style={{marginRight: iconStyle[0].marginRight}} />}`

This feature is an answer to issue #113


lumotroph:
Well done - PLEASE PUT THIS IN DOCUMENTATION! Took me an age to find this, and it makes rnuilib work very well with a very popular library!

ethanshar:
@lumotroph 
I updated the docs

PR #348, opened 2019-01-06
Fix/view hierarchy

Inbal-Tish:


src/screensComponents/modal/index.js
```diff
@@ -56,7 +57,7 @@ export default class Modal extends BaseComponent {
 
     return (
       <RNModal visible={Boolean(visible)} {...others}>
-        <Container style={{flex: 1}} blurType="light">
+        <Container style={{flex: 1}} blurType='light'>
```
ethanshar: isn't it a lint issue?

Inbal-Tish: What do you mean?

ethanshar: I see u disabled it in our eslintrc -> `'jsx-quotes': 'off',`
we need to turn it on, it was alway on.

Inbal-Tish: I don't understand how it is related to this PR?

ethanshar: it's not, I noticed it cause you use single quote instead of double quote in a JSX tag.. 
nvm, Ill merge, we'll take it later..

PR #350, opened 2019-01-08
Feat/Avatar animation

ArnonZ:


src/animations/animatedImage/index.js
```diff
@@ -56,21 +86,15 @@ export default class AnimatedImage extends BaseComponent {
...
+    const {containerStyle, loader} = this.props;
     return (
-      <View style={containerStyle} testID={testId}>
-        <Animated.Image
-          style={[{opacity: this.state.opacity}, imageStyle]}
-          source={imageSource}
-          onLoad={this.onLoad}
-        />
-        {this.state.isLoading && loader &&
+      <View testID={this.testID} style={containerStyle}>
```
ethanshar: We have too many views going on here. 
Try to see how we can reduce View Hierarchy here. 
Since Avatar appears in Lists it's important to reduce hierarchy here.

src/components/avatar/index.js
```diff
@@ -28,6 +29,14 @@ export default class Avatar extends BaseComponent {
...
   static modes = STATUS_MODES;
   static propTypes = {
+    /**
+     * Adds fade in animation when Avatar image loads
+     */
+    animate: PropTypes.bool,
+    /**
+     * Duration of the fade in animation when the Avatar image loads
+     */
+    animationDuration: PropTypes.number,
```
ethanshar: I would expect this prop to be passed using `imageProps` since it's part of `AnimatedImage`'s props

src/animations/animatedImage/index.js
```diff
@@ -36,42 +44,67 @@ export default class AnimatedImage extends BaseComponent {
...
+          source={this.source}
+          onLoad={() => this.onLoad()}
         />
-        {this.state.isLoading && loader &&
-          <View style={{position: 'absolute', top: 0, left: 0, bottom: 0, right: 0, alignItems: 'center'}}>
-            <View style={{flex: 1, flexDirection: 'row', alignItems: 'center'}}>
-              {loader}
-            </View>
+        {this.state.isLoading && loader && (
+          <View style={{position: 'absolute', top: 0, left: 0, bottom: 0, right: 0, justifyContent: 'center'}}>
```
ethanshar: use `StyleSheet.absoluteFillObject` to get this preset : `position: 'absolute', top: 0, left: 0, bottom: 0, right: 0`

src/components/avatar/index.js
```diff
@@ -165,17 +171,30 @@ export default class Avatar extends BaseComponent {
...
+        <ImageContainer
+          animate={animate}
           style={[this.styles.image, imageStyle]}
           source={imageSource}
           onLoadStart={onImageLoadStart}
           onLoadEnd={onImageLoadEnd}
           onError={onImageLoadError}
           testID={`${testID}.image`}
+          containerStyle={this.styles.container}
+          imageProps={imageProps}
```
ethanshar: why are u passing the `imageProps` as a prop here?

src/animations/animatedImage/index.js
```diff
@@ -1,22 +1,30 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import {Animated, View} from 'react-native';
+import {Animated, View, StyleSheet} from 'react-native';
+import {Image} from '../../../src';
```
ethanshar: Don't import components from `src` when you are in `src/...` path. 
We already talked about it can create a circular dependency

src/animations/animatedImage/index.js
ethanshar: @ArnonZ  Did you try going over the View hierarchy? is it possible to reduce it somehow?

PR #351, opened 2019-01-10
Feat/dialoge drag

Inbal-Tish:


src/components/dialog/index.js
```diff
@@ -53,85 +54,145 @@ class Dialog extends BaseComponent {
...
+  }
+
+  onModalShow = () => {
+    const {animationConfig} = this.getThemeProps();
+    const {deltaY} = this.state;
+ 
+    Animated.timing(deltaY, {
+      toValue: 0,
+      duration: _.get(animationConfig, 'duration', 280),
+      delay: _.get(animationConfig, 'delay', 200),
```
ethanshar: the default delay feels like a bug

src/components/dialog/index.js
```diff
@@ -53,85 +54,145 @@ class Dialog extends BaseComponent {
...
+    );
+  }
+
+  renderAnimationContainer() {
+    const {animationConfig, top} = this.getThemeProps();
+    const {alignments, deltaY} = this.state;
+    const centerByDefault = _.isEmpty(alignments);
     const animation = top ? AnimatableManager.presets.slideInDown : AnimatableManager.presets.slideInUp;
 
+    if (animationConfig && animationConfig.animation) {
```
ethanshar: The two renders are almost identical...
maybe we can do the trick when we if-else just the type of container and then render it.

demo/src/screens/componentScreens/DialogScreen.js
```diff
@@ -91,9 +104,21 @@ class DialogScreen extends Component {
           centerH
           onDismiss={() => this.setState({showDialog5: false})}
           animationConfig={{duration: 1000}}
```
ethanshar: let's remove the animationConfig here so the example will look better

src/components/panGestureView/index.js
```diff
@@ -0,0 +1,170 @@
...
+  };
+  handlePanResponderMove = (e, gestureState) => {
+    const {direction} = this.getThemeProps();
+    let newValue = 0;
+    
+    // VERTICAL
+    const up = (direction === DIRECTIONS.UP);
+    const {deltaY} = this.state;
+
+    if (Math.abs(gestureState.vy) >= 1.8) {
```
ethanshar: extract the weird values like `gestureState.vy` to constants and give them better naming also they repeat themselves

src/components/panGestureView/index.js
```diff
@@ -0,0 +1,170 @@
...
+      if ((up && gestureState.vy < 0) || (!up && gestureState.vy > 0)) {
+        // Swipe
+        this.swipe = true;
+      }
+    } else if ((up && gestureState.dy < 0) || (!up && gestureState.dy > 0)) {
+      // Drag
+      newValue = gestureState.dy;
+      Animated.spring(deltaY, {
```
ethanshar: `Animated.spring` being called twice almost the same, I would move to a function with a param to control the `toValue`

src/components/panGestureView/index.js
ethanshar: `1.8` is to arbitrary, maybe we can put in a const if it has a meaning

demo/src/screens/componentScreens/DialogScreen.js
Inbal-Tish: done

src/components/panGestureView/index.js
Inbal-Tish: done

Inbal-Tish: done

PR #352, opened 2019-01-10
Feat/hints

ethanshar:
This is phase1 of Hints, added TODOs at the top of the component with what's left todo there.


src/components/hint/index.js
```diff
@@ -0,0 +1,335 @@
+// TODO: Add support to custom hint rendering
+// TODO: Add animation
+import React from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+import {View as AnimatableView} from 'react-native-animatable';
```
Inbal-Tish: what is the purpose of this?

src/components/hint/index.js
```diff
@@ -0,0 +1,335 @@
...
+     */
+    useSideTip: PropTypes.bool,
+    /**
+     * The hint's border radius
+     */
+    borderRadius: PropTypes.number,
+    /**
+     * Hint space from screen edges
+     */
+    edgeSpace: PropTypes.number,
```
Inbal-Tish: you used 'distance' in the next prop, why use 'space' here?

src/components/hint/index.js
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  getTargetPositionOnScreen() {
+    const {targetLayout} = this.state;
+    const targetMidPosition = targetLayout.x + targetLayout.width / 2;
+    if (targetMidPosition > Constants.screenWidth * (2 / 3)) {
+      return 'right';
+    } else if (targetMidPosition < Constants.screenWidth * (1 / 3)) {
+      return 'left';
+    } else {
```
Inbal-Tish: no need for the 'else' here

src/components/hint/index.js
```diff
@@ -0,0 +1,335 @@
...
+      case 'left':
+        tipPositionStyle.left = this.useSideTip ? targetLayout.x : targetLayout.x + targetMidWidth - tipMidWidth;
+        break;
+      case 'right':
+        tipPositionStyle.right = this.useSideTip
+          ? Constants.screenWidth - targetLayout.x - targetLayout.width
+          : Constants.screenWidth - targetLayout.x - targetMidWidth - tipMidWidth;
+        break;
+      case 'center':
+      default: {
```
Inbal-Tish: why wrapping in an object here?

src/components/hint/index.js
ethanshar: What do u mean? For adding entrance animation to the hint.

src/components/hint/index.js
ethanshar: cause this is the horizontal space from the edges, we can change it maybe to margins. 
The other prop has more explicit meaning, it's the distance of the hint from the target, it's not just a space

src/components/hint/index.js
ethanshar: it's not an object, it's just the case's scope curly brackets. 
Anyway, I'll remove it

src/components/hint/index.js
Inbal-Tish: I mean why use AnimatableView instead of Animatable.View? Is the reason is that it is better to import just the View instead the * as Animatable?

ethanshar: yea, the `AnimatableView` is just an alias for `Animatable.View`. 
I didn't want to import everything.

Inbal-Tish: That's what I thought. We should use this in other imports of Animatable...

ethanshar: agree, can we merge? or is there anything else?

src/components/hint/index.js
Inbal-Tish: and 'hintDistance' is vertical distance/space/margin from the target view. I don't see much difference, I just think it is unclear when you name props that are quite similar in synonyms...

ethanshar: Yes it's vertical, What would you call these props?

PR #353, opened 2019-01-14
removing TestID from prop list; deprecating testId

Inbal-Tish:


src/components/stepper/StepperButton.js
```diff
@@ -34,6 +30,7 @@ StepperButton.propTypes = {
    * Handler function to receive updates when the value changes
    */
   onPress: PropTypes.func,
+  testID: PropTypes.string
```
ethanshar: no need to declare the prop, it's a common prop

Inbal-Tish: I know. wired, I removed all these declarations...

PR #354, opened 2019-01-14
fix for animated button - animate layout only for button

Inbal-Tish:


src/components/button/index.js
```diff
@@ -175,8 +175,8 @@ export default class Button extends BaseComponent {
     }
   }
 
-  componentDidUpdate() {
-    if (this.props.animateLayout) {
+  componentWillReceiveProps(nextProps) {
+    if (!_.isEqual(nextProps, this.props) && this.props.animateLayout) {
```
ethanshar: You can still keep using `componentDidUpdate`
`componentDidUpdate` param is `prevProps` which you can compare to `this.props`.
I rather keep using `componentDidUpdate` since it considered best practice

src/components/button/index.js
```diff
@@ -175,8 +175,8 @@ export default class Button extends BaseComponent {
     }
   }
 
-  componentDidUpdate() {
-    if (this.props.animateLayout) {
+  componentDidUpdate(prevProps) {
+    if (!_.isEqual(prevProps, this.props) && this.props.animateLayout) {
```
ethanshar: One more thing I just noticed...
Better check `this.props.animateLayout` before checking the equality of the props
since the equal check is more expensive and can be avoided if the `animateLayout` prop is off.

Inbal-Tish: done.

PR #359, opened 2019-01-22
adding fix type component_name 

Inbal-Tish:
for deprecated components and their import specifier


ethanshar:
let's add unit test

PR #364, opened 2019-02-02
Add labelStyle prop to Badge component

btoo:


src/components/badge/index.js
```diff
@@ -136,7 +140,7 @@ export default class Badge extends BaseComponent {
   renderLabel() {
     return (
       <Text
-        style={[this.styles.label, this.isSmallBadge() && this.styles.labelSmall]}
+        style={[this.styles.label, this.isSmallBadge() && this.styles.labelSmall, this.props.labelStyle]}
```
ArnonZ: Please use `{labelStyle}` in shorthand form to match our code convention:
`{labelStyle} = this.props`
and use it inside style prop as follows: `style={[this.styles.label, this.isSmallBadge() && this.styles.labelSmall, labelStyle]}`

btoo: @ArnonZ ok i’ve made the necessary updates. let me know what you think

PR #370, opened 2019-02-12
Bug/Avatar badge position

ArnonZ:
Fixed badge position

### DIFF
--- a/src/components/avatar/index.js
+++ b/src/components/avatar/index.js
@@ -142,10 +142,11 @@ export default class Avatar extends BaseComponent {
 
   getBadgePosition() {
     const {size} = this.props;
+    const badgeSize = 13.5;
     const radius = size / 2;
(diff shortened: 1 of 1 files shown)


src/components/avatar/index.js
```diff
@@ -141,14 +140,23 @@ export default class Avatar extends BaseComponent {
     return badgeColor;
   }
 
+  getBadgePosition() {
+    const {size} = this.props;
+    const radius = size / 2;
+    const x = Math.sqrt(radius ** 2 * 2);
+    const y = x - radius;
+    const shift = Math.sqrt(y ** 2 / 2) - 13.5 / 2; // 13.5 = hard coded badge size
```
ethanshar: Move the size of the pimple to a const at the top of the file and reuse it. 
this way you don't need to add comments explaining what you're doing...

PR #372, opened 2019-02-14
Adding generateColorPalette method from given color

Inbal-Tish:


src/style/colors.js
```diff
@@ -81,22 +77,52 @@ class Colors {
...
+
+  generateColorPalette(color) {
+    const hsl = Color(color).hsl();
+    const lightness = Math.round(hsl.color[2]);
+    
+    const ls = [lightness];
+    let l = lightness;
+    do {
+      l -= 10;
+      if (l > 20) {
```
ethanshar: what's the point of having the condition `l > 20` twice? in the `while` and in the `if` statements
instead, you can initialize `l` with `lightness - 10`  and do a simple while

src/style/colors.js
```diff
@@ -81,22 +77,52 @@ class Colors {
...
+
+    l = lightness;
+    do {
+      l += 10;
+      if (l < 100) {
+        ls.push(l);
+      }
+    } while (l < 100);
+
+    ls.sort((a, b) => a - b);
```
ethanshar: You can use `ls.unshift(l)` instead of using `push` and this way avoid sorting the list afterward. 
`unshift` will add items to the array from the beginning.

src/style/colors.js
```diff
@@ -81,22 +77,52 @@ class Colors {
...
-      } else if (tintLevel <= BASE_COLOR_LEVEL) {
-        const darkRatio = darkRatios[tintLevel - 1];
-        return Color(color).darken(darkRatio).hex();
-      } else {
-        const lightRatio = lightRatios[tintLevel - 4];
-        return Color(color).mix(Color('#ffffff'), lightRatio).hex();
-      }
+    } else { 
+      // Handles dynamic colors (non uilib colors)
+      let tintLevel = Math.floor((Number(tintKey) / 10) + 1);
```
ethanshar: why the `+ 1` here?

Inbal-Tish: to match the 10 - 80 values to my 20 - 90 range

src/style/colors.js
Inbal-Tish: I will still have to sort since I don't know the position of my given value and I add it in the beginning.

src/style/colors.js
Inbal-Tish: but it will still do the last loop iteration, which I don't want in my array...

ethanshar: are you sure?  if you do the following
```
let l = lightness -10;
while (l > 20) {
  ls.push(l);
  l -= 10;
}
```
This doesn't do the same?

src/style/colors.js
ethanshar: but the given position is the middle (not really middle, but between the low and the high values) of the array and you add numbers before it and numbers after it, this way it should already be sorted, no?

Inbal-Tish: oh, you mean to use it only for the first do-while? That can work...

Inbal-Tish: Done.

src/style/colors.js
Inbal-Tish: Done.

PR #374, opened 2019-02-18
Feat/Avatar to use badge

ArnonZ:


src/components/avatar/index.js
```diff
@@ -39,6 +50,14 @@ export default class Avatar extends BaseComponent {
      * Background color for Avatar
      */
     backgroundColor: PropTypes.string,
+    /**
+     * Badge location on Avatar
+     */
+    badgePosition: PropTypes.oneOf(Object.keys(BADGE_POSITIONS)),
```
ethanshar: although they are the same, it should be `Object.values` instead of keys. 
since what's being passed eventually is the value 
Theoretically, if someone will change the values, your code will be broken

src/components/avatar/index.js
```diff
@@ -5,20 +5,29 @@ import {StyleSheet, ViewPropTypes, TouchableOpacity} from 'react-native';
...
   AWAY: 'AWAY',
   NONE: 'NONE',
 };
 
+export const BADGE_POSITIONS = {
+  TOP_RIGHT: 'TOP_RIGHT',
+  BOTTOM_RIGHT: 'BOTTOM_RIGHT',
+};
+
+const DEFAULT_BADGE_SIZE = 'pimpleBig';
```
ethanshar: get this value from `Badge.badgeSizes`

src/components/badge/index.js
```diff
@@ -35,7 +45,7 @@ export default class Badge extends BaseComponent {
     /**
      * the badge size (default, small)
      */
-    size: PropTypes.oneOf(['default', 'small', 'pimpleBig', 'pimpleSmall']),
+    size: PropTypes.oneOf(['default', 'small', 'pimpleSmall', 'pimpleBig', 'pimpleHuge']),
```
ethanshar: consider using  `PropTypes.oneOf(Object.keys(BADGE_SIZE))` here, do we dont have to update this propType all the time. 
(btw, here it's right to use `Object.keys`)

PR #379, opened 2019-02-21
add animation to radio button

M-i-k-e-l:


src/components/radioButton/RadioButton.js
```diff
@@ -48,7 +48,61 @@ class RadioButton extends BaseComponent {
...
+    this.selected = undefined;
+    this.selectedPrevState = undefined;
+  }
+
+  componentDidMount() {
+    this.animate();
+  }
+
+  componentDidUpdate(prevProps) {
+    if (!this.props.selected) { // will always be radio group
```
Inbal-Tish: It will also be called for unselected individual

src/components/radioButton/RadioButton.js
```diff
@@ -48,7 +48,61 @@ class RadioButton extends BaseComponent {
...
+        this.animate();
+      }
+    } else if (prevProps.selected !== this.props.selected) { // will always be individual
+      this.selected = this.props.selected;
+      this.animate();
+    }
+  }
+
+  animate() {
+    const ANIMATION_TIME = 150;
```
Inbal-Tish: You can move these consts to the top or not use uppercase

src/components/radioButton/RadioButton.js
```diff
@@ -58,7 +112,14 @@ class RadioButton extends BaseComponent {
     const {value, disabled} = this.props;
     if (!disabled) {
       _.invoke(context, 'onValueChange', value);
-      _.invoke(this.props, 'onPress', this.isSelected(this.props, context));
+      const selected = this.isSelected(this.props, context);
```
Inbal-Tish: Why you're not using `this.selected` here instead of calling `this.isSelected()` again?
Did you check if the value sent to `onPress` here is now the correct one?

PR #381, opened 2019-02-26
Feat/add support for radio button label

M-i-k-e-l:


src/components/radioButton/RadioButton.js
```diff
@@ -165,21 +191,45 @@ class RadioButton extends BaseComponent {
...
+  getIconStyle() {
+    const {iconStyle} = this.getThemeProps();
+    const style = [this.styles.image];
+
+    return [style, iconStyle];
+  }
+
+  renderLabel() {
+    return (
+      this.props.label && (<Text marginL-10 style={this.props.labelStyle} >{this.props.label}</Text>)
```
ethanshar: Destruct the props you're using at the beginning of the method. 
Other than readability,  it will also save you from writing `this.props.label` twice

src/components/radioButton/RadioButton.js
```diff
@@ -165,21 +191,45 @@ class RadioButton extends BaseComponent {
     return style;
   }
 
+  getIconStyle() {
```
ethanshar: I don't think there's a need to extract the functionality of getting the icon style into a separate method. 
Basically, this code can be part of `renderIcon`

src/components/radioButton/RadioButton.js
```diff
@@ -133,9 +159,9 @@ class RadioButton extends BaseComponent {
     return value === selectedValue;
   }
 
-  getContainerStyle() {
+  getCircleStyle() {
```
ethanshar: `circle` is a little bit abstract, you can call it radio button since it's the actual radio button. 
before when it was `container` it was enough since all this component rendered is a radio button.

M-i-k-e-l: But this is only the outer circle, not the whole radio button

src/components/radioButton/RadioButton.js
ethanshar: @M-i-k-e-l you missed this one (:

PR #384, opened 2019-02-28
Feat/checkbox animation

ArnonZ:


src/components/checkbox/index.js
```diff
@@ -50,37 +69,46 @@ class Checkbox extends BaseComponent {
...
      */
-    iconColor: PropTypes.string
+    iconColor: PropTypes.string,
   };
 
   generateStyles() {
     this.styles = createStyles(this.getThemeProps());
   }
 
+  check(value) {
```
ethanshar: let's give it a better name so we'll know we're animating something here. 
`toggleCheckAnimation`

src/components/checkbox/index.js
```diff
@@ -50,37 +69,46 @@ class Checkbox extends BaseComponent {
...
-      if (disabled) {
-        style.push({backgroundColor: Colors.dark70, borderColor: Colors.dark70});
-      } else {
-        style.push(color ? {backgroundColor: color, borderColor: color} : this.styles.containerSelected);
-      }
-    } else if (disabled) {
-      style.push({borderColor: Colors.dark70});
-    } else if (color) {
-      style.push({borderColor: color});
+  getColors(element) {
```
ethanshar: something with function looks over complicated.I believe we can do something simpler. 

the colors we use are either the passed color or the disabled color. and we use them both for the border and the background. 
create a simple function that returns the color (no need to pass element and create dynamic objects, it's making it less readable) 

now create a function for getting border style and for getting the background style. 
I would even call them getOuterStyle and getInnerStyle.

PR #386, opened 2019-02-28
Feat/allow customizing android wheel picker style

M-i-k-e-l:


lib/components/WheelPicker/index.js
```diff
@@ -32,14 +44,27 @@ class WheelPicker extends Component {
   constructor(props) {
     super(props);
     this.onValueChange = this.onValueChange.bind(this); //eslint-disable-line
-
     this.getItems();
+    DeviceEventEmitter.addListener('log', this.onLogReceived); // TODO: consider moving to a more generic place (base class?)
```
Inbal-Tish: That's a good idea moving it to a base class. Open a task for it.

PR #388, opened 2019-03-04
Feat/refactor radio button and use as radio group child hoc

M-i-k-e-l:


demo/src/screens/componentScreens/RadioButtonScreen.js
```diff
@@ -39,27 +62,51 @@ export default class RadioButtonScreen extends Component {
...
-                <Text marginT-10>You chose: {this.state.textSide}</Text>
-              </RadioGroup>
-            </View>
+        <View flex>
+          <ScrollView style={{padding: 20}}>
+            <RadioGroup value={this.state.color} onValueChange={value => this.setState({color: value})}>
+              <Text marginB-20 text60 dark10>
+                Select a color
+              </Text>
+              {this.renderRadioButton('orange', 'Orange')}
```
ethanshar: since we now support the label, it's easier to render a radiobutton with a label with a one liner
`<RadioButton value={value} label={text} marginB-5/>`
we can clean the example screen a little to remove the `renderRadioButton` function since it doesn't do much now. 

In general, we should keep the example screen clean and readable since that's where we address people to learn how to use our components.

demo/src/screens/componentScreens/RadioButtonScreen.js
```diff
@@ -39,27 +62,51 @@ export default class RadioButtonScreen extends Component {
...
+            </RadioGroup>
+
+            <RadioGroup
+              value={this.state.messageType || null} onValueChange={value => this.setState({messageType: value})}
+            >
+              <Text marginV-20 text60 dark10>
+                Select Message type{'\n'}
+                (enum with default value)
+              </Text>
+              {this.renderRadioButton(null, 'Default')}
```
ethanshar: I would include the default example with null with the first example in the screen, no need to create too many of them.

demo/src/screens/componentScreens/RadioButtonScreen.js
```diff
@@ -39,27 +62,51 @@ export default class RadioButtonScreen extends Component {
...
+              <Text marginT-10>You chose: {this.state.textSide}</Text>
+            </RadioGroup>
+
+            <RadioGroup
+              value={this.state.transportationType} onValueChange={value => this.setState({transportationType: value})}
+            >
+              <Text marginV-20 text60 dark10>
+                Select transportation type (enum)
+              </Text>
+              {this.renderRadioButtonForEnum(RadioButtonScreen.transportationTypes.CAR)}
```
ethanshar: What's the purpose of this example? it's pretty similar to the first one

demo/src/screens/componentScreens/RadioButtonScreen.js
M-i-k-e-l: This is the section title, the `RadioButton`s have been changed to include the label

PR #389, opened 2019-03-04
added an option to show avatar as selected

doronpr:
image


ArnonZ:
https://github.com/wix-private/wix-react-native-ui-lib/pull/471

PR #400, opened 2019-03-07
adjusting alignments

Inbal-Tish:


src/components/inputs/TextField.js
```diff
@@ -319,19 +308,19 @@ export default class TextField extends BaseInput {
             !centered && {
               top: floatingPlaceholderState.interpolate({
                 inputRange: [0, 1],
-                outputRange: [multiline ? 30 : 28, multiline ? 7 : 0],
+                outputRange: [multiline ? 30 : 28, multiline ? (Constants.isAndroid ? 0 : 5) : 0]
```
ethanshar: Now when I look at this line again, not sure why I didn't do like this
`multiline ? [30, (Constants.isAndroid ? 0 : 5)] : [28 : 0]`
do you think it's more readable?

Inbal-Tish: Looks the same to me...

PR #404, opened 2019-03-10
Feat/text field right icon

Inbal-Tish:


src/components/inputs/TextField.js
```diff
@@ -587,30 +585,35 @@ function createStyles({placeholderTextColor, centered, multiline}) {
...
+      borderColor: Colors.dark70
     },
     innerContainerWithoutUnderline: {
       borderBottomWidth: 0
     },
     input: {
-      flexGrow: 1,
-      marginBottom: Constants.isIOS ? 10 : 5,
-      padding: 0,
+      flex: 1,
```
ethanshar: I checked again regarding the change with `flexGrow`, when I implemented it I've compared it with the behavior of RN's TextInput component, cause I wanted to provide a consistent behavior. 

The following examples should render the same
```
<View row bg-red50>
          <TextField hideUnderline enableErrors={false} _containerStyle={{flex: 1}} placeholder="enter text..." />
          <Text>text</Text>
        </View>
        <View row bg-blue50>
          <TextInput _containerStyle={{flex: 1}} placeholder="enter text..." />
          <Text>text</Text>
        </View>
```

but after the change, they're not. 

if we're still having trouble with supporting rightIcon without changing the `flexGrow` maybe we can to the least, condition it. 
if `rightIcon` passed, use `flex`, otherwise use `flexGrow`

Inbal-Tish: The condition won't solve the issue with a View row wrapper...

Inbal-Tish: I've added changes and design request for this feature.

demo/src/screens/componentScreens/InputsScreen.js
```diff
@@ -242,6 +239,32 @@ export default class InputsScreen extends Component {
             transformer={transformPrice}
           />
 
+          <TextField
+            text70
+            containerStyle={{marginBottom: INPUT_SPACING, width: 160}}
```
ethanshar: I would leave one example with no width limitation

src/components/inputs/TextField.js
```diff
@@ -642,6 +679,19 @@ function createStyles({placeholderTextColor, centered, multiline}) {
...
+    rightIcon: {
+      alignSelf: 'center', 
+      marginLeft: 8
+    },
+    rightButton: {
+      position: 'absolute',
+      right: 3,
+      alignSelf: 'center'
+    },
+    rightButtonImage: {
```
ethanshar: I wouldn't transform the size of the image, it might break the image proportion.

src/components/inputs/TextField.js
```diff
@@ -480,6 +497,7 @@ export default class TextField extends BaseInput {
     const {lineHeight, ...typographyStyle} = typography;
     const color = this.getStateColor(this.props.color || this.extractColorValue());
     const inputStyle = [
+      this.shouldDisplayRightButton() && {paddingRight: 33},
```
ethanshar: The 33 is a random number? what if the image is wider?

src/components/inputs/TextField.js
```diff
@@ -642,6 +679,19 @@ function createStyles({placeholderTextColor, centered, multiline}) {
     label: {
       ...LABEL_TYPOGRAPHY,
       height: LABEL_TYPOGRAPHY.lineHeight
+    },
+    rightIcon: {
+      alignSelf: 'center', 
+      marginLeft: 8
+    },
+    rightButton: {
+      position: 'absolute',
```
ethanshar: I don't think the alignment in the example looks good, maybe, it's worth checking with Yulia if that what she really wants. 

Anyway, even If I don't use floatingPlaceholder, I would expect the button/image to be aligned nicely to the text, but it's not.

Inbal-Tish: I told her that and asked her if that what's she wants. She said yes

src/components/inputs/TextField.js
Inbal-Tish: the image is limited to a certain size. The 33 is the image width + the gap between the image and the text

src/components/inputs/TextField.js
Inbal-Tish: It is not designed to contain just any image, the point is to have an icon of size 24 there. Or not?

demo/src/screens/componentScreens/InputsScreen.js
Inbal-Tish: Ok. Done

src/components/inputs/TextField.js
Inbal-Tish: Made it more explicit.

src/components/inputs/TextField.js
ethanshar: but it's open source, people can pass whatever.

PR #405, opened 2019-03-10
support selection mode in Card

ethanshar:
I had to fix my recent change in Card component and to always allow overflow to be visible (for Android and iOS). this is in order to support the overflowing of the selection mode. 
regardless, I guess it makes sense to avoid hiding card's overflow content.

### DIFF
--- a/demo/src/screens/componentScreens/CardsScreen.js
+++ b/demo/src/screens/componentScreens/CardsScreen.js
@@ -11,15 +11,12 @@ const cardImage2 = require('../../assets/images/empty-state.jpg');
 
 export default class CardsScreen extends Component {
   state = {
-    selected: true,
-  };
-
-  selectCard = () => {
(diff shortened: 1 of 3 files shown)


demo/src/screens/componentScreens/CardsScreen.js
```diff
@@ -43,43 +52,47 @@ export default class CardsScreen extends Component {
               })}
             </ScrollView>
 
-            <Card height={120} width={120} marginV-20>
-              <Card.Image height={'100%'} imageSource={cardImage} />
-            </Card>
+            <View row>
+              <Card height={120} flex marginV-20 marginR-20>
```
Inbal-Tish: I would give this one `selected` as well since they are in the same row...

ethanshar: done

PR #407, opened 2019-03-10
Feat/avatar icon badge

ArnonZ:
- Refactored Avatar component to properly enable icon badge.
- Includes general refactor (+ deprecation warning) to enable future deprecation of badgeColor logic, props: isOnline, status.
- Refactored tests to match new Avatar Badge logic.


src/components/avatar/index.js
```diff
@@ -116,7 +131,7 @@ export default class Avatar extends BaseComponent {
     /**
      * AWAY, ONLINE, OFFLINE or NONE mode (if set to a value other then 'NONE' will override isOnline prop)
      */
-    status: PropTypes.oneOf(Object.keys(STATUS_MODES)),
+    status: PropTypes.oneOfType([PropTypes.string, PropTypes.oneOf(Object.keys(STATUS_MODES))]),
```
ethanshar: why did u add `PropTypes.string` here?

src/components/avatar/index.js
```diff
@@ -142,28 +156,48 @@ export default class Avatar extends BaseComponent {
...
   }
-  
+
   getBadgeSize = () => _.get(this.props, 'badgeProps.size', DEFAULT_BADGE_SIZE);
-  
-  getBadgeBorderWidth = () => _.get(this.props, 'badgeProps.borderWidth', DEFAULT_BADGE_BORDER_WIDTH);
+
+  getBadgeBorderWidth() {
+    const {badgeProps} = this.props;
+    const borderWidths = _.get(this.props, 'badgeProps.borderWidth', DEFAULT_BADGE_BORDER_WIDTH);
```
ethanshar: why plural?

src/components/avatar/index.js
```diff
@@ -142,28 +156,48 @@ export default class Avatar extends BaseComponent {
...
+
+    if (_.get(this.props, 'badgeProps.icon')) {
+      return badgeProps.borderWidth || 0;
```
ethanshar: you've extracted the `borderWidth` couple of lines before, why not use it here? 
maybe I don't understand the logic

src/components/avatar/index.js
```diff
@@ -142,28 +156,48 @@ export default class Avatar extends BaseComponent {
...
+    }
+    return borderWidths;
+  }
+
+  getBorderColor() {
+    const {badgeProps} = this.props;
+    const borderColors = _.get(this.props, 'badgeProps.borderColor', DEFAULT_BADGE_BORDER_COLOR);
```
ethanshar: again, plural. 
also, not sure I understand the logic of both `getBadgeBorderWidth` and `getBorderColor` 
why not just pass the badgeProps as is to the `Badge` component.

src/components/avatar/index.js
ArnonZ: Resolved

src/components/avatar/index.js
ArnonZ: all badge types except icon badge should DEFAULT_BADGE_BORDER_WIDTH. So for Icon badge is you don't pass borderWidth you should get 0. Why not undefined || null. I use border width for calculating badge location on Avatar. so undefined is out of the question. Null seemed less "safe" than 0

src/components/avatar/index.js
ArnonZ: It's meant to give icon badge a specific treatment but in the case of border color since in `getBorderWidth()` icon badge gets 0 if undefined I don't see a harm in passing  DEFAULT_BADGE_BORDER_COLOR (which is white) since it won't show anyway.

src/components/avatar/index.js
ethanshar: it feels like this logic should move to our private repo, no? it's all very specific to our product's use case unless it feels like good UX guidelines we should enforce.

eslint-rules/tests/component_deprecation.json
```diff
@@ -7,6 +7,14 @@
       {
         "prop": "url",
         "message": "Please use the 'imageSource' prop instead."
+      },
```
ethanshar: btw, component depreciation here is merely for example, no need to update it

ArnonZ: 🤣 revereted

src/components/avatar/index.js
```diff
@@ -142,28 +156,40 @@ export default class Avatar extends BaseComponent {
...
   }
-  
+
   getBadgeSize = () => _.get(this.props, 'badgeProps.size', DEFAULT_BADGE_SIZE);
-  
-  getBadgeBorderWidth = () => _.get(this.props, 'badgeProps.borderWidth', DEFAULT_BADGE_BORDER_WIDTH);
+
+  getBadgeBorderWidth() {
+    const {badgeProps} = this.props;
+    const borderWidth = _.get(this.props, 'badgeProps.borderWidth', DEFAULT_BADGE_BORDER_WIDTH);
```
ethanshar: let's stop supporting the DEFAULT_BADGE_BORDER_WIDTH which will clean the code

PR #412, opened 2019-03-12
use Modal in Hint when onBackgoundPress was passed

ethanshar:


src/components/hint/index.js
```diff
@@ -132,9 +133,19 @@ class Hint extends BaseComponent {
     }
   };
 
+  get targetLayout() {
```
Inbal-Tish: This method is called 54 times but the targetLayout only updated twice...

src/components/hint/index.js
```diff
@@ -2,14 +2,15 @@
...
 import {View as AnimatableView} from 'react-native-animatable';
 import {Typography, Spacings, Colors, BorderRadiuses, AnimatableManager} from '../../style';
 import {Constants} from '../../helpers';
 import {BaseComponent} from '../../commons';
 import View from '../view';
 import Text from '../text';
 import Image from '../image';
+import Modal from '../../screensComponents/modal';
 
 const sideTip = require('./assets/hintTipSide.png');
```
Inbal-Tish: The side tip doesn't render nicely with the borderRadius of the hint...

demo/src/screens/componentScreens/HintsScreen.js
```diff
@@ -10,6 +10,7 @@ export default class HintsScreen extends Component {
       showBottomHint: false,
       showIcon: false,
       targetPosition: 'flex-start',
+      // useTargetFrame: true,
```
Inbal-Tish: on Android, all the redioButtons are selected in the screen

src/components/hint/index.js
ethanshar: it's just a getter to wrap the logic that returns the relevant `targetLayout`
cause each scenario uses a different value

src/components/hint/index.js
ethanshar: fixed

demo/src/screens/componentScreens/HintsScreen.js
ethanshar: what do u mean?

Inbal-Tish: When I run it in the device all the buttons are marked blue - selected.

ethanshar: not sure why, when I run on different emulators it looks good. 
anyway, if anything it's a radioButton issue, let's create a separate ticket to check it.

PR #415, opened 2019-03-12
WheelPicker - handle undefined labelStyle

M-i-k-e-l:


lib/components/WheelPicker/index.js
```diff
@@ -85,7 +85,7 @@ class WheelPicker extends Component {
 
   render() {
     const {style, color, labelStyle, itemHeight} = this.props;
-    const {fontSize, fontFamily} = labelStyle;
+    const {fontSize, fontFamily} = labelStyle || {fontSize: undefined, fontFamily: undefined};
```
Inbal-Tish: Why not set default labelStyle? also, if you open it from FormScreen the labels are align right instead of center

PR #417, opened 2019-03-17
Feat/rtl support

Inbal-Tish:


src/components/switch/index.js
```diff
@@ -124,9 +126,11 @@ class Switch extends BaseComponent {
   renderThumb() {
     const {thumbStyle} = this.getThemeProps();
     const {thumbPosition} = this.state;
+    const output = this.calcThumbOnPosition() * (Constants.isRTL ? -1 : 1);
```
ethanshar: I think it would be nicer to move the RTL logic to the `calcThumbOnPosition`, makes it cleaner to read.

src/components/inputs/TextField.js
```diff
@@ -508,6 +508,7 @@ export default class TextField extends BaseInput {
       this.styles.input,
       hideUnderline && this.styles.inputWithoutUnderline,
       {...typographyStyle},
+      Constants.isRTL && {minHeight: lineHeight + 3},
```
ethanshar: better confirm there're no conflicts with the expandable refactor

src/components/switch/index.js
Inbal-Tish: Done

PR #420, opened 2019-03-19
expandable to render textInput instead of text

Inbal-Tish:


src/components/inputs/TextField.js
```diff
@@ -436,51 +433,19 @@ export default class TextField extends BaseInput {
...
     }
 
-    const shouldShowPlaceholder = _.isEmpty(value) && !floatingPlaceholder;
-
     return (
       <TouchableOpacity
         style={this.styles.expandableInput}
         activeOpacity={1}
         onPress={() => !this.isDisabled() && this.toggleExpandableModal(true)}
-        testID={testID}
```
ethanshar: I think the testID here was meant for things like detox and triggering the press of `TouchableOpacity`
let's add testID with a suffix like we usually do - `${testID}.expandable` or something like that

Inbal-Tish: Yes. But in this case, this is where you trigger the click, equivalent to the input when you render just a textInput, so they should be treated the same in my opinion.

PR #421, opened 2019-03-19
Feat/ProgressBar - Add support for border radius

ArnonZ:
- Added support for BorderRadius.

### DIFF
--- a/src/components/animatedScanner/index.js
+++ b/src/components/animatedScanner/index.js
@@ -160,20 +160,12 @@ export default class AnimatedScanner extends BaseComponent {
 function createStyles() {
   return StyleSheet.create({
     container: {
-      position: 'absolute',
-      top: 0,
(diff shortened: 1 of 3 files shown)


src/components/animatedScanner/index.js
```diff
@@ -103,14 +103,15 @@ export default class AnimatedScanner extends BaseComponent {
   }
 
   renderNew() {
-    const {opacity, backgroundColor, hideScannerLine} = this.props;
+    const {opacity, backgroundColor, borderRadius, hideScannerLine} = this.props;
```
ethanshar: I think it's preferable to allow passing `style` instead of exposing specific prop. 
I guess the thumb rule for whether to use a `style` or an explicit prop is if we're using the prop's value for internal component logic. 
In this case, since it's just cosmetics, let's use `style`.

PR #422, opened 2019-03-20
Feat/Added accessibilityLabel

ArnonZ:
- Added accessibilityLabel to Switch, Checkbox, RadioButton

### DIFF
--- a/src/components/checkbox/index.js
+++ b/src/components/checkbox/index.js
@@ -115,11 +115,11 @@ class Checkbox extends BaseComponent {
 
   render() {
     const {value, selectedIcon, style, color, iconColor, disabled, testID, ...others} = this.getThemeProps();
-    const checkboxValueForDetox = value ? 'checked' : 'unchecked';
+    const accessibilityLabel = value ? 'checked' : 'unchecked';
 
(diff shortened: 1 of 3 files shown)


src/components/checkbox/index.js
```diff
@@ -115,8 +115,11 @@ class Checkbox extends BaseComponent {
 
   render() {
     const {value, selectedIcon, style, color, iconColor, disabled, testID, ...others} = this.getThemeProps();
+    const checkboxValueForDetox = value ? 'checked' : 'unchecked';
```
ethanshar: I'm not a fan of having `detox` in the variable name.
maybe just call it `accessibilityLabel`.

src/components/switch/index.js
```diff
@@ -138,8 +138,11 @@ class Switch extends BaseComponent {
 
   render() {
     const {value, style, ...others} = this.getThemeProps();
+    const switchValueForDetox = value ? 'switchedOn' : 'switchedOff';
```
ethanshar: I prefer `on` & `off`
or `switchOn` & `switchOff`

PR #424, opened 2019-03-24
WheelPicker - add color to labelStyle

M-i-k-e-l:


lib/components/WheelPicker/index.js
```diff
@@ -24,7 +24,9 @@ class WheelPicker extends Component {
      */
     style: PropTypes.oneOfType([PropTypes.object, PropTypes.number]),
     /**
-     * pass custom label style: fontSize, fontFamily
+     * pass custom label style: fontSize, fontFamily, color<br>
+     * Note: label's color will override the WheelPicker's color<br>
```
Inbal-Tish: Why is the label color overrides the picker color?

lib/components/WheelPicker/index.js
```diff
@@ -91,6 +93,11 @@ class WheelPicker extends Component {
   render() {
     const {style, color, labelStyle, itemHeight} = this.props;
     const {fontSize, fontFamily} = labelStyle;
+    let {color: labelColor} = labelStyle;
```
Inbal-Tish: What is `labelColor` here?

lib/components/WheelPicker/index.js
M-i-k-e-l: That was the requirement as I understood it

lib/components/WheelPicker/index.js
M-i-k-e-l: The color of the text

PR #426, opened 2019-03-31
Feat/custom action sheets

M-i-k-e-l:


src/components/actionSheet/index.js
```diff
@@ -59,6 +59,27 @@ export default class ActionSheet extends BaseComponent {
      * When passed (only with useNativeIOS), will display a cancel button at the bottom (overrides cancelButtonIndex)
      */
     showCancelButton: PropTypes.bool,
+    /**
+     * Add or override style of the action sheet
+     */
+    sheetStyle: ViewPropTypes.style,
```
Inbal-Tish: Is sheelStyle is a containerStyle?

src/components/actionSheet/index.js
```diff
@@ -59,6 +59,27 @@ export default class ActionSheet extends BaseComponent {
...
+    /**
+     * Add or override style of the dialog wrapping the action sheet
+     */
+    dialogStyle: ViewPropTypes.style,
+    /**
+     * Add or override style of the options list
+     */
+    listStyle: ViewPropTypes.style,
```
Inbal-Tish: is listStyle is action/optionStyle?

src/components/actionSheet/index.js
```diff
@@ -59,6 +59,27 @@ export default class ActionSheet extends BaseComponent {
...
+    /**
+     * Render custom title
+     */
+    renderTitle: PropTypes.func,
+    /**
+     * Render custom action<br>
```
Inbal-Tish: <br>???

src/components/actionSheet/index.js
```diff
@@ -134,12 +155,15 @@ export default class ActionSheet extends BaseComponent {
...
   renderActions() {
-    const {title, options, cancelButtonIndex} = this.props;
+    const {title, options, cancelButtonIndex, renderAction, listStyle} = this.props;
     const optionsToRender = _.filter(options, (option, index) => index !== cancelButtonIndex);
     
     return (
-      <View paddingB-8 paddingT-8={_.isEmpty(title)}>
-        {_.map(optionsToRender, this.renderAction)}
+      <View style={[_.isEmpty(title) ? styles.listNoTitle : styles.listWithTitle, listStyle]}>
+        {renderAction ?
```
Inbal-Tish: You are not checking if renderAction is a function before you call it?

src/components/actionSheet/index.js
```diff
@@ -158,17 +182,19 @@ export default class ActionSheet extends BaseComponent {
...
 
-  renderSheet() {    
+  renderSheet() {
+    const {renderTitle} = this.props;
+    const {sheetStyle} = this.getThemeProps();
     return (
-      <View bg-white>
-        {this.renderTitle()}
+      <View style={[styles.sheet, sheetStyle]} >
+        {renderTitle ? renderTitle() : this.renderTitle()}
```
Inbal-Tish: You are not checking if renderTitle is a function before you call it?

src/components/actionSheet/index.js
```diff
@@ -188,3 +214,19 @@ export default class ActionSheet extends BaseComponent {
...
+  },
+  dialog: {
+    backgroundColor: Colors.white
+  },
+  listWithTitle: {
+    paddingBottom: 8
+  },
+  listNoTitle: {
+    paddingTop: 8,
+    paddingBottom: 8
```
Inbal-Tish: Notice the duplication of paddingBottom. I like to pass one style to both cases (ex. action) and pass another one to the other case (ex. actionNoTitle). 
Also, it is better to move this value (8) to a const so it will be one source for the bunch.

src/components/actionSheet/index.js
M-i-k-e-l: It's a container of the title and actions, but it is a view inside the dialog, do you think containerStyle is a better name?

src/components/actionSheet/index.js
M-i-k-e-l: It is the style of all the actions, not a single one, do you like actionsListStyle more?

src/components/actionSheet/index.js
M-i-k-e-l: Not sure what is the problem here, is it the naming? - It is the same convention as in Picker

src/components/actionSheet/index.js
M-i-k-e-l: I get what you are saying, but in this case I feel like the code is more readable this way.
I'll add a const.

src/components/actionSheet/index.js
Inbal-Tish: I do. We use 'containerStyle' when the style implied on the wrapping views of the component

src/components/actionSheet/index.js
Inbal-Tish: I think using the word 'list' here is confusing (even if it eventually renders List item). We are referring to these as 'options' and 'actions' so we should stick to one and not add another one to the mix.

src/components/actionSheet/index.js
Inbal-Tish: The tag `br`?

src/components/actionSheet/index.js
Inbal-Tish: My role in these situations is if I want to change the one value, would I need to change the other one as well? if so, I'm using a const. What is the case here?

src/components/actionSheet/index.js
M-i-k-e-l: Done

src/components/actionSheet/index.js
M-i-k-e-l: Done

src/components/actionSheet/index.js
M-i-k-e-l: Done

src/components/actionSheet/index.js
M-i-k-e-l: Done

src/components/actionSheet/index.js
M-i-k-e-l: Done

src/components/actionSheet/index.js
```diff
@@ -134,12 +155,15 @@ export default class ActionSheet extends BaseComponent {
...
   renderActions() {
-    const {title, options, cancelButtonIndex} = this.props;
+    const {title, options, cancelButtonIndex, renderAction, optionsStyle} = this.props;
     const optionsToRender = _.filter(options, (option, index) => index !== cancelButtonIndex);
     
     return (
-      <View paddingB-8 paddingT-8={_.isEmpty(title)}>
-        {_.map(optionsToRender, this.renderAction)}
+      <View style={[_.isEmpty(title) ? styles.listNoTitle : styles.listWithTitle, optionsStyle]}>
+        {renderAction && _.isFunction(renderAction) ?
```
Inbal-Tish: _.isFunction will return undefined if renderAction is undefined, no need to check renderAction before that

src/components/actionSheet/index.js
```diff
@@ -158,17 +182,19 @@ export default class ActionSheet extends BaseComponent {
...
 
-  renderSheet() {    
+  renderSheet() {
+    const {renderTitle} = this.props;
+    const {containerStyle} = this.getThemeProps();
     return (
-      <View bg-white>
-        {this.renderTitle()}
+      <View style={[styles.sheet, containerStyle]} >
+        {renderTitle && _.isFunction(renderTitle) ? renderTitle() : this.renderTitle()}
```
Inbal-Tish: same here

src/components/actionSheet/index.js
```diff
@@ -188,3 +214,21 @@ export default class ActionSheet extends BaseComponent {
     );
   }
 }
+
+const verticalPadding = 8;
```
Inbal-Tish: Please move const up and use uppercase

M-i-k-e-l: Done

PR #430, opened 2019-04-18
Feat/avatar ribbon refactor

ArnonZ:
- Refactored Avatar's ribbon to allow rendering a custom ribbon

### DIFF
--- a/src/components/avatar/index.js
+++ b/src/components/avatar/index.js
@@ -124,7 +124,7 @@ export default class Avatar extends BaseComponent {
     /**
      * custom ribbon
      */
-    customRibbon: PropTypes.oneOfType([PropTypes.object, PropTypes.func, PropTypes.element]),
+    customRibbon: PropTypes.element,
     /**
      * Determine if to show online badge
      */
--- a/src/components/avatar/index.js
+++ b/src/components/avatar/index.js
@@ -122,9 +122,9 @@ export default class Avatar extends BaseComponent {
      */
     ribbonLabelStyle: Text.propTypes.style,
     /**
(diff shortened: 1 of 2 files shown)


src/components/avatar/index.js
```diff
@@ -122,6 +121,10 @@ export default class Avatar extends BaseComponent {
      * ribbon label custom style
      */
     ribbonLabelStyle: Text.propTypes.style,
+    /**
+     * custom ribbon
+     */
+    customRibbon: PropTypes.oneOfType([PropTypes.object, PropTypes.func, PropTypes.element]),
```
ethanshar: you need to decide on the type of `customRibbon`. 
first of all, `object` is irrelevant, you can render a plain object. 
second, there's a difference between `element` and `func`. 
if you're going with element then you just place it in your render `{customRibbon}`
if you're going with a function you actual need to invoke it - `{customRibbon()}`

Usually going with `func` type is good if you want to pass things to the function (for example the `ribbonLabel`)

src/components/avatar/index.js
```diff
@@ -217,9 +219,11 @@ export default class Avatar extends BaseComponent {
   }
 
   renderRibbon() {
-    const {ribbonLabel, ribbonStyle, ribbonLabelStyle} = this.props;
+    const {ribbonLabel, ribbonStyle, ribbonLabelStyle, customRibbon} = this.props;
     if (ribbonLabel) {
-      return (
+      return customRibbon ? (
+        <View style={this.styles.customRibbon}>{customRibbon}</View>
```
ethanshar: according to what u decide on the proptype of `customRibbon` you need to update this code. 
currently if someone passed a function, this code will not work

PR #431, opened 2019-04-22
Bug/getColorTint support for single Ui-Lib colors

ArnonZ:
- Fixed exception where ui-lib Colors without tint variants returned original color instead of tinted one.
- Added test coverage

### DIFF
--- a/src/style/colors.js
+++ b/src/style/colors.js
@@ -74,8 +74,6 @@ class Colors {
   }
 
   getColorTint(color, tintKey) {
-    const dynamicHexColor = this.getTintedColorForDynamicHex(color, tintKey);
-
     if (_.isUndefined(tintKey) || isNaN(tintKey) || _.isUndefined(color)) {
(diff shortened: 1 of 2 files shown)


src/style/colors.js
```diff
@@ -74,6 +74,8 @@ class Colors {
   }
 
   getColorTint(color, tintKey) {
+    const dynamicHexColor = this.getTintedColorForDynamicHex(color, tintKey);
```
ethanshar: I would say this one `getTintedColorForDynamicHex` is a little expensive. 
so first, I wouldn't call it first thing. it's possible the validations will exit the method even before we need to use it. 

also, in this case it's best to call it twice in each condition just to avoid calling it for nothing. 
basically what it does is creating a palette based on the given color and then return the right tint - and generating this color might be heavy (imagine a list of 100 items with a thumbnail that use this function internally)

ArnonZ: Yes, I agree, in this case performance outweighs cleanliness.
Thanks. Fixed.

PR #432, opened 2019-04-22
Feat/drawer v2

ethanshar:
- Use Swipeable (gesture-handler) for the version of Drawer
- Passing width to items is not mandatory anymore. 
- Item object has new props - `keepOpen` & `style`

### DIFF
--- a/demo/src/screens/interactableComponentScreens/DrawerScreen.js
+++ b/demo/src/screens/interactableComponentScreens/DrawerScreen.js
@@ -7,7 +7,7 @@ import conversations from '../../data/conversations';
 
 const collectionsIcon = require('../../assets/icons/collections.png');
 const starIcon = require('../../assets/icons/star.png');
-const sharIcon = require('../../assets/icons/share.png');
+const shareIcon = require('../../assets/icons/share.png');
 const videoIcon = require('../../assets/icons/video.png');
 
 class DrawerScreen extends Component {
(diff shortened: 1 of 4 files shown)


src/interactableComponents/drawer/index.js
```diff
@@ -462,6 +467,12 @@ export default class Drawer extends BaseComponent {
     );
   }
   render() {
+    const {migrate} = this.props;
+
+    if (migrate) {
```
Inbal-Tish: Why use 'migrate' prop and not replace the component?

demo/src/screens/interactableComponentScreens/DrawerScreen.js
```diff
@@ -130,54 +135,59 @@ export default class DrawerScreen extends Component {
...
           itemsTextStyle={{fontSize: 18}}
           onPress={this.onPress}
-          ref={r => this.secondDrawer = r}
+          ref={r => (this.secondDrawer = r)}
         >
           {this.renderContent('1', conversations[1])}
         </Drawer>
 
-        <Drawer
+        {/* <Drawer
```
Inbal-Tish: This example is working well, why hide it?

src/interactableComponents/drawer/index.js
ethanshar: to allow us control the migration. 
Also this way we can migrate only in private uilib. 
for the public it will be better to migrate when we upgrade major version

PR #433, opened 2019-04-22
Change ActionSheet's showCancelButton defaultProp to undefined

M-i-k-e-l:


src/components/actionSheet/index.js
```diff
@@ -93,7 +93,7 @@ export default class ActionSheet extends BaseComponent {
   static defaultProps = {
     title: undefined,
     message: undefined,
-    showCancelButton: false,
+    showCancelButton: undefined,
```
Inbal-Tish: Just remove this line. No need to set it to undefined

PR #437, opened 2019-05-05
Feat/add cusomization for wheel picker

M-i-k-e-l:


src/components/picker/index.js
```diff
@@ -50,7 +50,7 @@ class Picker extends BaseComponent {
      */
     enableModalBlur: PropTypes.bool,
     /**
-     * Render custom picker
+     * Render custom picker - input will be value (see above)
```
ethanshar: better show a small snippet 
`Render custom picker (e.g. (selectedValue) => {})`

src/components/picker/PickerDialog.android.js
```diff
@@ -17,6 +17,48 @@ class PickerDialog extends BaseComponent {
     onDone: PropTypes.func,
     onCancel: PropTypes.func,
     children: PropTypes.array,
+    /**
+     * Pass props for the WheelPicker (Android only)
+     */
+    wheelPickerProps: PropTypes.shape({
```
ethanshar: If those are the WheelPicker props why are you re-defining them here?
Why not just use the WheelPicker.propTypes as we usually do?

src/components/picker/index.js
M-i-k-e-l: Added example

PR #442, opened 2019-05-19
Fix Carousel on Android not responding to onPress (goToPage)

M-i-k-e-l:


src/components/carousel/index.js
```diff
@@ -69,6 +69,15 @@ export default class Carousel extends BaseComponent {
     
     if (this.carousel) {
       this.carousel.scrollTo({x, animated});
+      if (Constants.isAndroid) {
+        // this is done to handle onMomentumScrollEnd not being called in Android:
+        // https://github.com/facebook/react-native/issues/11693
+        // https://github.com/facebook/react-native/issues/19246
+        const {currentStandingPage, currentPage} = this.state;
```
ethanshar: you've copied the code from `onMomentumScrollEnd`, but you can just call the function instead.

M-i-k-e-l: It's not exactly the same code (setState is only in onMomentumScrollEnd).
Do you want me to create something like:
checkForPageChange = () => {
  const {currentStandingPage, currentPage} = this.state;
    if (currentStandingPage !== currentPage) {
      _.invoke(this.props, 'onChangePage', currentPage, currentStandingPage);
    }
}

ethanshar: you're right, it's not exactly the same. 
but why actually in the new code we don't do this also?
`this.setState({currentStandingPage: currentPage});`

I mean the new code covers Android for not triggering `onMomentumScrollEnd` event, right?
I would expect us to do the same we do in `onMomentumScrollEnd` code.

M-i-k-e-l: Because the new code (inside `updateOffset`) is called from `onScroll`; it might work, but then it (`setState`) will be called twice, once from the `onScroll` and later from `onMomentumScrollEnd`

ethanshar: Maybe we do want to include this
`this.setState({currentStandingPage: currentPage});`
since we never update `currentStandingPage` elsewhere..

M-i-k-e-l: Done

PR #444, opened 2019-05-20
Add testID to ActionSheet

M-i-k-e-l:


src/components/dialog/index.js
```diff
@@ -138,12 +138,12 @@ class Dialog extends BaseComponent {
...
 
   renderContent() {
-    const {bottom} = this.getThemeProps();
+    const {bottom, testID} = this.getThemeProps();
     const bottomInsets = Constants.getSafeAreaInsets().paddingBottom;
 
     return (
       <TouchableWithoutFeedback>
-        <SafeAreaView style={{flexGrow: 1}}>
+        <SafeAreaView style={{flexGrow: 1}} testID={testID}>
```
ethanshar: I would pass it to top possible View you can, I think in this case it's the one that is rendered in `renderAnimationContainer`

M-i-k-e-l: Done

PR #447, opened 2019-06-03
Drawer add support for minWidth

M-i-k-e-l:


src/interactableComponents/drawer/newDrawer.js
```diff
@@ -53,6 +53,10 @@ export default class NewDrawer extends BaseComponent {
      * Whether to give the items equal width (the max width)
      */
     equalWidths: PropTypes.bool,
+    /**
+     * Set a different minimum width
+     */
+    minWidth: PropTypes.number,
```
ethanshar: I would change to `itemsMinWidth` to keep it similar to other props (`itemsTintColor`, `itemsIconSize`, etc..)

src/interactableComponents/drawer/newDrawer.js
```diff
@@ -118,8 +122,8 @@ export default class NewDrawer extends BaseComponent {
 
   // eslint-disable-next-line react/prop-types
   renderAction = ({item, index, progress, itemsCount}) => {
-    const {itemsTintColor, itemsIconSize, itemsTextStyle} = this.getThemeProps();
-
+    const {itemsTintColor, itemsIconSize, itemsTextStyle, minWidth} = this.getThemeProps();
+    const width = _.max([item.width, minWidth]);
```
ethanshar: no need for that, react native's style support the `minWidth` prop. 

Also, If the user doesn't pass a `width` to an item but does pass the `minWidth` prop, the item width will be fixed and wouldn't stretch according to its content.

src/interactableComponents/drawer/newDrawer.js
M-i-k-e-l: Do we want all items to have the same minimum width? - Or maybe add item.minWidth?

M-i-k-e-l: Done

src/interactableComponents/drawer/newDrawer.js
M-i-k-e-l: Done

PR #455, opened 2019-06-12
Fix ActionSheet not passing useSafeArea

M-i-k-e-l:


src/components/actionSheet/index.js
```diff
@@ -198,12 +198,13 @@ export default class ActionSheet extends BaseComponent {
   }
 
   render() {
-    const {useNativeIOS, visible, onDismiss, useModal, dialogStyle, onModalDismissed, testID} = this.getThemeProps();
+    const {useNativeIOS, visible, onDismiss, useModal, dialogStyle, onModalDismissed, testID, useSafeArea} = this.getThemeProps();
```
ethanshar: I think we should define this as a prop type, so people be aware it exists

M-i-k-e-l: Done

PR #456, opened 2019-06-13
PanGestureView add specific swipeable layout

M-i-k-e-l:


src/components/panGestureView/index.js
```diff
@@ -53,9 +58,37 @@ export default class PanGestureView extends BaseComponent {
...
+      }
+    }
+  };
+
   handleMoveShouldSetPanResponder = (e, gestureState) => {
     // return true if user is swiping, return false if it's a single click
-    const {dy} = gestureState;
+    const {dy, moveX, moveY} = gestureState;
+    if (this.swipeBox) {
+      const swipeBox = this.swipeBox;
```
Inbal-Tish: Is this const necessary?

src/components/panGestureView/index.js
```diff
@@ -28,7 +28,12 @@ export default class PanGestureView extends BaseComponent {
     /**
      * The direction of the allowed pan (default is down)
      */
-    direction: PropTypes.oneOf(Object.values(DIRECTIONS))
+    direction: PropTypes.oneOf(Object.values(DIRECTIONS)),
+    /**
+     * Specify a layout reference, if specified only touches in this layout will trigger a swipe (iOS only)
+     * You must use React.createRef() and send it (not the current)
+     */
+    swipeLayoutRef: PropTypes.object
```
Inbal-Tish: Here as well, use 'pan' instead of 'swipe'

src/components/dialog/index.js
```diff
@@ -41,6 +41,11 @@ class Dialog extends BaseComponent {
      * The direction of the swipe to dismiss the dialog (default is 'down')
      */
     dismissSwipeDirection: PropTypes.oneOf(Object.values(SWIPE_DIRECTIONS)), // DEFRECATED
+    /**
+     * Specify a layout reference, if specified only touches in this layout will trigger a swipe
+     * You must use React.createRef() and send it (not the current)
+     */
+    swipeLayoutRef: PropTypes.object,
```
Inbal-Tish: 'swipe' is not the right term, it should be 'pan'.

PR #458, opened 2019-06-23
Create DialogView

M-i-k-e-l:


src/components/dialogView/index.js
```diff
@@ -0,0 +1,221 @@
...
+      <View style={[styles.header, headerStyle]}>
+        {showKnob && this.renderKnob()}
+        {hasTitle && this.renderTitle()}
+        {renderBelowTitle && renderBelowTitle()}
+        {showDivider && this.renderDivider()}
+      </View>
+    );
+  };
+
+  hasHeader = () => {
```
ethanshar: let's try to simplify the API for the `DialogView` since it's a public component and move the complex UI logic to the private lib. 
So, in my opinion, we should have basic props for 
- knob
- title, titleStyle
- subtitle, subtitleStyle
- renderHeader (custom header)

and we can avoid too specific props like: `titleOuterContainerStyle`, `titleInnerContainerStyle`, `renderRightOfTitle`, `renderBelowTitle`

M-i-k-e-l: Done

PR #462, opened 2019-07-08
TextField - Right Icon

Inbal-Tish:
Move right icon to text input render to allow on both expandable and non-expandable inputs


src/components/inputs/TextField.js
```diff
@@ -509,15 +510,17 @@ export default class TextField extends BaseInput {
   }
 
   renderRightButton() {
-    const {rightButtonProps} = this.getThemeProps();
-    const tintColor = _.get(rightButtonProps, 'style.tintColor', Colors.blue30);
-
     if (this.shouldDisplayRightButton()) {
+      const {rightButtonProps} = this.getThemeProps();
+      const tintColor = _.get(rightButtonProps, 'style.tintColor', Colors.blue30);
```
ethanshar: this is kinda weird. 
first, accessing to style as an object is not so great, cause it can be a number or an array
second, tintColor is an image style prop, not a button. 
maybe to add an explicit prop to `rightButtonProps` for `tintColor` or `iconColor`

Inbal-Tish: Done. Thanks!

PR #463, opened 2019-07-09
add RTL support for TextField and Text components

ethanshar:


src/components/inputs/TextField.js
```diff
@@ -616,7 +617,7 @@ function createStyles({placeholderTextColor, centered, multiline}) {
     },
     input: {
       flexGrow: 1,
-      textAlign: centered ? 'center' : undefined,
+      textAlign: centered ? 'center' : inputTextAlign,
```
Inbal-Tish: You don't need to do that. If you just switch the undefined with 'left', it will turn to 'right' automatically in RTL layout.

src/components/inputs/TextField.js
```diff
@@ -643,7 +644,8 @@ function createStyles({placeholderTextColor, centered, multiline}) {
       backgroundColor: 'transparent'
     },
     placeholder: {
-      color: placeholderTextColor
+      color: placeholderTextColor,
+      writingDirection: Constants.isRTL ? 'rtl' : 'auto'
```
Inbal-Tish: Here 'auto' should be enough, no? It should recognize an RTL layout and flip it automatically.

src/components/text/index.js
```diff
@@ -74,6 +75,7 @@ function createStyles() {
   return StyleSheet.create({
     container: {
       backgroundColor: 'transparent',
+      writingDirection: Constants.isRTL ? 'rtl' : 'auto'
```
Inbal-Tish: Here too. Worth checking just in case

src/components/inputs/TextField.js
Inbal-Tish: Also worth checking in Android, as I understand this is an iOS prop

ethanshar: Android works fine, iOS needs it. 
when using `forceRTL` that's what actually makes it work.

src/components/text/index.js
ethanshar: did check, Android is fine, ios works only after this change.

src/components/inputs/TextField.js
ethanshar: it actually better when explicitly setting `left` or `right` 
it also moves the cursor of the input to the right (on RTL) 
unfortunately, when just setting `left`, everything works but the cursor

src/components/text/index.js
ethanshar: ok, I doubled checked. on Android it wasn't fine, is it possible there's a difference between the public and private?

src/components/inputs/TextField.js
Inbal-Tish: Got it. Thanks

PR #465, opened 2019-07-10
memoize generateColorPalette and use palette for getColorTint

ethanshar:


src/style/colors.js
```diff
@@ -99,13 +99,15 @@ class Colors {
...
-    let tintLevel = Math.floor(Number(tintKey) / 10 + 1);
-    tintLevel = Math.max(2, tintLevel);
-    tintLevel = Math.min(9, tintLevel);
-    return generateColorTint(color, tintLevel * 10);
+    let tintLevel = Math.floor(Number(tintKey) / 10);
+    tintLevel = Math.max(1, tintLevel);
+    tintLevel = Math.min(8, tintLevel);
+
+    const colorsPalette = this.generateColorPalette(color);
+    return colorsPalette[tintLevel - 1];
```
Inbal-Tish: This returns the same color, not the tinted color.

src/style/colors.js
```diff
@@ -130,7 +132,7 @@ class Colors {
 
     const sliced = tints.slice(0, 8);
     return sliced;
-  }
+  })
```
Inbal-Tish: semicolon missing

src/style/colors.js
ethanshar: are you sure? 
there's a test that shows it works (: 
https://github.com/wix/react-native-ui-lib/blob/master/src/style/__tests__/colors.spec.js#L81

src/style/colors.js
ethanshar: done

src/style/colors.js
Inbal-Tish: I've called it and printed it. I keep getting the same color I sent

Inbal-Tish: Ok. My bad. Seems to work

PR #468, opened 2019-07-14
newDrawer - support only rightItems in drawer

roiberlin:


src/interactableComponents/drawer/newDrawer.js
```diff
@@ -128,7 +128,8 @@ class NewDrawer extends PureBaseComponent {
   // TODO: enable support for rendering more than one left item
   renderLeftActions = (progress, dragX) => {
     const {leftItem} = this.getThemeProps();
-    return this.renderActions([leftItem], progress, dragX);
+    const leftItems = leftItem ? [leftItem] : [];
```
Inbal-Tish: I would return 'undefined' instead of an empty array since it has no meaning, while 'undefined' is more explicit and will not pass the condition in renderActions() at all.

roiberlin: Done

PR #469, opened 2019-07-14
newDrawer - add useNativeAnimations

roiberlin:


src/interactableComponents/drawer/Swipeable.js
```diff
@@ -59,9 +59,9 @@ export default class Swipeable extends Component<PropType, StateType> {
   static defaultProps = {
     friction: 1,
     overshootFriction: 1,
-    // useNativeAnimations: true // issue in iPhone 5
+    useNativeAnimations: true
```
Inbal-Tish: Did you test it on iPhone5?

roiberlin: for react-native-ui-lib we will enable it for all devices
for our private library i will disable it for iPhone5

Inbal-Tish: But the bug exists in the public repo as well. I can't leave it like this

ethanshar: @roiberlin keep the default value false here and just expose the prop thru Drawer. 
in the private repo we'll turn it on only if it's not iPhone5

btw, @Inbal-Tish in my opinion, the performance here is preferable on the iPhone5 issue. 
your call

Inbal-Tish: I didn't see a hit on performance when I turned it off. We need to find a way to actually measure it

roiberlin: Done

src/interactableComponents/drawer/newDrawer.js
```diff
@@ -234,6 +238,7 @@ class NewDrawer extends PureBaseComponent {
         leftActionsContainerStyle={this.leftActionsContainerStyle}
         onSwipeableWillOpen={this.onSwipeableWillOpen}
         onSwipeableWillClose={this.onSwipeableWillClose}
+        useNativeAnimations={useNativeAnimations}
```
Inbal-Tish: you don't need to pass it, it will be passed in 'others' if you don't extract it from the props object.

roiberlin: Done

PR #470, opened 2019-07-15
Performance improvements

emilisb:
- Refactored modifiers (saves up to 1ms per each base component)
- Do not render selection view on Card component if `selected` prop is undefined

### DIFF
--- a/src/commons/modifiers.js
+++ b/src/commons/modifiers.js
@@ -153,13 +153,11 @@ export function extractFlexStyle(props) {
     flexG: 'flexGrow',
     flexS: 'flexShrink',
   };
-  const flexPropKey = _.chain(props)
-    .keys(props)
-    .filter(key => FLEX_KEY_PATTERN.test(key))
-    .last()
-    .value();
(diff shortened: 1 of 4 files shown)


src/commons/modifiers.js
```diff
@@ -18,14 +18,15 @@ export function extractColorValue(props) {
...
 export function extractBackgroundColorValue(props) {
   let backgroundColor;
-  _.forEach(Colors, (value, key) => {
-    if (props[`background-${key}`] === true || props[`bg-${key}`] === true) {
-      backgroundColor = value;
-    }
-  });
+
+  const keys = Object.keys(props);
+  const bgProp = keys.find((prop) => Colors.getBackgroundKeysPattern().test(prop));
```
ethanshar: I checked in master and I see it did take the last backgroundColor modifier that was passed (even though we don't have a test specifically for it).
I think it's still better to keep parity with the previous behavior, just in case...

emilisb: Done

PR #472, opened 2019-07-15
Optimize typography and spacing modifier key patterns

emilisb:
Very small changes, but in screens with a lot of text render time decreased by 50ms (with debugger connected).

### DIFF
--- a/src/style/spacings.js
+++ b/src/style/spacings.js
@@ -13,6 +13,8 @@ class Spacings {
   s9 = Constants.isIOS ? 27 : 36;
   s10 = Constants.isIOS ? 30 : 40;
 
+  keysPattern = this.generateKeysPattern();
+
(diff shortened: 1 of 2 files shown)


src/style/spacings.js
```diff
@@ -20,12 +20,7 @@ class Spacings {
   }
 
   getKeysPattern() {
-    return new RegExp(
-      _.chain(this)
-        .keys()
-        .join('|')
-        .value(),
-    );
+    return /^(s[0-9]+)/;
```
ethanshar: for `Spacings` we also have "semantic" names (we load them in the private code) like `page`, `innerCard`, etc.. 
so we still need to base our regex on the keys name. 
I guess you could do the same thing u did for the Typography

emilisb: Fixed, thanks

PR #475, opened 2019-07-18
Feat/carousel landscape

Inbal-Tish:


src/components/carousel/index.js
```diff
@@ -15,7 +16,7 @@ import * as presenter from './CarouselPresenter';
  * @extendsLink: https://facebook.github.io/react-native/docs/scrollview 
  * @notes: This is screed width Component
  */
-export default class Carousel extends BaseComponent {
+export default class NewCarousel extends BaseComponent {
```
ethanshar: better call this `Carousel` and the old `CarouselDeprecated`

Inbal-Tish: Done.

PR #476, opened 2019-07-18
Feat/orientation support

Inbal-Tish:


src/helpers/Constants.js
```diff
@@ -1,30 +1,79 @@
...
+function updateConstants() {
+  const {height, width} = Dimensions.get(dimensionsScope.WINDOW);
+  screenWidth = width;
+  screenHeight = height;
+  isSmallScreen = screenWidth <= 340;
+  isShortScreen = screenHeight <= 600;
+  setStatusBarHeight();
+}
+
+/* Orientation */
```
ethanshar: you just pass the callback, instead of wrapping it since you don't pass any params
`Dimensions.addEventListener('change', updateConstants);`

src/helpers/Constants.js
```diff
@@ -1,30 +1,79 @@
...
+Dimensions.addEventListener('change', () => {
+  updateConstants();
+});
+
+export function getOrientation() {
```
ethanshar: What about keep an orientation var (like we do for screen width and height) that get update when orientation changes? 
Theoretically, the orientation changes only when rotating the device, no need to calculate it each time, right?

Inbal-Tish: Sure

src/helpers/Constants.js
Inbal-Tish: I don't understand...

Inbal-Tish: ah. Got it. Ok

PR #488, opened 2019-07-28
PickerModal - fix orientation change bug

M-i-k-e-l:
(when there's a component above the FlatList)


ethanshar:
@M-i-k-e-l 
Can you verify if this issue still occurs.
I tried to reproduce and did see an issue. but maybe I don't remember the issue so well.
can please check yourself.

M-i-k-e-l:
Looks like RN61 fixed this.

PR #491, opened 2019-07-30
Add saturation adjustment for generated color palette

Inbal-Tish:


src/style/colors.js
```diff
@@ -123,11 +123,43 @@ class Colors {
...
 
-    const sliced = tints.slice(0, 8);
-    return sliced;
+    let sliced = tints.slice(0, 8);
+    const adjusted = adjustSaturation(sliced, color);
+    return adjusted || sliced;
   });
 }
 
+function adjustSaturation(colors, color) {
```
ethanshar: Let's add some test to cover these logic

Inbal-Tish: Ok

src/style/colors.js
```diff
@@ -123,11 +123,43 @@ class Colors {
...
+  let array;
+  const lightnessLevel = 80;
+  const saturationLevel = 60;
+  const hsl = Color(color).hsl();
+  const lightness = Math.round(hsl.color[2]);
+
+  if (lightness > lightnessLevel) {
+    const saturation = Math.round(hsl.color[1]);
+    if (saturation > saturationLevel) {
+      array = [];
```
ethanshar: sorry, should have wrote it before
let's shorten this a little..
you can
`array = _.map(colors, e => e !== color ? addSaturation(e, saturationLevel) : e)`

Inbal-Tish: Cool

PR #494, opened 2019-08-01
Fix/Avatar - Refactor to enable dynamic prop change

ArnonZ:
- Moved all props from StyleSheet to class

### DIFF
--- a/src/components/avatar/index.js
+++ b/src/components/avatar/index.js
@@ -284,7 +284,7 @@ export default class Avatar extends PureBaseComponent {
       return (
         <ImageContainer
           animate={animate}
-          style={[this.getContainerStyle(), {...StyleSheet.absoluteFillObject}, imageStyle]}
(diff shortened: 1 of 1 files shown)


src/components/avatar/index.js
```diff
@@ -246,17 +279,18 @@ export default class Avatar extends PureBaseComponent {
     } = this.props;
     const hasImage = !_.isUndefined(imageSource);
     const ImageContainer = animate ? AnimatedImage : Image;
+
     if (hasImage) {
       return (
         <ImageContainer
           animate={animate}
-          style={[this.styles.image, imageStyle]}
+          style={[this.getContainerStyle(), {...StyleSheet.absoluteFillObject}, imageStyle]}
```
ethanshar: `{...StyleSheet.absoluteFillObject}`
spreading an object inside an empty object is good as just passing the object.

ArnonZ: Fixed.

PR #498, opened 2019-08-07
Feat/image overlay

Inbal-Tish:
Adding overlay option to Image component


src/components/overlay/index.js
```diff
@@ -0,0 +1,94 @@
...
+ * @extendsLink: https://facebook.github.io/react-native/docs/image
+ */
+export default class Overlay extends PureBaseComponent {
+  static displayName = 'Overlay';
+
+  static propTypes = {
+    /** 
+    * The type of overlay to set on top of the image 
+    */
+    type: PropTypes.oneOfType([PropTypes.oneOf(Object.keys(OVERLY_TYPES)), PropTypes.string])
```
ethanshar: This is not correct, it should be 
`type: PropTypes.oneOf(_.values(OVERLY_TYPES))`

src/components/image/index.js
```diff
@@ -39,13 +41,19 @@ class Image extends PureBaseComponent {
     /**
      * Show image as a cover, full width, image (according to aspect ratio, default: 16:8)
      */
-    cover: PropTypes.bool
+    cover: PropTypes.bool,
+    /**
+     * the type of overly to place on top of the image
+     */
+    overlayType: PropTypes.oneOfType([PropTypes.oneOf(Object.keys(Overlay.overlayTypes)), PropTypes.string])
```
ethanshar: you can just use the definition you have in Overlay component to avoid duplicity 
`overlayType: Overlay.propTypes.type`

src/components/overlay/index.js
```diff
@@ -0,0 +1,94 @@
...
+        source={image}
+      />
+    );
+  }
+
+  render() {
+    const {style, type} = this.props;
+    const width = _.get(style, 'width');
+    
+    if (type === OVERLY_TYPES.DEFAULT) {
```
ethanshar: I would assume the default behavior of overlay is the top overlay.
this is probably what most people want/use.
WDYT?

src/components/image/index.js
```diff
@@ -88,6 +96,25 @@ class Image extends PureBaseComponent {
       />
     );
   }
+
+  onLayout = (event) => {
```
ethanshar: Why do we need that?
The overlay will be stretched anyway because you're using absoluteFillObject in the overlay component.

src/components/image/index.js
```diff
@@ -88,6 +96,25 @@ class Image extends PureBaseComponent {
...
+  }
+
+  render() {
+    const {style, overlayType} = this.getThemeProps();
+  
+    if (overlayType) {
+      return (
+        <View onLayout={this.onLayout}>
+          {this.renderImage()}
+          <Overlay style={[style, {width: this.state.width}]} type={overlayType}/>
```
ethanshar: Passing the same style you pass to the image might be risky here. 
I wouldn't pass anything else to the overlay and keep is simple at this point.
but if we do want to support custom style for the overlay, we need a separate prop.

src/components/overlay/index.js
Inbal-Tish: I set the default for the top+bottom. Please note that there is no real default if you don't pass an overlay type you get no overlay

src/components/image/index.js
Inbal-Tish: Since it's not stretching as you expect

src/components/image/index.js
Inbal-Tish: I'm passing the style to position the overlay properly on the image. Not for actual styling purposes.

src/components/overlay/index.js
ethanshar: true, but passing DEFAULT will give u both, which is kinda weird. 
I mean the name is weird, not the behavior necessarily . 
maybe worth calling it `BOTH` ? as both top and bottom?

src/components/image/index.js
ethanshar: It looks the same with or without the width

src/components/overlay/index.js
```diff
@@ -0,0 +1,94 @@
...
+        return styles.bottom;
+      case OVERLY_TYPES.SOLID: 
+        return styles.solid
+      default:
+        break;
+    }
+  }
+
+  renderImage(typeStyle) {
+    const {type, style} = this.props;
```
ethanshar: if someone passed `tintColor` (for example) inside the style for the actual image, wouldn't it be passed to the gradient of the overlay too?

src/components/overlay/index.js
```diff
@@ -0,0 +1,94 @@
...
+        style={[style, styles.container, typeStyle, {width: undefined}]}
+        resizeMode={'stretch'}
+        source={image}
+      />
+    );
+  }
+
+  render() {
+    const {style, type} = this.props;
+    const width = _.get(style, 'width');
```
ethanshar: You can't extract `width` or other props from style, cause you can't tell if style is a number/object/array. this potentially can return undefined.

ethanshar: checking again, it actually does return undefined, cause you pass an array from the `Image` component. 

and if it works without it, it means the `onLayout` in the `Image` is redundant.

src/components/overlay/index.js
Inbal-Tish: I don't like the name BOTH, that's why 'default' seemed better. I don't think it's weird, but if you have a better name I don't mind changing it.

src/components/overlay/index.js
Inbal-Tish: So what are you suggesting? I need the layout style from the image

src/components/overlay/index.js
Inbal-Tish: Not really. The width is passed in the style object to the renderImage() as well

PR #499, opened 2019-08-08
Add isTablet to Constants

M-i-k-e-l:


src/helpers/Constants.js
```diff
@@ -40,6 +40,8 @@ export let screenWidth = width;
 export let screenHeight = height;
 export let isSmallScreen = screenWidth <= 340;
 export let isShortScreen = screenHeight <= 600;
+export const ratio = screenWidth < screenHeight ? screenHeight / screenWidth : screenWidth / screenHeight;
+export const isTablet = Platform.isPad || (ratio < 1.6 && Math.max(screenWidth, screenHeight) >= 900);
```
Inbal-Tish: Where did you get the 1.6 ratio limit?

src/helpers/Constants.js
```diff
@@ -40,6 +40,8 @@ export let screenWidth = width;
 export let screenHeight = height;
 export let isSmallScreen = screenWidth <= 340;
 export let isShortScreen = screenHeight <= 600;
+export const ratio = screenWidth < screenHeight ? screenHeight / screenWidth : screenWidth / screenHeight;
```
Inbal-Tish: Why are you exporting ratio?

src/helpers/Constants.js
M-i-k-e-l: It was a guesstimation in a Stack-Overflow answer.

src/helpers/Constants.js
M-i-k-e-l: I don't need it exported, but thought it might be useful.
Do you want me to remove that? - Or maybe change the name?

Inbal-Tish: Let's remove the export till we need it and change the name to be more descriptive.

src/helpers/Constants.js
Inbal-Tish: Did you test it?

M-i-k-e-l: Yes, it worked on the devices I've tested; and I've added the screen size for good measure

Inbal-Tish: Cool

src/helpers/Constants.js
M-i-k-e-l: Done

PR #503, opened 2019-08-12
Feat/image overlay

Inbal-Tish:


src/components/overlay/index.js
```diff
@@ -0,0 +1,111 @@
...
+      .pickBy((value, key) => _.includes(key, attName))
+      .value();
+    }
+    return attributes;
+  }
+
+  renderImage(typeStyle) {
+    const {type, style} = this.props;
+    const image = type !== OVERLY_TYPES.SOLID ? gradientImage : undefined;
+    const marginStyle = type !== OVERLY_TYPES.VERTICAL ? this.getAttributeStyles(style, 'margin') : undefined;
```
ethanshar: Just so I understand, for what use case is it?

demo/src/screens/componentScreens/OverlaysScreen.js
```diff
@@ -0,0 +1,93 @@
+import React, {Component} from 'react';
+import {StyleSheet, ScrollView} from 'react-native';
+import {View, Text, Image} from 'react-native-ui-lib'; // eslint-disable-line
+
+
+const image = require('../../assets/images/card-example.jpg');
```
ethanshar: I think it will better to show the overlay on big assets like the placeholder illustrations from the private repo (one is enough)

Inbal-Tish: Not sure I understand what you mean...

src/components/overlay/index.js
Inbal-Tish: Since the overlay image is a sibling to the image it should get the margin style sent to the image in over to layout correctly. In case of the VERTICAL type the margin should go to the container of the two overlay images and not to the overlay image itself.

PR #506, opened 2019-08-19
Feat/featur highlight extras

Inbal-Tish:
Adding borderRadius, headline and pageControlProps props.

### DIFF
--- a/demo/src/screens/componentScreens/FeatureHighlightScreen.js
+++ b/demo/src/screens/componentScreens/FeatureHighlightScreen.js
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import React, {Component} from 'react';
-import {Colors, View, Text, Button, FeatureHighlight} from 'react-native-ui-lib'; // eslint-disable-line
+import {Colors, Typography, View, Text, Button, FeatureHighlight} from 'react-native-ui-lib'; // eslint-disable-line
 
 
 const titles = [
@@ -9,7 +9,7 @@ const titles = [
   'Title number three',
   'Title number four',
   'Title number five',
-  'Only a title and headline'
+  'Welcome to Uilib demo!'
 ];
 const messages = [
   'Important notifications appear right on your clubs and groups. Tap them to get more information about the most' +
@@ -20,10 +20,7 @@ const messages = [
   ' warning about it',
   'Very short message',
   'Short message with information about the below highlighted feature',
-  undefined
-];
-const headlines = [
-  'Welcome to uilib demo!'
+  'Here is where you can view demos of all Uilib components'
 ];
 
 class FeatureHighlightScreen extends Component {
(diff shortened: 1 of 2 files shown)


src/components/featureHighlight/index.js
```diff
@@ -63,6 +64,10 @@ class FeatureHighlight extends BaseComponent {
      * Message to be displayed
      */
     message: PropTypes.string,
+    /**
+     * The headline to be displayed above the title
+     */
+    headline: PropTypes.string,
```
ethanshar: I think this is very specific to what Growth team wants, why not allowing styling the title and the message so thy can achieve the typography they want?

src/components/featureHighlight/index.js
```diff
@@ -268,9 +296,16 @@ class FeatureHighlight extends BaseComponent {
         strokeWidth={borderWidth || defaultStrokeWidth}
         minimumRectSize={minimumRectSize}
         innerPadding={innerPadding}
+        borderRadius={borderRadius}
       >
         <TouchableWithoutFeedback style={styles.touchableOverlay} onPress={onBackgroundPress}>
+          {pageControlProps ?
+          <View flex bottom>
```
ethanshar: maybe we can allow passing pageControl position to allow changing the position somehow? 
or pass `pageControlerProps.containerStyle` together with yours

demo/src/screens/componentScreens/FeatureHighlightScreen.js
```diff
@@ -61,29 +60,53 @@ class FeatureHighlightScreen extends Component {
...
         visible={showFTE}
         title={titles[currentTargetIndex]}
         message={messages[currentTargetIndex]}
+        headline={currentTargetIndex === lastPage ? headlines[0] : undefined}
         confirmButtonProps={{label: 'Got It', onPress: this.moveNext}}
         onBackgroundPress={this.closeHighlight}
         getTarget={() => this.targets[currentTargetIndex]}
         // highlightFrame={{x: 30, y: 70, width: 150, height: 30}}
         // highlightFrame={{x: 160, y: 336, width: 150, height: 56}}
+        // borderRadius={4}
```
ethanshar: you can set for the example the first item to be round and the reset be square so users can see it's customizable

src/components/featureHighlight/index.js
Inbal-Tish: Your call. You want me to change it?

src/components/featureHighlight/index.js
Inbal-Tish: I don't see why it's necessary since it will look pretty bad somewhere else

demo/src/screens/componentScreens/FeatureHighlightScreen.js
Inbal-Tish: I know. I didn't want to change the example. Done.

src/components/featureHighlight/index.js
Inbal-Tish: The point was to control the text design though

ethanshar: yea, but it will never good enough, especially if it's an open source component. 
someone will always want to customize it

src/components/featureHighlight/index.js
ethanshar: If someone will want to feature highlight  elements that are at the bottom of the screen? 
anyway, your call, just an option, if someone will need they will probably request it

src/components/featureHighlight/index.js
Inbal-Tish: Ok. I'll change it

src/components/featureHighlight/index.js
Inbal-Tish: Yes, if someone will ask it we can change it.

src/components/featureHighlight/index.js
Inbal-Tish: So you think they will never use the three of them together?

ethanshar: if the idea was to add a third type of content so people can have 3 different contents then I guess this prop make sense
if we're just adding another content prop with the exact same styling of the `title` prop just so people can render it under the `message` then it's weird, if that's the case it makes more sense to have custom styling over the those content 

Since the current requirement was to still have two contents just in a different order I assume letting them customize the style feels more right

PR #510, opened 2019-08-25
support passing screen width for the hint container

ethanshar:
this is for the tablet support

### DIFF
--- a/src/components/hint/index.js
+++ b/src/components/hint/index.js
@@ -107,6 +107,10 @@ class Hint extends BaseComponent {
      */
     onBackgroundPress: PropTypes.func,
     /**
+     * The Screen width for the hint container
+     */
+    screenWidth: PropTypes.number,
+    /**
      * The hint's test identifier
      */
     testID: PropTypes.string,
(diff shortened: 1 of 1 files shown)


src/components/hint/index.js
```diff
@@ -106,6 +106,10 @@ class Hint extends BaseComponent {
      * Callback for the background press
      */
     onBackgroundPress: PropTypes.func,
+    /**
+     * The Screen width for the hint container
+     */
+    screenWidth: PropTypes.number,
```
Inbal-Tish: I don't understand why you need screenWidth prop? You don't want to listen to orientation change in the components?

ethanshar: This is for the tablet support with the custom margins
because we define a custom screen width (in private, the margin make the screen shorter), I need a way to declare this width cause the hint container is based on that.

Inbal-Tish: We read the screen width from the Constants. Why not make the treatment there? That changes the way other components layout for landscape.

ethanshar: because in the private code the screen width in tablets is different because of the custom margins we have. I need a way to pass this value from the private repo

Inbal-Tish: Why not get it in onLayout? It should be called in orientation change as well.

ethanshar: renamed screenWidth to containerWidth

PR #526, opened 2019-09-04
Feat/Accessibility phase 1

ArnonZ:
- Added accessibility for Avatar and Badge

### DIFF
--- a/src/components/avatar/index.js
+++ b/src/components/avatar/index.js
@@ -61,7 +61,7 @@ export default class Avatar extends PureBaseComponent {
     animate: PropTypes.bool,
     /**
      * Background color for Avatar
-     */ 
+     */
     backgroundColor: PropTypes.string,
     /**
      * Badge location on Avatar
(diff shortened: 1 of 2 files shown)


src/components/avatar/index.js
```diff
@@ -294,6 +294,7 @@ export default class Avatar extends PureBaseComponent {
 
   render() {
     const {
+      accessibilityLabel,
```
ethanshar: I think you forgot to remove it, it should also fail lint

src/components/badge/index.js
```diff
@@ -97,6 +97,17 @@ export default class Badge extends PureBaseComponent {
     }
   }
 
+  getAccessibilityProps() {
+    const {onPress, icon, label} = this.props;
+    
+    return {
+    accessibilityLabel: icon ? 'badge' : `badge ${label}` || 'badge',
```
ethanshar: what is this
```
`badge ${label}` || 'badge'
```
you're using pipe for a fallback value?

ethanshar: also consider using `imagebutton` role for when the badge has onPress with an icon

ArnonZ: Fixed.
Regarding `imagebutton`, it can have onPress in all badge variation (icon, text, no text - just bg color).

PR #528, opened 2019-09-05
Feat/add accessibility support

ethanshar:


src/components/picker/index.js
```diff
@@ -233,6 +242,7 @@ class Picker extends BaseComponent {
           getItemValue: child.props.getItemValue || getItemValue,
           onSelectedLayout: this.onSelectedItemLayout,
           renderItem: child.props.renderItem || renderItem,
+          accessibilityHint: 'Double click to select this suggestion'
```
Inbal-Tish: ... to select this option

ethanshar: it's not my content, it's according to the UX guidelines

PR #529, opened 2019-09-08
Feat/accessibility part 1

M-i-k-e-l:


src/components/radioButton/RadioButton.js
```diff
@@ -136,6 +136,19 @@ class RadioButton extends BaseComponent {
     }
   };
 
+  getAccessibilityProps = () => {
+    const {label, selected, disabled} = this.getThemeProps();
+    const selectedAccessibilityText = selected ? 'selected. ' : 'unselected. ';
+    const accessibilityLabel = selectedAccessibilityText + (label ? label : '');
```
ethanshar: I recommend using template strings instead of concating with `+` 
also regarding this `(label ? label : '')`
first, pipe with default value will be cleaner here
second, even better solution is to extract the `label` with a default value of empty string, this way you don't need this whole condition.

M-i-k-e-l: Done

PR #538, opened 2019-09-18
Fix some whyDidYouRender issues

M-i-k-e-l:


demo/src/screens/componentScreens/DialogScreen.js
```diff
@@ -37,9 +37,15 @@ export default class DialogScreen extends Component {
     {value: Colors.yellow70, label: 'Yellow70'}
   ];
 
+  static pannableTitle = {title: 'This is a pannable header Dialog'};
```
ethanshar: any reason to keep these statics?

M-i-k-e-l: No, changes to `this`

PR #546, opened 2019-10-07
BaseInput - adding 'markRequired' prop

Inbal-Tish:


src/components/inputs/TextField.js
```diff
@@ -617,13 +616,13 @@ export default class TextField extends BaseInput {
...
 
-function createStyles({placeholderTextColor, centered, multiline, expandable}) {
+function createStyles({centered, multiline, expandable}) {
   const inputTextAlign = Constants.isRTL ? 'right' : 'left';
 
   return StyleSheet.create({
     container: {},
     innerContainer: {
-      flexGrow: 1,
+      // flexGrow: 1, // create bugs with lineHeight
```
ethanshar: I don't think we can remove it, I almost certain there was an important reason for it.

Inbal-Tish: To be honest I don't remember the issue nor see any difference with or without it now. I reverted the change, we'll revisit it if we need to.

PR #550, opened 2019-10-28
Allow NativePicker to change items

M-i-k-e-l:


src/components/picker/NativePicker.js
```diff
@@ -14,6 +14,13 @@ class NativePicker extends BaseComponent {
     showDialog: false
   };
 
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
ethanshar: I'm not so thrilled to add UNSAFE lifecycle methods when we plan to deprecate them soon
Is it possible to use one of the new lifecycle methods ?
also, we need to understand carefully if this component is a "controlled" one or an "uncontrolled"
we should aim to keep our component as "controlled" as possible. 
I wrote a comment on the ticket, please see what I wrote there, this is the recommended solution in my opinion.

M-i-k-e-l: The problem with the new methods is they do not go together the the old ones and since we inherit from `BaseComponent` we have one of those (we're using `getThemeProps` so it looks like we cannot change the inheritance).
I think the `Picker` is controlled (like you suggested), so that seems like a valid solution.

PR #551, opened 2019-10-29
Selected value should not look disabled in picker

M-i-k-e-l:


src/components/picker/index.js
```diff
@@ -336,5 +340,11 @@ class Picker extends BaseComponent {
   }
 }
 
+const styles = StyleSheet.create({
```
ethanshar: this is a general comment on the PR. 
not sure fixing the Picker is the right solution in this case. 
Picker is based on TextField, therefore the color, and the UI in general, should be derived from the TextField. 
Now, I've checked if the TextField has the same issue and it does. so if anything we should fix the source of the problem. 

Having said that, it's possible the fix should part of our the private uilib since maybe our color configuration is there. 
There's currently an open PR by Inbal in the private that is related to the TextField. 
so let's put this one on hold till we migrate her PR and then I would recommend investigate the issue in the public code.

M-i-k-e-l: We can put this on hold, but I think the solution should be for the public lib as well since I think it's a bad UX in general, not just for us.

src/components/picker/index.js
```diff
@@ -321,10 +323,12 @@ class Picker extends BaseComponent {
     }
 
     const textInputProps = TextField.extractOwnProps(this.props);
+    const style = textInputProps.style;
```
ethanshar: style is a common prop, all components has it, it has no meaning getting it from `textInputProps`.

src/components/picker/index.js
```diff
@@ -321,10 +323,12 @@ class Picker extends BaseComponent {
...
     const label = this.getLabel();
     return (
       <TextField
         {...textInputProps}
+        style={[styles.textField, style]}
```
ethanshar: instead of using style for coloring the inner TextField 
I would have use `TextField`'s `color` prop.
this way we can then override it in our private code in our `Picker.config`

src/components/picker/index.js
M-i-k-e-l: Removed

src/components/picker/index.js
M-i-k-e-l: Fixed

PR #556, opened 2019-11-11
Fix new Dialog's extractOwnProps

M-i-k-e-l:


src/components/dialog/index.js
```diff
@@ -88,6 +89,14 @@ class Dialog extends BaseComponent {
     width: '90%'
   };
```
ethanshar: let's just migrate the Dialog in places we're using it 
like `PickerDialog.android` and `PickerDialog.ios`

M-i-k-e-l: Done

PR #557, opened 2019-11-11
support dynamic width for Carousel

ethanshar:


demo/src/screens/componentScreens/CarouselScreen.js
```diff
@@ -62,6 +62,20 @@ class CarouselScreen extends Component {
         <View margin-20 center /*style={{...StyleSheet.absoluteFillObject}} */ pointerEvents="none">
           <Text text10>{this.state.currentPage}</Text>
         </View>
+
+        <View padding-20>
+          <Carousel test migrate containerStyle={{height: 160}}>
+            {_.times(5, index => {
+              return (
+                <View key={index} flex padding-20>
+                  <Card flex padding-20>
```
Inbal-Tish: I think it's not clear enough that this is a "full width" Carousel example because of the Cards...

src/components/carousel/index.js
```diff
@@ -145,11 +145,31 @@ export default class Carousel extends BaseComponent {
     return index;
   }
 
+  getSnapToOffsets = () => {
+    const {itemSpacings} = this.props;
+    const {containerWidth, pageWidth} = this.state;
+
+    if (containerWidth) {
```
Inbal-Tish: This condition means that even if the user passed 'pageWidth' it needs to wait for the onLayout to get the snapToOffset.

src/components/carousel/index.js
```diff
@@ -145,11 +145,31 @@ export default class Carousel extends BaseComponent {
...
+    }
+  }
+
   shouldUsePageWidth() {
     const {loop, pageWidth} = this.props;
     return !loop && pageWidth;
   }
 
+  onContainerLayout = ({nativeEvent: {layout: {width: containerWidth}}}) => {
+    const update = {containerWidth};
```
Inbal-Tish: I would declare 'containerWidth' in the state object in the c'tor.

demo/src/screens/componentScreens/CarouselScreen.js
ethanshar: yea, I wanted to still look nice.
but maybe I can add a title for the example.
WDYT?

src/components/carousel/index.js
ethanshar: it will have to wait for the `snapToOffset` because the code of the offsets depends on the containerWidth. 
That's another issue we had there, we always assume the carousel is full-width, so the "breakpoints" were in-correct sometimes. 
But it doesn't delay the render. I checked it.

src/components/carousel/index.js
ethanshar: This is how I started it, and them removed it on purpose. 
I want it to be undefined at the beginning so it can block the initial render (otherwise there's a flicker)

Inbal-Tish: Yes. Of course, as undefined. I only like to declare all state variables at the top for the sack of clarity.

demo/src/screens/componentScreens/CarouselScreen.js
Inbal-Tish: How about something like an image gallery that looks less like the other example?

src/components/carousel/index.js
Inbal-Tish: Yes sure, you can clearly see that it doesn't delay the render, especially with the other carousel example on the screen.

PR #560, opened 2019-11-14
Picker - allow error state

M-i-k-e-l:


src/components/inputs/TextField.js
```diff
@@ -175,6 +175,10 @@ export default class TextField extends BaseInput {
     if (prevState.value !== this.state.value || prevProps.focused !== this.state.focused) {
       this.updateFloatingPlaceholderState();
     }
+
+    if (prevProps.value !== this.props.value) {
```
M-i-k-e-l: Unsure if this should be here or in `BaseInput`

ethanshar: why do we need it at all?

src/components/picker/index.js
```diff
@@ -128,6 +128,7 @@ class Picker extends BaseComponent {
 
   static defaultProps = {
     ...TextField.defaultProps,
+    enableErrors: undefined,
```
ethanshar: if anything define the default to be false, all props by default are undefined.

src/components/picker/index.js
```diff
@@ -302,7 +303,7 @@ class Picker extends BaseComponent {
   };
 
   render() {
-    const {useNativePicker, renderPicker, customPickerProps, testID} = this.props;
```
ethanshar: is this whole change really necessary? 
till now we override whatever the user pass for `enableErrors` with false. 
you can quickly just enable the user use this prop with a much smaller change

src/components/inputs/TextField.js
```diff
@@ -175,6 +175,10 @@ export default class TextField extends BaseInput {
     if (prevState.value !== this.state.value || prevProps.focused !== this.state.focused) {
       this.updateFloatingPlaceholderState();
     }
+
```
ethanshar: the PR has conflicts

src/components/inputs/TextField.js
M-i-k-e-l: If we do not add this the error is not cleared when the selection is changed in the Picker. Since validation only occurs onChangeText until now, a prop change does not affect it.

src/components/picker/index.js
M-i-k-e-l: Removed

src/components/picker/index.js
M-i-k-e-l: Done

src/components/inputs/TextField.js
M-i-k-e-l: Merged and removed most of the code

src/components/inputs/TextField.js
ethanshar: the question is - is it an issue? 
If we aim to have a controlled component that means the user is responsible for clearing the error by passing undefined to `error` prop when the value has changed.

ethanshar: I actually now looked at the related private PR.
I think I get where you're going with this. 
Let's take this offline.

M-i-k-e-l: Done

Issue #539, opened 2019-09-20
Picker - Invariant Violation: [1347,"RCTView",51,{"flexDirection":"row","height":"<<NaN>>"}]


francopetra:
I upgraded to the latest version and this isssue started to happen in Android (works fine in iOS).

<img width="310" alt="Screen Shot 2019-09-20 at 12 48 44 PM" src="https://user-images.githubusercontent.com/20055787/65349509-f0638f00-dbb9-11e9-9a6a-a6380418fdc0.png">

mouthzipper:
got the same issue

vdino96:
I have the same problem, addressed it in an issue https://github.com/wix/react-native-ui-lib/issues/536, still waiting for response. So tune in at that issue, too.

schieck0:
I have the same problem on all my projects

grrrgpratik:
I have the same problem when updating my react-native to 0.60.4

Issue #536, opened 2019-09-17
Invariant Violation issue/Malformed calls from JS


vdino96:
Hello,

upon upgrading my react-native project to 0.59.10, I have received the following error when opening Picker:

![rnuilib1](https://user-images.githubusercontent.com/51232663/65022773-dca5f780-d931-11e9-850d-721a7c277351.png)

Debugger: 
![rnuilib2](https://user-images.githubusercontent.com/51232663/65024109-7c648500-d934-11e9-87c4-c16cbe3a73d6.png)

I have searched the other answers from different libraries and it seems that this is the common error when setting the height according to StatusBar height, somehow. 

I have searched through the code and found that the issue is in "Constants.statausBarHeight" that is set in "TopBar.js". When I comment out that part of code, the Picker starts working again.

![rnuilib3](https://user-images.githubusercontent.com/51232663/65023209-a5841600-d932-11e9-9ee8-54244e860d7d.png)

I am not sure if anyone else is having the same problem and I am sure this is not a solution that would be accepted, just pointing out the issue.

Epick362:
Having the same problem. Only seems to happen after I reload the app, though. When I re-run `react-native run-android` the picker modal works again until the next reload.

PR #562, opened 2019-11-20
Bug Fix: Invariant Violation issue/Malformed calls from JS

ahmadkhudeish:
It seems like the value of **`StatusBarManager.HEIGHT`** inside 
> `/react-native-ui-lib/src/helpers/Constants.js`
 is always resulting to **`undefined`** when called inside the container stylesheet of the following file: 
> `/react-native-ui-lib/src/screensComponents/modal/TopBar.js`

That's why it's showing up the error height: "NaN" as referenced by this #539 and #536 

You have a problem with the way you are exporting the **setStatusBarHeight** method inside the **Constants.js** file.

Fix: 

```
function setStatusBarHeight() {
  let height = 0;
  height = isIOS ? 20 : StatusBarManager.HEIGHT;
  if (isIOS) {
    // override guesstimate height with the actual height from StatusBarManager
    StatusBarManager.getHeight(data => (height = data.height));
  }
  return height;
}

```


ahmadkhudeish:
created the change in another pull request

PR #567, opened 2019-11-27
fix: update avatar component

cajuuh:
Remove the 241 line so the label on the avatar component don't be overwritten.

### DIFF
--- a/src/components/avatar/index.js
+++ b/src/components/avatar/index.js
@@ -238,7 +238,6 @@ export default class Avatar extends PureBaseComponent {
           size={this.getBadgeSize()}
           {...badgeProps}
           containerStyle={this.getBadgePosition()}
-          label={undefined}
           testID={`${testID}.onlineBadge`}
         />
       );


src/components/avatar/index.js
```diff
@@ -238,7 +238,6 @@ export default class Avatar extends PureBaseComponent {
           size={this.getBadgeSize()}
           {...badgeProps}
           containerStyle={this.getBadgePosition()}
-          label={undefined}
```
ArnonZ: 👋 , thank you for the pull request!
I'm not sure I understand why this line deletion will achive what you descriped: "so the label on the avatar component don't be overwritten."
what this deleted line (241) does is control the **Avatar's badge** so it will always show as a "pimple badge" (without text) as can be seen in the Avatar example screen. 
This line does not control the Avatar's label...

Please specify your intentions since currently the deletion of this line will not achieve what you seek.

cajuuh: hi @ArnonZ what happens is, instead of the value passed on the {...badgeProps} prop being showed on the badge inside the `Avatar` component the value of the label is always undefined, is being overwritten by this line, so if you're changing the `badgeProps` on the Avatar component, the `label` string value will never be showed since it's being overwritten on the component itself, i'm not sure if i'm being clear, my english is a little bit rush, otherwise since the `label` prop of badge is already set on the `badgeProps` even if the user does not set a value to the label it'll become undefined, this line does not affect the use of the `badgeProps` it overwrites it

ArnonZ: Agree, this is the case regarding the Badges label (not the Avatar's).
Should you need the look of a badge with textual label please use `ribbonLabel` prop can also be seen in our demo app:

`
<Avatar
  ribbonLabel={'label'}
  ribbonStyle={{backgroundColor: Colors.green30}}
/>
`

Hope this helps.

cajuuh: Well, i need the label to show a number as string, if i create the component that way:
```
<Avatar
            size={32}
            badgePosition="TOP_LEFT"
            badgeProps={{
              size: "small",
              label: "2",
              backgroundColor: "red",
              borderWidth: 1.5,
              borderColor: "white",
              labelStyle: { color: "white", fontSize: 10 }
            }}
            imageSource={{
              uri: "https://randomuser.me/api/portraits/women/24.jpg"
            }}
            animate={true}
          />
```
Since the Avatar component is overwritting the label prop and setting the value to `undefined` it'll never assign the `"2"` value to the badge inside the avatar. Otherwise, removing the 241 line, even if i don't set a value to the `label` inside `badgeProps` it'll be considered `undefined` so the line can be removed, unless it'll be used by another component inside the Avatar component.

By the way, using the ribbon i can't set the position to top left, the ribbon becomes all twisted, using the badge and assigning the amount of, lets suppose, notifications, fits better my approach. 

Thanks for your patience @ArnonZ

cajuuh: @ArnonZ maybe i'm not understanding the purpose of the badge itself...

Inbal-Tish: @cajuuh Hi. What @ArnonZ is trying to say is that the Badge on Avatar will never have a label BY DESIGN, that's why the label is set to undefined to override the Badge's label prop intentionally (on purpose). The purpose of the Avatar's Badge is to be displayed as a Pimple (only an indicator, not text), and if you want to have an indicator with text on the Avatar you should use the Ribbon props instead.

cajuuh: Hi @Inbal-Tish so i was right about myself, i didn't understood the purpose of the badge right, i was hoping it looked like a notification badge, or something like that, i tried to style the ribbon so it looked like a notification badge on the top left corner of the avatar but i've failed, like i said the ribbon got all twisted, anyway, thank you for the explanation, and sorry @ArnonZ  for being so incisive on my mistake, sorry man.

ArnonZ: All good ✌️

PR #568, opened 2019-11-27
Carousel - fix initialPage for dynamic width

Inbal-Tish:


src/components/carousel/index.js
```diff
@@ -229,11 +227,6 @@ export default class Carousel extends BaseComponent {
   };
 
   renderChildren() {
-    const {containerWidth} = this.state;
```
ethanshar: By removing this, aren't u getting a flicker with the children width?

Inbal-Tish: No. Do you?

ethanshar: I can check again, but the reason I added this code was to prevent this flicker. 
without it the children will render twice. The second time will correct the children width and its noticeable..

PR #569, opened 2019-11-28
Fix image overlay (and cover) with absolute position

M-i-k-e-l:


src/components/overlay/index.js
```diff
@@ -44,34 +49,59 @@ export default class Overlay extends PureBaseComponent {
...
+    if (flatStyle) {
+      attributes = _.chain(flatStyle)
+        .pickBy((value, key) => _.includes(attNames, key))
         .value();
     }
+
     return attributes;
   }
 
+  getStyle = typeStyle => {
```
ethanshar: I'm not a fan of this change, let's try exploring the ImageBackground component for when rendering overlay on top of an image
https://facebook.github.io/react-native/docs/imagebackground#docsNav

M-i-k-e-l: I’ve been working on this for some time (using `ImageBackground`), I’ve reached about the same result.
It seemed simpler at start but it had a problem with images that fill their container, because of:

> Note that you must specify some width and height style attributes.
(https://facebook.github.io/react-native/docs/imagebackground#docsNav)

I think it has about the same complexity and both versions have a small bug that I think we should take offline to look at.

Link to the other option: https://github.com/wix/react-native-ui-lib/tree/fix/image-overlay-with-absolute-position-take-2

PR #576, opened 2019-12-04
Support useSafeArea in Button

M-i-k-e-l:


src/components/button/index.js
```diff
@@ -505,6 +515,21 @@ export default class Button extends PureBaseComponent {
         </View>
       </TouchableOpacity>
     );
+  };
+
+  render() {
+    const {useSafeArea} = this.getThemeProps();
+    const backgroundColor = this.getBackgroundColor();
+
+    if (useSafeArea) {
```
ethanshar: this change of layout feel risky... 
Currently I found an issue with the border-radius being broken

Did u consider maybe implement the code in `TouchableOpacity` ?

M-i-k-e-l: You found an issue with or without `SafeArea`?
Do you mean doing the layout change in `TouchableOpacity`?

ethanshar: > You found an issue with or without SafeArea?
With `useSafeArea`. You assume someone will render a rectangle button but that's not necessarily the case. 
> Do you mean doing the layout change in TouchableOpacity?
yes, like we're solving the safe area in `View`. I guess it make sense to fix it the most base component.

M-i-k-e-l: I managed to do it in `TouchableOpacity` but it feels a bit hacky.
We cannot just add a `View` with `useSafeArea` prop because the `backgroundColor` might be nested in the `style` (and I assumed we do not want to use `flatten`).
I'll push the changed, let me know what you think.

ethanshar: The solution is still not good. try rendering the following in Playground
```
<View bg-dark80 flex bottom>
        <TouchableOpacity bg-red40 useSafeArea style={{backgroundColor: 'red'}}>
        </TouchableOpacity>
      </View>
```

Why not render as a child (inside TouchableOpacity) a SafeAreView so it will be add the necessary padding.

src/components/touchableOpacity/index.js
```diff
@@ -117,7 +119,28 @@ export default class TouchableOpacity extends PureBaseComponent {
     );
   }
 
+  render() {
+    const {useSafeArea, safeAreaBackgroundColor} = this.getThemeProps();
+
+    if (useSafeArea && Constants.isIphoneX) {
```
ethanshar: try instead rendering a `SafeAreaView` inside as a child. all you need here is to add the extra padding

src/components/button/index.js
M-i-k-e-l: Fixed

src/components/touchableOpacity/index.js
M-i-k-e-l: Done

PR #580, opened 2019-12-05
Fix Avatar and Badge accessibility

M-i-k-e-l:


src/components/avatar/index.js
```diff
@@ -314,17 +314,17 @@ export default class Avatar extends PureBaseComponent {
...
+        accessible={!_.isUndefined(onPress)}
         accessibilityLabel={'Avatar'}
         accessibilityRole={onPress ? 'button' : 'image'}
         {...this.extractAccessibilityProps()}
       >
         <View
           style={[this.getInitialsContainer(), {backgroundColor}, hasImage && this.styles.initialsContainerWithInset]}
         >
-          <Text numberOfLines={1} style={[{fontSize}, this.styles.initials, {color}]}>
+          {label && <Text numberOfLines={1} style={[{fontSize}, this.styles.initials, {color}]}>
```
ethanshar: this is what we talked about last week.
I believe that if you'll pass `''` (empty string) to label it will throw an error, can u please verify

M-i-k-e-l: You are right 🤦‍♂ 
I'll fix this tomorrow since I cannot test this on iOS ATM.

ethanshar: You can test on a simulator, the fix is simple, as discussed we need to use _.isUndefined

M-i-k-e-l: Fixed

src/components/badge/index.js
```diff
@@ -236,7 +237,7 @@ export default class Badge extends PureBaseComponent {
           activeOpacity={activeOpacity}
           {...animationProps}
         >
-          {icon ? this.renderIcon() : this.renderLabel()}
+          {icon ? this.renderIcon() : label ? this.renderLabel() : undefined}
```
ethanshar: regarding this `label ? this.renderLabel() : undefined`
why shouldn't you move this logic to be part of `renderLabel` 
ternary inside a ternary is never readable...

M-i-k-e-l: Done

PR #587, opened 2019-12-10
Adding carousel support to tab controller

ethanshar:
In general, I added the option for the user to wrap `TabPage`s with `PageCarousel` component.
this will change the behavior of the TabController to render the pages as carousel instead of single page each time.

I made some changes in `Carousel` component to allow it to accept `Animated.event` for `onScroll` instead of a function. This way I can intercept the carousel offset position to use in the TabController.

### DIFF
--- a/demo/src/screens/incubatorScreens/TabControllerScreen/index.js
+++ b/demo/src/screens/incubatorScreens/TabControllerScreen/index.js
@@ -6,18 +6,19 @@ import Tab1 from './tab1';
 import Tab2 from './tab2';
 import Tab3 from './tab3';
 
-const TABS = ['tab1', 'tab2' , 'tab3', 'account', 'groups', 'blog'];
+const USE_CAROUSEL = true;
+const TABS = ['tab1', 'tab2', 'tab3', 'account', 'groups', 'blog'];
 
 class TabControllerScreen extends Component {
   state = {
     selectedIndex: 0,
     items: [
       ..._.map(TABS, tab => ({label: tab, key: tab})),
-      {key: 'addTabs', icon: Assets.icons.settings, ignore: true, onPress: this.addTab},
+      {key: 'addTabs', icon: Assets.icons.settings, ignore: true, onPress: this.addTab}
     ],
 
     tabsCount: 3,
-    key: Date.now(),
+    key: Date.now()
   };
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


Inbal-Tish:
@ethanshar I have general questions about the design: 
First, In `PageCarousel` component, why use `Carousel` instead of `Animated.ScrollView` directly? I see that the `TabController` already has all the features for selecting index and callback for index change, so what is the `Carousel` benefit here? 
Second, why pass every `TabPage` the `asCarouselPage` prop instead of having something like `useCarousel` option on the `TabController` that, when true, renders the children (the `TabPage`s) in the `PageCarousel` component. 
Third, why expose the `PageCarousel` component as a container and not use it internally from `TabController`? Or moreover, why to have the `TabPage`s in a container and not use something like `FlatList`'s `renderItem` function?

ethanshar:
@Inbal-Tish 
> First, In PageCarousel component, why use Carousel instead of Animated.ScrollView directly? I see that the TabController already has all the features for selecting index and callback for index change, so what is the Carousel benefit here?

I tried using `Animated.ScrollView` directly, but the layout didn't work well so I thought I might need to duplicate the code that we do for children, with wrapping the children. But I agree, it should have been enough. I'll try again. I also prefer avoiding changing things in Carousel.


> Second, why pass every TabPage the asCarouselPage prop instead of having something like useCarousel option on the TabController that, when true, renders the children (the TabPages) in the PageCarousel component.

Agree, that was the main intention, I'll fix it.

> Third, why expose the PageCarousel component as a container and not use it internally from TabController? Or moreover, why to have the TabPages in a container and not use something like FlatList's renderItem function?

I don't believe it's impossible. The current solution allows the users to render the TabPage anywhere in the hierarchy. I can't control it. and also can't wrap the pages for them. it work similar to Carousel, when you have to render elements inside it.

demo/src/screens/incubatorScreens/TabControllerScreen/index.js
```diff
@@ -6,18 +6,19 @@ import Tab1 from './tab1';
...
+const TABS = ['tab1', 'tab2', 'tab3', 'account', 'groups', 'blog'];
 
 class TabControllerScreen extends Component {
   state = {
     selectedIndex: 0,
     items: [
       ..._.map(TABS, tab => ({label: tab, key: tab})),
-      {key: 'addTabs', icon: Assets.icons.settings, ignore: true, onPress: this.addTab},
+      {key: 'addTabs', icon: Assets.icons.settings, ignore: true, onPress: this.addTab}
     ],
```
Inbal-Tish: extra line

src/incubator/TabController/index.js
```diff
@@ -71,18 +80,31 @@ class TabController extends Component {
 
   render() {
     const {itemStates, ignoredItems} = this.state;
+    
+    // Rounding on Android, cause it cause issues when comparing values
+    const screenWidth = Constants.isAndroid ? Math.round(Constants.screenWidth) : Constants.screenWidth;
```
Inbal-Tish: Is landscape support is on the todo list?

ethanshar: yes, it won't work on Landscape at the moment I'm aware.

demo/src/screens/incubatorScreens/TabControllerScreen/index.js
ethanshar: extra line where?

Inbal-Tish: line 19. Nevermind.

PR #594, opened 2019-12-17
Add Wizard (new component)

M-i-k-e-l:


src/components/wizard/WizardStep.js
```diff
@@ -0,0 +1,159 @@
...
+import {StyleSheet} from 'react-native';
+import View from '../view';
+import Text from '../text';
+import Image from '../image';
+import TouchableOpacity from '../touchableOpacity';
+import {PureBaseComponent} from '../../commons';
+import Colors from '../../style/colors';
+import BorderRadiuses from '../../style/borderRadiuses';
+import Spacings from '../../style/spacings';
+import {States, StatesConfig} from './WizardStates';
```
ethanshar: States is not being used, it will fail lint

src/components/wizard/WizardStep.js
```diff
@@ -0,0 +1,159 @@
...
+    } else {
+      return config[propName];
+    }
+  }
+
+  renderCircle(config, activeConfig) {
+    const {testID, index, onPress, indexLabelStyle} = this.getThemeProps();
+    const hitSlopSize = Spacings.s2;
+    const color = this.getProp(activeConfig, config, 'color');
+    const circleColor = this.getProp(activeConfig, config, 'circleColor');
```
ethanshar: I would use the `color` value as default for the `circleColor`.

src/components/wizard/WizardStep.js
```diff
@@ -0,0 +1,159 @@
...
+
+    return (
+      <TouchableOpacity
+        testID={`${testID}.circle`}
+        style={[styles.circle, {borderColor: circleColor}]}
+        onPress={enabled ? onPress : undefined}
+        hitSlop={{top: hitSlopSize, bottom: hitSlopSize, left: hitSlopSize, right: hitSlopSize}}
+        disabled={!enabled}
+      >
+        {_.isUndefined(icon) ? (
```
ethanshar: this is one of the cases that checking with `_.isUndefined` is redundant. `icon ? ... : ....` is enough

src/components/wizard/WizardStep.js
```diff
@@ -0,0 +1,159 @@
...
+          <Image testID={`${testID}.image`} source={icon} tintColor={color}/>
+        )}
+      </TouchableOpacity>
+    );
+  }
+
+  render() {
+    const {
+      testID,
+      state,
```
ethanshar: `state` is missing from prop types

src/components/wizard/WizardStep.js
```diff
@@ -0,0 +1,159 @@
...
+     */
+    indexLabelStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
+    /**
+     * Whether the step should be enabled
+     */
+    enabled: PropTypes.bool
+  };
+
+  getProp(activeConfig, config, propName) {
+    const props = this.getThemeProps();
```
ethanshar: why extracting the props like that?
also `index` and `activeIndex` can be taken directly from `this.props`

ethanshar: ok I see why u did that. 

anyway, you can shorten the way you get the config values. 
the way I see it there are 3 sets of config. 
the state config, the active config and the props (which should are identical to the config structure)

you can create some sort of merged config out of those three and just take the final values from the merge config. 
it should turn this whole code two 2-3 lines of code (:

src/components/wizard/WizardStep.js
M-i-k-e-l: I was unsure if it can be 0

src/components/wizard/WizardStep.js
M-i-k-e-l: Done.

src/components/wizard/WizardStep.js
M-i-k-e-l: Added state to props, it is now being used

src/components/wizard/WizardStep.js
M-i-k-e-l: Done.

src/components/wizard/WizardStep.js
M-i-k-e-l: Done

PR #595, opened 2019-12-18
Feat/date time picker

Inbal-Tish:
Adding DateTimePicker wrapper component


android/settings.gradle
```diff
@@ -6,4 +6,7 @@ include ':react-native-navigation'
 project(':react-native-navigation').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-navigation/lib/android/app/')
 
 include ':react-native-ui-lib'
-project(':react-native-ui-lib').projectDir = new File(rootProject.projectDir, '../lib/android')
\ No newline at end of file
+project(':react-native-ui-lib').projectDir = new File(rootProject.projectDir, '../lib/android')
```
ethanshar: this is also should be unnecessary

src/components/dateTimePicker/index.js
```diff
@@ -0,0 +1,207 @@
...
+    _.invoke(this.props, 'onChange');
+  }
+
+  toggleExpandableModal = () => {
+    this.setState({showExpandableModal: !this.state.showExpandableModal});
+  };
+
+  onDonePressed = () => {
+    this.toggleExpandableModal();
+    setTimeout(() => {
```
ethanshar: maybe instead of triggering  setState twice you can pass to `toggleExpandableModal` an extra state to update with

src/components/dateTimePicker/index.js
```diff
@@ -0,0 +1,207 @@
...
+  static propTypes = {
+    ...TextField.propTypes,
+    /**
+     * The type of picker to display ('date' or 'time')
+     */
+    mode: PropTypes.oneOf(Object.values(MODES)),
+    /**
+     * The initial value to set the picker to. Defaults to device's date / time
+     */
+    value: PropTypes.instanceOf(Date),
```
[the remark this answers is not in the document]
Inbal-Tish: `value` is the name of the RNDateTimePicker prop

src/components/dateTimePicker/index.js
Inbal-Tish: Not sure I understand. Why do you say I'm triggering setState twice? toggleExpandableModal is not triggered for Android...

android/app/build.gradle
```diff
@@ -80,6 +80,7 @@ dependencies {
     implementation project(':react-native-ui-lib')
     implementation 'com.facebook.fresco:fresco:2.0.0'
     implementation 'com.facebook.fresco:animated-gif:2.0.0'
+    implementation project(':react-native-datetimepicker')
```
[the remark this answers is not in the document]
Inbal-Tish: I didn't try it... I will and fix the rest of the native files

package.json
```diff
@@ -38,6 +38,7 @@
     "build:typings": "rm -rf ./typings && ./scripts/build-typings.sh && tsc"
   },
   "dependencies": {
+    "@react-native-community/datetimepicker": "^2.1.0",
```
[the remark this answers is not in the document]
Inbal-Tish: Moved to dev-dependencies.

android/app/build.gradle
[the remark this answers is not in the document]
Inbal-Tish: Done.

ios/Podfile
```diff
@@ -4,6 +4,8 @@ require_relative '../node_modules/@react-native-community/cli-platform-ios/nativ
 target 'rnuilib' do
   pod 'ReactNativeUiLib', :path => "../lib/ReactNativeUiLib.podspec"
   # Pods for rnuilib
+  pod 'RNDateTimePicker', :path => '../node_modules/@react-native-community/datetimepicker/RNDateTimePicker.podspec'
```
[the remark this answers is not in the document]
Inbal-Tish: Done.

android/settings.gradle
Inbal-Tish: Done.

package.json
```diff
@@ -57,6 +57,7 @@
     "@babel/runtime": "^7.4.2",
     "@react-native-community/async-storage": "^1.6.2",
     "@react-native-community/blur": "^3.4.1",
+    "@react-native-community/datetimepicker": "^2.1.0",
```
ethanshar: also add as peerDep

src/components/dateTimePicker/index.js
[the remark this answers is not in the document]
ethanshar: The question is - should "fix" their API or stick to what they have... 
Their value act as an initial value ?

src/components/dateTimePicker/index.js
ethanshar: because `this.toggleExpandableModal` calls setState
and right after that you call it again for setting the `chosenDate`

Inbal-Tish: Yes, but that's intentional. I need to close the modal and only then update the value of the TextField. That is also why I added the timeout around the setState() on the chosen date. Do you think there is a better way to time it?

Inbal-Tish:
Regarding the 'value' prop. Yes, their 'value' acts as an initial value, but I don't think we should fix their API. First, because it will be easier for those who want to migrate from their API to ours. Second, in our case, it also acts as the TextField 'value' prop so it's more intuitive in my opinion.

package.json
Inbal-Tish: Done.

src/components/dateTimePicker/index.js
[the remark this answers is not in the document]
Inbal-Tish: Regarding the 'value' prop. Yes, their 'value' acts as an initial value, but I don't think we should fix their API. First, because it will be easier for those who want to migrate from their API to ours. Second, in our case, it also acts as the TextField 'value' prop so it's more intuitive in my opinion.

Inbal-Tish:
Regarding the 'value' prop. Yes, their 'value' acts as an initial value, but I don't think we should fix their API. First, because it will be easier for those who want to migrate from their API to ours. Second, in our case, it also acts as the TextField 'value' prop so it's more intuitive in my opinion.

src/components/dateTimePicker/index.js
ethanshar: you can pass a callback to `this.toggleExpandableModal` that will be called after its `setState`. this way it will happen after for sure.

Inbal-Tish: Done.

PR #600, opened 2019-12-24
Infra/migrate keyboard tracking view

M-i-k-e-l:


package.json
```diff
@@ -76,8 +76,6 @@
     "react-dom": "^15.4.2",
     "react-js-to-ts": "^1.4.0",
     "react-native": "0.61.4",
-    "react-native-keyboard-aware-scrollview": "^2.0.0",
-    "react-native-keyboard-tracking-view": "^5.6.1",
```
ethanshar: let's try to make the library work while `react-native-keyboard-tracking-view` is installed because that's how it's going to be with the engine. 
as discussed, try to rename the native class names, it might fix the duplicate symbol issue that fail the build.

src/index.js
```diff
@@ -219,6 +219,12 @@ module.exports = {
   get SafeAreaInsetsManager() {
     return require('./nativeComponents').SafeAreaInsetsManager;
   },
+  get KeyboardAwareInsetsView() {
```
ethanshar: why one is commented and the other isn't..

ethanshar: Also didn't we decide to export those under the `Keyboard` namespace

demo/src/screens/MenuStructure.js
```diff
@@ -75,6 +75,7 @@ export const navigationData = {
       },
       {title: 'Wheel Picker', tags: 'wheel picker', screen: 'unicorn.nativeComponents.WheelPickerViewScreen'},
       {title: 'SafeArea Spacer', tags: 'native safe area', screen: 'unicorn.nativeComponents.SafeAreaSpacerViewScreen'}
+      // {title: 'KeyboardTracking (iOS)', tags: 'KeyboardTracking', screen: 'unicorn.nativeComponents.KeyboardTrackingViewScreen'},
```
ethanshar: Why the screen is comment out?

package.json
M-i-k-e-l: This will be fixed in one of the other PRs

demo/src/screens/MenuStructure.js
M-i-k-e-l: This will be fixed in one of the other PRs

src/index.js
M-i-k-e-l: Not sure, the first might be required because it is being used.
Both will be fixed in one of the other PRs.

PR #605, opened 2019-12-29
Carousel - enablePagination

Inbal-Tish:
Adding 'enablePagination' prop to allow blocking multiple pages scroll (full-width pages only)

### DIFF
--- a/demo/src/screens/componentScreens/CarouselScreen.js
+++ b/demo/src/screens/componentScreens/CarouselScreen.js
@@ -73,7 +73,7 @@ class CarouselScreen extends Component {
         </View>
 
         <View padding-20>
-          <Carousel migrate containerStyle={{height: 160}} initialPage={INITIAL_PAGE} loop>
+          <Carousel migrate containerStyle={{height: 160}} initialPage={INITIAL_PAGE} loop usePagination>
             {_.map(IMAGES, (image, index) => {
               return (
                 <View key={index} flex padding-10 bottom>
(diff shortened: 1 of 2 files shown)


demo/src/screens/componentScreens/CarouselScreen.js
```diff
@@ -73,7 +73,7 @@ class CarouselScreen extends Component {
         </View>
 
         <View padding-20>
-          <Carousel migrate containerStyle={{height: 160}} initialPage={INITIAL_PAGE} loop>
+          <Carousel migrate containerStyle={{height: 160}} initialPage={INITIAL_PAGE} loop usePagination>
```
ethanshar: what's `usePagination`? I don't you defined this prop anywhere..

src/components/carousel/index.js
```diff
@@ -164,6 +172,11 @@ export default class Carousel extends BaseComponent {
     return !loop && pageWidth;
   }
 
+  shouldEnablePagination() {
```
ethanshar: I don't think the user needs to control if the paging is enabled or not. 
in a full-width page carousel it should always come with paging. 
I think it's a bug, not a feature..

demo/src/screens/componentScreens/CarouselScreen.js
Inbal-Tish: opps. Fixed

src/components/carousel/index.js
Inbal-Tish: Why? In ScrollView the user controls it, why not in the Carousel? We can default it to true and the user can disable it...

ethanshar: Cause the whole point of Carousel is the paging. 
Anyway, I'm fine with setting the default value to true.

Inbal-Tish: Done.

PR #614, opened 2020-01-09
Feat/ColorPicker updates

Inbal-Tish:


src/components/colorPicker/ColorPalette.js
```diff
@@ -39,6 +36,10 @@ export default class ColorPalette extends PureBaseComponent {
      * Style to pass the palette container
      */
     containerStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
+    /**
+     * The container margins
+     */
+    containerMargins: PropTypes.number,
```
ethanshar: Is this for our tablet margins? 
can't we accomplish this with containerStyle? 
with TabController.TabBar i used the containerStyle for the tablet margins.. 
I think it will be better than defining one-app specific props

Inbal-Tish: How did you extract the margins if, say, user passed 'marginLeft', 'marginRight', 'marginHorizontal', as opposed to 'margins'? Did you check each separately to determine the container width?

Inbal-Tish: Changed containerMargins to containerWidth.

PR #616, opened 2020-01-12
Feat/Image - custom overlay

ArnonZ:
- Added support for rendering a custom overlay for Image component (can be stacked with already supported overlays).


src/components/overlay/index.js
```diff
@@ -26,7 +26,11 @@ export default class Overlay extends PureBaseComponent {
     /**
      * The type of overlay to set on top of the image
      */
-    type: PropTypes.oneOf(_.values(OVERLY_TYPES))
+    type: PropTypes.oneOf(_.values(OVERLY_TYPES)),
+    /**
+     * Custom overlay content to be rendered on top of the image
+     */
+    customContent: PropTypes.func
```
ethanshar: better define it as an element prop type

src/components/overlay/index.js
```diff
@@ -44,22 +48,32 @@ export default class Overlay extends PureBaseComponent {
     }
   }
 
+  renderCustomContent = () => {
+    const {customContent} = this.props;
+    return <View style={styles.customContent}>{customContent()}</View>;
```
ethanshar: Better pass pointerEvents to the wrapping View to avoid touch block

src/components/overlay/index.js
```diff
@@ -44,22 +48,32 @@ export default class Overlay extends PureBaseComponent {
...
-    const {type} = this.props;
+    const {type, customContent} = this.props;
 
     if (type === OVERLY_TYPES.VERTICAL) {
-
       return (
         <View style={[styles.container]}>
           {this.renderImage(styles.top)}
           {this.renderImage(styles.bottom)}
+          {customContent && this.renderCustomContent()}
```
ethanshar: you render the custom content too many times, in this example 
```
<Image
          customOverlay={() => <View><Text>A</Text></View> }
          overlayType={'vertical'}
          style={{width: 100, height: 100}}
          source={{
            uri:
              'https://images.pexels.com/photos/3326362/pexels-photo-3326362.jpeg?auto=compress&cs=tinysrgb&dpr=1&w=500'
          }}
        />
```

it will be rendered 3 times.

demo/src/screens/componentScreens/OverlaysScreen.js
```diff
@@ -1,22 +1,53 @@
...
+        />
+      </View>
+    );
+  }
+
+  renderCustomOverlay() {
+    return (
+      <Image
+        style={{
+          position: 'absolute',
```
ethanshar: Consider moving this big style chunk to a stylesheet

src/components/image/index.js
```diff
@@ -49,7 +49,11 @@ class Image extends PureBaseComponent {
      * The type of overly to place on top of the image. Note: the image MUST have proper size, see examples in:
      * https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/OverlaysScreen.js
      */
-    overlayType: Overlay.propTypes.type
+    overlayType: Overlay.propTypes.type,
+    /**
+     * Render an overlay with custom content
+     */
+    customOverlay: PropTypes.func
```
ethanshar: Consider renaming to `customOverlayContent` cause `customOverlay` sounds like you can provide custom gradient overlay

src/components/overlay/index.js
ArnonZ: ooooooh weeeeee...

src/components/overlay/index.js
```diff
@@ -50,35 +50,37 @@ export default class Overlay extends PureBaseComponent {
...
-
-  renderImage(typeStyle) {
-    const {type, customContent} = this.props;
-    const image = type !== OVERLY_TYPES.SOLID ? gradientImage : undefined;
-
     return (
-      <View style={styles.container}>
-        {type && <Image style={[styles.container, typeStyle]} resizeMode={'stretch'} source={image}/>}
-        {customContent && this.renderCustomContent()}
+      <View pointerEvents="none" style={styles.customContent}>
```
ethanshar: use box-none so the next custom content won't be blocked

src/components/overlay/index.js
```diff
@@ -50,35 +50,37 @@ export default class Overlay extends PureBaseComponent {
...
+    return <Image style={[styles.container, style]} resizeMode={'stretch'} source={source}/>;
+  };
 
   render() {
     const {type, customContent} = this.props;
+    const image = type !== OVERLY_TYPES.SOLID ? gradientImage : undefined;
 
     if (type === OVERLY_TYPES.VERTICAL) {
       return (
-        <View style={[styles.container]}>
```
ethanshar: Consider using Fragment (`<>`) instead of the container View for both Vertical case and the other case

PR #618, opened 2020-01-13
improve TextField example screen

ethanshar:
I separated the text field screen to 3 
- General Inputs - what we have today
- Custom Inputs - showing different custom UI styling 
- Input Validation - show the (new) validation functionality in depth 

This is just the initial work, but from here we can start re-arrange the input examples better. 
It will also help when we'll add e2e tests


demo/src/screens/componentScreens/TextFieldScreen/InputsScreen.js
```diff
@@ -41,22 +41,22 @@ export default class InputsScreen extends Component {
       message = 'Please enter a valid text';
     }
     this.setState({error: message});
```
Inbal-Tish: If there is a dedicated screen for validations why not clean this screen from all error examples to keep it simple?

demo/src/screens/componentScreens/TextFieldScreen/InputValidationsScreen.js
```diff
@@ -0,0 +1,82 @@
+import React, {Component} from 'react';
+import {StyleSheet, ScrollView} from 'react-native';
+import {View, Text, TextField, Spacings} from 'react-native-ui-lib';
+import {Navigation} from 'react-native-navigation';
+import {KeyboardAwareInsetsView} from 'react-native-keyboard-tracking-view';
+
+class InputValidationsScreen extends Component {
```
Inbal-Tish: Don't you want to add an `useTopErrors` example? Maybe a toggle button to switch between the top and the bottom error like in the input screen?

demo/src/screens/componentScreens/TextFieldScreen/InputsScreen.js
ethanshar: yes, that's the plan.. that what I will do afterwards.. 
I just wanted to start simple and create the different screens without changing the current one.. I plan to clean the general screen (it has too many examples)

Inbal-Tish: Cool. So you want to merge it like that or after the cleanup?

ethanshar: Done. I'll make it more simple in the future, for now I move the error/validation examples

demo/src/screens/componentScreens/TextFieldScreen/InputValidationsScreen.js
ethanshar: done

PR #631, opened 2020-01-26
Fix TextField color by state logic

ethanshar:
We had some issues (only in public) with the TextField colors 
I went over it and decided to do a little refactor there (also refactor the tests) 

In general, the logic should make more sense now and follows closely are UX guidelines

### DIFF
--- a/src/components/inputs/TextField.js
+++ b/src/components/inputs/TextField.js
@@ -19,12 +19,14 @@ import Text from '../text';
 import TouchableOpacity from '../touchableOpacity';
 
 const DEFAULT_COLOR_BY_STATE = {
-  default: Colors.dark40,
-  focus: Colors.blue30,
-  error: Colors.red30
+  default: Colors.grey10,
+  focus: Colors.grey10,
+  error: Colors.grey10,
+  disable: Colors.grey50
 };
+
 const DEFAULT_UNDERLINE_COLOR_BY_STATE = {
-  default: Colors.dark70,
+  default: Colors.grey50,
   focus: Colors.blue30,
   error: Colors.red30
 };
@@ -52,7 +54,7 @@ export default class TextField extends BaseInput {
      */
     floatingPlaceholder: PropTypes.bool,
     /**
-     * floating placeholder color as a string or object of states, ex. {default: 'black', error: 'red', focus: 'blue'}
+     * floating placeholder color as a string or object of states, ex. {default: 'black', error: 'red', focus: 'blue', disable: 'grey'}
      */
     floatingPlaceholderColor: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
     /**
@@ -65,7 +67,7 @@ export default class TextField extends BaseInput {
      */
     hideUnderline: PropTypes.bool,
     /**
-     * underline color as a string or object of states, ex. {default: 'black', error: 'red', focus: 'blue'}
+     * underline color as a string or object of states, ex. {default: 'black', error: 'red', focus: 'blue', disable: 'grey'}
      */
     underlineColor: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
     /**
@@ -115,7 +117,7 @@ export default class TextField extends BaseInput {
      */
     title: PropTypes.string,
     /**
-     * The title's color as a string or object of states, ex. {default: 'black', error: 'red', focus: 'blue'}
+     * The title's color as a string or object of states, ex. {default: 'black', error: 'red', focus: 'blue', disable: 'grey'}
      */
     titleColor: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
     /**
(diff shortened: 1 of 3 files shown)


src/components/inputs/TextField.js
```diff
@@ -19,12 +19,14 @@ import Text from '../text';
 import TouchableOpacity from '../touchableOpacity';
 
 const DEFAULT_COLOR_BY_STATE = {
-  default: Colors.dark40,
-  focus: Colors.blue30,
-  error: Colors.red30
+  default: Colors.grey10,
+  focus: Colors.grey10,
+  error: Colors.grey10,
+  disable: Colors.grey50
```
Inbal-Tish: 'disabled' - you're missing the 'd'

src/components/inputs/TextField.js
```diff
@@ -330,7 +326,7 @@ export default class TextField extends BaseInput {
     const {expandable, placeholder, placeholderTextColor, floatingPlaceholderColor, multiline} 
     = this.getThemeProps();
     const typography = this.getTypography();
-    const placeholderColor = this.getStateColor(placeholderTextColor || DEFAULT_COLOR_BY_STATE.default);
+    const placeholderColor = this.getStateColor(placeholderTextColor || Colors.grey30);
```
Inbal-Tish: Why to hard-code the color and not use the DEFAULT_COLOR_BY_STATE.default (that should be set to 'grey30' instead of 'grey10')?

src/components/inputs/TextField.js
```diff
@@ -351,7 +347,7 @@ export default class TextField extends BaseInput {
               }),
               color: floatingPlaceholderState.interpolate({
                 inputRange: [0, 1],
-                outputRange: [placeholderColor, this.getStateColor(floatingPlaceholderColor)]
+                outputRange: [placeholderColor, this.getStateColor(floatingPlaceholderColor || {focus: Colors.blue30, default: Colors.grey30})]
```
Inbal-Tish: why to set all DEFAULT_COLOR_BY_STATE colors to 'grey10' if these are the defaults you want? Why to hard-code them here (and in line #368)?

src/components/inputs/TextField.js
ethanshar: I know, I tried to follow the same format you did with `focus`(and not `focused`)

ethanshar: I actually prefer disabled, I will change it

src/components/inputs/TextField.js
ethanshar: cause DEFAULT_COLOR_BY_STATE.default should be `grey10` and not `grey30`

src/components/inputs/TextField.js
ethanshar: That was part of the bug with the colors. 

The component assumed we have the same color states for placeholder and the input text color, but they're not the same according to the guidelines. 
The color states for the input text is different than the color states of the placeholder 

So basically, as I wrote in the previous comment `DEFAULT_COLOR_BY_STATE` represent the color states for the input text
and I pass a custom object for each part (placeholder, underline, etc..) according to our guidelines.

src/components/inputs/TextField.js
Inbal-Tish: So why not hold a DEFAULT_PLACEHOLDER_COLOR_BY_STATE enum (like the DEFAULT_UNDERLINE_COLOR_BY_STATE) and get it from there instead of changing in 3 places?

ethanshar: Cause technically the behavior is different. 
in line 329 I have 
`const placeholderColor = this.getStateColor(placeholderTextColor || Colors.grey30);`
in line 350 I have 
`outputRange: [placeholderColor, this.getStateColor(floatingPlaceholderColor || {focus: Colors.blue30, default: Colors.grey30})]`
which is different

and in line 500 I have this
`const placeholderColor = this.getStateColor(placeholderTextColor || Colors.grey30);`

The 2nd case is different than the other cause we want the color transition when the state changes from focused to not-focused. 
In the other cases (1+3), we don't want a different color for the different states. so I pass the `Colors.grey30`. If anything I can defined a single const for `DEFAULT_PLACEHOLDER_COLOR`

ethanshar: I've defined `DEFAULT_PLACEHOLDER_COLOR` const

src/components/inputs/TextField.js
```diff
@@ -351,7 +347,7 @@ export default class TextField extends BaseInput {
               }),
               color: floatingPlaceholderState.interpolate({
                 inputRange: [0, 1],
-                outputRange: [placeholderColor, this.getStateColor(floatingPlaceholderColor)]
+                outputRange: [placeholderColor, this.getStateColor(floatingPlaceholderColor || {focus: Colors.blue30, default: DEFAULT_PLACEHOLDER_COLOR})]
```
Inbal-Tish: I don't understand, why DEFAULT_PLACEHOLDER_COLOR is not an enum with states if you have 'focus' and 'default' cases?

src/components/inputs/TextField.js
```diff
@@ -369,7 +365,7 @@ export default class TextField extends BaseInput {
 
   renderTitle() {
     const {floatingPlaceholder, title, titleColor, titleStyle} = this.getThemeProps();
-    const color = this.getStateColor(titleColor);
+    const color = this.getStateColor(titleColor || {focus: Colors.blue30, default: Colors.grey30});
```
Inbal-Tish: Why not use DEFAULT_PLACEHOLDER_COLOR here as well?

src/components/inputs/TextField.js
ethanshar: done

src/components/inputs/TextField.js
ethanshar: done

PR #634, opened 2020-01-28
Fix/accss fixes

ArnonZ:
Switch AccessibilityRole fix

### DIFF
--- a/src/components/switch/index.js
+++ b/src/components/switch/index.js
@@ -82,13 +82,12 @@ class Switch extends BaseComponent {
 
   getAccessibilityProps() {
     const {accessibilityLabel, disabled, value} = this.getThemeProps();
-    const switchState = value ? 'on' : 'off';
 
     return {
       accessible: true,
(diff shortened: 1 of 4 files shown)


src/components/switch/index.js
```diff
@@ -81,14 +81,12 @@ class Switch extends BaseComponent {
   }
 
   getAccessibilityProps() {
-    const {accessibilityLabel, disabled, value} = this.getThemeProps();
-    const switchState = value ? 'on' : 'off';
+    const {disabled, value} = this.getThemeProps();
 
     return {
       accessible: true,
-      accessibilityLabel: accessibilityLabel ? `${accessibilityLabel} ${switchState}` : `switch ${switchState}`, //TODO: RN60 fix label and role and convert to accessibilityActions
```
drorbiran: @ethanshar @ArnonZ - This change broke one of our tests. We were counting on the accessibility label to check if toggle is on or off

ArnonZ: Is the prop `accessibilityStates` visible in Detox? if so please use it. It indicates:
`accessibilityStates:` disabled ? ['disabled'] : value ? ['checked'] : ['unchecked']`.
Otherwise, we need to find a creative solution

PR #639, opened 2020-02-02
Fix DateTimePicker placeholder, allow undefined as initial value

ethanshar:
There was an issue where the component couldn't have an undefined for the initial value so the placeholder never really worked. 

This is related to https://jira.wixpress.com/browse/WOAUILIB-432

### DIFF
--- a/src/components/dateTimePicker/index.js
+++ b/src/components/dateTimePicker/index.js
@@ -91,12 +91,12 @@ class DateTimePicker extends BaseComponent {
   constructor(props) {
     super(props);
 
-    const initialValue = props.value || new Date();
+    const initialValue = props.value;
     this.chosenDate = initialValue;
 
     this.state = {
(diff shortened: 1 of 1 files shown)


src/components/dateTimePicker/index.js
```diff
@@ -115,20 +115,35 @@ class DateTimePicker extends BaseComponent {
...
   };
 
   onDonePressed = () =>
     this.toggleExpandableOverlay(() => {
       _.invoke(this.props, 'onChange', this.chosenDate);
-      this.setState({chosenDate: this.chosenDate});
+      this.setState({value: this.chosenDate});
     });
 
+  getValue = () => {
```
Inbal-Tish: value is a Date object, I would call this method 'getStringValue()'.

ethanshar: done

PR #645, opened 2020-02-05
Fix/wizard minor fixes accessibility and export screens

M-i-k-e-l:
Wizard - minor fixes and add accessibility.
Export screens.


src/components/text/index.js
```diff
@@ -35,7 +35,11 @@ export default class Text extends PureBaseComponent {
     /**
      * Custom highlight style for highlight string
      */
-    highlightStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number])
+    highlightStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number]),
+    /**
+     * Turn off accessibility for this view and its nested children
+     */
+    inaccessible: PropTypes.bool
```
ethanshar: I thought we said we won't add it for now?

src/components/wizard/WizardStep.js
```diff
@@ -102,7 +129,13 @@ export default class WizardStep extends PureBaseComponent {
         {index > activeIndex && <View flex style={[styles.connector, connectorStyle]}/>}
         {this.renderCircle(props)}
         {index === activeIndex && (
-          <Text text80 testID={`${testID}.label`} numberOfLines={1} style={[styles.label, {maxWidth}, labelStyle]}>
+          <Text
+            text80
+            testID={`${testID}.label`}
+            numberOfLines={1}
+            style={[styles.label, {maxWidth}, labelStyle]}
+            inaccessible
```
ethanshar: Why aren't you using accessible={false}

src/components/text/index.js
M-i-k-e-l: Removed

src/components/wizard/WizardStep.js
M-i-k-e-l: Changed

PR #647, opened 2020-02-06
ExampleScreenPresenter - allow row in RadioGroup

M-i-k-e-l:


demo/src/screens/ExampleScreenPresenter.js
```diff
@@ -14,16 +14,26 @@ export function renderBooleanOption(title, key) {
   );
 }
 
-export function renderRadioGroup(title, key, options) {
+export function renderRadioGroup(title, key, options, isRow) {
```
ethanshar: can you make the last argument an object in case we want to pass other stuff in the future. 
```suggestion
export function renderRadioGroup(title, key, options, {isRow} = {}) {
```

PR #648, opened 2020-02-06
Fix signing

M-i-k-e-l:


Inbal-Tish:
@M-i-k-e-l  Why would you do that?

M-i-k-e-l:
@Inbal-Tish 
Did you try to install on a real device lately? - it fails for us.

Inbal-Tish:
I don't think this is the solution.

PR #649, opened 2020-02-07
Scrollbar features and fixes

ethanshar:
Add `focusIndex` prop for "scrollToSelected" capability
Add a dedicate example screen (so it's easy to test functionality)
Fix issue with fixed height, instead height should be determine by the content (as default)

*These changes are preparation so I can use the component in our TabController. 
Also the scrollToSelected functionality started to repeat itself as well.


src/components/scrollBar/index.js
```diff
@@ -126,40 +146,61 @@ class ScrollBar extends BaseComponent {
...
         horizontal
         showsHorizontalScrollIndicator={false}
         onScroll={this.onScroll}
         onContentSizeChange={this.onContentSizeChange}
-      />
+      >
+        {children &&
+          React.Children.map(children, (child, index) => {
+            return (
+              <Item onLayout={this.onItemLayout} index={index}>
```
Inbal-Tish: Why do we need to wrap the child with another View? Can't we just alter its props and return it?

demo/src/screens/componentScreens/ScrollBarScreen.js
```diff
@@ -0,0 +1,42 @@
+import React, {Component} from 'react';
+import {View, Text, ScrollBar, Button} from 'react-native-ui-lib';
+class ScrollBarScreen extends Component {
+  state = {
+    selectedIndex: 0
```
Inbal-Tish: 'focusIndex' doesn't work if I set the initial value to 6, for example, since you only apply it in ScrollBar's componentDidUpdate()

src/components/scrollBar/index.js
ethanshar: I wanted to avoid cloning, similar to what we did in the Carousel. 
Also, cloning might be problematic when I'll use it in TabController

demo/src/screens/componentScreens/ScrollBarScreen.js
ethanshar: you mean it doesn't work as an initial value, only changing value?

Inbal-Tish: Yes

src/components/scrollBar/index.js
Inbal-Tish: I meant something like: `
React.Children.map(children, (child, index) => {
      child.props.onLayout = this.onItemLayout;
      child.props.index = index;
      return child;
});
`
Meaning alter the child's props but return it without a wrapper view. Can it work?

demo/src/screens/componentScreens/ScrollBarScreen.js
ethanshar: fixed

src/components/scrollBar/index.js
ethanshar: you can do that, that's why there's the cloneElement api 
We'll have to actually clone the children in order to do that..

Inbal-Tish: ok

PR #650, opened 2020-02-09
Dialog - dismiss is not always called in accessibility mode

M-i-k-e-l:


src/components/dialog/index.js
```diff
@@ -105,10 +105,17 @@ class Dialog extends BaseComponent {
 
   componentDidMount() {
     Constants.addDimensionsEventListener(this.onOrientationChange);
+    AccessibilityInfo.addEventListener('screenReaderChanged', this.handleScreenReaderChanged);
```
ethanshar: Maybe we can add a listener in our `Constants` instead of the dialog component and expose a new constants. 
Since this is a mode that is alway turned on (for blind people) you can assume this is alway on or off. 
This will be helpful for this component and others that need a similar implementation.

M-i-k-e-l: Moved

PR #660, opened 2020-02-13
PageControl - accessibility - remove focus

M-i-k-e-l:


src/components/pageControl/index.js
```diff
@@ -206,8 +206,9 @@ class PageControl extends PureComponent {
     return (
       <View
         style={[styles.container, containerStyle]}
-        accessible
-        accessibilityLabel={`page control page ${this.props.currentPage + 1}`}
+        accessible={false}
+        accessibilityElementsHidden
```
ethanshar: why aren't you using our `inaccessible` prop we added for View

M-i-k-e-l: Fixed

PR #665, opened 2020-02-18
Feat/ExampleScreenPresenter - Small additions

ArnonZ:
Added: 
- Checkbox post-state update callback arguement 
- RadioGroup row flexWrap styling

### DIFF
--- a/demo/src/screens/ExampleScreenPresenter.js
+++ b/demo/src/screens/ExampleScreenPresenter.js
@@ -1,9 +1,9 @@
 import _ from 'lodash';
 import React from 'react';
 import {StyleSheet} from 'react-native';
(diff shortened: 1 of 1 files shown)


demo/src/screens/ExampleScreenPresenter.js
```diff
@@ -1,15 +1,22 @@
+import _ from 'lodash';
 import React from 'react';
+import {StyleSheet} from 'react-native';
 import {View, Text, Checkbox, RadioGroup, RadioButton, ColorPalette, Colors, Slider} from 'react-native-ui-lib';
-import _ from 'lodash';
 
-export function renderBooleanOption(title, key) {
+export function renderBooleanOption(title, key, stateCallback) {
```
ethanshar: let's see if we can avoid adding the stateCallback

PR #668, opened 2020-02-20
Feat/carousel allow selected card

M-i-k-e-l:


src/components/carousel/index.js
```diff
@@ -243,12 +268,20 @@ export default class Carousel extends BaseComponent {
...
   renderChild = (child, key) => {
     if (child) {
-      const paddingLeft = this.shouldUsePageWidth() ? this.props.itemSpacings : undefined;
+      const paddingLeft = this.shouldUsePageWidth() ? this.getItemSpacings(this.getThemeProps()) : undefined;
       const index = Number(key);
       const length = presenter.getChildrenLength(this.props);
+      const containerMarginHorizontal = this.getContainerMarginHorizontal();
+      const marginLeft = index === 0 ? containerMarginHorizontal : 0;
+      const marginRight = index === length - 1 ? containerMarginHorizontal : 0;
+      const paddingVertical = this.getContainerPaddingVertical();
```
ethanshar: Why not add the vertical padding to the ScrollView instead of each child?

M-i-k-e-l: Because overflow does not work on Android and it has to be done on the child.

ethanshar: I understand, but instead of passing each child the vertical padding you can pass to the Carousel's ScrollVIew the `contentContainerStyle` prop with the vertical padding once

M-i-k-e-l: If you mean `contentContainerStyle={[scrollContainerStyle, paddingVertical]}` then it does not work correctly - the `View` is still cut.

ethanshar: it should work, I played with a simple horizontal ScrollView and it works. 
The reason it doesn't work in Carousel seem to be related to the way we render the children. 
in `renderChild` 
each child wrapped with a View with `collapsable={false}`. 
removing the `collapsable={false}` fix it.

src/components/carousel/index.js
```diff
@@ -367,6 +395,7 @@ export default class Carousel extends BaseComponent {
           onContentSizeChange={this.onContentSizeChange}
           onScroll={this.onScroll}
           onMomentumScrollEnd={this.onMomentumScrollEnd}
+          overflow={'visible'}
```
ethanshar: I don't think overflow it's a prop but a style prop. 
you can remove this one

M-i-k-e-l: You are right, removed

src/components/carousel/index.js
M-i-k-e-l: Added a TODO

PR #669, opened 2020-02-20
Infra/keyboard tracking view allow package temp 2

M-i-k-e-l:


ethanshar:
@M-i-k-e-l 
Are you sure we need to rename all the files here.
In the private uilib everything works fine after the last PR merge.

M-i-k-e-l:
I think we need this so there's no conflict with react-native-keyboard-tracking-view from the engine.
I'll test to be sure.

PR #679, opened 2020-03-01
Pass avatar container ref

EdenGottlieb:
We need this to enable an image preview of an avatar


EdenGottlieb:
@ethanshar Is it OK now?

PR #686, opened 2020-03-06
ColorName tool

Inbal-Tish:
Adding ColorName tool (example in ColorPickerScreen) to find the name of the nearest color using a provided color name map.
Adding accessibility label to ColorSwatch component based on the ColorName tool.


src/style/index.js
```diff
@@ -25,5 +25,8 @@ module.exports = {
   },
   get AnimatableManager() {
     return require('./animatableManager').default;
+  },
+  get ColorName() {
```
ethanshar: What about keeping this class private and expose this functionality thru our Colors class.

src/components/colorPicker/ColorSwatch.js
```diff
@@ -143,6 +143,7 @@ export default class ColorSwatch extends PureBaseComponent {
         onPress={this.onPress}
         style={[this.styles.container, style]}
         onLayout={this.onLayout}
+        accessibilityLabel={ColorName.name(color)[1]}
```
ethanshar: can we simplify the functionality so it will return the color name instead of an array. 
I think the array is a little confusing as a return value, and since we're exporting this functionality I think it will be more clear if it would just return a string name

Inbal-Tish: The thing is that the color returned is not necessarily the original color so I think it's important to return the hex as well. The last value is an indication if the original color is part of the map and will allow adding colors to the map when we'll want this functionality in the future. If we go to the mask this in our Colors class we can return only the name string.

src/style/colorNameMap.js
```diff
@@ -0,0 +1,1585 @@
+const colorNameMap = [
```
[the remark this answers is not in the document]
Inbal-Tish: I copied it from 'name that color' library.

src/style/index.js
Inbal-Tish: I can do that.

src/style/colorNameMap.js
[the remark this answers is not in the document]
Inbal-Tish: credits added

src/style/index.js
Inbal-Tish: I added `getColorName` to Colors class but I didn't turn `ColorName` to private in case someone will need the full array returned from `name()` method.

src/components/colorPicker/ColorSwatch.js
Inbal-Tish: Moved example to use `Colors.getColorName(color)`

src/style/index.js
ethanshar: 👍

PR #690, opened 2020-03-10
Modal - change Android StatusBar color

M-i-k-e-l:


src/screensComponents/modal/index.js
```diff
@@ -1,12 +1,13 @@
...
 import PropTypes from 'prop-types';
 import React from 'react';
-import {StyleSheet, Modal as RNModal, TouchableWithoutFeedback} from 'react-native';
+import {StyleSheet, Modal as RNModal, TouchableWithoutFeedback, StatusBar} from 'react-native';
 import {BlurView} from '@react-native-community/blur';
 import {Constants} from '../../helpers';
 import {BaseComponent} from '../../commons';
 import TopBar from './TopBar';
 import View from '../../components/view';
+import Colors from '../../style/colors';
```
ethanshar:
```suggestion
import {Colors} from '../../style';
```

src/screensComponents/modal/index.js
```diff
@@ -58,17 +59,22 @@ export default class Modal extends BaseComponent {
...
 
     return (
-      <RNModal visible={Boolean(visible)} {...others}>
-        <Container style={{flex: 1}} blurType="light">
-          {this.renderTouchableOverlay()}
-          {this.props.children}
-        </Container>
-      </RNModal>
+      <View>
+        {Constants.isAndroid && Boolean(visible) && !!overlayBackgroundColor && (
```
ethanshar: There's an issue, after dismissing the modal the status bar color stays

M-i-k-e-l: It is a bit weird, it works fine for me in regular condition, but if I kill the application and opens it again it has this bug. Do you have an idea?

M-i-k-e-l: It seems that the <StatusBar/> component is buggy:
https://github.com/react-native-community/react-native-statusbar/issues/20
I'll upload a fix.

M-i-k-e-l: Two notes to consider:
1. I've changed the Modal to extend Component and not BaseComponent (I am using a new lifecycle method).
2. I'm using ThemeManager.primatyColor, it looks ok on both apps, but it might need further testing.

src/screensComponents/modal/index.js
```diff
@@ -37,6 +36,24 @@ export default class Modal extends BaseComponent {
...
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      visible: props.visible
+    };
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
```
ethanshar: Why not using `componentDidUpdate` for that?

src/screensComponents/modal/index.js
```diff
@@ -37,6 +36,24 @@ export default class Modal extends BaseComponent {
...
+    const {visible, overlayBackgroundColor} = nextProps;
+    if (Constants.isAndroid && !!overlayBackgroundColor && prevState.visible !== visible) {
+      const statusBarColor = visible ? Colors.getOpaqueColor(overlayBackgroundColor, 0.8) : ThemeManager.primaryColor;
```
ethanshar: I prefer to add a new prop since we might be able to add the theme logic for the color. 
by using `ThemeManager.primaryColor` we're always going to have a static color which is incorrect..

src/screensComponents/modal/index.js
```diff
@@ -1,13 +1,12 @@
 import _ from 'lodash';
```
ethanshar: a general comment on this whole issue..
I noticed that on a non-RNN project (our expo app for that matter) there's no issue. 
If you'll run the expo app on Android and open a dialog you will see that it covers the whole screen. 

I think that's the reason it was all connected to RNN. 
It's possible we need to find a way to fix it specifically for RNN.

M-i-k-e-l: If it only happens when RNN is there, then I think we should close this PR, and wait for RN's update - see our Jira ticket (595), or maybe open a ticket fir RNN.

ethanshar: I think the recent feature of RNN, that allows on Android to `drawBehind` status bar is the solution
Can u try setting that and see if it works. 
If it does we need to figure how to make the Modal set it automatically.

M-i-k-e-l: Isn't the `drawBehind` for the RNN's navigation bar? - I think Guy said it's a known bug, and we tried to find a workaround for it.

M-i-k-e-l:
We'll revisit this in RN62

Issue #561, opened 2019-11-16
TextField placeholder resizing multiple times onFocus and onBlur


maeldz:
<img src="https://user-images.githubusercontent.com/54718471/68986165-7cc9b100-07fb-11ea-98b3-45f2671e0ee2.gif" width="50%">

I am using the example TextField:

```
<TextField
             text70
             containerStyle = {{marginBottom: INPUT_SPACING}}
             floatingPlaceholder
             placeholder = "FloatingPlaceholder & validator"
             onChangeText = {this.onChangeText}
             validate = {'required'}
             errorMessage = {'Required field!'}
             useTopErrors = {this.state.topError}
             floatOnFocus
           />
```

PR #694, opened 2020-03-12
fix TextField floating placeholder animation

ethanshar:
- Fix issue with animation triggered every time value has changed
- Make the animation useNativeDriver for better performance


ethanshar:
fix #561

PR #695, opened 2020-03-14
Fix ConnectionStatusBar NetInfo

Peynn:
Since `NetInfo` moved from `react-native` to `@react-native-community/netinfo`, the component `ConnectionStatusBar` is not working anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@react-native-community/blur": "^3.4.1",
     "@react-native-community/datetimepicker": "^2.1.0",
     "@react-native-community/eslint-config": "latest",
+    "@react-native-community/netinfo": "^5.6.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
     "build:typings": "rm -rf ./typings && ./scripts/build-typings.sh && tsc --emitDeclarationOnly"
   },
   "dependencies": {
+    "@react-native-community/netinfo": "^5.6.2",
```
ethanshar: Can you please move the dependency to `devDependencies` and `peerDependencies`. 
As you can see we avoid installing native dependencies so users can decide on the version that fit them.

Peynn: Oh ! Done :+1:

PR #699, opened 2020-03-16
Carousel autoplay

yginton:


ethanshar:
@yginton The PR looks great ! 👍 

Just one comment, there's a use-case when the user manually scroll the page while the auto-play is on. 
In that case I think it's best to reset the autoPlay timer. 
currently, if the user scroll to a different page, if they did it a a second before the interval ends, the current page will immediately scroll again to the next page.

PR #709, opened 2020-03-19
TextField now triggers validation on `value` prop change if `validateOnChange` is set

ykravv:


src/components/inputs/TextField.js
```diff
@@ -181,7 +181,12 @@ export default class TextField extends BaseInput {
 
   UNSAFE_componentWillReceiveProps(nextProps) {
     if (nextProps.value !== this.props.value) {
-      this.setState({value: nextProps.value}, this.updateFloatingPlaceholderState);
+      this.setState({value: nextProps.value}, () => {
+        this.updateFloatingPlaceholderState();
+        if (nextProps.validateOnChange) {
+          this.validate();
+        }
```
ethanshar: I don't think there's a reason to re-validate only when passing the `validateOnChange` prop
Technically, the `validateOnChange` meant for typing. when a user programmatically changes the value of the field, I think we should alway trigger the validation. 

```suggestion
        this.validate();
```

ykravv: Makes sense

PR #717, opened 2020-03-23
create commons/new.ts for typescript migration

ethanshar:
in order to import with destruct all `commons` components I created a a separate file with the relevant components (that were already migrated to ts)


generatedTypes/commons/new.d.ts
```diff
@@ -0,0 +1,5 @@
+export { default as UIComponent } from './UIComponent';
```
Niryo: why do you need this file?

PR #720, opened 2020-03-25
Carousel - allow animating container

M-i-k-e-l:


src/components/carousel/index.js
```diff
@@ -441,14 +445,15 @@ export default class Carousel extends BaseComponent {
...
+    const {containerStyle, animate, ...others} = this.getThemeProps();
     const {initialOffset} = this.state;
     const scrollContainerStyle = this.shouldUsePageWidth() ? {paddingRight: this.getItemSpacings(this.getThemeProps())} : undefined;
     const snapToOffsets = this.getSnapToOffsets();
     const marginBottom = Math.max(0, this.getContainerPaddingVertical() - 16);
+    const Container = animate ? Animated.View : View;
 
     return (
-      <View style={[{marginBottom}, containerStyle]} onLayout={this.onContainerLayout}>
+      <Container style={[{marginBottom}, containerStyle]} onLayout={this.onContainerLayout}>
```
M-i-k-e-l: Should I always use Animated.View and remove the. new prop?

ethanshar: What is this change for exactly?

M-i-k-e-l: This is for our private #1065

src/components/carousel/index.js
```diff
@@ -60,6 +60,10 @@ export default class Carousel extends BaseComponent {
      * callback for onScroll event of the internal ScrollView
      */
     onScroll: PropTypes.func,
+    /**
+     * Should the container be animated (send the animation style via containerStyle)
+     */
+    animate: PropTypes.bool,
```
ethanshar: keep the name consistent with the prop we have for the `View` and `Text` components. 
```suggestion
    animated: PropTypes.bool,
```

src/components/carousel/index.js
```diff
@@ -441,14 +445,15 @@ export default class Carousel extends BaseComponent {
   }
 
   renderCarousel() {
-    const {containerStyle, ...others} = this.getThemeProps();
+    const {containerStyle, animate, ...others} = this.getThemeProps();
     const {initialOffset} = this.state;
     const scrollContainerStyle = this.shouldUsePageWidth() ? {paddingRight: this.getItemSpacings(this.getThemeProps())} : undefined;
     const snapToOffsets = this.getSnapToOffsets();
     const marginBottom = Math.max(0, this.getContainerPaddingVertical() - 16);
+    const Container = animate ? Animated.View : View;
```
ethanshar: Instead of the that you can just pass the `animated` prop to the `View` since it also supports this prop.

src/components/carousel/index.js
M-i-k-e-l: Fixed

src/components/carousel/index.js
M-i-k-e-l: Fixed

PR #721, opened 2020-03-26
Fix Android auto-link for KeyboardInput

M-i-k-e-l:


react-native.config.js
```diff
@@ -7,11 +7,13 @@ module.exports = {
       android: {
         sourceDir: './lib/android/',
         packageImportPath: `import com.wix.reactnativeuilib.highlighterview.HighlighterViewPackage;
+import com.wix.reactnativeuilib.keyboardinput.KeyboardInputPackage;
```
ethanshar: I actually removed this a while ago because it broke the Android build. 
you can see the commit here. 
https://github.com/wix/react-native-ui-lib/commit/ad099f86c245a112671eab032ccf304b8f34b8ea#diff-d62f334f16fa218e0d46106a63d89c17

It's my bad, I should have update you about this. 

I wrote you in the ticket it caused an issue, did u verify it works?

M-i-k-e-l: I remember, yes, I copied KeyboardInputViewScreen (and the dependencies) to a new project and it worked.

PR #722, opened 2020-03-27
Feat/drawer full left swipe

ethanshar:
WOAUILIB-522
Support full left swipe in `Drawer` component


src/components/drawer/index.js
```diff
@@ -68,7 +68,23 @@ class NewDrawer extends PureBaseComponent {
     /**
      * Perform the animation in natively
      */
-    useNativeAnimations: PropTypes.bool
+    useNativeAnimations: PropTypes.bool,
```
M-i-k-e-l, line 71: At first I thought this was added by mistake because we have `...Swipeable.PropTypes` at the top, but it looks like the `...Swipeable.PropTypes` should be removed, not sure if you want to do it in this PR.

ethanshar: removed

ethanshar:
@M-i-k-e-l 
The users can do whatever they want once the full swipe is completed. They have a callback for it. 
In the example screen I just show the recommended animation to use, but technically they can also remove the item completely.

PR #723, opened 2020-03-29
UILib Packages

ethanshar:
This PR include a script for generating custom, partial packages for uilib. 

**What does it mean?** 
It means the user will be able to import a single component in the following way
```
import View from 'react-native-ui-lib/view';
```

**Why is it important?** 
By not importing the whole library and only specific components you
1) Don't have to install peer native dependencies that you don't need (You won't get an error)
2) The total bundle size of the app will be reduced only to the component files you import. 
(Today uilib is 33% of bundle size of a new empty project)

- [x] Create 'react-native-ui-lib/core' package for basic components (View, Text, Image, TouchableOpacity, etc..)
- [x] Create 'react-native-ui-lib/keyboard' package for all Keyboard components we migrated from other repos (KeyboardTrackingView, KeyboardAwareInsetsView, etc...)
- [x] Create 'react-native-ui-lib/xxx' package per component to allow single component imports.


scripts/buildPackages.js
```diff
@@ -0,0 +1,47 @@
+const fs = require('fs');
+const _ = require('lodash');
+
+const packages = [
+  {
+    filename: 'keyboard.js',
+    content: `module.exports = require('./lib/components/Keyboard').default;\n`
```
Inbal-Tish, line 7: Why not use 'components: [keyboard]'?

Inbal-Tish:
Very cool!!!

scripts/buildPackages.js
ethanshar: Take a look at `./lib/components/Keyboard`
It already contains all the various Keyboard components we have. 
The user will be able to import it like this
```
import {KeyboardTrackingView, KeyboardAwareInsetsView, KeyboardRegistry, ...} from 'react-native-ui-lib/keyobard';
```

Inbal-Tish: I see. mmm... I still think it's clearer to use the same object format, seems less random. Your call.

PR #724, opened 2020-03-29
Fix/Badge - add hitSlop support to badge container

ArnonZ:
Some use cases of Badge require enlarging hit area.
Added specific support for `hitSlop` prop instead of spreading `{...others}` prop for it being the safer and more controlled option.


src/components/badge/index.js
```diff
@@ -212,6 +212,7 @@ export default class Badge extends PureBaseComponent {
       borderWidth,
       backgroundColor,
       containerStyle,
+      hitSlop,
```
ethanshar, line 215: please declare this prop in prop types

ArnonZ: Done

PR #729, opened 2020-04-01
Fix Carousel jerks in some cases (Jira 372)

M-i-k-e-l:
This was added at some point for "carousel page width".
I did not find an example that had a problem without these lines - perhaps one of the later commits fixed this issue?
Any chance you remember this? Or any specific cases we had problems with?


M-i-k-e-l:
> @M-i-k-e-l
> looking at this change commit message
> it says `fixes the initialPage for dynamic width`
> so my guess..
> if u have a carousel with an initialPage (I think that’s greater than zero)
> and with a page with a custom width (not full-width) there was an issue

I've re-tested this and it looks like it's working well with the change.

PR #730, opened 2020-04-01
Fix/TextField - Refactored TextField accessibility props logic

ArnonZ:
- Refactored TextField accessibility props logic


src/components/inputs/TextField.js
```diff
@@ -217,21 +217,21 @@ export default class TextField extends BaseInput {
...
-      accessibilityLabel = `${accessibilityLabel || ''}. ${value || ''}`;
+      accessibilityLabel = `${accessibilityLabel}`;
     }
 
-    const accessibilityStates = this.isDisabled() ? ['disabled'] : [];
     return {
       accessibilityLabel,
       // on Android accessibilityStates cause issues with expandable input
-      accessibilityStates: Constants.isIOS ? accessibilityStates : undefined
+      accessibilityState: Constants.isIOS ? accessibilityState : undefined
```
ethanshar, line 234: I think `accessibilityState` (singular) is relevant to the new RN version (rn62)
I think in rn61 they use `accessibilityStates` (plural) 
All other places we use in our code are plural. 
Did u check it works? 
Also why are u changing it.. I believe we have just fixed it recently to
`const accessibilityStates = this.isDisabled() ? ['disabled'] : [];`
on purpose, it solved a bug.

src/components/inputs/TextField.js
```diff
@@ -217,21 +217,21 @@ export default class TextField extends BaseInput {
...
 
   getAccessibilityInfo() {
-    const {floatingPlaceholder, placeholder, expandable, value} = this.getThemeProps();
+    const {floatingPlaceholder, placeholder, expandable} = this.getThemeProps();
+    const accessibilityState = this.isDisabled() ? {disabled: true} : undefined;
+    let accessibilityLabel = floatingPlaceholder ? this.props.accessibilityLabel || placeholder : '';
 
-    let accessibilityLabel = floatingPlaceholder ? placeholder : undefined;
     if (this.isRequiredField()) {
-      accessibilityLabel = `${accessibilityLabel || ''}. Mandatory`;
```
ethanshar, line 224: I think we did it on purpose ``${accessibilityLabel || ''}`, otherwise the reader will read `undefined` 
Why are u fixing it?

ArnonZ: it's added at the end of the ternary in line 222

src/components/inputs/TextField.js
ArnonZ: this: `const accessibilityStates = this.isDisabled() ? ['disabled'] : [];` doesn't work now...
Why would I change it if it doesn't work 🤷‍♂.

ethanshar: ok

src/components/inputs/TextField.js
```diff
@@ -217,21 +217,21 @@ export default class TextField extends BaseInput {
...
+      accessibilityLabel = `${accessibilityLabel}. Mandatory`;
     }
     if (expandable) {
-      accessibilityLabel = `${accessibilityLabel || ''}. ${value || ''}`;
+      accessibilityLabel = `${accessibilityLabel}`;
     }
```
ethanshar, line 229: isn't that become redundant ?
We're basically assigning a const to its own value..

ethanshar: @ArnonZ  ?

PR #744, opened 2020-04-12
Feat/dialog add on close with reason

M-i-k-e-l:


M-i-k-e-l:
> @M-i-k-e-l
> Do we really need to add the new 'onClose` prop? according to the code I see we might need to go thru a long migration here.
> 
> Correct me if I'm wrong, our problem is that `onModalDismiss` is triggered **only** for iOS?
> 
> If so, why not use the prop we already have (`onModalDismiss`) and as you did with `onClose` for Android just invoke the `onModalDismiss` instead.
> Shouldn't that be enough? Do we really need the new dismissal reasons?
> 
> the way I see it.
> `onDismiss` - is for when the dismiss is triggered, usually by changing the visible prop.
> `onModalDismiss` - is for when the modal has completely dismissed / dismiss animation has ended.

Yes, I know it will be an annoying migration, but I think this is the way to go.
No one actually needs two methods and `onModalDismissed` has the one behaviour that someone will want: "when can I do something in the UI, that will not cause other bugs?".
This is called `onDismiss` in RN-Modal, but unfortunately: 1. it supports only iOS. 2. We renamed it to `onModalDismissed`.
Because I think we would want, in the end, to have the name `onDismiss` as our API, we unfortunately need to do it in stages.
This is because if someone is currently using `onModalDismissed` only for iOS, and we change it right away, he might have a bug on Android.

My original thought was to do:
Migration 1: `onModalDismissed` --> `onClose`.
Migration 2: remove `onDismiss`.
Migration 3: `onClose` --> `onDismiss`.

We might be able to do - needs some (a lot of?) work:
Migration 1: `onModalDismissed` --> `onClose` & `onDismiss` --> `onClose`.
Migration 2: `onClose` --> `onDismiss`.

And I feel like this is too risky:
Migration 1: `onModalDismissed` --> (new) `onDismiss` & `onDismiss` --> (new) `onDismiss`.

What do you think?

PR #745, opened 2020-04-13
Carousel - fix various bugs

M-i-k-e-l:
Carousel - fix various bugs related to tablet, orientation change and Android's screen VS window size.

### DIFF
--- a/src/components/carousel/index.js
+++ b/src/components/carousel/index.js
@@ -116,8 +116,7 @@ export default class Carousel extends BaseComponent {
 
     const themeProps = this.getThemeProps();
     this.carousel = React.createRef();
-    const defaultPageWidth = themeProps.loop ?
-      Constants.screenWidth : themeProps.pageWidth + this.getItemSpacings(themeProps) || Constants.screenWidth;
+    const defaultPageWidth = (themeProps.loop || !themeProps.pageWidth) ? Constants.screenWidth : themeProps.pageWidth;
     
     this.state = {
       containerWidth: undefined,
--- a/src/components/carousel/index.js
+++ b/src/components/carousel/index.js
@@ -114,23 +114,24 @@ export default class Carousel extends BaseComponent {
   constructor(props) {
     super(props);
 
+    const themeProps = this.getThemeProps();
     this.carousel = React.createRef();
-    const defaultPageWidth = props.loop ? 
-      Constants.screenWidth : props.pageWidth + this.getItemSpacings(props) || Constants.screenWidth;
+    const defaultPageWidth = themeProps.loop ?
+      Constants.screenWidth : themeProps.pageWidth + this.getItemSpacings(themeProps) || Constants.screenWidth;
     
     this.state = {
       containerWidth: undefined,
-      currentPage: this.shouldUsePageWidth() ? this.getCalcIndex(props.initialPage) : props.initialPage,
-      currentStandingPage: props.initialPage,
+      currentPage: this.shouldUsePageWidth() ? this.getCalcIndex(themeProps.initialPage) : themeProps.initialPage,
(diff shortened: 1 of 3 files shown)


src/components/carousel/index.js
```diff
@@ -151,7 +152,10 @@ export default class Carousel extends BaseComponent {
...
   onOrientationChanged = () => {
     const {pageWidth, loop} = this.getThemeProps();
-    if (!pageWidth || loop) {
+    if (pageWidth) {
+      this.orientationChange = true;
+      this.setState({pageWidth});
+    } else if (!pageWidth || loop) {
       this.orientationChange = true;
       // HACK: setting to containerWidth for Android's call when view disappear
       this.setState({pageWidth: this.state.containerWidth || Constants.screenWidth});
```
ethanshar, line 161: something with the condition here seem off.
first of all `this.orientationChange = true;` always being set to true, so not sure why it's inside the condition twice. 
second.. the `... || loop` seem redundant now. 
if `pageWidth` is defined it will go to the first condition, otherwise it will always enter the second condition without even checking the `loop` value.. 

maybe we're missing something here?

src/components/carousel/index.js
```diff
@@ -261,7 +265,13 @@ export default class Carousel extends BaseComponent {
       update.initialOffset = {
         x: presenter.calcOffset(this.getThemeProps(), {currentPage: this.state.currentPage, pageWidth: containerWidth})
       };
+    } else {
+      update.pageWidth = pageWidth;
+      update.initialOffset = {
+        x: presenter.calcOffset(this.getThemeProps(), {currentPage: this.state.currentPage, pageWidth})
+      };
     }
+
```
ethanshar, line 274: Is that necessary? What is it for?
The `onContainerLayout` was only meant for full-width pages
when the user passes his own `pageWidth` it's technically redundant. 
also in this case we already initializing the `state.pageWidth` in the constructor (with a little different value, btw).

src/components/carousel/index.js
M-i-k-e-l: I'm not sure what this fixed, but it seems it's not needed anymore - removed

src/components/carousel/index.js
M-i-k-e-l: This fixes the offset and width when orientation changes (maybe this is why the `onOrientationChanged` code is redundant).
The width changes because the screen's width changes.
I'm not sure why the code in the constructor is that way (specifically the ` + this.getItemSpacings(themeProps)`), but I'm not sure if I should change it.

ethanshar: but it's just because that in the example screen you decided to change pageWidth when orientation change, no? 
It doesn't necessarily relevant to the orientation change more than it's relevant to the pageWidth change. 
I think it will make more sense to handle this case when the user changes the pageWidth. WDYT?

M-i-k-e-l: I see your point, but I tried to move the code to `UNSAFE_componentWillReceiveProps` and it actually had a bug when the initial page was set to something other than zero.
I'm thinking it might need both, or just the `onContainerLayout` if we want to handle just the orientation change and not actual width change. WDYT?

ethanshar: Why using `UNSAFE_componentWillReceiveProps`? 
why not using `getDerivedStateFromProps` to save the new state values for `pageWidth` and `initialOffset`?

M-i-k-e-l: Because that would require changing from `extends BaseComponent` to `asBaseComponent`, not sure I want to do that in this PR.

ethanshar: ok.. acceptable.. 
So let's just add a TODO there..

M-i-k-e-l: Done

PR #749, opened 2020-04-20
Added validation error feature to TagsInput

guywix:
added new onTagNotValidate prop for the user to write is validation error message, and UI support as asked here: [https://app.zeplin.io/project/5934252871cf6fc586d47aeb/screen/5e41125dfaf8a8949a886bf8](url)

### DIFF
--- a/src/components/tagsInput/index.js
+++ b/src/components/tagsInput/index.js
@@ -30,9 +30,9 @@ export default class TagsInput extends BaseComponent {
 
   static propTypes = {
     /**
-     * list of tags. can be string or custom object when implementing getLabel
+     * list of tags. can be string boolean or custom object when implementing getLabel
      */
-    tags: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.object, PropTypes.string])),
+    tags: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.object, PropTypes.string, PropTypes.bool])),
     /**
      * callback for extracting the label out of the tag item
      */
@@ -53,6 +53,10 @@ export default class TagsInput extends BaseComponent {
      * callback for when pressing a tag in the following format (tagIndex, markedTagIndex) => {...}
      */
     onTagPress: PropTypes.func,
+    /**
+     * validation message error appears when tag isn't validate
+     */
+    onTagNotValidate: PropTypes.string,
     /**
      * if true, tags *removal* Ux won't be available
      */
@@ -98,11 +102,14 @@ export default class TagsInput extends BaseComponent {
     this.state = {
       value: props.value,
       tags: _.cloneDeep(props.tags) || [],
-      tagIndexToRemove: undefined
+      tagIndexToRemove: undefined,
+      isTagNotValidate: false,
     };
   }
 
   componentDidMount() {
+    const {tags} = this.state;
+    this.updateInvalidTags(tags);
     if (Constants.isAndroid) {
       const textInputHandle = ReactNative.findNodeHandle(this.input);
       if (textInputHandle && NativeModules.TextInputDelKeyHandler) {
@@ -140,6 +147,7 @@ export default class TagsInput extends BaseComponent {
     const newTag = _.isFunction(onCreateTag) ? onCreateTag(value) : value;
     const newTags = [...tags, newTag];
 
+    this.updateInvalidTags(newTags);
     this.setState({
       value: '',
       tags: newTags
@@ -155,6 +163,7 @@ export default class TagsInput extends BaseComponent {
       const removedTag = tags[tagIndexToRemove];
 
       tags.splice(tagIndexToRemove, 1);
+      this.updateInvalidTags(tags);
       this.setState({
         tags,
         tagIndexToRemove: undefined
@@ -198,6 +207,12 @@ export default class TagsInput extends BaseComponent {
     return isLastTagMarked;
   }
 
+  updateInvalidTags(tags) {
+    _.filter(tags, (tag) => {return tag.invalid}).length > 0 ?
+    this.setState({isTagNotValidate: true}) :
+    this.setState({isTagNotValidate: false});
+  }
+
   onKeyPress(event) {
     _.invoke(this.props, 'onKeyPress', event);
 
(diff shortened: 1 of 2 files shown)


src/components/tagsInput/index.js
```diff
@@ -198,6 +207,14 @@ export default class TagsInput extends BaseComponent {
     return isLastTagMarked;
   }
 
+  updateInvalidTags(tags) {
```
ethanshar, line 210: instead of going through all of the tags to search if there's an invalid tag. 
You can determine if you are in an invalid state by checking if the user passed the error message prop. (what you call `onTagNotValidate`) 
You don't even need to save the `isTagNotValidate` in your state, you can base everything on the prop the user pass.

src/components/tagsInput/index.js
```diff
@@ -310,16 +351,21 @@ export default class TagsInput extends BaseComponent {
...
-    const {tags} = this.state;
+    const {tags, isTagNotValidate, tagIndexToRemove} = this.state;
 
     return (
       <View style={[!hideUnderline && styles.withUnderline, containerStyle]}>
         <View style={styles.tagsList}>
           {_.map(tags, tagRenderFn)}
           {this.renderTextInput()}
         </View>
+        <View>
```
ethanshar, line 364: I think the error message should appear under the underline, in the design they don't show it. but better check with the UX to make sure.

src/components/tagsInput/index.js
```diff
@@ -30,9 +30,9 @@ export default class TagsInput extends BaseComponent {
 
   static propTypes = {
     /**
-     * list of tags. can be string or custom object when implementing getLabel
+     * list of tags. can be string boolean or custom object when implementing getLabel
      */
-    tags: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.object, PropTypes.string])),
+    tags: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.object, PropTypes.string, PropTypes.bool])),
```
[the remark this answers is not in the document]
guywix: ok, removed

src/components/tagsInput/index.js
```diff
@@ -53,6 +53,10 @@ export default class TagsInput extends BaseComponent {
      * callback for when pressing a tag in the following format (tagIndex, markedTagIndex) => {...}
      */
     onTagPress: PropTypes.func,
+    /**
+     * validation message error appears when tag isn't validate
+     */
+    onTagNotValidate: PropTypes.string,
```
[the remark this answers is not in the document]
guywix: changed

src/components/tagsInput/index.js
guywix: but what if the user is passing the error message , but not passing an invalid tag in tags,
an error message shouldn't appear ?

src/components/tagsInput/index.js
guywix: that 's how it shows in the zeplin, who exactly to check with ?

ethanshar: talk to Or Gertman

src/components/tagsInput/index.js
ethanshar: It should.. 
the `validationErrorMessage` controls the error message
the `tag.invalid` controls the tag error UI

if the user didn't pass both it's his problem 
Also, it's possible we want to show an error message without marking any tag as invalid .
for example, if I allow only 5 tags and user add 6 tag. 
In that case I will **only** pass the error message without marking any tag as invalid.

guywix: ok , done

src/components/tagsInput/index.js
guywix: asked Or and she said its supposed to be above

ethanshar: 👍

src/components/tagsInput/index.js
```diff
@@ -236,24 +239,48 @@ export default class TagsInput extends BaseComponent {
     return _.get(item, 'label');
   }
 
-  renderLabel(tag, shouldMarkTag) {
+  checkTagValidate(tag, validationErrorMessage) {
+    return validationErrorMessage ? tag.invalid : false;
+  }
+
+  renderLabel(tag, shouldMarkTag, isTagNotValidate) {
```
ethanshar, line 246: also here. 
you don't need the `isTagNotValidate` argument. 
you can use the `tag.invalid` instead

src/components/tagsInput/index.js
```diff
@@ -236,24 +239,48 @@ export default class TagsInput extends BaseComponent {
...
+          <Image
+            style={[isTagNotValidate ? styles.inValidTagRemoveIcon : styles.removeIcon]}
+            source={Assets.icons.x}
+          />)
+        }
+        <Text
+          style={[isTagNotValidate ? (shouldMarkTag ? styles.inValidMarkedTagLabel : styles.inValidTagLabel)
+            : styles.tagLabel, typography]} accessibilityLabel={`${label} tag`}
+        >
+          {isTagNotValidate ? label : (shouldMarkTag ? 'Remove' : label)}
```
ethanshar, line 262: you can simplify the ternary condition
```suggestion
          {shouldMarkTag && !isTagNotValidate ? 'Remove' : label}
```

src/components/tagsInput/index.js
```diff
@@ -310,16 +337,21 @@ export default class TagsInput extends BaseComponent {
...
+    const {tags, tagIndexToRemove} = this.state;
 
     return (
       <View style={[!hideUnderline && styles.withUnderline, containerStyle]}>
         <View style={styles.tagsList}>
           {_.map(tags, tagRenderFn)}
           {this.renderTextInput()}
         </View>
+        <View>
+          <Text style={[validationErrorMessage && tagIndexToRemove ? styles.inValidMarkedTagLabel : styles.inValidTagLabel]}>
```
ethanshar, line 351: regarding the the style 
we usually have a base style for the element and we only condition the change. What does it mean? 
In this case I would have a base style for the error message, for example: `styles.errorMessage`, and it will contain the base props, in our case it's 
```
errorMessage: {
  ...Typography.text80,
  color: Colors.red30
}
```
Now, I will add a style for the "markedTagLabel" case, and it will contain only the difference 
```
errorMessageWhileMarked: {
  color: Colors.red10
}
```
And the way we assign the style is like this
```
<Text style={[styles.errorMessage, tagIndexToRemove && styles. errorMessageWhileMarked]}
```
The reason for that is so we don't have duplicates of styles (like the Typography.text80)
if the design changes in the future, we change it only at one place and not several places.

src/components/tagsInput/index.js
```diff
@@ -310,16 +337,21 @@ export default class TagsInput extends BaseComponent {
...
-    const {tags} = this.state;
+    const {tags, tagIndexToRemove} = this.state;
 
     return (
       <View style={[!hideUnderline && styles.withUnderline, containerStyle]}>
         <View style={styles.tagsList}>
           {_.map(tags, tagRenderFn)}
           {this.renderTextInput()}
         </View>
+        <View>
```
ethanshar, line 350: add condition to the error message so it will render it only if the validationErrorMessage is not undefined

src/components/tagsInput/index.js
```diff
@@ -236,24 +239,48 @@ export default class TagsInput extends BaseComponent {
...
   }
 
   renderTag(tag, index) {
-    const {tagStyle, renderTag} = this.getThemeProps();
+    const {tagStyle, renderTag, validationErrorMessage} = this.getThemeProps();
     const {tagIndexToRemove} = this.state;
     const shouldMarkTag = tagIndexToRemove === index;
+    const isTagNotValidate = this.checkTagValidate(tag, validationErrorMessage);
+
+    if (isTagNotValidate) {
```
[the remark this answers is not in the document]
guywix: done

src/components/tagsInput/index.js
guywix: changed

src/components/tagsInput/index.js
guywix: done

src/components/tagsInput/index.js
guywix: done , did it also for the invalid tag & removeIcon styles

src/components/tagsInput/index.js
guywix: done, not sure if that's what you meant

PR #750, opened 2020-04-21
Fix/custom modifiers typings

ethanshar:
As discussed.. 
I made some changes, please take a look. 
I think it looks pretty ok.


Niryo:
looks good

PR #755, opened 2020-04-25
Fix/dialog on dismiss

ethanshar:
-  Revert recent changes with onDismiss functionality 
- Add parity to Android with Modal's onDismiss callback that was supported only for iOS

### DIFF
--- a/src/components/dialog/index.js
+++ b/src/components/dialog/index.js
@@ -99,7 +99,9 @@ class Dialog extends BaseComponent {
 
     this.setAlignment();
 
-    LogService.deprecationWarn({component: 'Dialog', oldProp: 'onModalDismissed', newProp: 'onDialogDismissed'});
+    if (!_.isUndefined(props.onModalDismissed)) {
+      LogService.deprecationWarn({component: 'Dialog', oldProp: 'onModalDismissed', newProp: 'onDialogDismissed'});
+    }
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/components/dialog/index.js
```diff
@@ -137,7 +137,13 @@ class Dialog extends BaseComponent {
   onDismiss = () => {
     this.setState({modalVisibility: false}, () => {
       const props = this.getThemeProps();
-      _.invoke(props, 'onDismiss', props);
+      if (props.visible) {
+        _.invoke(props, 'onDismiss', props);
+      }
+      // Parity with iOS Modal's onDismiss
+      if (Constants.isAndroid) {
+        _.invoke(props, 'onModalDismissed', props);
```
M-i-k-e-l, line 145: I feel this is risky, a user could have counted on this *not* working on Android, and now it will, this is a big potential for bugs.

ethanshar: as discussed, I added a new prop - `onDialogDismissed`

src/components/dialog/index.js
```diff
@@ -59,9 +59,13 @@ class Dialog extends BaseComponent {
      */
     useSafeArea: PropTypes.bool,
     /**
-     * Called once the modal has been dissmissed (iOS only)
+     * Called once the modal has been dismissed (iOS only) - Deprecated, use onDialogDismissed instead
```
M-i-k-e-l, line 62: Do we want to add a warning?

ethanshar: done ✔️

Issue #759, opened 2020-04-29
TypeScript - RadioButton value prop typing error


girishjjain:
I see one typing error with latest version (v5.7.0) of library, all other have resolved. Here's the error with `RadioButton`'s `value` property:
```
error TS2769: No overload matches this call.
  Overload 1 of 2, '(props: Readonly<RadioGroupChild<RadioButtonProps> & Partial<Record<"fixme", boolean>> & Partial<Record<AlignmentLiterals, boolean>> & Partial<...> & Partial<...> & Partial<...>>): RadioButton', gave the following error.
    Type '{ value: string; label: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<RadioButton> & Readonly<RadioGroupChild<RadioButtonProps> & ... 4 more ... & Partial<...>> & Readonly<...>'.
      Property 'value' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<RadioButton> & Readonly<RadioGroupChild<RadioButtonProps> & ... 4 more ... & Partial<...>> & Readonly<...>'.
  Overload 2 of 2, '(props: RadioGroupChild<RadioButtonProps> & Partial<Record<"fixme", boolean>> & Partial<Record<AlignmentLiterals, boolean>> & Partial<...> & Partial<...> & Partial<...>, context?: any): RadioButton', gave the following error.
    Type '{ value: string; label: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<RadioButton> & Readonly<RadioGroupChild<RadioButtonProps> & ... 4 more ... & Partial<...>> & Readonly<...>'.
      Property 'value' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<RadioButton> & Readonly<RadioGroupChild<RadioButtonProps> & ... 4 more ... & Partial<...>> & Readonly<...>'.

174                 <RadioButton value={Strings.MonthsString} label={'Months'} />
```

PR #766, opened 2020-05-05
Typescript - RadioButton and RadioGroup

M-i-k-e-l:
I'm adding some comments in the PR where there's stuff that I'm unsure how to solve and if it should be changed.

Hopefully fixes https://github.com/wix/react-native-ui-lib/issues/759


src/components/radioButton/RadioButton.tsx
```diff
@@ -140,21 +107,8 @@ class RadioButton extends BaseComponent {
...
   };
 
-  getAccessibilityProps = () => {
-    const {label = '', selected, disabled} = this.getThemeProps();
-    const selectedAccessibilityText = selected ? 'selected' : 'unselected';
-    const accessibilityLabel = `${selectedAccessibilityText}. ${label}`;
-
-    return {
-      accessible: true,
-      accessibilityStates: disabled ? ['disabled'] : undefined,
```
M-i-k-e-l, line 150: I was unable to have this method with accessibilityStates & accessibilityRole - it had errors of unrecognised types that I did not find a solution to.
I've created `getSomeAccessibilityProps` a bit lower in the code to at least avoid too much code in the render.

src/components/radioButton/RadioButton.tsx
```diff
@@ -216,7 +181,9 @@ class RadioButton extends BaseComponent {
         {...others}
         style={undefined}
         onPress={this.onPress}
-        {...this.getAccessibilityProps()}
+        accessibilityStates={disabled ? ['disabled'] : undefined}
+        accessibilityRole={'button'} // 'radio', TODO: uncomment when switching to RN60
```
M-i-k-e-l, line 185: Technically we could change this to 'radio', are users with the current version should all be using RN61?

src/components/radioButton/asRadioGroupChild.tsx
```diff
@@ -1,20 +1,9 @@
...
-       * The identifier value of the radio button. must be different than other RadioButtons in the same group
-       */
-      value: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
-      /**
-       * When using RadioButton without a RadioGroup, use this prop to toggle selection
-       */
-      selected: PropTypes.bool
-    };
-
+export default function asRadioGroupChild(WrappedComponent: React.ComponentType<any>) {
```
M-i-k-e-l, line 5: I'm not sure this can be changed to something other than `React.ComponentType<any>` (and `any` below), it might be possible to make it `RadioButton` if we unify the files.

src/components/radioButton/RadioButton.tsx
```diff
@@ -1,81 +1,108 @@
...
+   */
+  label?: string;
+  /**
+   * Label style
+   */
+  labelStyle?: TextStyle;
+  /**
+   * Icon image source
+   */
+  iconSource?: object | number;
```
ethanshar, line 65: Use `ImageSourcePropType` type

src/components/radioButton/RadioButton.tsx
ethanshar: I prefer to have `//@ts-ignore` above it instead of changing the code

src/components/radioButton/RadioButton.tsx
ethanshar: I prefer to separate non typescript related changes from this PR. 
The todo is enough for now

src/components/radioButton/asRadioGroupChild.tsx
ethanshar: that's fine, those internal typings are less important because they not what we expose to our users

src/components/radioButton/RadioGroup.tsx
```diff
@@ -0,0 +1,65 @@
+import _ from 'lodash';
```
ethanshar, line 1: For next migration, make sure to not delete and create new file but to make sure git just rename the files 
(similar to how u did for `RadioButton` file)
It's important for git log history of the files

src/components/radioButton/RadioButton.tsx
```diff
@@ -1,81 +1,108 @@
...
+  ViewStyle,
+  ViewProps
+} from 'react-native';
 import {Colors} from '../../style';
-import {BaseComponent} from '../../commons';
+import {asBaseComponent, forwardRef, BaseComponentInjectedProps, ForwardRefInjectedProps} from '../../commons/new';
 import TouchableOpacity from '../touchableOpacity';
 import View from '../view';
 import Text from '../text';
-import Image from '../image';
```
ethanshar, line 10: I think u can import as usual but just add `//@ts-ignore`

src/components/radioButton/RadioGroup.tsx
M-i-k-e-l: I reverted and renamed, but I think it just looks different in GitHub, I see the history locally, can you verify you do to?

src/components/radioButton/RadioButton.tsx
M-i-k-e-l: `import Image from '../Image';` gives an error

src/components/radioButton/RadioButton.tsx
M-i-k-e-l: Fixed.

src/components/radioButton/RadioButton.tsx
M-i-k-e-l: Reverted

src/components/radioButton/RadioButton.tsx
M-i-k-e-l: ok

PR #769, opened 2020-05-11
Lint - fix undefined bug when importing all (*)

M-i-k-e-l:


eslint-rules/lib/rules/typography-deprecation.js
```diff
@@ -47,11 +47,15 @@ module.exports = {
       if (source === importSource) {
         const specifiers = node.specifiers;
         if (specifiers) {
```
ethanshar, line 49: if this code repeat itself maybe create a util function (in `utils.js`)

M-i-k-e-l: Done

PR #775, opened 2020-05-14
Typescript/image

ethanshar:
-  Migrate `Image` component to typescript
- Migrate `Overlay` component to typescript


ethanshar:
@ArnonZ ping?

src/components/image/index.tsx
```diff
@@ -158,4 +167,5 @@ const styles = StyleSheet.create({
 });
 
 hoistNonReactStatic(Image, RNImage);
-export default Image;
+export {Image};
+export default asBaseComponent<ImageProps>(forwardRef(Image));
```
ArnonZ, line 171: Don't you want to declare that you receive the `Props` type and not only the `ImageProps`?

src/components/image/index.tsx
```diff
@@ -1,75 +1,80 @@
...
+   * Pass a custom color for the overlay
+   */
+  overlayColor?: string;
+  /**
+   * Render an overlay with custom content
+   */
+  customOverlayContent?: JSX.Element;
+};
+
+type Props = ImageProps & ForwardRefInjectedProps;
```
ArnonZ, line 56: Since you spread `others` in the render block shouldn't your Props include `ViewProps`. and if so export should change to `export default asBaseComponent<Props>(forwardRef(Image));`

PR #782, opened 2020-05-18
ColorSliderGroup - allow sending translation to the labels

M-i-k-e-l:


src/components/slider/ColorSliderGroup.js
```diff
@@ -32,9 +32,14 @@ export default class ColorSliderGroup extends PureBaseComponent {
...
     /**
-     * Show the sliders labels (Hue, Lightness, Saturation)
+     * Show the sliders labels (defaults are: Hue, Lightness, Saturation)
      */
     showLabels: PropTypes.bool,
+    /**
+     * In case you would like to change the default labels (translations etc.), you can provide
+     * a function that received the type (one of GradientSlider.types without DEFAULT) and returns a string.
+     */
+    getLabel: PropTypes.func,
```
ethanshar, line 42: Wouldn't be simpler to allow passing a map of labels
`{hue: ..., lightness:..., Saturation: ..}` ?
You can set here the default map in default props
in our private config you can override the defaultProps (**not** using ThemeManager)

M-i-k-e-l: I am not sure what the best solution.
In your example you have {**h**ue: ..., **l**ightness:..., **S**aturation: ..} - lower case 'h', lower case 'l' and upper case 's' (which will cause something not to work), that is less likely to happen if you are using `types` (and I can't set the key to that in a simple readable way).
I started with:
`labels = {};`
`labels[GradientSlider.types.HUE] = 'Hue';`
`...`
But that seemed less clean.

What do you think?

ethanshar: This is a very small feature and I think it's best to got with the minimum changes possible. 

In my example I meant everything to be lower case, it's clearly a mistake. 
I don't see any reason to add types and extra function prop just for that. 
I would usually use functions for dynamic values, the title texts are static, you set them once and don't need to touch them again

M-i-k-e-l: Changed to defaultProps

src/components/slider/ColorSliderGroup.js
```diff
@@ -62,18 +71,29 @@ export default class ColorSliderGroup extends PureBaseComponent {
     _.invoke(this.props, 'onValueChange', value);
   }
 
+  getSlider = (type) => {
```
ethanshar, line 74: I think it's better to rename the  function renderSlider like we usually do

M-i-k-e-l: Fixed :S

PR #785, opened 2020-05-19
Fix/Picker + TextField Ui fixes

ArnonZ:
- Removed unneeded Picker value default typography. TextField defaults to that same typography.
In addition, this is a blocker for (WOAUILIB-816).
- Updated TextField's rightIcon bottom padding condition
- Updated DateTimePicker example screen.

### DIFF
--- a/demo/src/screens/componentScreens/DateTimePickerScreen.js
+++ b/demo/src/screens/componentScreens/DateTimePickerScreen.js
@@ -3,24 +3,38 @@ import {DateTimePicker, View, Text} from 'react-native-ui-lib'; // eslint-disabl
 
 
 export default class DateTimePickerScreen extends Component {
-  
+
   render() {
     return (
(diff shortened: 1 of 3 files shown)


demo/src/screens/componentScreens/DateTimePickerScreen.js
```diff
@@ -1,27 +1,42 @@
...
-  
   render() {
     return (
-      <View flex padding-20>
-        <Text text40 marginB-40>Date Time Picker</Text>
-        
-        <DateTimePicker 
-          title={'Date'} 
-          placeholder={'Select a date'} 
+      <View flex padding-page>
```
ethanshar, line 7: public repo doesn't have the `page` spacing 
so the example screen will look good in our private environment but in the OSS :( 
```suggestion
      <View flex padding-s5>
```

PR #789, opened 2020-05-21
Infra/TS - Checkbox

ArnonZ:


src/components/checkbox/index.tsx
```diff
@@ -1,64 +1,98 @@
...
+   */
+  selectedIcon?: number;
+  /**
+   * The selected icon color
+   */
+  iconColor?: string;
+  /**
+   * Additional styling
+   */
+  style?: StyleProp<ViewStyle>
```
ArnonZ, line 51: Didn't see any other ts including this type but without it example screen shouts:
`Type '{ value: boolean; onValueChange: (value2: any) => void; borderRadius: number; size: number; color: string; selectedIcon: number; style: { marginBottom: number; }; }' is not assignable to type 'IntrinsicAttributes & CheckboxProps & { children?: ReactNode; }'.
  Property 'style' does not exist on type 'IntrinsicAttributes & CheckboxProps & { children?: ReactNode; }'.ts(2322)`

Is there another solution for this?

src/components/checkbox/index.tsx
```diff
@@ -1,64 +1,98 @@
...
+}
+
+type CheckboxState = {
+  isChecked: Animated.Value;
+};
+
+type Props = CheckboxProps & BaseComponentInjectedProps & TouchableOpacityProps;
```
ArnonZ, line 58: Not a 100% sure about this.

src/components/checkbox/index.tsx
```diff
@@ -177,4 +209,4 @@ function createStyles({color = DEFAULT_COLOR, iconColor = DEFAULT_ICON_COLOR, si
   });
 }
 
-export default Checkbox;
+export default asBaseComponent<CheckboxProps>(forwardRef(Checkbox));
```
ArnonZ, line 212: and this...In general, the whole use of the offered  HOCs from `'../../commons/new'` is not crystal clear to me when it comes to knowing what to use and when :)

PR #794, opened 2020-05-26
Add asScrollableContent

M-i-k-e-l:


src/commons/asScrollableContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+//@ts-ignore
+import hoistStatics from 'hoist-non-react-statics';
+
+interface ScrollableContentProps {
+  scrollableContentProps?: ScrollViewProps;
+  maximumContentSize?: number;
+}
+
+interface ScrollableContentState {
+  maximumContentSize?: number;
```
ethanshar, line 13: Why should the user need to pass the `maximumContentSize`? 
The whole idea is to measure (using onLayout) the content size vs the container size and do the whole work

src/commons/asScrollableContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+export default function asScrollableContent(WrappedComponent: React.ComponentType<any>) {
+  class ScrollableContent extends Component<PropTypes, ScrollableContentState> {
+    static displayName: string | undefined;
+
+    constructor(props: PropTypes) {
+      super(props);
+      this.state = {
+        maximumContentSize: props.maximumContentSize,
+        contentSize: undefined,
+        scrollEnabled: false
```
ethanshar, line 29: we have the scrollEnable state to lock/unlock the scroll that's only internal. 
How would you use this HOC to decide if the to render a gradient at the bottom? or anything at all. 
You don't pass the user back this value in order to make a use of it.

src/commons/asScrollableContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+      };
+    }
+
+    static isScrollEnabled = (contentSize?: number, maximumContentSize?: number) => {
+      return !_.isUndefined(contentSize) && !_.isUndefined(maximumContentSize) && contentSize > maximumContentSize;
+    };
+
+    static getDerivedStateFromProps(nextProps: any, prevState: any) {
```
ethanshar, line 37: If you're writing a new component and already has to use `getDerivedStateFromProps` it should already raise a flag. Do we really want this component has to be uncontrolled?

src/commons/asScrollableContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface ScrollableContentState {
+  maximumContentSize?: number;
+  contentSize?: number;
+  scrollEnabled: boolean;
+}
+
+type PropTypes = ScrollableContentProps;
+
+export default function asScrollableContent(WrappedComponent: React.ComponentType<any>) {
+  class ScrollableContent extends Component<PropTypes, ScrollableContentState> {
```
ethanshar, line 21: Consider experimenting with functions and hooks

src/commons/index.js
```diff
@@ -16,6 +16,9 @@ module.exports = {
   get UIComponent() {
     return require('./UIComponent').default;
   },
+  get asScrollableContent() {
```
ethanshar, line 19: move it to commons/new
This file will be replaced by commons/new eventually

ethanshar:
@M-i-k-e-l 
Is this PR still relevant?

M-i-k-e-l:
> @M-i-k-e-l
> Is this PR still relevant?

No

PR #802, opened 2020-06-02
[Feature] Added handling of array of undefined validations

Effanuel:
#### Issues:
* Passing `validate={[undefined, undefined]}` throws red screen;

#### Motivation:
* When having feature toggles and code that looks like this:
```
validate={[
   condition1 ? validateFunc1 : undefined,
   condition2 ? validateFunc2 : undefined
   ]}
```
The only solution is to use `() => true`, instead of `undefined`

#### Changes:
* Added handling of array of undefined;

**This only handles the situation where mutliple false returning values are the same**

### DIFF
--- a/src/components/inputs/BaseInput.js
+++ b/src/components/inputs/BaseInput.js
@@ -6,7 +6,6 @@ import {Colors, Typography} from '../../style';
 import {BaseComponent} from '../../commons';
 import Validators from './Validators';
 
-
 const VALIDATORS = {
   REQUIRED: 'required',
   EMAIL: 'email',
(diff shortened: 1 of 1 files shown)


ethanshar:
Hi @Effanuel 
When I render 
```
        <TextField
          placeholder="Placeholder"
          validate={[undefined, undefined]}
        />
```

I don't get any error. 
Can you paste the code that throw an exception for u

Effanuel:
> Hi @Effanuel
> When I render
> 
> ```
>         <TextField
>           placeholder="Placeholder"
>           validate={[undefined, undefined]}
>         />
> ```
> 
> I don't get any error.
> Can you paste the code that throw an exception for u

I have `validateOnChange` enabled, so when that triggers, I get an error.

You can basically see that in the code since `[undefined, undefined]` will evaluate to true and `undefined` to false.

ethanshar:
Ok, cool it reproduce now. 
Why not just add here a check for undefined here 
https://github.com/wix/react-native-ui-lib/blob/master/src/components/inputs/BaseInput.js#L166
```
if (validatorFunction && !validatorFunction(value)) {
```

PR #804, opened 2020-06-03
Feat/ColorPicker - preview text style prop

ArnonZ:
- Added previewTextStyle prop to allow adding styling to the preview text 
(WOAUILIB-823)


src/components/colorPicker/index.js
```diff
@@ -62,6 +62,10 @@ export default class ColorPicker extends PureBaseComponent {
      * Props to pass the Dialog component
      */
     dialogProps: PropTypes.object,
+    /**
+     * Additional styling for the color preview text.
+     */
+    previewTextStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
```
ethanshar, line 68: the style is meant for the input. call it `previewInputStyle`

src/components/colorPicker/index.js
```diff
@@ -320,7 +324,7 @@ export default class ColorPicker extends PureBaseComponent {
               onChangeText={this.onChangeText}
               style={[
                 this.styles.input,
-                {color: textColor, width: (value.length + 1) * 16.5 * fontScale},
+                {color: textColor, width: (value.length + 1) * 16.5 * fontScale, ...previewTextStyle},
```
ethanshar, line 327: Don't spread the style, technically it can be a number that's pointing to a style. 
the Input style is already constructed of an array, you can add it as another item of the array. 
Also, consider passing it last, to give full control to the user.

PR #806, opened 2020-06-07
Feat/card section

M-i-k-e-l:


src/components/card/CardImage.js
```diff
@@ -35,29 +35,32 @@ export default class CardImage extends BaseComponent {
...
 
-  static defaultProps = {
-    borderRadius: BorderRadiuses.br40
-  };
+  constructor(props) {
+    super(props);
+
+    if (props.borderRadius) {
+      LogService.warn('uilib: Please stop passing borderRadius to Card.Image, it will get the borderRadius from the Card');
+    }
```
ethanshar, line 48: I don't think people pass borderRadius to CardImage. 
I think it's an internal prop, we injected it in the Card code.

src/components/card/CardSection.js
```diff
@@ -0,0 +1,70 @@
...
+  static displayName = 'Card.Section';
+
+  static propTypes = {
+    /**
+     * Text content for the CardSection.
+     * Example: content={[{text: 'You’re Invited!', props: {text70: true, dark10: true}}]}
+     */
+    content: PropTypes.arrayOf(PropTypes.shape({
+      text: PropTypes.string,
+      props: Text.propTypes
```
ethanshar, line 24: Do you think that maybe we can flat this shape and allow the user to pass `text` and directly other text props? 

so instead of passing 
```
content: {text: 'title', props: {numberOfLines: 2, color: 'red'}}
```
the user can pass 
```
content: {text: 'title', numberOfLines: 2, color: 'red'}
```

src/components/card/asCardChild.js
```diff
@@ -0,0 +1,20 @@
+import React, {Component} from 'react';
+import CardContext from './CardContext';
+
+function asCardChild(WrappedComponent) {
+  class CardChild extends Component {
```
ethanshar, line 5: consider simplify it by replacing class with a function

src/components/card/index.js
```diff
@@ -205,16 +212,20 @@ class Card extends PureBaseComponent {
     );
   }
 
-  renderChildren() {
+  renderChildren = () => {
     const children = React.Children.map(this.props.children, (child, index) => {
```
ethanshar, line 216: instead of keeping indices of null items, maybe we can just filter those children and ignore them. 
and then the `calcChildPosition` can be simpler..

src/components/card/CardImage.js
M-i-k-e-l: Unfortunately they do, I've seen usages, but we might be able to ignore because the two usages I've seen pass the same radius as the card (I did not look for more usages).
What would you like me to do?

src/components/card/CardSection.js
M-i-k-e-l: Done

src/components/card/asCardChild.js
M-i-k-e-l: Done

src/components/card/index.js
M-i-k-e-l: Done

src/components/card/CardImage.js
ethanshar: Let's ignore them. 
But still keep the warning so people remove it.

ethanshar: I see u already do that. so nvm (:

PR #807, opened 2020-06-09
Feat/with scroll enabler

M-i-k-e-l:
There are still a few TS errors that I'm not sure how to solve.


demo/src/screens/commonScreens/index.js
```diff
@@ -0,0 +1,6 @@
+
```
ethanshar, line 1: I recently did some work and reducing hierarchy of our folders 
let's put this example screen with the rest of the components

src/commons/withScrollEnabler.tsx
```diff
@@ -0,0 +1,79 @@
...
+        forwardedRef.current = scrollViewRef;
+      }
+    }, [scrollViewRef]);
+
+    useEffect(() => {
+      const isScrollEnabled = contentSize > layoutSize;
+      if (isScrollEnabled !== scrollEnabled) {
+        setScrollEnabled(isScrollEnabled);
+        if (Constants.isAndroid && scrollViewRef.current && scrollViewRef.current) {
+          if (_.isFunction(scrollViewRef.current.scrollTo)) {
```
ethanshar, line 31: What is that for?

src/commons/withScrollEnabler.tsx
```diff
@@ -0,0 +1,79 @@
...
+export type WithScrollEnablerProps = FlatListProps<any> | ScrollViewProps;
+function withScrollEnabler<PROPS extends WithScrollEnablerProps>(WrappedComponent: React.ComponentType<WithScrollEnablerProps>): React.ComponentType<PROPS> {
+  const ScrollEnabler = (props: WithScrollEnablerProps) => {
+    const [scrollEnabled, setScrollEnabled] = useState(true);
+    const [contentSize, setContentSize] = useState(0);
+    const [layoutSize, setLayoutSize] = useState(0);
+
+    const scrollViewRef = useRef<WithScrollEnablerProps>(null);
+
+    useEffect(() => {
```
ethanshar, line 17: Why do u need an access to the Scroll element ref?

src/commons/withScrollEnabler.tsx
```diff
@@ -0,0 +1,79 @@
...
+        setLayoutSize(size);
+      }
+      _.invoke(props, 'onLayout', event);
+    },
+    [props]);
+
+    return (
+      <WrappedComponent
+        {...props}
+        onContentSizeChange={onContentSizeChange}
```
ethanshar, line 68: There might be a little risk with passing the Scroll props as is. with the same names. 
They can be passed in a spread by mistake to places you don't want. also I think we might want to have a more explicit name for them so the user should understand where they are coming from . 
Another option is to group them as an object that the user can spread intentionally 

for instance 
```
<WrappedComponent scrollEnablerProps={{onLayout, scrollEnabled, onContentSizeChange}}
```

demo/src/screens/commonScreens/index.js
M-i-k-e-l: Moved

src/commons/withScrollEnabler.tsx
M-i-k-e-l: Moved to screen

src/commons/withScrollEnabler.tsx
M-i-k-e-l: Moved to screen

src/commons/withScrollEnabler.tsx
M-i-k-e-l: Done, I've created two use classes, 1 for FlatList and 1 for ListView, they are very similar but I'm not sure how easy they are to combine.

src/commons/withScrollEnabler.tsx
```diff
@@ -1,74 +1,67 @@
...
-    },
-    [props]);
+    const onContentSizeChange = useCallback(
+      (contentWidth: number, contentHeight: number) => {
+        const size = props.horizontal ? contentWidth : contentHeight;
+        if (size !== contentSize) {
+          setContentSize(size);
+        }
+      },
+      [props]
```
ethanshar, line 42: Is there a reason to pass all props in the dependency array?

src/commons/withScrollEnabler.tsx
```diff
@@ -1,74 +1,67 @@
...
-      } = event;
-      const size = props.horizontal ? width : height;
-      if (size !== layoutSize) {
-        setLayoutSize(size);
-      }
-      _.invoke(props, 'onLayout', event);
-    },
-    [props]);
+      },
+      [props]
```
ethanshar, line 57: same here? 
why do u pass all the props?

demo/src/screens/componentScreens/WithScrollEnablerScreen/AutoLockScrollView.tsx
```diff
@@ -0,0 +1,89 @@
...
+export type AutoLockScrollViewProps = ScrollViewProps & {
+  numberOfItems: number;
+};
+
+const AutoLockScrollView = (props: AutoLockScrollViewProps) => {
+  const numberOfItems = props.numberOfItems;
+
+  const WithScrollEnabler = withScrollEnabler(
+    (props: WithScrollEnablerProps) => {
+      function renderItem(index: number) {
```
ethanshar, line 26: You must use `useCallback` in order to define methods inside a function component. 
Otherwise what will happen is that on every render you will re-create those functions.

demo/src/screens/componentScreens/WithScrollEnablerScreen/AutoLockFlatList.tsx
```diff
@@ -0,0 +1,98 @@
...
+export type AutoLockScrollViewProps = FlatListProps<number> & {
+  numberOfItems: number;
+};
+
+const AutoLockFlatList = (props: AutoLockScrollViewProps) => {
+  const numberOfItems = props.numberOfItems;
+
+  const WithScrollEnabler = withScrollEnabler(
+    (props: WithScrollEnablerProps) => {
+      const getData = memoize((numberOfItems: number) => {
```
ethanshar, line 27: even though you memoize this function you still re-creating it on each render.

src/commons/withScrollEnabler.tsx
M-i-k-e-l: Fixed and removed - does not need to use state here

src/commons/withScrollEnabler.tsx
M-i-k-e-l: Fixed and removed - does not need to use state here

demo/src/screens/componentScreens/WithScrollEnablerScreen/AutoLockScrollView.tsx
M-i-k-e-l: Changed to useCallback

demo/src/screens/componentScreens/WithScrollEnablerScreen/AutoLockFlatList.tsx
M-i-k-e-l: Changed to useCallback

src/commons/withScrollEnabler.tsx
```diff
@@ -22,24 +22,27 @@ function withScrollEnabler<PROPS extends SupportedViews>(
...
 
-    useEffect(() => {
-      const isScrollEnabled = contentSize > layoutSize;
+    const checkScroll = useCallback(() => {
+      const isScrollEnabled = contentSize.current > layoutSize.current;
       if (isScrollEnabled !== scrollEnabled) {
         setScrollEnabled(isScrollEnabled);
       }
-    }, [contentSize, layoutSize]);
+    }, []);
```
ethanshar, line 33: If you're using something inside the callback it should be part of the dependencies array of the useCallback
in this case you're using `scrollEnabled` state, so it should be included in the array.

M-i-k-e-l: Done

src/commons/withScrollEnabler.tsx
```diff
@@ -42,7 +42,7 @@ function withScrollEnabler<PROPS extends SupportedViews>(
           }
         }
       },
-      [props.horizontal]
+      [props.horizontal, checkScroll]
```
ethanshar, line 45: I don't think it's recommended to pass to the dependency array a callback. 
and actually I don't think you even need to, I think `[props.horizontal]` is enough in this case. 

same goes for the onLayout callback

M-i-k-e-l: That was recommended by the new plugin.

PR #810, opened 2020-06-10
typography-deprecation - JSX should use short version

M-i-k-e-l:
typography-deprecation - JSX should use short version (no Typography prefix) - fix


M-i-k-e-l:
> @M-i-k-e-l
> Do we have a test for it?

I did not find a way to test `fix` other than to add logs and run the tests, so no.

PR #823, opened 2020-06-19
Feat/Chip

ArnonZ:
- New Chip component + exmaple screen.
- Added small checkmark icon to demo/assets


src/components/chip/index.tsx
```diff
@@ -0,0 +1,345 @@
...
+        tintColor={labelColor}
+        style={[styles.label, getMargins('label'), labelStyle]}
+        testID={`${testID}.label`}
+      >
+        {label}
+      </Text>
+    );
+  };
+
+  const getMargins = (element: string): object | undefined => {
```
ethanshar, line 251: I'm not a fan of the those huge hub functions, I think we can do a little refactor here: 

First, I see that all elements excepts `label` barely have any logic, so I think we can just set these margins inline in each appropriate render function and consider making this function solely for label. 

Now regarding the label margins logic. 
Since you have a separate render function for each component, wouldn't it be cleaner to check in each function if we have a label and add the margin accordingly to the element (avatar, badge, etc..) and not the label.

src/components/chip/index.tsx
```diff
@@ -0,0 +1,345 @@
...
+   */
+  onPress?: (props: any) => void;
+  /**
+   * Chip's background color
+   */
+  backgroundColor?: string;
+  /**
+   * Sets size to use minWidth and minHeight - default is true
+   */
+  minSize?: boolean;
```
ethanshar, line 34: The name sounds a little like this prop act exactly as the `size` prop (meaning it accept number or an object) 
which is not such a bad idea. Either the user pass `width` or `minWidth` 

But if this prop is a boolean for another prop, I think the name should be a little different, maybe `useSizeAsMinimum` or something like that, I know it's a little long, but I guess it's clearer. 
WDYT?

src/components/chip/index.tsx
```diff
@@ -0,0 +1,345 @@
...
+   */
+  labelColor?: string;
+  /**
+   * Label's style
+   */
+  labelStyle?: StyleProp<TextStyle>;
+  /**
+   * Text to show to the right of the label or inside the Badge.
+   */
+  counterLabel?: string;
```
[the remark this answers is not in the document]
ArnonZ: Done. it had its costs on the cleanliness of the code (as expected), I think it's an ok price to pay...let me know if you have any improvements.

src/components/chip/index.tsx
ArnonZ: In general, by refactoring (shortening it) we lose all its purpose.

Which is to avoid specific margin annotations that are disconnected from the rest of the margin logic making it hard for to track, debug and override
Main advantages:
- Readability (all in one place): you know where to look and what effects do the inner elements have on each other in terms of spacings. Instead of hopping from render to render.
- Code cleanliness: no more messy inline styling! (was a fan).
- Supports how the "margins api" is built: we supply basic ui (margins included) and 2 override "switches": you can override specific margin with styling or burn them all with fire and pass `resetMargins`. The hub makes it manageable to see instantly what happens when you have elements together.
Disadvantage: its long.

*I made this choice because to write this logic, I had to decrypt all the margins of the 3 components we want to render in Private using this base component (chip, label, tag). Their margin logic was hell to keep track of.

As I see it, it's either this or going back to inline margins and one function for the label margins `getLabelMargin()`.  Once you add label margins to each component's render function you get 3 and sometimes 4 margin variations (like for counter)

src/components/chip/index.tsx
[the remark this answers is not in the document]
ethanshar: I actually aimed for having a single badge prop (where you can pass badgeProps), similar to what we have in Avatar and TabBar.

src/components/chip/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+        numberOfLines={1}
+        // @ts-ignore
+        tintColor={labelColor}
+        style={[styles.label, getMargins('label'), labelStyle]}
+        testID={`${testID}.label`}
+      >
+        {label}
+      </Text>
+    );
+  }, [label]);
```
ethanshar, line 231: labelColor and labelStyle should also be included in the `useCallback` dependencies since they affect the rendered output.

PR #827, opened 2020-06-25
Keyboard accessory view safe area api support

mendyEdri:


lib/components/Keyboard/KeyboardInput/CustomKeyboardViewBase.js
```diff
@@ -6,11 +6,13 @@ export default class CustomKeyboardViewBase extends Component {
   static propTypes = {
     initialProps: PropTypes.object,
     component: PropTypes.string,
-    onItemSelected: PropTypes.func
+    onItemSelected: PropTypes.func,
+    useSafeArea: PropTypes.bool
```
M-i-k-e-l, line 10: This is shared with Android, it should not be here (and the default should be in ios)

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.js
```diff
@@ -110,7 +110,15 @@ class KeyboardAccessoryView extends Component {
      *
      * default: false
      */
-    allowHitsOutsideBounds: PropTypes.bool
+    allowHitsOutsideBounds: PropTypes.bool,
+
+    /**
+     * iOS only.
+     * Allow KeyboardAccessoryView to reach to the bottom of the screen (on iPhone x and above)
```
M-i-k-e-l, line 117: 1. Doesn't it do the opposite? - Maybe "Set to false in order to allow..."?
2. Is there a new iPhone without SafeArea? (I don't know, maybe SE or something?)

mendyEdri: `* iOS only.
     * Set false to allow the KeyboardAccessoryView to reach to the bottom of the screen,
     * true to save space to screen's safe area.
     * Supported in all iOS devices that has notch.
     * Other devices (including android) will avoid it and won't have space at the bottom.`

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.js
```diff
@@ -110,7 +110,18 @@ class KeyboardAccessoryView extends Component {
      *
      * default: false
      */
-    allowHitsOutsideBounds: PropTypes.bool
+    allowHitsOutsideBounds: PropTypes.bool,
+
+    /**
+     * iOS only.
+     * Set false to allow the KeyboardAccessoryView to reach to the bottom of the screen,
```
M-i-k-e-l, line 117: The language is a bit cumbersome, I don't think you need to explain it in depth.
We have these two for other components:
1. "Whether or not to handle SafeArea"
2. "Use safe area, in case action bar attached to the bottom (default: true)"
Maybe something like:
"Whether or not to handle SafeArea

default: true"

PR #828, opened 2020-06-28
withScrollReached (new HOC) - notifies when start\end of scroll (ScrollView\FlatList) has been reached

M-i-k-e-l:
withScrollReached (new HOC) - notifies when start\end of scroll (ScrollView\FlatList) has been reached.


ethanshar:
@M-i-k-e-l 
Please add a description or something that can help understand what's this PR is for.. 
also, as discussed, it would help if the PR title will have a better description that fits a change log

src/commons/withScrollReached.tsx
```diff
@@ -0,0 +1,70 @@
...
+  isScrollAtEnd?: boolean;
+};
+
+type SupportedViews = FlatListProps<any> | ScrollViewProps;
+
+export type WithScrollReachedProps = SupportedViews & {
+  scrollReachedProps: ScrollReachedProps;
+  ref?: any;
+};
+type PropTypes = ForwardRefInjectedProps & SupportedViews;
```
ethanshar, line 23: Not sure about the type here? 
Basically you're saying that the Wrapped component must be a ScrollView/FlatList? 
But isn't it possible the Wrapped component is just a simple View that wraps ScrollView/FlatList ?

demo/src/screens/componentScreens/WithScrollReachedScreen/index.tsx
```diff
@@ -0,0 +1,70 @@
+import React, {Component} from 'react';
```
ethanshar, line 1: Personally I think that the 3 example screen files makes it a little difficult to understand what's going on. 
I think it's a great feature, but still a pretty small one, as a user I'm getting lost in all of these files when I'm trying to figure out how the API works.

Let's simplify it. I don't think there's really a need for showing how it works both on FlatList and ScrollView, the principles are similar, showing only on ScrollView it's good enough.

src/commons/withScrollReached.tsx
```diff
@@ -0,0 +1,70 @@
...
+            contentOffset: {x: offsetX, y: offsetY},
+            contentSize: {width: contentWidth, height: contentHeight}
+          }
+        } = event;
+
+        const horizontal = props.horizontal;
+        const layoutSize = horizontal ? layoutWidth : layoutHeight;
+        const offset = horizontal ? offsetX : offsetY;
+        const contentSize = horizontal ? contentWidth : contentHeight;
+        const closeToStart = offset <= 0; // TODO: consider adding padding (user input)
```
ethanshar, line 45: Let's add a `threshold` option or something.. 
I see how we already have use cases for it.. (I do, In TabController (: )

demo/src/screens/componentScreens/WithScrollReachedScreen/FadedScrollView.tsx
```diff
@@ -0,0 +1,96 @@
...
+          style={styles.fadeOutImage}
+          testID={`${props.testID}.fadeOut`}
+          source={fadeImage}
+        />
+      );
+    }
+  }, [
+    props.scrollEnabled,
+    props.scrollReachedProps.isScrollAtEnd,
+    props.testID
```
ethanshar, line 53: why do we depend on  `props.scrollEnabled` (I think it's a left over of copy-paste)
And also `props.testID` how does it affect the render in any way?

src/commons/withScrollReached.tsx
M-i-k-e-l: I think that what matters is this:
`WrappedComponent: React.ComponentType<WithScrollReachedProps>`
maybe I should rename this to something else?
BTW, this section is changed because of other warnings\errors, but please let me know if the name (or my thinking) is wrong.

demo/src/screens/componentScreens/WithScrollReachedScreen/index.tsx
M-i-k-e-l: ok, do you want it to be in a single file or a screen + component?

src/commons/withScrollReached.tsx
M-i-k-e-l: That was the TODO, although threshold is much better :)
Done

demo/src/screens/componentScreens/WithScrollReachedScreen/FadedScrollView.tsx
M-i-k-e-l: `testID` is a mistake, `scrollEnabled` will technically be needed in a real example.
When I think about it, maybe a better solution for the screen(s) is having a single screen for both HOCs, although it will make it more complicated, it will be a "real" example.
What do you think?

demo/src/screens/componentScreens/WithScrollReachedScreen/index.tsx
ethanshar: Yes, let's simplify it. It's important that the example screen is easy to understand

demo/src/screens/componentScreens/WithScrollReachedScreen/FadedScrollView.tsx
ethanshar: For now keep the example screens separated. 
Specifically here, I don't there there's a `props.scrollEnabled` so it's redundant...

src/commons/withScrollReached.tsx
ethanshar: it's not about the name. 

I think there's might be something wrong with the types, but maybe i'm missing something, let's take it offline

demo/src/screens/componentScreens/WithScrollReachedScreen/index.tsx
M-i-k-e-l: Merged files and removed `FlatList`

demo/src/screens/componentScreens/WithScrollReachedScreen/FadedScrollView.tsx
M-i-k-e-l: Done

PR #833, opened 2020-06-30
Fix some types, add exports for components `Props` types

ykravv:


src/commons/asBaseComponent.tsx
```diff
@@ -14,7 +14,7 @@ export interface BaseComponentInjectedProps {
   modifiers: ReturnType<typeof Modifiers.generateModifiersStyle>;
 }
 
-function asBaseComponent<PROPS, STATICS = {}>(WrappedComponent: React.ComponentType<any>): React.ComponentType<PROPS> & STATICS {
+function asBaseComponent<PROPS, STATICS = {}>(WrappedComponent: React.ComponentType<any>): React.ComponentClass<PROPS> & STATICS {
```
ethanshar, line 17: Doesn't it limit us to only Class components? 
what about Function components?

ykravv: Inside of `asBaseComponent` we create class and use it to wrap `WrappedComponent` which still could be either class or function. Result of the function will be a class though. 

Without this change `View` has type 
```
const View: React.ComponentClass<ViewPropTypes, any> | React.FunctionComponent<ViewPropTypes>
``` 
which then leads to errors when we try for instance 
```
View.contextType = ThemeContext;
``` 
as functional components does not have `contextType`

ykravv: Inside of `asBaseComponent` we create class and use it to wrap `WrappedComponent` which still could be either class or function. Result of the function will be a class though. 

Without this change `View` has type 
```
const View: React.ComponentClass<ViewPropTypes, any> | React.FunctionComponent<ViewPropTypes>
``` 
which then leads to errors when we try for instance 
```
View.contextType = ThemeContext;
``` 
as functional components does not have `contextType`

ethanshar: Got ya 👍

PR #837, opened 2020-07-02
Migrate to RN62

ethanshar:
- iOS migrate is completed
- Android also, only having issue with adding our native packages if can take a look please


android/app/src/main/java/com/rnuilib/MainApplication.java
```diff
@@ -1,112 +1,122 @@
...
+                    return BuildConfig.DEBUG;
+                }
+
+                @Override
+                protected List<ReactPackage> getPackages() {
+                    @SuppressWarnings("UnnecessaryLocalVariable")
+                    List<ReactPackage> packages = new PackageList(this).getPackages();
+                    // Packages that cannot be autolinked yet can be added manually here, for example:
+                    // packages.add(new MyReactNativePackage());
+                    // packages.addAll(new UiLibPackageList(this).getPackageList());
```
ethanshar, line 34: @M-i-k-e-l 
I left this in comment cause it fail the build. 
Not sure exactly how to solve it, i'm sure you'll have an idea.

PR #839, opened 2020-07-05
Infra/tabbar typescript

Inbal-Tish:
TabBar and TabBarItem to TS


src/components/tabBar/TabBarItem.tsx
```diff
@@ -1,14 +1,87 @@
...
+   */
+  label?: string,
+  /**
+   * custom label style
+   */
+  labelStyle: TextStyle,
+  /**
+   * Badge component props to display next the item label
+   */
+  badge: BadgeProps,
```
ethanshar, line 38: shouldn't these be optional as well ?

Inbal-Tish: Yes, of course. Done.

PR #843, opened 2020-07-06
TS/Avatar

ArnonZ:
WOAUILIB-626


src/components/avatar/index.tsx
```diff
@@ -324,7 +341,7 @@ export class Avatar extends PureBaseComponent {
         accessible={!_.isUndefined(onPress)}
         accessibilityLabel={'Avatar'}
         accessibilityRole={onPress ? 'button' : 'image'}
-        {...this.extractAccessibilityProps()}
+        {...extractAccessibilityProps(this.props)}
```
ArnonZ, line 344: 🤔

src/components/avatar/index.tsx
```diff
@@ -42,11 +141,20 @@ const DEFAULT_BADGE_POSITION = BADGE_POSITIONS.TOP_RIGHT;
...
  */
-export class Avatar extends PureBaseComponent {
-  constructor(props) {
+class Avatar extends PureComponent<AvatarPropTypes> {
+
+  styles: {
+    initials: StyleProp<ViewStyle>
+    initialsContainerWithInset: StyleProp<ViewStyle>,
+    ribbon: StyleProp<ViewStyle>
+  }
```
ethanshar, line 150: Try this instead
```suggestion
  styles: ReturnType<typeof createStyles>;
```

src/components/avatar/index.tsx
```diff
@@ -34,6 +35,104 @@ export const BADGE_POSITIONS = {
 const DEFAULT_BADGE_SIZE = 'pimpleBig';
 const DEFAULT_BADGE_POSITION = BADGE_POSITIONS.TOP_RIGHT;
 
+type StatusModes = keyof typeof STATUS_MODES
+type BadgePosition = keyof typeof BADGE_POSITIONS
```
ethanshar, line 39: Consider using `enum`. See references in the private repo

src/components/avatar/index.tsx
```diff
@@ -181,8 +195,8 @@ export class Avatar extends PureBaseComponent {
     };
   }
 
-  getRibbonStyle() {
-    const {size} = this.props;
+  getRibbonStyle(): StyleProp<ViewStyle> {
+    const {size = Avatar.defaultProps.size} = this.props;
```
ethanshar, line 199: It feels kinda weird that we need to set the default here also. 
I assume TS throws an error here. 
I would check in the guild channel if someone know how to solve that. 
A lot of people know TS better than us. maybe there's a better solution

ArnonZ: This is a major issue with TS and React. A known issue with lots of weird and weirder solutions all very extravagant for this small issue we have here.
I asked around, it was suggested either `@ts-ignore` this, use constructor to initialize size and not use `defaultProps` at all...
The internet suggests even weirder gymnastics to solve this.
I don't feel. either suits this small `size` issue. IMO, we shouldn't impose unlogical changes in code to cater to TS. I can suggest leaving it like or ignoring it, or if you have a better solution...wdyt?

ethanshar: Yes I agree. 
Anyway, I discussed with someone from the guild, he recommended changing the `size` prop to be not optional
This way we're telling it that we got the value covered with default props. 
Try it, it should work..

ArnonZ: Cool, seems to work 👍

PR #844, opened 2020-07-07
Feat/drawer left toggle

Inbal-Tish:
Adding left toggle functionality

### DIFF
--- a/src/components/drawer/index.js
+++ b/src/components/drawer/index.js
@@ -128,9 +128,8 @@ class NewDrawer extends PureBaseComponent {
         : this.renderRightActions
       : undefined;
 
-    this.state = {
-      leftX: new Animated.Value(0)
-    };
+    
+    this.leftActionX = new Animated.Value(0);
   }
 
   /** Actions */
@@ -189,7 +188,7 @@ class NewDrawer extends PureBaseComponent {
   };
 
   onToggleSwipeLeft = ({rowWidth, leftWidth, dragX}) => {
-    Animated.timing(this.state.leftX, {
+    Animated.timing(this.leftActionX, {
       toValue: dragX ? dragX - leftWidth : rowWidth * 0.6 - leftWidth,
       easing: Easing.linear,
       duration: 100,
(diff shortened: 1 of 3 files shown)


ethanshar:
@Inbal-Tish 
I didn't go over all the code cause I found some UI issues I think it's better to fix first.

First, the UI for right buttons is broken in this PR (the actions are not displayed horizontally but vertically) 

Second, 
I don't think the toggle works so good, something there is a little laggy. 
If we need to compromise on the UX in order to make it work slicker so let's do it. 
I think you can find a good, simple reference for toggle drawer in Gmail app

And one last thing, in the example screen, now it's impossible to see how the basic functionality works (when just opening the drawer and it stays open)

Inbal-Tish:
oh my. Checking

src/components/drawer/index.js
```diff
@@ -119,6 +127,10 @@ class NewDrawer extends PureBaseComponent {
         ? this.renderLeftActions
         : this.renderRightActions
       : undefined;
+
+    this.state = {
+      leftX: new Animated.Value(0)
+    };
```
ethanshar, line 133: No need to keep values that don't affect the state in state (like animated values)
Also, the name is a little too general, since there's a lot of things that refer to left/right, it would help to have a better name that hint what's this animation is responsible for. (e.g leftActionX, leftContentX, etc..)

src/components/drawer/index.js
```diff
@@ -225,17 +259,17 @@ class NewDrawer extends PureBaseComponent {
   renderActions(items, progress, dragX) {
     if (items) {
       return (
-        <View row>
+        <Animated.View style={[styles.actionsContainer, {transform: [{translateX: this.state.leftX}]}]}>
```
ethanshar, line 262: No need to replace View with Animated.View anymore. 
You can just render 
```
<View row animated>
```

the `animated` prop will make sure to use an `Animated.View`

src/components/drawer/index.js
```diff
@@ -341,6 +378,10 @@ class NewDrawer extends PureBaseComponent {
 export default NewDrawer;
 
 const styles = StyleSheet.create({
+  actionsContainer: {
+    flexDirection: 'row', 
+    justifyContent: 'center'
+  },
```
ethanshar, line 384: Once you'll use the `animated` prop on View, you can go back to using the `row` modifier instead of this

src/components/drawer/index.js
Inbal-Tish: Done

src/components/drawer/index.js
Inbal-Tish: Done

src/components/drawer/index.js
Inbal-Tish: Done

PR #847, opened 2020-07-09
Fader - a component to easily add fade look to other components

M-i-k-e-l:
Notes about the images:
1. I've tried at first using the original `gradient` image and rotating it, I only managed doing it with `transform` and that required `onLayout` so I'd know the size of the view, this worked all right on iOS but was slower on Android for some reason - it did not look good enough, so I chose to go with 4 images for the different orientations.
* The code was much more complex in the older version.
2. If we merge this, we should deprecate the original `gradient` image in the next version.


src/assets/images/index.js
```diff
@@ -1,5 +1,18 @@
 export const images = {
+  // TODO: deprecate
```
ethanshar, line 2: Since the gradient image can changed according to many different requirements I think it's better that the user will provide the image instead. 
If we want to provide defaults we can keep these images, but keep them as part of the component assets and don't expose them in our Assets

src/components/withScrollFader/index.tsx
```diff
@@ -0,0 +1,173 @@
+import React, {useCallback} from 'react';
+import {
+  StyleSheet,
+  // eslint-disable-next-line no-unused-vars
```
ethanshar, line 4: It's impossible to read the code with all the lint comments. 
First, there's a way to ignore a block of code with eslint. 
Second, because of TS, this lint error won't break the build for now, I know it's annoying to see the red marks but for now ignore them, these things will be fixed/handled all together once we complete the TS migration.

src/components/withScrollFader/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+import View from '../view';
+import Image from '../image';
+import Assets from '../../assets';
+
+export type ScrollFaderProps = {
+  onScroll: (event: NativeSyntheticEvent<NativeScrollEvent>) => void;
+  /**
+   * Use to render the fade
+   */
+  renderFader: () => React.ReactNode;
```
ethanshar, line 30: If the whole idea of this HOC is to provide a render function that renders a component based on props. wouldn't it be easier to create a component? 
I think it would be a lot simpler if the use can just render a "ScrollFader" component. WDYT?

src/components/withScrollFader/index.tsx
M-i-k-e-l: Removed

src/assets/images/index.js
M-i-k-e-l: Moved internally

src/components/withScrollFader/index.tsx
M-i-k-e-l: Maybe if there has to be a **single** child that has to have `onScroll` (or has to be a `supportedView`).

M-i-k-e-l: Changed to Fader, changed the title (but not the branch name :) )

src/components/fader/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+    return {
+      containerStyle,
+      imageStyle,
+      imageSource
+    };
+  }, [fadeSize, props.location]);
+
+  const isVisible = useCallback(() => {
+    return props.visible || _.isUndefined(props.visible);
+  }, [props.visible]);
```
ethanshar, line 79: Do we need the callback? 
Can't we check directly `props.visible` in the render?

src/components/fader/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export type FaderProps = {
+  /**
+   * Whether the fader is visible (default is true)
+   */
+  visible?: boolean;
+  /**
+   * The location of the fader (the image is different), defaults to Fader.location.RIGHT
+   */
+  location?: FaderLocation;
```
ethanshar, line 22: What do you think of renaming to `position`? to be more consistent with other components (toast position, badge position, etc..)

demo/src/screens/componentScreens/FaderScreen.tsx
```diff
@@ -0,0 +1,89 @@
+import _ from 'lodash';
```
ethanshar, line 1: Looks great. Very simple and easy to understand the component 👍

src/components/fader/index.tsx
M-i-k-e-l: Moved

src/components/fader/index.tsx
M-i-k-e-l: Changed

demo/src/screens/componentScreens/FaderScreen.tsx
M-i-k-e-l: 10x

PR #850, opened 2020-07-12
add numeric size support to Avatar's badge + unit-tests

mendyEdri:
Added numeric value support for Avatar's badge size.


src/components/avatar/__tests__/index.spec.js
```diff
@@ -58,4 +59,40 @@ describe('Avatar Badge', () => {
...
+    it('should return 876 as the size number given', () => {
+      const uut = new Avatar({badgeProps: {size: 876}});
+      expect(uut.getBadgeSize()).toEqual(876);
+    });
+
+    it('should return 0 as the given size number', () => {
+      const uut = new Avatar({badgeProps: {size: 0}});
+      expect(uut.getBadgeSize()).toEqual(0);
+    });
+
```
ArnonZ, line 78: Not the end of the world obviously but having 3 tests testing numeric input might be 2 tests too much 😸 .
If one of these tests fail, all of them will fail for the same reason.
For your consideration...

src/components/avatar/index.tsx
```diff
@@ -222,14 +222,21 @@ class Avatar extends PureComponent<AvatarPropTypes> {
     return _.get(this.props, 'badgeProps.backgroundColor') || statusColor || onlineColor;
   }
 
-  getBadgeSize = () => _.get(this.props, 'badgeProps.size', DEFAULT_BADGE_SIZE);
+  getBadgeSize = () => {
+    const badgeSize = _.get(this.props, 'badgeProps.size', DEFAULT_BADGE_SIZE);
+    
+    if (_.isString(badgeSize)) {
+      return BADGE_SIZES[badgeSize];
```
ArnonZ, line 229: When you split the logic between enum and numeric size you eliminated the fallback of `DEFAULT_BADGE_SIZE` for string-based sizes (currently returns undefined if sending a bad enum). Consider adding `DEFAULT_BADGE_SIZE` fallback if supplied enum size isn't included in `BADGE_SIZES` - to maintain old fallback logic.

mendyEdri: Right, added.

src/components/avatar/__tests__/index.spec.js
mendyEdri: right, keeping two for zero/undefined check + some number

src/components/avatar/__tests__/index.spec.js
```diff
@@ -58,4 +59,35 @@ describe('Avatar Badge', () => {
       expect(uut.renderBadge()).toEqual(undefined);
     });
   });
+
+  describe('badgeProps.size, supports enum or number', () => {
+    it('should return 10 as the size number given', () => {
```
ArnonZ, line 64:
```suggestion
    it('should return 99 as the size number given', () => {
```

PR #854, opened 2020-07-15
The new TextField (Incubator)

ethanshar:
Recreate the TextField component basing on these [guidelines](https://uxdesign.cc/text-fields-forms-design-ui-components-series-2b32b2beebd0)

- [x] Add parity with the old TextField
- [ ] Use ThemeManager to apply our private UI 
- [ ] QA the component for missing functionality and UI issues
- [ ] Start planning a migration process 

**Features**
- [x] General structure
- [x] Support floating placeholder
- [x] Refine floating placeholder behavior (support custom typography, add scale animation) 
- [x] Support accessory buttons
- [x] Support style per state (focused, invalid, disabled, blur)
- [x] Allow styling Label
- [x] Allow styling ValidationMessage
- [x] Connect to validation logic
- [x] Support checking validation in blur, change, state
- [ ] Consider adding presets for different UI looks
- [x] Fix Android spacing issues, make sure UI is aligned between the two Platforms
- [x] Support char counter
- [x] Support helper text (renamed to `hint`)
- [x] Support render validation message on top/bottom
- [x] Support multiline
- [x] Verify controlled behavior 
- [x] Support Accessibility
- [x] Fix ref warning
- [x] Fix typescript errors
- [x] Support theme props
- [x] RTL

**Notes for Review**
- The reason for this reimplementation is the fact this component contains legacy code which is impossible to maintain today
- There are lots of files to go over, no need to review generatedTypes files 
- In the new implementation I broke the component into very smaller components, each with a specific purpose.  
It should be much more readable to go over the code. And in general that's how I see we should write complex components. 
- I decided to omit specific UI/Functionality decisions that are internal to the Wix App in order to make the component more generic. We should be able to customize everything using ThemeManager.
- There are spacing differences between Android and iOS which are impossible to unify (height of the input for example) and  
I decided not to fix them - I believe fixing them lead us to lots of repeating spacings issues.


egorshulga:
Hey guys! I found out TextField is typed as any in the current release. Would this PR just replace the old one, would new TextField have correct typings, so it would be unnecessary to fix the old one?

ethanshar:
Hi @egorshulga 
Yes, once released the new TextField will include typings and better API overall.

demo/src/screens/componentScreens/IncubatorTextFieldScreen.tsx
```diff
@@ -0,0 +1,180 @@
...
+            validationMessagePosition={errorPosition}
+            validate={'email'}
+            validateOnChange
+            // validateOnStart
+            // validateOnBlur
+            fieldStyle={styles.withUnderline}
+          />
+
+          <Text h3 blue50 marginV-s4>
+            Colors By State
```
Inbal-Tish, line 117: This example is not working for me (testing iOS only for now)

src/incubator/TextField/CharCounter.tsx
```diff
@@ -0,0 +1,32 @@
+import React, {useContext} from 'react';
+import {TextStyle, StyleSheet} from 'react-native';
+import _ from 'lodash';
+import Text from '../../components/text';
+import FieldContext from './FieldContext';
+
+export interface CharCounterProps {
+  showCharCounter?: boolean;
```
Inbal-Tish, line 8: Why do you need this prop here? If it was passed to TextField you render it and if not you don't, no?

src/incubator/TextField/CharCounter.tsx
```diff
@@ -0,0 +1,32 @@
...
+  maxLength?: number;
```
Inbal-Tish, line 9: maxLength can be required (without the question mark), no?

src/incubator/TextField/AccessoryButton.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import Button, {ButtonPropTypes} from '../../components/button';
+
+export default (props: ButtonPropTypes) => {
+  return (
+    <Button link grey10 activeOpacity={props.onPress ? 0.6 : 1} {...props} />
```
Inbal-Tish, line 6: here you can basically override any of the default configs. Did you mean to give that much flexibility?

src/incubator/TextField/Label.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {TextStyle} from 'react-native';
+import {Colors} from '../../style';
+import Text, {TextPropTypes} from '../../components/text';
+import {ColorType, ValidationMessagePosition} from './types';
+import {getColorByState} from './Presenter';
+import FieldContext from './FieldContext';
+
+export interface LabelProps {
+  label?: string;
+  labelColor?: ColorType;
```
Inbal-Tish, line 11: Why not try to deprecate the 'labelColor' and 'labelStyle' props and stay with 'labelProps' only? (since 'style' can be passed as a prop and 'color' as a style prop...)

src/incubator/TextField/Label.tsx
```diff
@@ -0,0 +1,43 @@
+import React, {useContext} from 'react';
+import {TextStyle} from 'react-native';
+import {Colors} from '../../style';
+import Text, {TextPropTypes} from '../../components/text';
+import {ColorType, ValidationMessagePosition} from './types';
+import {getColorByState} from './Presenter';
+import FieldContext from './FieldContext';
+
+export interface LabelProps {
+  label?: string;
```
Inbal-Tish, line 10: Few things: 1) is 'label' is the best name for the COMPONENT? It is the 'title', right? The name 'title' gives the user a better clue to what this component is (a display of some text on top), while 'label' is very generic and a bit confusing in my opinion, like where this label is being displayed?
2) The 'label' PROP name - also a bit confusing as 'label' is the component, and what you're asking for is a string. I think 'text' is more suitable here.
3) Shouldn't the 'label' prop be required (not optional)?

src/incubator/TextField/CharCounter.tsx
ethanshar: No. If we'll make it require people will always have to pass the `maxLength` prop.

src/incubator/TextField/AccessoryButton.tsx
ethanshar: yes, completely. 
I don't see any reason to block the users from doing anything they want. 
In our private lib we will enforce it to our guidelines.

src/incubator/TextField/CharCounter.tsx
ethanshar: Are you talking about the `maxLength` prop, right? 
I need to declare it here so it's recognized (typescript wise) when I use it in the component.

src/incubator/TextField/Label.tsx
ethanshar: I always prefer to flatten and separate these two. 

In general, I believe specific props should be added for common cases and I think it's pretty common for the user to change the label style (without passing any other extra prop). 
Let's say it was very common to change the label letter spacings (it's clearly not), then I would consider adding a specific prop for it.

src/incubator/TextField/Label.tsx
ethanshar: Yes. `label` is the right name for this prop. I'm personally more familiar with `label` for this element. 
I verified with Yulia and the original guidelines used `label`. 
And the guidelines I used to structure the component also use this name
https://uxdesign.cc/text-fields-forms-design-ui-components-series-2b32b2beebd0

Last, most important, this is the name people use for it in the community, see here: 
https://callstack.github.io/react-native-paper/text-input.html
https://react-native-elements.github.io/react-native-elements/docs/input#props

Regarding the component name, I just name it like the prop (like in all the other internal components). It's an internal component, why is that a problem?

Regarding
> and what you're asking for is a string. I think 'text' is more suitable here.

Yes, I understand what you mean. The reason for that is so I can extend the TextField types nicely and use all the internal component to get the typings. 
This way  (almost) each component responsible to declare its part of the API

3) No. not all fields require label. You don't have to render it if you don't want to.

demo/src/screens/componentScreens/IncubatorTextFieldScreen.tsx
ethanshar: Fixed. There were couple of issues there. 
- I didn't implement the "error" state
- The default color was wrong

I also remove one of the example for "Color By State". 
One is good enough.

src/incubator/TextField/index.tsx
```diff
@@ -129,11 +130,12 @@ const TextField = (
               retainSpace
             />
           )}
-          <CharCounter
-            showCharCounter={showCharCounter}
-            maxLength={props.maxLength}
-            charCounterStyle={charCounterStyle}
-          />
+          {showCharCounter && !_.isUndefined(props.maxLength) &&
```
Inbal-Tish, line 133: I meant 'maxLength' should be required in CharCounter component so you don't need to check it here, the chack is in the component. 'showCharCounter' is of the TextField: 
{showCharCounter && 
  <CharCounter
    maxLength={props.maxLength}
    charCounterStyle={charCounterStyle}
 />}

src/incubator/TextField/CharCounter.tsx
```diff
@@ -6,22 +6,17 @@ import FieldContext from './FieldContext';
...
-  maxLength?: number;
+  maxLength: number;
   charCounterStyle?: TextStyle;
 }
 
-export default ({showCharCounter, maxLength, charCounterStyle}: CharCounterProps) => {
+export default ({maxLength, charCounterStyle}: CharCounterProps) => {
   const {value} = useContext(FieldContext);
-
-  if (showCharCounter && !_.isUndefined(maxLength)) {
```
Inbal-Tish: !_.isUndefined(maxLength) should stay in the CharCounter component.

generatedTypes/incubator/TextField/index.d.ts
```diff
@@ -8,7 +8,7 @@ import { LabelProps } from './Label';
 import { FieldStateProps } from './withFieldState';
 import { FloatingPlaceholderProps } from './FloatingPlaceholder';
 import { CharCounterProps } from './CharCounter';
-interface TextFieldProps extends InputProps, LabelProps, FloatingPlaceholderProps, FieldStateProps, ValidationMessageProps, CharCounterProps {
+interface TextFieldProps extends InputProps, LabelProps, FloatingPlaceholderProps, FieldStateProps, ValidationMessageProps, Omit<CharCounterProps, 'maxLength'> {
```
Inbal-Tish, line 11: I'm not too familiar with this usage, why do you need to omit 'maxLength'? Because it exists on both TextField and CharCounter components? Or because in one it's required and in the other is optional?

src/incubator/TextField/CharCounter.tsx
ethanshar: But we declared maxLength is required.. so what's the point in checking it..

ethanshar: when I do that.. 
I get a TS error in TextField code. 
Because I try to pass an optional value (maxLength) to a required prop (the maxLength in CharCounter)

src/incubator/TextField/Presenter.ts
```diff
@@ -0,0 +1,21 @@
...
+    if (context?.disabled) {
+      finalColor = color?.disabled;
+    } else if (context?.isFocused) {
+      finalColor = color?.focus;
+    } else {
+      finalColor = color?.default;
+    }
+  }
+
+  return finalColor || Colors.green10;
```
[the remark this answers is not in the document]
ethanshar: leftovers. Fixed and fixed.

src/incubator/TextField/Input.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const context = useContext(FieldContext);
+  const placeholder = !context.isFocused ? props.placeholder : (hint || props.placeholder);
+  return (
+    <TextInput
+      style={[styles.input, style]}
+      {...props}
+      placeholder={placeholder}
+      ref={forwardedRef}
+      underlineColorAndroid="transparent"
+      accessibilityState={props.editable === false ? {disabled: true} : undefined}
```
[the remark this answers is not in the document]
ethanshar: It used to be plural in rn59, [see here](https://reactnative.dev/docs/0.59/touchablewithoutfeedback#accessibilitystates)
Not it's singular, since rn61 [see here](https://reactnative.dev/docs/0.62/touchablewithoutfeedback#accessibilitystate)

Anyway, checked it and it reads the correct trait.

src/incubator/TextField/Input.tsx
```diff
@@ -0,0 +1,42 @@
...
+    />
+  );
+};
+
+export default Input;
+
+const styles = StyleSheet.create({
+  input: {
+    textAlign: Constants.isRTL ? 'right' : 'left',
```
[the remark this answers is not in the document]
ethanshar: It's an RN issue and will be fixed in RN63

generatedTypes/incubator/TextField/index.d.ts
ethanshar: because in one it's required and in the other is optional?

demo/src/screens/componentScreens/IncubatorTextFieldScreen.tsx
```diff
@@ -0,0 +1,170 @@
...
+            Char Counter
+          </Text>
+
+          <TextField
+            label="Description"
+            placeholder="Enter text..."
+            multiline
+            showCharCounter
+            charCounterStyle={{color: Colors.blue30}}
+            maxLength={225}
```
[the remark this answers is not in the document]
ethanshar: done

src/incubator/TextField/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }: InternalTextFieldProps
+) => {
+  const context = useMemo(() => {
+    return {...fieldState, disabled: props.editable === false};
+  }, [fieldState, props.editable]);
+
+  return (
+    <FieldContext.Provider value={context}>
+      <View style={containerStyle}>
+        <Label
```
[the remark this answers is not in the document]
ethanshar: I believe render methods is less clean in a function components, that's why I moved to breaking each render to an actual component. 

In function component you usually need to wrap internal methods with the `useCallback` hook and I wanted to avoid that..

src/incubator/TextField/CharCounter.tsx
Inbal-Tish: Make sense... Ok, let's just leave it as it was. If it is not meant to be standalone component then it doesn't really matter...

generatedTypes/incubator/TextField/index.d.ts
Inbal-Tish: I see. So let's leave it optional then. It's not worth it.

src/incubator/TextField/Input.tsx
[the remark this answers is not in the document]
Inbal-Tish: But does it update if say it starts as disabled=true and then turned to disabled=false?

src/incubator/TextField/CharCounter.tsx
```diff
@@ -0,0 +1,27 @@
+import React, {useContext} from 'react';
+import {TextStyle, StyleSheet} from 'react-native';
+import _ from 'lodash';
+import Text from '../../components/text';
+import FieldContext from './FieldContext';
+
+export interface CharCounterProps {
+  showCharCounter?: boolean;
+  maxLength: number;
```
Inbal-Tish, line 9: Do you want to revert the change here and keep it optional?

src/incubator/TextField/Input.tsx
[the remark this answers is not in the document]
ethanshar: Ok, I fixed it. Now it updates accordingly.

src/incubator/TextField/CharCounter.tsx
ethanshar: Changed it back to optional
and moved the check for `maxLength` to CharCounter

src/incubator/TextField/ValidationMessage.tsx
```diff
@@ -0,0 +1,36 @@
+import React, {useContext} from 'react';
+import {TextStyle, StyleSheet} from 'react-native';
+import Text from '../../components/text';
+import FieldContext from './FieldContext';
+
+export interface ValidationMessageProps {
+  enableErrors?: boolean;
+  validationMessage?: string;
+  validationMessageStyle?: TextStyle;
+  retainSpace?: boolean;
```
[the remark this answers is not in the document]
ethanshar: This how it started and I ended up with this solution. 
If i'll move the logic out, I will also need to move out the check for `context.isValid` (cause they go together). 
And the whole idea was to keep the logic of each internal component inside and remove complexity from the main TextField component. 

It's not that it's not possible, but It's cleaner way. 

Also, Why `enableErrors` is redundant? 
That's the only way for the user to decided whether to show the validation message or not.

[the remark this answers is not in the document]
Inbal-Tish: The 'enableErrors' prop issue is the same as we discussed before with TextField props that affect the render of the subcomponent. Since you're taking a different approach it doesn't matter anymore. I'll resolve this comment as well.

PR #855, opened 2020-07-16
Demo screen support injected ui

mendyEdri:
Added support for the demo screen to have injected UI as extra settings.

### DIFF
--- a/demo/src/screens/MainScreen.js
+++ b/demo/src/screens/MainScreen.js
@@ -37,10 +37,12 @@ class MainScreen extends Component {
     super(props);
 
     const data = props.navigationData || navigationData;
-
+    const extraSettingsUI = props.extraSettingsUI;
+    
     this.state = {
       currentPage: 0,
-      filteredNavigationData: data
+      filteredNavigationData: data,
+      extraSettingsUI
     };
 
     this.filterExplorerScreens = _.throttle(this.filterExplorerScreens, 300);
@@ -64,7 +66,7 @@ class MainScreen extends Component {
     if (buttonId === 'uilib.settingsButton') {
       this.pushScreen({
         name: 'unicorn.Settings',
-        passProps: {navigationData: data, playground: this.props.playground}
+        passProps: {navigationData: data, playground: this.props.playground, extraSettingsUI: this.state.extraSettingsUI}
       });
     }
   };
--- a/demo/src/screens/MainScreen.js
+++ b/demo/src/screens/MainScreen.js
@@ -15,9 +15,10 @@ class MainScreen extends Component {
   static propTypes = {
     containerStyle: ViewPropTypes.style,
     renderItem: PropTypes.func,
-    pageStyle: ViewPropTypes.style,
-    showRefreshAppMessage: PropTypes.bool
+    pageStyle: ViewPropTypes.style    
   };
+
+  settingsScreenName = 'unicorn.Settings';
 
   static options() {
     return {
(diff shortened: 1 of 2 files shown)


ethanshar:
@mendyEdri When and if this ready, assign me as a reviewer 👍

src/components/avatar/__tests__/index.spec.js
```diff
@@ -1,5 +1,6 @@
 import {Avatar} from '../index';
 import {Colors} from '../../../style';
+import {BADGE_SIZES} from '../../badge';
```
ethanshar, line 3: how this is related to the PR?

src/components/avatar/index.tsx
```diff
@@ -222,14 +222,21 @@ class Avatar extends PureComponent<AvatarPropTypes> {
     return _.get(this.props, 'badgeProps.backgroundColor') || statusColor || onlineColor;
   }
 
-  getBadgeSize = () => _.get(this.props, 'badgeProps.size', DEFAULT_BADGE_SIZE);
```
ethanshar, line 225: Same here, I think your PRs got mixed up
make sure when create a new branch to always create it out of `master`

src/components/avatar/__tests__/index.spec.js
mendyEdri: it's not, fixed.

src/components/avatar/index.tsx
mendyEdri: sure, fixed

demo/src/screens/SettingsScreen.js
```diff
@@ -5,19 +5,21 @@ import {StyleSheet, I18nManager} from 'react-native';
...
     super(props);
 
     const data = props.navigationData || navigationData;
     const playground = props.playground || playgroundScreen;
-    
+    const extraSettingsUI = props.extraSettingsUI;
+
     this.state = {
       showRefreshMessage: false,
+      extraSettingsUI,
```
ethanshar, line 22: Why are you keeping `extraSettingsUI` in the component state? Why not use it directly from prop in your render function ?

demo/src/screens/MainScreen.js
```diff
@@ -33,21 +34,32 @@ class MainScreen extends Component {
...
     this.filterExplorerScreens = _.throttle(this.filterExplorerScreens, 300);
 
     Navigation.events().bindComponent(this);
   }
 
+  shouldComponentUpdate = (nextProps, nextState) => {  
+    this.updateShowRefreshAppMessage(nextProps.showRefreshAppMessage);
+    return true;
+  }
+
```
ethanshar, line 62: This will be called before each render. 
Not sure that's the right way to do that.. 

`shouldComponentUpdate` is meant to have calculation regarding whether the component should render or not. 
If anything you probably want to move this part of code to `componentDidUpdate` .

Anyway, maybe there's a better way to trigger a toast from the private repo.. 
I think the wiring here is a little more complicated then is should be.

demo/src/screens/SettingsScreen.js
```diff
@@ -99,10 +117,13 @@ class SettingsScreen extends Component {
               <Text text80 dark20>Force RTL</Text>
             </View>
           </View>
+          
+          {extraSettingsUI?.()}
```
ethanshar, line 121: Why do you render extraSettingsUI twice?

demo/src/screens/SettingsScreen.js
mendyEdri: fixed

demo/src/screens/SettingsScreen.js
mendyEdri: fixed.
(public/private copy mistake)

demo/src/screens/MainScreen.js
mendyEdri: `componentDidUpdate` - gotcha. 

@ethanshar  Triggering the toast from the private repo - you mean that the toast should be rendered in the private repo?
or some different - easier wiring?

ethanshar: Yes, something like that. 
In the private repo we have an infrastructure for trigger Toast from anywhere in the code, we call it global toast. 
It's a functionality we offer to One App and we use it ourselves in the demo app. so it might easier for you to use that.. 
Search in the private repo for "toggleToast" in the MainScreen.

demo/src/screens/SettingsScreen.js
ethanshar: 👍

demo/src/screens/SettingsScreen.js
```diff
@@ -15,9 +14,11 @@ class SettingsScreen extends Component {
 
     const data = props.navigationData || navigationData;
     const playground = props.playground || playgroundScreen;
-    
+    const extraSettingsUI = props.extraSettingsUI;
```
ethanshar, line 17: This part is also redundant, no need to store this prop in state cause we're rendering it directly from props.

PR #856, opened 2020-07-16
TS/fix carousel-related typing errors

VernonHawk:
- 'limitShownPages' does not exist in type 'PageControlProps'
- 'key' does not exist in type 'PageControlProps'
- Property 'pageControlPositions' does not exist on type 'typeof Carousel'

---
This change is [Reviewable](https://reviewable.io/reviews/wix/react-native-ui-lib/856)

### DIFF
--- a/typings/components/PageControl.d.ts
+++ b/typings/components/PageControl.d.ts
@@ -3,7 +3,8 @@ import {GestureResponderEvent, StyleProp, ViewStyle} from 'react-native';
 import {BaseComponent} from '../commons';
 import {ColorValue} from '../style/colors';
 
-export interface PageControlProps {
+export interface PageControlProps extends JSX.IntrinsicAttributes {
+  limitShownPages: boolean;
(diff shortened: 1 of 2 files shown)


typings/components/PageControl.d.ts
```diff
@@ -3,7 +3,8 @@ import {GestureResponderEvent, StyleProp, ViewStyle} from 'react-native';
 import {BaseComponent} from '../commons';
 import {ColorValue} from '../style/colors';
 
-export interface PageControlProps {
+export interface PageControlProps extends JSX.IntrinsicAttributes {
+  limitShownPages: boolean;
```
ethanshar, line 7: I believe this should be an optional prop

VernonHawk: Hm, I didn't notice that all of these are optional, sorry

VernonHawk, line 7:
```suggestion
  limitShownPages?: boolean;
```

PR #860, opened 2020-07-20
Drawer - trigger onToggleSwipeLeft on releases

Inbal-Tish:
Fix for trigger event on release instead of just threshold

### DIFF
--- a/src/components/drawer/Swipeable.tsx
+++ b/src/components/drawer/Swipeable.tsx
@@ -90,7 +90,7 @@ export default class Swipeable extends Component<PropType, StateType> {
     // 1 -> closing to the left
     // -1 -> closing to the right
     this.rowState = 0;
-    this.dragFired = false;
+    this.dragThresholdReached = false;
 
     this.state = {
       dragX,
@@ -112,12 +112,12 @@ export default class Swipeable extends Component<PropType, StateType> {
   _handleDrag = (e) => {
     const {onToggleSwipeLeft} = this.props;
     
-    if (onToggleSwipeLeft && !this.dragFired) {
+    if (onToggleSwipeLeft && !this.dragThresholdReached) {
       const {rowWidth, leftWidth} = this.state;
       const x = e.nativeEvent.translationX;
       const threshold = rowWidth * LEFT_TOGGLE_THRESHOLD;
       if (x >= threshold && x < threshold + 10) {
-        this.dragFired = true;
+        this.dragThresholdReached = true;
         onToggleSwipeLeft({rowWidth, leftWidth, dragX: x});
       }
     }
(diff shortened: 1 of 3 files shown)


src/components/drawer/Swipeable.tsx
```diff
@@ -303,8 +303,8 @@ export default class Swipeable extends Component<PropType, StateType> {
       }
     });
 
-    if (toValue === rowWidth * LEFT_TOGGLE_THRESHOLD && onToggleSwipeLeft) {
-      onToggleSwipeLeft({rowWidth, leftWidth});
+    if ((toValue === rowWidth * LEFT_TOGGLE_THRESHOLD || this.dragFired) && onToggleSwipeLeft) {
+      onToggleSwipeLeft({rowWidth, leftWidth, released: true});
```
ethanshar, line 307: You're triggering `onToggleSwipeLeft` twice. In `_handleDrag` and in `_animateRow`
Not sure if that intentional, I doesn't break anything, but it might be a possible future bug unless I'm missing something

src/components/drawer/Swipeable.tsx
```diff
@@ -303,8 +303,8 @@ export default class Swipeable extends Component<PropType, StateType> {
       }
     });
 
-    if (toValue === rowWidth * LEFT_TOGGLE_THRESHOLD && onToggleSwipeLeft) {
-      onToggleSwipeLeft({rowWidth, leftWidth});
+    if ((toValue === rowWidth * LEFT_TOGGLE_THRESHOLD || this.dragFired) && onToggleSwipeLeft) {
```
ethanshar, line 306: At first I wasn't sure where this comes from (`this.dragFired`) I think the PR is missing some of the code changes. 
Anyway, the name of the variable was a little confusing to me. 
Is the flag purpose is to mark when the drag exceed threshold? if so, consider renaming it..

src/components/drawer/Swipeable.tsx
Inbal-Tish: Yes. I'm triggering `onToggleSwipeLeft` twice, one with release = false and once with release = true so I can first change the item's position and only on release fire the `onToggleSwipeLeft` callback and the haptic trigger.

ethanshar: Got ya!

src/components/drawer/Swipeable.tsx
Inbal-Tish: Done.

Inbal-Tish:
@ethanshar This is a second PR for this ticket. The first was merged and this is the changes UX ask for so maybe that's why you think some of the changes are not showing in the code?

PR #871, opened 2020-07-29
Picker multi select support

mendyEdri:
Support Picker multi-selection


src/components/picker/index.js
```diff
@@ -183,10 +183,20 @@ class Picker extends BaseComponent {
...
+    return mode === Picker.modes.MULTI && showExpandableModal;
+  }
+
   getLabel() {
     const {value} = this.state;
     const {getLabel, getItemLabel} = this.props;
 
+    if (this.isDialogAndMultiSelectInEditing()) {
+      return undefined;
+    }
```
ethanshar, line 198: Why are we returning undefined here..?

mendyEdri: We don't want to update the outer-textfield with the selection while the user selecting, until they tap on 'Done'

ethanshar: Ohh ok. 
I think that we need to solve the root cause of this issue. 
By returning "undefined" we're just hiding the value, but technically it's still being store in the state's value. 

I think the right place to handle this issue is in the `toggleItemSelection` function.

mendyEdri: I'm thinking out loud:
there is a single state to represent the selected items.
then callbacks are called upon user events and then, it invokes with the current state value.

The only place reacting to those state changes is this label, and I didn't want to make extra state value to hold the same data.

I can make it more intuitive by passing to `getLabel()` the source it should work on (props.value / state.value).

WDYT?

ethanshar: But fixing the `getLabel` doesn't fix the source of truth in this case. the value in state. 

Technically if in the future we'll add a new feature that depends on the state value, this feature will be broken because it will depend on a wrong value. 

I believe that by fixing what we store in the state's value we're fixing the root problem 
and not worry about whatever logic depends on it.

PR #876, opened 2020-08-02
Added accessibility prop to TextField's right button

mendyEdri:
TextField component now accepts right button accessibilityLabel

### DIFF
--- a/src/components/inputs/TextField.js
+++ b/src/components/inputs/TextField.js
@@ -171,7 +171,8 @@ export default class TextField extends BaseInput {
       iconSource: RNImage.propTypes.source,
       iconColor: PropTypes.string,
       onPress: PropTypes.func,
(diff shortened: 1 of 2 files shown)


src/components/inputs/TextField.js
```diff
@@ -609,7 +610,7 @@ export default class TextField extends BaseInput {
       const {style, iconSource, iconColor} = rightButtonProps;
```
ethanshar, line 610: instead of passing all the props included in `rightButtonProps`, pass on the "rest" of the props that were not used specifically. 
```
const {style, iconSource, iconColor, ...others} = rightButtonProps;
```

```
<TouchableOpacity {...others} style={[this.styles.rightButton, style]} onPress={this.onPressRightButton}>

```

PR #880, opened 2020-08-04
add testID

lidord-wix:
add testID on the Modal that renders when onBackgroundPress prop sent in Hint

### DIFF
--- a/src/components/hint/index.js
+++ b/src/components/hint/index.js
@@ -398,7 +398,7 @@ class Hint extends BaseComponent {
   }
 
   render() {
-    const {visible, onBackgroundPress} = this.props;
+    const {visible, onBackgroundPress, testID} = this.props;
 
     if (!visible) {
       return this.props.children;
@@ -414,6 +414,7 @@ class Hint extends BaseComponent {
             transparent
             onBackgroundPress={onBackgroundPress}
             onRequestClose={onBackgroundPress}
+            testID={testID}
           >
             {this.renderHintContainer()}
           </Modal>
--- a/src/components/hint/index.js
+++ b/src/components/hint/index.js
@@ -358,7 +358,7 @@ class Hint extends BaseComponent {
             this.getHintPadding()
           ]}
           pointerEvents="box-none"
-          testID={testID}
+          testID={`${testID}.animatableView`}
         >
           {this.renderHintTip()}
           <View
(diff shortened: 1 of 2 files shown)


src/components/hint/index.js
```diff
@@ -398,7 +398,7 @@ class Hint extends BaseComponent {
   }
 
   render() {
-    const {visible, onBackgroundPress} = this.props;
+    const {visible, onBackgroundPress, testID} = this.props;
```
ethanshar, line 401: couple of things regarding testID (: 

testID is an identifier we use for e2e tests and it has to be unique throughout the render tree. 
That means that if two elements will have the same testID (in a screen) the test will fail. 

If you'll search in the code of this component, you'll see we are already passing the testID. 
In these cases, when we need to tag more than one element in the component with a testID. we just concatenate the user's `testID` with the element type/name/id. 
For instance in your case it can be 
`${testID}.modal` or `${testID}.background`, etc... 

Another thing that is worth noticing in this case is that you're passing the testID to the `Modal` component and I'm not sure the Modal handles the `testID`. 
see this: https://stackoverflow.com/questions/50117265/detox-testing-visibility-of-modal-in-react-native
So we might want to treat the Modal component as well.

src/components/hint/index.js
```diff
@@ -358,7 +358,7 @@ class Hint extends BaseComponent {
             this.getHintPadding()
           ]}
           pointerEvents="box-none"
-          testID={testID}
+          testID={`${testID}.animatableView`}
         >
```
ethanshar, line 362: Don't change this. 
There's always a main element that gets the original testID that the user pass. 
Also, changing this is a breaking change cause it will break tests that depend on that.

src/components/hint/index.js
```diff
@@ -414,6 +414,7 @@ class Hint extends BaseComponent {
             transparent
             onBackgroundPress={onBackgroundPress}
             onRequestClose={onBackgroundPress}
+            testID={`${testID}.modal`}
```
ethanshar, line 417: This is good 👍

src/components/modal/index.js
```diff
@@ -33,7 +33,11 @@ export default class Modal extends BaseComponent {
     /**
      * the background color of the overlay
      */
-    overlayBackgroundColor: PropTypes.string
+    overlayBackgroundColor: PropTypes.string,
+    /**
+     * The modal's end-to-end test identifier
+     */
+    testID: PropTypes.string
```
ethanshar, line 40: Just declaring the prop doesn't really do anything. 
You will actually need to use the testID inside the component. (See the link for I added in the previous comment to stack overflow)

React Native's Modal component doesn't support testID internally, so we will need to pass it ourselves to the wrapper view inside the Modal.

PR #883, opened 2020-08-05
Fix Picker right item position side-effect

mendyEdri:


src/components/inputs/TextField.js
```diff
@@ -361,6 +361,10 @@ export default class TextField extends BaseInput {
     return enableErrors && error;
   }
 
+  getErrorMessage() {
```
mendyEdri, line 364: Couldn't use the parent's original `getErrorMessage` since `generateStyles` getting called before the state created (in BaseComponent) and it produces error.

ethanshar:
Tests are failing for some reason :/

PR #891, opened 2020-08-12
Fix/dialog animation

ethanshar:
WOAUILIB-972
The purpose of this fix is to improve the dialog entrance/exist animation. 
On the way I fixed small stuff I found
- In `PanResponderView` I'm now using `View`'s internal `animation` prop instead of the Container ternary. 
- In `DialogDismissibleView` I removed unnecessary state update (`isAnimating` doesn't affect render)

In the `Dialog` component
- I changed the Modal `animationType` to "none" so we can have better control over the background fading animation 
- I created a separate component, `OverlayFadingBackground`, with a single purpose - the handle the background fading transition. (Technically, now  the background fade animation and the dialog entrance/exit animation happen together.)
- Also, minor tweaks with the animation duration, easing, etc..

### DIFF
--- a/src/components/dialog/DialogDismissibleView.js
+++ b/src/components/dialog/DialogDismissibleView.js
@@ -54,16 +54,6 @@ class DialogDismissibleView extends PureComponent {
     this.width = Constants.screenWidth;
     this.height = Constants.screenHeight;
     this.hiddenLocation = this.getHiddenLocation(0, 0);
-    this.animationStyle = {
-      transform: [
-        {
-          translateX: this.hiddenLocation.left
-        },
-        {
-          translateY: this.hiddenLocation.top
-        }
-      ]
-    };
   }
 
   componentDidUpdate(prevProps) {
@@ -192,7 +182,6 @@ class DialogDismissibleView extends PureComponent {
     this.thresholdY = this.height / 2;
     this.ref.measureInWindow((x, y) => {
       this.hiddenLocation = this.getHiddenLocation(x, y);
-      this.animationStyle = this.getAnimationStyle();
       this.animateTo(1, this.onAnimationEnd);
     });
   };
--- a/src/components/dialog/DialogDismissibleView.js
+++ b/src/components/dialog/DialogDismissibleView.js
@@ -159,7 +159,7 @@ class DialogDismissibleView extends PureComponent {
     Animated.timing(this.animatedValue, {
       toValue,
       duration: 400,
-      easing: Easing.bezier(0.65, 0, 0.35, 1),
+      easing: Easing.bezier(0.2, 0, 0.35, 1),
       useNativeDriver: true
     }).start(animationEndCallback);
   };
--- a/src/components/dialog/DialogDismissibleView.js
+++ b/src/components/dialog/DialogDismissibleView.js
@@ -40,10 +40,10 @@ class DialogDismissibleView extends PureComponent {
     super(props);
 
     this.setInitialValues();
+    this.isAnimating = false;
     this.state = {
       visible: props.visible,
-      hide: false,
-      isAnimating: false
+      hide: false
     };
   }
 
@@ -69,7 +69,7 @@ class DialogDismissibleView extends PureComponent {
   componentDidUpdate(prevProps) {
     const {isPanning, dragDeltas, swipeDirections} = this.props.context; // eslint-disable-line
     const {dragDeltas: prevDragDeltas, swipeDirections: prevSwipeDirections} = prevProps.context; // eslint-disable-line
-    const {hide, isAnimating} = this.state;
+    const {hide} = this.state;
 
     if (
       isPanning &&
@@ -87,7 +87,7 @@ class DialogDismissibleView extends PureComponent {
       this.onSwipe(swipeDirections);
     }
 
-    if (hide && !isAnimating) {
+    if (hide && !this.isAnimating) {
       this.hide();
     }
   }
@@ -127,7 +127,7 @@ class DialogDismissibleView extends PureComponent {
   };
 
   onAnimationEnd = () => {
-    this.setState({isAnimating: false});
+    this.isAnimating = false;
   };
 
   getHiddenLocation = (left, top) => {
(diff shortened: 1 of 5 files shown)


M-i-k-e-l:
> Technically, now the background fade animation and the dialog entrance/exit animation happen together

I believe we had a bug with that, I cannot remember ATM what it was (I think it was on iOS), I hope our e2e will find it (or it doesn't happen :) ).

src/components/dialog/OverlayFadingBackground.tsx
```diff
@@ -0,0 +1,46 @@
...
+  modalVisibility,
+  overlayBackgroundColor
+}: Props) => {
+  const fadeAnimation = useRef(new Animated.Value(0)).current;
+
+  useEffect(() => {
+    if (!dialogVisibility) {
+      animateFading(0);
+    }
+  }, [dialogVisibility]);
```
M-i-k-e-l, line 22: The hooks plugin says you are missing `animateFading`

src/components/dialog/OverlayFadingBackground.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  useEffect(() => {
+    if (modalVisibility) {
+      animateFading(1);
+    }
+  }, [modalVisibility]);
```
M-i-k-e-l, line 28: The hooks plugin says you are missing `animateFading`

src/components/dialog/OverlayFadingBackground.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const animateFading = useCallback((toValue) => {
+    Animated.timing(fadeAnimation, {
+      toValue,
+      duration: 400,
+      useNativeDriver: true
+    }).start();
+  }, []);
```
M-i-k-e-l, line 36: The hooks plugin says you are missing `fadeAnimation`

src/components/dialog/OverlayFadingBackground.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const style = useMemo(() => {
+    return {
+      opacity: fadeAnimation,
+      backgroundColor: overlayBackgroundColor
+    };
+  }, [overlayBackgroundColor]);
```
M-i-k-e-l, line 43: The hooks plugin says you are missing `fadeAnimation`

src/components/dialog/DialogDismissibleView.js
```diff
@@ -156,14 +156,12 @@ class DialogDismissibleView extends PureComponent {
...
+    Animated.timing(this.animatedValue, {
       toValue,
-      duration: 400,
-      easing: Easing.bezier(0.65, 0, 0.35, 1),
+      duration: 300,
+      easing: Easing.bezier(0.2, 0, 0.35, 1),
       useNativeDriver: true
-    });
-
-    this.setState({isAnimating: true}, () => animation.start(animationEndCallback));
```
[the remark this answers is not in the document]
ethanshar: DialogDismissibleView is an internal component of dialog and being used only with isAnimated = true
Maybe it's indeed redundant.. should I remove it ?

src/components/dialog/DialogDismissibleView.js
```diff
@@ -241,7 +239,7 @@ class DialogDismissibleView extends PureComponent {
       <View ref={r => (this.ref = r)} style={containerStyle} onLayout={this.onLayout}>
         <PanResponderView
           // !visible && styles.hidden is done to fix a bug is iOS
-          style={[style, this.animationStyle, !visible && styles.hidden]}
+          style={[style, this.getAnimationStyle(), !visible && styles.hidden]}
```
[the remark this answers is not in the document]
ethanshar: yes, and correct.. I will remove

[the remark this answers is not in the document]
ethanshar: done

src/components/dialog/OverlayFadingBackground.tsx
ethanshar: fixed the warnings..
But I gotta say, the plugin is not that smart. 
It just blindly check what's being used and ask to add it as a dependency, which sometimes is not really necessary.

src/components/dialog/OverlayFadingBackground.tsx
ethanshar: fixed.

src/components/dialog/OverlayFadingBackground.tsx
ethanshar: fixed.

src/components/dialog/OverlayFadingBackground.tsx
ethanshar: fixed.

src/components/dialog/DialogDismissibleView.js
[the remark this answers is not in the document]
M-i-k-e-l: Not sure if it's a typo or not, I wrote `isAnimating` and not `isAnimated` as in `this. isAnimating`.
It is checked in:
```
if (hide && !this.isAnimating) {
  this.hide();
}
```
But since it is always false, it seems it can be removed.

src/components/dialog/OverlayFadingBackground.tsx
M-i-k-e-l: I know, and it has bugs sometimes (see [AutoLockScrollView](https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/WithScrollEnablerScreen/AutoLockScrollView.tsx#L41), it cannot recognize that not all the `props` need to be checked), but it did find these which are bugs, I think, so it's worth using IMHO.

src/components/dialog/DialogDismissibleView.js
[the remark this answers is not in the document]
ethanshar: My previous comment was about something else.. I got confused. nvm..
Anyway, I will remove this.isAnimating

PR #898, opened 2020-08-17
Fix/button RTL support

lidord-wix:
Button is now support RTL which allows users to flip icon inside button.

### DIFF
--- a/src/components/button/index.tsx
+++ b/src/components/button/index.tsx
@@ -459,7 +459,7 @@ class Button extends PureComponent<Props, ButtonState> {
       if (typeof iconSource === 'function') {
         return iconSource(iconStyle);
       } else {
-        return <Image source={iconSource} supportRTL={supportRTL} style={[iconStyle]}/>;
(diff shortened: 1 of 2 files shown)


src/components/button/index.tsx
```diff
@@ -406,10 +410,12 @@ class Button extends PureComponent<Props, ButtonState> {
...
   getIconStyle() {
-    const {disabled, iconStyle: propsIconStyle, iconOnRight} = this.props;
+    const {disabled, iconStyle: propsIconStyle, iconOnRight, supportRTL} = this.props;
     const size = this.props.size || DEFAULT_SIZE;
+    const shouldFlipRTL = supportRTL && Constants.isRTL;
     const iconStyle: ImageStyle = {
-      tintColor: this.getLabelColor()
+      tintColor: this.getLabelColor(),
+      transform: shouldFlipRTL ? [{scaleX: -1}] : undefined
     };
```
ethanshar, line 419: The Image already supports this feature, why not just pass forward the `supportRTL` prop

ethanshar (approved):

PR #905, opened 2020-08-19
Drawer - testID

Inbal-Tish:
Passing the testID to the appropriate Drawer's subcomponent; Adding testIDs to demo screen


M-i-k-e-l (approved):
Approved, merge if the usage is valid with detox 17: please verify that if you are `tapping` on a view it is indeed touchable itself (and not a parent\child of the touchable), since it will break in detox 17.

PR #916, opened 2020-08-26
FloatingButton - Remove UNSAFE method

Inbal-Tish:
FloatingButton - Turning component to a fully controlled component by removing state management.

No breaking changes!

### DIFF
--- a/src/components/floatingButton/index.tsx
+++ b/src/components/floatingButton/index.tsx
@@ -121,6 +121,9 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
     if (this.firstLoad === true && !this.initialVisibility) {
       return false;
     }
+    if (!visible && withoutAnimation) {
+      return false;
+    }
 
     return (
       <Container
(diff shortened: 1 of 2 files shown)


src/components/floatingButton/index.tsx
```diff
@@ -57,44 +51,21 @@ const gradientImage = () => require('./gradient.png');
...
+  
   constructor(props: FloatingButtonProps) {
     super(props);
 
-    this.state = {
-      shouldAnimateHide: false,
-      isVisible: props.visible,
-      animating: false
-    };
+    this.initialState = props.visible;
```
M-i-k-e-l, line 63: Maybe `initialVisibility`?

M-i-k-e-l (commented):
Hiding does not work when `withoutAnimation = true`

src/components/floatingButton/index.tsx
Inbal-Tish: Done

M-i-k-e-l:
> Hiding does not work when `withoutAnimation = true`

I think you've missed this one.

Inbal-Tish:
@M-i-k-e-l Thanks! Good catch! Fixed.

M-i-k-e-l (approved):

Issue #874, opened 2020-07-30
There is no way to change the background Color of TabBar


RyanCloudFM:
Hi, 

My app is using the TabBar component, it is also has a dark and light theme. 

From what I can tell, the TabBar background is always white and there's simply nothing I can do about it. I've tried style, indicatorStyle, setting the backgroundColor prop, nothing works! 

![image](https://user-images.githubusercontent.com/33055882/88932795-73438480-d276-11ea-86d6-31e3f76f352a.png)

RyanCloudFM:
Hello, 

Thanks for the work on 795528d90b7f1d01a939e5d8c85d99367db594f1, however, this only partially solves the problem, this has revealed another problem in that the gradient color cannot be changed.

![image](https://user-images.githubusercontent.com/33055882/91310463-cfcc8d80-e7a9-11ea-964e-dfec3041c2f5.png)

I assume it would be the same solution for the ScollBar component which has the gradientColor prop?

Inbal-Tish:
@RyanCloudFM Hi. The plan is to move TabBar component to use ScrollBar component. Once that will happen you'll be able to pass 'gradientColor' prop.

RyanCloudFM:
@Inbal-Tish awesome, thanks for the update!

ethanshar:
Released in v5.14.0

jamesjara:
This won't work for the tabController tabbar https://github.com/wix/react-native-ui-lib/blob/master/src/components/tabController/TabBar.js#L426

ethanshar:
@jamesjara 
There's a separate upcoming PR for supporting TabController.TabBar 
It should be released next version

jamesjara:
thanks @ethanshar 
I'm using the following post-install script (it's quicker than forking)
```
fix-rn-ui-lib.sh
# This patch will fix the background color in the tabbar
sed -i '' '432s/white/transparent/g' node_modules/react-native-ui-lib/src/components/tabController/TabBar.js
```

arnabJ:
@ethanshar I am still facing the same issue.
Changing the backgroundColor of TabController.TabBar shows a White/Light colored gradient on the far side end of the TabBar.
gradientColor prop doesn't works to hide that.
This is my code:
`<TabController.TabBar
                    backgroundColor={
                      Appearance.getColorScheme() === 'dark'
                        ? BG_COMPONENT_DARK
                        : BG_COMPONENT_LIGHT
                    }
                    gradientColor={
[...]

PR #917, opened 2020-08-26
TabBar - Adding 'backgroundColor' prop

Inbal-Tish:
Passing 'backgroundColor' prop to adjust the entire background and gradient color.


Inbal-Tish:
#874 gradient color issue

ethanshar (approved):

PR #920, opened 2020-08-31
Feat/modal migrate to typescript

M-i-k-e-l:
Migrate Modal (and dependency) to typescript


src/components/modal/index.tsx
```diff
@@ -91,3 +101,10 @@ const styles = StyleSheet.create({
 });
 
 Modal.TopBar = TopBar;
+
+export default asBaseComponent<
+  ModalPropTypes,
+  {
+    TopBar: typeof TopBar;
+  }
```
ethanshar, line 109: Consider changing to `typeof Modal` it will automatically pass the type of Modal class with all of its static member. 
This is what we do in our Button component, let's see if it works here as well.

src/components/modal/index.tsx
```diff
@@ -1,44 +1,51 @@
...
- * @extends: Modal
- * @extendslink: https://facebook.github.io/react-native/docs/modal.html
- * @gif: https://media.giphy.com/media/3oFzmfSX8KgvctI4Ks/giphy.gif
- * @example: https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/ModalScreen.js
- */
-export default class Modal extends BaseComponent {
-  static displayName = 'Modal';
-  static propTypes = {
+export {ModalTopBarPropTypes};
+export interface ModalPropTypes extends ModalProps {
```
ethanshar, line 11: Let's rename to `ModalProps`
In general, I would like to unify the naming of our component typings and make it shorter. 
I know there are already some component that the typings are export as `...PropTypes`, we'll handle them later.

src/components/modal/index.tsx
```diff
@@ -1,44 +1,51 @@
...
-import React from 'react';
-import {StyleSheet, Modal as RNModal, TouchableWithoutFeedback} from 'react-native';
+import React, {Component} from 'react';
+import {StyleSheet, Modal as RNModal, ModalProps, TouchableWithoutFeedback, GestureResponderEvent} from 'react-native';
 import {BlurView} from '@react-native-community/blur';
 import {Constants} from '../../helpers';
-import {BaseComponent} from '../../commons';
-import TopBar from './TopBar';
+import {asBaseComponent} from '../../commons/new';
+import TopBar, {ModalTopBarPropTypes} from './TopBar';
```
ethanshar, line 7: rename `ModalTopBarPropTypes` to `ModalTopBarProps` 
(see comment below)

src/components/modal/index.tsx
M-i-k-e-l: Done

src/components/modal/index.tsx
M-i-k-e-l: Done

src/components/modal/index.tsx
M-i-k-e-l: Done

M-i-k-e-l:
I've just now noticed that we send `cancelIcon={null}` in our `ModalScreen` to remove the default icon, but `null` is not actually allowed, should I simply add `null` as an option with an explanation or do you want to think on a better solution?

ethanshar (approved):

PR #923, opened 2020-09-02
Migrate panning views to typescript

M-i-k-e-l:
This is a big one, sorry...
There are some TS errors that I'd appreciate if you take a look at, most of them is simply TS not recognising that there is a default value for a prop and I did not find a way to solve that (and still use defaultProps and not ignore).


ethanshar (approved):

ethanshar:
@M-i-k-e-l 
Approved, but need to fix conflicts first.

PR #928, opened 2020-09-07
Picker - add support for layout modifiers (margin, paddings)

Inbal-Tish:


ethanshar:
@Inbal-Tish 
There's some conflicts, probably from merging the previous Picker PR

Inbal-Tish:
Done.

src/components/picker/index.js
```diff
@@ -385,10 +386,12 @@ class Picker extends PureComponent {
 
     const textInputProps = TextField.extractOwnProps(this.props);
     const label = this.getLabelValueText();
-    
+    const layoutProps = {...extractMarginValues(this.props), ...extractPaddingValues(this.props)};
```
ethanshar, line 389: no need for that.. 
by wrapping a component with asBaseComponent you get injected with the `modifiers` prop. 
From there you can get whatever modifiers values you need (margins, paddings, colors, etc..) 

I would see how `View` is implemented for a reference.

ethanshar (changes requested):

src/components/picker/index.js
```diff
@@ -385,10 +392,18 @@ class Picker extends PureComponent {
...
+
     return (
       <TextField
         {...textInputProps}
+        containerStyle={[
+          containerStyle,
+          paddings,
+          margins,
+          alignments,
+          positionStyle
```
ethanshar, line 405: `containerStyle` should be prioritized over modifiers style. 
`alignments` style is redundant in this case, it does't affect the component so there's no need to support it.

ethanshar (changes requested):

src/components/picker/index.js
Inbal-Tish: Done.

src/components/picker/index.js
Inbal-Tish: Done.

ethanshar (approved):

ethanshar:
@Inbal-Tish 
I approved this, but I see now build has failed. 
Can you check? 
You can merge once build is fixed..

PR #930, opened 2020-09-08
Dialog - migrate to typescript (and some fixes)

M-i-k-e-l:
Please note that I've made small changes to `View` and `Constants`


src/components/view/index.tsx
```diff
@@ -32,7 +32,7 @@ export interface ViewPropTypes extends Omit<ViewProps, 'style'>, ContainerModifi
    * Set background color
    */
   backgroundColor?: string;
-  style?: StyleProp<ViewStyle> | Animated.AnimatedProps<StyleProp<ViewStyle>>;
+  style?: StyleProp<ViewStyle | Animated.AnimatedProps<ViewStyle>>;
```
ethanshar, line 35: I almost sure I changed that because of an issue I had in the private repo. 
Is there a way to verify it doesn't break anything in the private repo?

M-i-k-e-l: Do you mean break the TS or usage?

ethanshar: break TS, I don't think it can break functionality..

M-i-k-e-l: I pasted the change into `view/index.d.ts` and it didn't break `npm run build:dev`, not sure if that's enough of a check

ethanshar: That's should be enough. merging.

ethanshar (approved):

PR #932, opened 2020-09-10
RadioGroup - Replace unsafe method with getDerivedStateFromProps

mendyEdri:
Using now `getDerivedStateFromProps` instead of `UNSAFE_componentWillReceiveProps`

### DIFF
--- a/src/components/radioButton/RadioGroup.tsx
+++ b/src/components/radioButton/RadioGroup.tsx
@@ -1,6 +1,11 @@
 import _ from 'lodash';
-import React, {PureComponent} from 'react';
-import {asBaseComponent, forwardRef, BaseComponentInjectedProps, ForwardRefInjectedProps} from '../../commons/new';
+import React, {PureComponent, GetDerivedStateFromProps} from 'react';
+import {
+  asBaseComponent,
+  forwardRef,
+  BaseComponentInjectedProps,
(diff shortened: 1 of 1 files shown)


src/components/radioButton/RadioGroup.tsx
```diff
@@ -35,10 +42,24 @@ class RadioGroup extends PureComponent<Props, RadioGroupState> {
...
-    if (this.props.initialValue !== nextProps.initialValue) {
-      this.setState({value: nextProps.initialValue});
+  static getUpdatedState = (
+    nextProps: Props,
+    prevState: RadioGroupState
+  ): RadioGroupState | null => {
+    const {value} = prevState;
+    const {initialValue} = nextProps;
+    
+    if (_.isUndefined(nextProps.initialValue) || value === initialValue) {
```
Inbal-Tish, line 52: Why no turn the component to uncontrolled component? Where the `initialValue` is literally **initial** value and not updated through a prop change?

mendyEdri: @Inbal-Tish, Yeah, I though about it, but then we going to lose the use case to let users 'reset' the component value. 
I consult with Ethan to have a second opinion and we said that this use-cases should be kept

Inbal-Tish (approved):

PR #935, opened 2020-09-14
add backgroundColor to tabBar

lidord-wix:
add custom background color support to tabBar


src/components/tabController/TabBar.js
```diff
@@ -404,7 +408,7 @@ class TabBar extends PureComponent {
           snapToOffsets={this.getSnapBreakpoints()}
           decelerationRate={'fast'}
         >
-          <View style={[styles.tabBar, height && {height}, {paddingHorizontal: this.centerOffset}]}>
```
ethanshar, line 407: Not sure why I applied the white background color both on the ScrollView and the inner View. I guess it's I forgot to remove.
I think it's redundant and we can keep it only on the ScrollView or the inner View.

ethanshar (changes requested):

src/components/tabController/TabBar.js
lidord-wix: Actually it seems to work the same even when I remove the white background from them both..

ethanshar: But then it will transparent, because the background color is undefined. 
What do you mean exactly?

src/components/tabController/TabBar.js
```diff
@@ -54,6 +55,10 @@ class TabBar extends PureComponent {
      * Show Tab Bar bottom shadow
      */
     enableShadow: PropTypes.bool,
+    /**
+     * custom shadow style
+     */
+    shadowStyle: RNText.propTypes.style,
```
ethanshar, line 61: Why are you taking shadow style from RNText?

ethanshar (commented):
Looks great. 
see my only comment on the shadowStyle propType..

src/components/tabController/TabBar.js
lidord-wix: I just saw it on `labelStyle `and `selectedLabelStyle` so I thought it is the relevant type of style prop.. 
Should it be `ViewPropTypes.style`?

ethanshar: Yes 👍

ethanshar (approved):

PR #937, opened 2020-09-14
Migrate Badge to TS

ArnonZ:
WOAUILIB-1070


src/components/badge/index.tsx
```diff
@@ -1,109 +1,115 @@
...
+  pimpleHuge = 14,
+  small = 16,
+  default = 20,
+  large = 24
 };
 
+type LabelFormatterValues = typeof LABEL_FORMATTER_VALUES[number]
+export type BadgeSizes = keyof typeof BADGE_SIZES
+
+export type BadgePropTypes = ViewProps & TouchableOpacityProps & {
```
ethanshar, line 27: Let's keep the name as `BadgeProp` instead of `BadgePropTypes`, I wanted to have a shorter name 
(I know we already have some components with xxxPropTypes, we'll change them eventually).

src/components/badge/index.tsx
```diff
@@ -1,109 +1,115 @@
...
+   */
+  label?: string;
+  /**
+   * Color of the badge background
+   */
+  backgroundColor?: string;
+  /**
+   * the badge size (default, small)
+   */
+  size: BadgeSizes | number;
```
ethanshar, line 40: Should it be optional as the rest of the props? 
Otherwise it will require user to pass it (even though it has a default value)

ArnonZ: I've checked the example screen and it doesn't require since it knows to use the `defaultProp` (I left one example where size is not passed).
As far as I know this is the only way to make TS take into account `defaultProps`:
` size: 'default'`.
Also, TS doesn't like it as optional since yells at me for methods that use `size` and assume that it can be undefined and fall back to `defaultProp`. 

Same thing we did in Avatar btw: https://github.com/wix/react-native-ui-lib/blob/master/src/components/avatar/index.tsx#L121

ethanshar (approved):

Issue #83, opened 2018-02-08
react-native-interactable requirement


birkir:
It seems that react-native-interactable is required.

When compiling for android in CI, everything breaks.

```bash
Failed to capture snapshot of input files for task ':app:bundleReleaseJsAndAssets' property '$1' during up-to-date check.
> Could not list contents of '/home/travis/build/ueno-llc/react-native-starter/node_modules/react-native-interactable/{ios,android}'. Couldn't follow symbolic link.
```

When looking at the android setup, it is added in  `./android/settings.gradle` but has no project dependency in `./android/app/build.gradle`.

birkir:
Seems to be issue with react-native-interactable itself.

markhu:
Looking a little closer reveals that node_modules/react-native-interactable/ had a problem with assuming that `bash` was the default shell, so it tried a bash-specific shortcut. Sad thing that trying something clever to save just one line of a simple shell script, has caused problems for so many people. Hope this helps: https://github.com/wix/react-native-interactable/pull/209

PR #940, opened 2020-09-16
ColorPicker - Remove UNSAFE method

Inbal-Tish:
1) Separating the preview dialog from the component - ColorPickerDialog.
2) Changing component to 'asBaseComponent' HOC.
3) Removing UNSAFE method by passing a 'key' prop to the ColorPickerDialog component when 'initialColor' prop changes.
4) Adding TODO comments for V6 migrations.


src/components/colorPicker/index.js
```diff
@@ -118,6 +121,12 @@ export default class ColorPicker extends PureBaseComponent {
     }
   }
 
+  showMigrationMessage() {
+    console.warn(`ColorPicker component will soon become uncontrolled component.` +
+    ` Thus 'initialColor' prop will become INITIAL, meaning, no updates to this prop will take effect` +
+    ` unless passing 'key' prop.`);
+  }
```
ethanshar, line 128: I noticed this warnings pops every time we select a color in the palette (in the example screen) 
I think it might be a problem not supporting the current behavior. 

In One App we have this scenario when the user can select a color from a palette and it affect the ColorPicker as well..

Inbal-Tish: Yes, the warning is in componentWillRecieveProps() to alert when using 'initialColor' prop to update the picker. 
Regarding the functionality, it will be the same as long as the users will ADD the 'key' prop with the state color. This way we keep the component uncontrolled but let the user update using state getting a new instance of the component when the 'initialColor' (and 'key') changes. An example is on ColorPickerScreen line #83.

Inbal-Tish:
@ethanshar I don't see a way to update the' initialColor' without turning the component to an uncontrolled component and pass a 'key' prop. Do you have any other suggestions?

src/components/colorPicker/index.js
ethanshar: Ok, I went over the component implementation again.. 
The component is actually kind of controlled. which is good. 
We need to separate the two parts of the component... 

One part is the color picker field (the add button with the colors palette) - This part is controlled cause it depends on the user's `value` prop - which is good!

The second part is the color picker dialog (the "editor") which clearly needs to have an internal state. 
That means that we can separate the "editor" part to an internal component and always initialize it when opened, with the current `value` prop. 
The fact that the "editor" has internal state doesn't change the fact the ColorPicker is controlled. 
Eventually, we determine the "current" value according to what the user pass in props, and that's what matter. 

Anyway, we don't need both `value` and `initialValue` we can have only `value`. 

Let me know if what I wrote isn't clear, we can sync on this on Sunday.

Inbal-Tish: Love it. Done.

ethanshar:
@Inbal-Tish 
Just verifying first. Is this PR should be merged to `v6` branch or `master`? 
Also, do we still need the `initialColor` prop?

ethanshar (commented):
@Inbal-Tish 
Just verifying first. Is this PR should be merged to `v6` branch or `master`? 
Also, do we still need the `initialColor` prop?

Inbal-Tish:
To master. Yes, the initialColor is sent to the ColorPickerDialog as key to render the preview on the required color.

ethanshar (approved):

PR #941, opened 2020-09-16
Carousel - remove unsafe method

Inbal-Tish:


src/components/carousel/index.js
```diff
@@ -114,49 +114,75 @@ export default class Carousel extends BaseComponent {
...
+  componentWillUnmount() {
+    Constants.removeDimensionsEventListener(this.onOrientationChanged);
 
-    if (pageWidth !== nexPageWidth) {
-      const pageWidth = nexPageWidth;
-      this.setState({
+    clearInterval(this.autoplayTimer);
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
```
ethanshar, line 150: move static method before componentDidMount

src/components/carousel/index.js
```diff
@@ -114,49 +114,75 @@ export default class Carousel extends BaseComponent {
...
+    const {currentPage, prevProps} = prevState;
```
ethanshar, line 151: Is there a reason to keep the whole "prevProps" in state? 
Why not just keep the initialPageWidth in state and check against its value.

src/components/carousel/index.js
```diff
@@ -493,28 +535,26 @@ export default class Carousel extends BaseComponent {
...
-      marginVertical: 16 - containerPaddingVertical
-    },
-    hiddenText: {
-      position: 'absolute', 
-      width: 1
-    }
-  });
-}
+export {Carousel}; // For tests
+export default asBaseComponent(forwardRef(Carousel));
```
ethanshar, line 539: We don't want to forward ref in the case of the Carousel. 
Forwarding ref means the user will have access to the internal ScrollView ref instead of the Carousel component.
In our case the Carousel provide imperative methods to use with ref like `goToPage`

Anyway, even if we would have want to forward ref, notice that you didn't really forward it completely. 
See how it's done in the View component 
https://github.com/wix/react-native-ui-lib/blob/master/src/components/view/index.tsx#L117
You still need to pass the `forwardedRef` injected prop

ethanshar (changes requested):
I pushed a fix to the file code format, I think you formatted it before we fixed .prettierrc config. 

Anyway, try to add description to PRs even when seem trivial (unless it's a one line fix)
I won't necessarily be the one that review them and it's important to understand what you fix here and how.

src/components/carousel/index.js
Inbal-Tish: Ok

src/components/carousel/index.js
Inbal-Tish: I'm using 'pageWidth', 'containerMarginHorizontal', 'loop' for the prevProps and passing them to 'presenter.calcOffset()'

src/components/carousel/index.js
Inbal-Tish: Ok got it. Thanks

src/components/carousel/index.js
ethanshar: 👍

ethanshar (approved):

PR #943, opened 2020-09-17
Feat/expandable list item new component

lidord-wix:
## Description
Add generic expandableSection component that render headerSection element and children as a content of the component.


demo/src/screens/MenuStructure.js
```diff
@@ -31,6 +31,7 @@ export const navigationData = {
       {title: 'Cards', tags: 'cards feed', screen: 'unicorn.components.CardsScreen'},
       {title: 'Connection Status Bar', tags: 'connection status bar', screen: 'unicorn.components.ConnectionStatusBar'},
       {title: 'Chip', tags: 'chip', screen: 'unicorn.components.ChipScreen'},
+      {title: 'Expandable Section', tags: 'expandable section', screen: 'unicorn.components.ExpandableSectionScreen'},
```
ethanshar, line 34:
```suggestion
      {title: 'ExpandableSection', tags: 'expandable section', screen: 'unicorn.components.ExpandableSectionScreen'},
```

typings/components/index.d.ts
```diff
@@ -9,6 +9,7 @@ export * from './Carousel';
 export * from './ColorPicker';
 export * from './ConnectionStatusBar';
 export * from './Dialog';
+export * from './ExpandableSection';
```
ethanshar, line 12: This file will be removed eventually.
No need to add components here anymore..

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+   */
+  sectionHeader?: JSX.Element;
+  /**
+   * expandableSection expandable children
+   */
+  children?: React.ReactNode;
+  /**
+   * expandableSection icon color
+   */
+  iconColor?: string;
```
ethanshar, line 22: Don't support having a separate icon. 
Let the user pass the whole expand section element.

The user can decide how it looks and if they want an icon or not. 
See following comment, I explain better how it should integrate with private uilib.

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const chevronDown = require('../../assets/icons/chevronDown.png');
+const chevronUp = require('../../assets/icons/chevronUp.png');
```
ethanshar, line 26: The icons and the section UI should be part of the example screen.

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [expanded, setExpanded] = useState(false);
+  const {iconColor, sectionHeader, children} = props;
+
+  const onExpand = () => {
+    LayoutAnimation.configureNext(LayoutAnimation.Presets.spring);
+    setExpanded(!expanded);
+  };
+
+  return (
+    <View style={{borderWidth: 0, borderRadius: 0}}>
```
ethanshar, line 38: The style here looks like left overs..

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <View style={{marginLeft: 10, marginTop: 15}}>
```
ethanshar, line 39: Avoid setting default style here, it will force our users to adjust their element to this constraints. 
Better let the user decide on UI completely. 
They can style the element as they wish. 

This applies to us as well when we'll use this component in our private code.

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+          {sectionHeader}
+        </View>
+        <View style={{marginLeft: 380, marginTop: 25, position: 'absolute'}}>
+          <Button
+            iconSource={expanded ? chevronUp : chevronDown}
+            iconStyle={{tintColor: iconColor}}
+            style={{backgroundColor: 'transparent'}}
+            onPress={() => onExpand()}
+          ></Button>
```
ethanshar, line 48: It will be better to have the whole item trigger the expansion instead of the button. 

The main reason for that is that in our private lib we already have a ListItem with the appropriate UI for the expand section (and it's already includes the chevron icon)

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <View style={{borderWidth: 0, borderRadius: 0}}>
+        <View style={{marginLeft: 10, marginTop: 15}}>
+          {sectionHeader}
+        </View>
+        <View style={{marginLeft: 380, marginTop: 25, position: 'absolute'}}>
+          <Button
+            iconSource={expanded ? chevronUp : chevronDown}
+            iconStyle={{tintColor: iconColor}}
+            style={{backgroundColor: 'transparent'}}
+            onPress={() => onExpand()}
```
ethanshar, line 47: Notice that creating a inline callback here (`() => {}`) is not good for performance in general and in this case is redundant.

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const chevronDown = require('../../assets/icons/chevronDown.png');
+const chevronUp = require('../../assets/icons/chevronUp.png');
+
+function ExpandableSection(props: ExpandableSectionProps) {
+  const [expanded, setExpanded] = useState(false);
+  const {iconColor, sectionHeader, children} = props;
+
+  const onExpand = () => {
+    LayoutAnimation.configureNext(LayoutAnimation.Presets.spring);
```
ethanshar, line 33: I recommend toning down a little the spring animation (: 
The UX who'll review it will probably comment about it.. 

If you want a more specific specification on the animation talk to Or.
Maybe she can give some reference for how it should look.

ethanshar (commented):
Few comments on this PR
First, try avoid having an inline style. Always move styles into StyleSheet when possible. 
Passing inline style can affect performance. 

Second, Regarding  the structure of the component. I left some comments. 
Let me know if something is not clear.. 

One last thing. 
Notice that when you collapse the component, there's a brief moment where you can still see the children overflow the boundary. Try looking into the `overflow` style prop in order to fix that..

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+   * expandableSection icon color
+   */
+  expanded?: boolean;
+};
+
+function ExpandableSection(props: ExpandableSectionProps) {
+  const {expanded, sectionHeader, children} = props;
+
+  function onValueChange() {
+    LayoutAnimation.configureNext({...LayoutAnimation.Presets.easeInEaseOut, duration: 200});
```
ethanshar, line 25: We don't trigger animation or update things during a render. 
Render can be triggered for many reasons and that means you'll trigger an animation even when you don't intend to. 
For that we use hooks like `useEffect`. (or lifecycle methods in case you use classes)

src/components/expandableSection/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * expandableSection header element
+   */
+  sectionHeader?: JSX.Element;
+  /**
+   * expandableSection expandable children
+   */
+  children?: React.ReactNode;
+  /**
+   * expandableSection icon color
```
ethanshar, line 16: You forgot changing the prop description (:

lidord-wix: 🙊

ethanshar (approved):

PR #944, opened 2020-09-18
Infra/color slider group remove unsafe

Inbal-Tish:
## Description
GradientSlider, ColorSliderGroup, SliderGroup - removing UNSAFE method.

## Changelog
SliderGroup, GradientSlider, ColorSliderGroup - removing UNSAFE method.


src/components/slider/context/SliderGroup.js
```diff
@@ -21,22 +21,9 @@ export default class SliderGroup extends Component {
...
-  //   }
-  //   return null;
-  // }
-
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (nextProps.color !== Colors.getHexString(this.state.value)) {
-      return {value: Colors.getHSL(nextProps.color)};
-    }
-  }
-
```
ethanshar, line 36: Doesn't it break functionality?

Inbal-Tish: Not when I checked it. Did you have an issue?

ethanshar: No. 
But If I understand correctly.. 
It supported initializing the color value when passing a different value to the prop. No? 

Actually, I think it's better when removed....
merging.

ethanshar (approved):

PR #945, opened 2020-09-19
Drawer - Remove UNSAFE method

Inbal-Tish:
## Description
Swipeable - Remove 'UNSAFE_componentWillUpdate' method.

## Changelog
Drawer - Swipeable - Remove UNSAFE method.


src/components/drawer/Swipeable.tsx
```diff
@@ -420,6 +427,10 @@ export default class Swipeable extends Component<PropType, StateType> {
   };
 
   render() {
+    if (this._shouldUpdateAnimatedEvent()) {
+      this._updateAnimatedEvent(this.props, this.state);
+    }
```
ethanshar, line 432: We don't trigger updates or animations from the `render` method. 
There're other lifecycle methods for that..

Why don't keep this logic in `componentDidUpdate`?

src/components/drawer/Swipeable.tsx
```diff
@@ -132,18 +135,22 @@ export default class Swipeable extends Component<PropType, StateType> {
     }
   }
 
-  // TODO: change to componentDidUpdate
-  UNSAFE_componentWillUpdate(props: PropType, state: StateType) {
+  componentDidUpdate(prevProps, prevState) {
+    this.prevProps = prevProps;
+    this.prevState = prevState;
```
ethanshar, line 140: Why storing these? 
Why not just have the check here..

ethanshar (changes requested):

src/components/drawer/Swipeable.tsx
Inbal-Tish: This is not triggering animation, it just updates the variables and it happens only once, after the first render and before the second one that is triggered by a 'setState' in the 'onLayout'. If I'll move it to the 'componentDidUpdate' it won't take effect since it will happen after the second render, and since I can't trigger render in 'componentDidUpdate' from obvious reasons...

src/components/drawer/Swipeable.tsx
Inbal-Tish: I need to check after the first render

src/components/drawer/Swipeable.tsx
ethanshar: I think there's fundamental issue with the component and maybe it's a good opportunity to refactor it a little. 
This hack doesn't feel unnatural to React ecosystem. 

Currently we store derived animated values in the class instance (`this._transX`, `this._showLeftAction`, `this._leftActionTranslate`, etc..)

This forces us to store these values at the right moment so it will take into account when the next render happens. 

What if instead of sorting these values we'll just implement methods that return the derived animated value separately. 
For instance, we'll have a method for `getTransX` and `getShowLeftAction` and so on.. 
This way I believe the values should by dynamic and always derive from current state/props and might reduce us some unnecessary code

src/components/drawer/Swipeable.tsx
ethanshar: See my other comment, this may be redundant.

src/components/drawer/Swipeable.tsx
Inbal-Tish: Love it. Thanks. Done.

ethanshar (approved):

PR #951, opened 2020-09-23
Change how we extract a component props

ethanshar:
## Description
- There was an issue in PickerDialog components (ios and android) with the usage of `Dialog.extractOwnProps` due to the migration to `asBaseComponent` 
- I added a new methods in `Modifiers` class - `extractComponentProps` which should replace `extractOwnProps`

## Changelog
Fix issue with Picker when passing the `useNativePicker` prop an error was thrown.

### DIFF
--- a/src/commons/modifiers.ts
+++ b/src/commons/modifiers.ts
@@ -304,6 +304,7 @@ export function extractModifierProps(props: Dictionary<any>) {
   return modifierProps;
 }
 
+// TODO: should not be used anymore
 export function extractOwnProps(props: Dictionary<any>, ignoreProps: string[]) {
   //@ts-ignore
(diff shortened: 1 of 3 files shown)


src/commons/modifiers.ts
```diff
@@ -304,6 +304,7 @@ export function extractModifierProps(props: Dictionary<any>) {
   return modifierProps;
 }
 
+// TODO: should not be used anymore
```
M-i-k-e-l, line 307:
```suggestion
/**
 * TODO:
 * @deprecated switch to Modifiers#extractComponentProps
 */
```

M-i-k-e-l (approved):

PR #958, opened 2020-09-29
feat/slider enhancements

mendyEdri:
## Description
* Added scale animation to Slider's thumb.
* Support Slider-track tap to change value.


src/components/slider/index.js
```diff
@@ -297,6 +325,37 @@ export default class Slider extends PureBaseComponent {
...
   };
 
+  shouldDoubleSizeByDefault = () => {
+    const {activeThumbStyle, thumbStyle} = this.props;
+    return !activeThumbStyle || !thumbStyle;
+  }
+
+  calculatedThumbActiveScale = () => {
+    const {activeThumbStyle, thumbStyle, disabled} = this.props;
+    if (disabled) {
```
Inbal-Tish, line 335: Also if user passed 'disableActiveStyling', no?

src/components/slider/index.js
```diff
@@ -370,14 +438,43 @@ export default class Slider extends PureBaseComponent {
...
+    const {
+      thumbStyle,
+      disabled,
+      thumbTintColor
+    } = this.getThemeProps();
+    return (
+      <Animated.View
+        ref={this.setThumbRef}
+        onLayout={this.onThumbLayout}
+        {...this._panResponder.panHandlers}
```
Inbal-Tish, line 451: Moving the panHandlers here means you must touch and drag the thumb itself and not the track. Did you test this on a real device? Is it an easy thing to do? Is the hit slop big enough?

src/components/slider/index.js
mendyEdri: Right, fixed

src/components/slider/index.js
mendyEdri: Added.
You are right, although it's seems easy to drag it, i've added hitslop for better experience.

Inbal-Tish: One more thing about this. This is a UX change and I think we should pass it by our UX guys. Once they'll approve it we can merge.

mendyEdri: Approved by Yulia 🚀

Inbal-Tish (approved):

PR #959, opened 2020-09-29
Feat/add label support to checkbox

lidord-wix:
## Description
Add label and labelStyle props to checkbox to support passing label to checkbox

## Changelog
Add label support to checkbox


ethanshar (commented):
There's an issue with the new feature. 
The checkbox works only if pressing the label. 
Pressing the actual checkbox doesn't trigger value change..

src/components/checkbox/index.tsx
```diff
@@ -155,33 +172,38 @@ class Checkbox extends Component<CheckboxPropTypes, CheckboxState> {
...
-        onPress={this.onPress}
-      >
-        {
-          <Animated.View
-            style={[this.styles.container, {backgroundColor: this.getColor()}, {opacity: this.animationStyle.opacity}]}
-          >
-            <Animated.Image
+      <TouchableOpacity row onPress={this.onPress}>
+        <Animated.View
+          {...this.getAccessibilityProps()}
```
ethanshar, line 179: Some of these props belong to the most top container. 
For instance `activeOpacity` is relevant only for TouchableOpacity. 
Notice that now the behavior is a little different - pressing the checkbox comes with the default activeOpacity (0.6) and it doesn't look as before)

src/components/checkbox/index.tsx
```diff
@@ -155,33 +172,38 @@ class Checkbox extends Component<CheckboxPropTypes, CheckboxState> {
...
-        style={[this.getBorderStyle(), style]}
-        onPress={this.onPress}
-      >
-        {
-          <Animated.View
-            style={[this.styles.container, {backgroundColor: this.getColor()}, {opacity: this.animationStyle.opacity}]}
-          >
-            <Animated.Image
+      <TouchableOpacity row onPress={this.onPress}>
+        <Animated.View
```
ethanshar, line 178: Do you need an Animated.View here?

src/components/checkbox/index.tsx
```diff
@@ -155,33 +172,38 @@ class Checkbox extends Component<CheckboxPropTypes, CheckboxState> {
...
+                  {transform: this.animationStyle.transform},
+                  disabled && {tintColor: DEFAULT_ICON_COLOR}
+                ]}
+                source={selectedIcon || Assets.icons.checkSmall}
+                testID={`${testID}.selected`}
+              />
+            </Animated.View>
+          }
+        </Animated.View>
+        {label && <Text style={labelStyle}>{label}</Text>}
```
ethanshar, line 206: I notice that the text is not vertically aligned to the checkbox. 
Also consider adding default spacing to the make the default appearance look good

ethanshar (changes requested):

src/components/checkbox/index.tsx
lidord-wix: I added the spacing on the private config, but in second thought it is better to apply it in public..

src/components/checkbox/index.tsx
```diff
@@ -15,6 +14,9 @@ import {Colors} from '../../style';
 import Assets from '../../assets';
 import {asBaseComponent} from '../../commons/new';
 import TouchableOpacity from '../touchableOpacity';
+import Text from '../text';
+import View from '../view';
+import Spacing from '../../style/spacings';
```
ethanshar, line 19:
```suggestion
import {Spacings} from '../../style';
```

src/components/checkbox/index.tsx
```diff
@@ -174,13 +177,15 @@ class Checkbox extends Component<CheckboxPropTypes, CheckboxState> {
...
-        <Animated.View
+      <View row style={style}>
+        {/*@ts-ignore*/}
+        <TouchableOpacity
           {...this.getAccessibilityProps()}
           activeOpacity={1}
           testID={testID}
           {...others}
-          style={[this.getBorderStyle(), style]}
+          style={this.getBorderStyle()}
```
ethanshar, line 187: Notice that before this line was 
```
style={[this.getBorderStyle(), style]}
```

Moving the style to the top container might be a breaking change in some cases. 
For instance if I rendered a checkbox like this: 
```
<Checkbox style={{borderWidth: 3}} ..../>
```
The style now will apply on the whole container instead of only the checkbox and the user will get different results. 

In this case, we can keep passing the style to the actual checkbox and add a new prop called `containerStyle` for the whole thing. (checkbox+label)

src/components/checkbox/index.tsx
```diff
@@ -225,6 +234,10 @@ function createStyles(props: CheckboxPropTypes) {
       tintColor: iconColor,
       alignItems: 'center',
       justifyContent: 'center'
+    },
+    checkboxLabel: {
+      marginLeft: Spacing.s3,
```
ethanshar, line 239: Assuming you'll except the suggestion at the top comment, then..
```suggestion
      marginLeft: Spacings.s3,
```

ethanshar (changes requested):
We're getting there (: 
just few last fixes

ethanshar (approved):

PR #960, opened 2020-09-29
Move from KeyboardAwareListView to KeyboardAwareFlatList

M-i-k-e-l:
## Description
KeyboardAwareFlatList - add new type to the KeyboardAware view and deprecate KeyboardAwareListView
I think we need to do some more refactoring around this (fix warnings and code duplications), but maybe in another task\PR.

## Changelog
Move from KeyboardAwareListView to KeyboardAwareFlatList

### DIFF
--- a/src/components/KeyboardAwareScrollView/KeyboardAwareBase.js
+++ b/src/components/KeyboardAwareScrollView/KeyboardAwareBase.js
@@ -1,7 +1,7 @@
+import _ from 'lodash';
 import {Component} from 'react';
 import PropTypes from 'prop-types';
-import ReactNative, {DeviceEventEmitter, Keyboard, NativeModules} from 'react-native';
-const ScrollViewManager = NativeModules.ScrollViewManager;
+import ReactNative, {DeviceEventEmitter, Keyboard} from 'react-native';
 
 export default class KeyboardAwareBase extends Component {
   constructor(props) {
@@ -12,8 +12,11 @@ export default class KeyboardAwareBase extends Component {
       '_removeKeyboardListeners',
       '_scrollToFocusedTextInput',
       '_onKeyboardAwareViewLayout',
+      '_updateKeyboardAwareViewContentSize',
       'scrollToBottom',
       'scrollBottomOnNextSizeChange');
+    this.contentSize = undefined;
+    this.layoutSize = undefined;
     this.state = {keyboardHeight: 0};
     this._addKeyboardEventListeners();
   }
@@ -55,8 +58,9 @@ export default class KeyboardAwareBase extends Component {
     }
   }
 
-  _onKeyboardAwareViewLayout(layout) {
-    this._keyboardAwareView.layout = layout;
+  _onKeyboardAwareViewLayout(layoutEvent) {
+    const layout = layoutEvent.nativeEvent.layout;
+    this.layoutSize = layout;
     this._keyboardAwareView.contentOffset = {x: 0, y: 0};
     this._updateKeyboardAwareViewContentSize();
   }
@@ -66,17 +70,22 @@ export default class KeyboardAwareBase extends Component {
     this._updateKeyboardAwareViewContentSize();
   }
 
-  _updateKeyboardAwareViewContentSize() {
-    if (ScrollViewManager?.getContentSize) {
-      ScrollViewManager.getContentSize(ReactNative.findNodeHandle(this._keyboardAwareView), res => {
-        if (this._keyboardAwareView) {
-          this._keyboardAwareView.contentSize = res;
-          if (this.state.scrollBottomOnNextSizeChange) {
-            this.scrollToBottom();
-            this.setState({scrollBottomOnNextSizeChange: false});
-          }
-        }
-      });
+  _updateKeyboardAwareViewContentSize(width, height) {
+    let heightHasChanged = false;
+    if (width && height) {
+      if (this.contentSize && this.contentSize.height !== height) { // for FlatList
+        heightHasChanged = true;
+      }
+
+      this.contentSize = {width, height};
+    }
+
+    if (this._keyboardAwareView) {
+      if (this.state.scrollBottomOnNextSizeChange ||
+          (this.props.startScrolledToBottom && heightHasChanged)) {
+        this.scrollToBottom();
+        this.setState({scrollBottomOnNextSizeChange: false});
+      }
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/components/KeyboardAwareScrollView/KeyboardAwareBase.js
```diff
@@ -146,12 +146,16 @@ export default class KeyboardAwareBase extends Component {
...
     }
   }
   scrollTo(options) {
     if (this._keyboardAwareView) {
-      this._keyboardAwareView.scrollTo(options);
+      if (this._keyboardAwareView.scrollTo) {
+        this._keyboardAwareView.scrollTo(options);
+      } else if (this._keyboardAwareView.scrollToOffset) { // scroll to start in FlatList
+        this._keyboardAwareView.scrollToOffset({...options, offset: 0});
+      }
```
ethanshar, line 158: Why in the case of FlatList you scroll to offset: 0? 
Shouldn't you scroll to the y offset passed in options?

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
```diff
@@ -0,0 +1,50 @@
...
+      <FlatList
+        {...this.props}
+        {...this.style}
+        contentInset={{bottom: this.state.keyboardHeight}}
+        ref={(r) => {
+          this._keyboardAwareView = r;
+        }}
+        onLayout={(layoutEvent) => {
+          this._onKeyboardAwareViewLayout(layoutEvent.nativeEvent.layout);
+        }}
```
ethanshar, line 36: The inline callback here is probably not so good for performance. 
Why not have the KeyboardAwareBase accept the whole layoutEvent and extract what it needs.

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
```diff
@@ -0,0 +1,50 @@
...
+        onScroll={(event) => {
```
ethanshar, line 37: Same for the onScroll callback. 
maybe you can unify the logic in KeyboardAwareBase. 
Anyway, avoid passing inline callbacks.

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
```diff
@@ -0,0 +1,50 @@
...
+          this._onKeyboardAwareViewScroll(event.nativeEvent.contentOffset);
+          if (this.props.onScroll) {
+            this.props.onScroll(event);
+          }
+        }}
+        onContentSizeChange={() => {
+          this._updateKeyboardAwareViewContentSize();
+        }}
```
ethanshar, line 45: same here.. 
here it's easier, we don't pass any arguments. no need to wrap the callback at all.

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
```diff
@@ -0,0 +1,50 @@
...
+        scrollEventThrottle={200}
```
ethanshar, line 46: Consider moving `scrollEventThrottle` to the top so it'll set as default but not override. 
Maybe the user wants a different value

src/index.js
```diff
@@ -81,6 +81,9 @@ export default {
   get KeyboardAwareListView() {
     return require('./components/KeyboardAwareScrollView').KeyboardAwareListView;
   },
+  get KeyboardAwareFlatList() {
```
ethanshar, line 84: Now I notice we export KeyboardAwareFlatList, KeyboardAwareScroll and KeyboardAwareListView as standalone components and not as part of the Keyboard group. 
Why is that?

src/components/KeyboardAwareScrollView/KeyboardAwareBase.js
M-i-k-e-l: Fixed, some logic has changed and Android should also work now.

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
M-i-k-e-l: Done

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
M-i-k-e-l: Done

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
M-i-k-e-l: Done

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
M-i-k-e-l: Done

src/index.js
M-i-k-e-l: We did not start migrating all of the OS libs, only `react-native-keyboard-input` and `react-native-keyboard-tracking-view`

ethanshar: But still? 
Why not include them in the keyboard namespace? 
Which libraries we have left?

src/components/KeyboardAwareScrollView/KeyboardAwareFlatList.js
ethanshar: I see you didn't update `KeyboardAwareListView` with the same fixes.. 
Is that because we're deprecating the component ?

M-i-k-e-l: You are right, it should be aligned even if it's should not be used, fixed

src/index.js
M-i-k-e-l: I fear that someone will start using it, and I'm not sure there won't be conflicts (in native code).
We're using the `Keyboard` as is in the private lib and we're not exporting the `KeyboardAware` part (which is installed from the original by the engine).

M-i-k-e-l: Oh, these are the ones I remember:
https://github.com/wix/react-native-keyboard-aware-scrollview
https://github.com/wix/react-native-autogrow-textinput

I think I left my notebook with the notes in the office :D

ethanshar: >I fear that someone will start using it, and I'm not sure there won't be conflicts (in native code).
We're using the Keyboard as is in the private lib and we're not exporting the KeyboardAware part (which is installed from the original by the engine).

Im a little confused about this one.. 
let's have a quick talk about when you have time..

ethanshar (approved):

Issue #16, opened 2017-07-07
ListItem touch effect


safindr:
After change component in ListItem from TouchableHighlight to TouchableOpacity props onHideUnderlay and onShowUnderlay not available.

ethanshar:
Hey @safindr, 
the onHideUnderlay and onShowUnderlay were used internally. 
I don't think you could have pass this props before. 
Just to understand the issue better so I can fix/improve the component. 
What's exactly missing for you?

safindr:
In file /src/components/listItem/index.js used onHideUnderlay and onShowUnderlay on TouchableOpacity, but this props not available in TouchableOpacity and TouchableWithoutFeedback; 
This happened after the replacement of component TouchableHightlight by component TouchableOpacity;

ethanshar:
Yes, I saw that. 
I will remove onHideUnderlay and onShowUnderlay from the component, unless it's something you need.

safindr:
Yes, they can be removed, but this does not solve the problem. After replacing the component, the effect was lost when pressed, I would really like you to return it using TouchableOpacity

ethanshar:
@safindr, 
Ok instead of removing it I added a **containerElement** prop. 
You can use it to send a different Touchable wrapper, for instance TouchableHighlight and then the underlay behavior will work.
also added **underlayColor** prop to control the color

```
<ListItem height={50} containerElement={TouchableHighlight} underlayColor="red"/>
```

PR #961, opened 2020-09-29
ThemeManager - Adding deprecation warning for 'setTheme()'.

Inbal-Tish:
## Description
ThemeManager - Adding deprecation warning for 'setTheme()' to start the deprecation process before removing the method on the next major version.
Removing all usages of ThemeManager global constants - 'primaryColor', 'CTA', 'titleColor' etc.

## Changelog
ThemeManager - Starting deprecation process for 'setTheme()' method.

### DIFF
--- a/src/components/touchableOpacity/index.tsx
+++ b/src/components/touchableOpacity/index.tsx
@@ -62,6 +62,11 @@ type Props = BaseComponentInjectedProps &
 class TouchableOpacity extends PureComponent<Props, {active: boolean}> {
   static displayName = 'TouchableOpacity';
 
+  static defaultProps = {
+    throttleTime: 0,
+    throttleOptions: {leading: true, trailing: false}
+  }
+
   constructor(props: Props) {
     super(props);
 
--- a/src/components/touchableOpacity/index.tsx
+++ b/src/components/touchableOpacity/index.tsx
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import React, {PureComponent} from 'react';
 import {
   TouchableOpacity as RNTouchableOpacity,
@@ -6,7 +7,6 @@ import {
   ViewStyle,
   Animated
 } from 'react-native';
-import _ from 'lodash';
 import {
   asBaseComponent,
   forwardRef,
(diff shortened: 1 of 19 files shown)


src/style/themeManager.ts
```diff
@@ -25,6 +25,7 @@ export class ThemeManager {
   }
 
   setTheme(overrides: Dictionary<string>) {
+    console.warn('ThemeManager.setTheme() is deprecated. Please remove usage before next uilib major version update.');
```
ethanshar, line 28: I would also add:
"Consider using ThemeManager.setComponentTheme instead"

ethanshar (changes requested):

src/style/themeManager.ts
Inbal-Tish: Done.

Inbal-Tish:
@ethanshar I added to this PR the removal of all usages of the ThemeManager global constants.

src/style/themeManager.ts
Inbal-Tish: Done.

src/components/touchableOpacity/index.tsx
```diff
@@ -62,6 +62,11 @@ type Props = BaseComponentInjectedProps &
 class TouchableOpacity extends PureComponent<Props, {active: boolean}> {
   static displayName = 'TouchableOpacity';
 
+  static defaultProps = {
+    throttleTime: 0,
+    throttleOptions: {leading: true, trailing: false}
+  }
```
ethanshar, line 68: This will override our config in private uilib.. 
Why did you add it?

Inbal-Tish: This is to replace the configuration in ThemeManager (line #16).

ethanshar: Ohh I see now..
The thing is that setting it as `defaultProps` give it a different priority. 
Unfortunately, defaultProps are prioritized over theme props and override them. 
And the way it was defined before was considered as theme props..

Another option we can do is to use default Constants like we do sometimes when we extract the values 
WDYT?

Inbal-Tish: I bit messy but I guess that could be a solution. I'll push a fix

Inbal-Tish: Done

ethanshar (approved):

Issue #946, opened 2020-09-20
src/style/colors.ts has bug


447166939:
function validateHex(value: string) {
  if (!/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/.test(value)) {
    throw new Error(`${value} is invalid hex color`);
  }
  return value.replace('#', '');
}
the function when we enter the value eg. '#333'.this is valid,and it will return '333'
but in the rgba function , hex = validateHex(hex);
      red = parseInt(hex.substring(0, 2), 16);
      green = parseInt(hex.substring(2, 4), 16);
      blue = parseInt(hex.substring(4, 6), 16);
if we enter three bit string,the result will be wrong

PR #976, opened 2020-10-11
convert three digit hex to six

lidord-wix:
## Description
`Colors.rgba` returned wrong values when getting 3 digit hex,
I've added function to convert 3 digit hex to 6 digit hex inside `validateHex` 
Bug description: #946 

## Changelog
fix bug in `Colors.rgba`


src/style/colors.ts
```diff
@@ -203,7 +203,15 @@ function validateHex(value: string) {
   if (!/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/.test(value)) {
     throw new Error(`${value} is invalid hex color`);
   }
-  return value.replace('#', '');
+  value = value.replace('#', '');
+  if (value.length === 3) {
+    value = threeDigitHexToSix(value);
```
ethanshar, line 208: I know using regex to duplicate the string look fancy (: 
But maybe a more readable approach will be to concatenate value with itself.

src/style/colors.ts
```diff
@@ -203,7 +203,15 @@ function validateHex(value: string) {
   if (!/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/.test(value)) {
```
ethanshar, line 203: Let's add a unit test for this use case in `colors.spec.js`

src/style/colors.ts
```diff
@@ -211,7 +211,10 @@ function validateHex(value: string) {
 }
 
 function threeDigitHexToSix(value: string) { 
-  return value.replace(/./g, '$&$&');
+  const firstDigit = value[0];
+  const secondDigit = value[1];
+  const thirdDigit = value[2];
+  return firstDigit.concat(firstDigit, secondDigit, secondDigit, thirdDigit, thirdDigit);
```
ethanshar: Why not just return `value + value` ?

lidord-wix: We need to double each letter: 
`threeDigitHexToSix(123)` should return `112233` and not `123123`

ethanshar: You are completely right! my bad :( 
In that case the regex solution is definitely preferable... 
apologies for the trouble

ethanshar (approved):

PR #983, opened 2020-10-15
support render top section at expandableSection

lidord-wix:
## Description
support render top section at expandableSection
the user should pass the prop 'top' in the ExpandableSection component, default is bottom section


src/components/expandableSection/index.tsx
```diff
@@ -17,24 +17,29 @@ export type ExpandableSectionProps = {
    * should the expandableSection be expanded
    */
   expanded?: boolean;
+  /**
+   * should the expandableSection open above the sectionHeader
+   */
+  top?: boolean;
```
ethanshar, line 23: Let's add an example demonstrating it in the example screen..
you can add a toggle that switch between the two behaviors..

ethanshar (approved):

PR #986, opened 2020-10-18
Add appropriate deprecate warnings for picker API

ethanshar:
## Description
- Added migration warning for stop using object for Picker's value prop. 
- Added migration warning for passing value and label props separately for Picker.Item
- Added an example for a migrated picker in PickerScreen to verify we support correctly the new API and still doesn't break the old API (by keeping the old examples) 

## Changelog
Migrate Picker component's API. Should pass only string|number for `value` prop 
Migrate Picker.Item component's API. Should pass value and label props separately.

### DIFF
--- a/src/components/picker/PickerItem.js
+++ b/src/components/picker/PickerItem.js
@@ -1,4 +1,3 @@
-// TODO: deprecate passing an an object as a value, use label and value props separately
 import _ from 'lodash';
 import PropTypes from 'prop-types';
 import React from 'react';
--- a/src/components/picker/PickerItem.js
+++ b/src/components/picker/PickerItem.js
@@ -11,7 +11,6 @@ import TouchableOpacity from '../touchableOpacity';
 import Image from '../image';
 import Text from '../text';
 
-
 /**
  * @description: Picker.Item, for configuring the Picker's selectable options
  * @extends: TouchableOpacity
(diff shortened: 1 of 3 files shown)


src/components/picker/index.js
```diff
@@ -160,6 +160,9 @@ class Picker extends PureComponent {
     // if (props.useNativePicker && _.isPlainObject(props.value)) {
```
Inbal-Tish, line 160: Please remove the commented code and the relevant TODO at the top of the file to avoid confusion.

src/components/picker/PickerItem.js
```diff
@@ -23,20 +22,19 @@ class PickerItem extends BaseComponent {
 
   static propTypes = {
     /**
-     * [DEPRECATED - please include the label in the value prop] The item label
```
Inbal-Tish, line 26: Please remove the TODO at the top of the file to avoid confusion.

src/components/picker/index.js
ethanshar: Right! 
Removed

src/components/picker/PickerItem.js
ethanshar: Removed!

Inbal-Tish (approved):

PR #988, opened 2020-10-18
Use TS compiler to generate JS files

ethanshar:
## Description
Currently our JS files that are generated from TS files are creates with ES5 syntax which makes it very painful when investigating issues in private uilib when from node_modules. 

This PR fix how we generate the JS files and use `tcs` for that instead of `babel`. 

## Changelog
Change generated JS files from TS to have ES6 syntax.


M-i-k-e-l:
That's really nice!
Is there a possible drawback for anyone? Can anyone want it to remain in ES5?

M-i-k-e-l (approved):

PR #989, opened 2020-10-19
Fix/date time picker to controlled

Inbal-Tish:
## Description
DateTimePicker - turn to a controlled component. Fix issue #729 

## Changelog
DateTimePicker - turn to a controlled component.


src/components/textField/index.js
```diff
@@ -197,7 +197,8 @@ export default class TextField extends BaseInput {
       value: props.value, // for floatingPlaceholder functionality
```
ethanshar, line 197: You created the branch again from another branch. 
This change/file should be in this PR

src/helpers/Constants.ts
```diff
@@ -136,7 +136,10 @@ const constants = {
   /* Accessibility */
```
ethanshar, line 136: Also this file, I think it belong to another PR

src/components/dateTimePicker/index.js
```diff
@@ -90,17 +92,23 @@ class DateTimePicker extends BaseComponent {
...
-    this.styles = createStyles(this.props);
+  static getDerivedStateFromProps(nextProps, prevState) {
+    if (nextProps.value !== prevState.prevValue) {
+      return {
+        prevValue: prevState.value,
+        value: nextProps.value
+      };
+    }
+    return null;
   }
```
ethanshar, line 112: There's no need to keep the value in state if the component is controlled.

src/components/dateTimePicker/index.js
```diff
@@ -133,14 +141,14 @@ class DateTimePicker extends BaseComponent {
...
         this.chosenDate = new Date();
       }
-      
+
       _.invoke(this.props, 'onChange', this.chosenDate);
       this.setState({value: this.chosenDate});
     });
 
   getStringValue = () => {
     const {value} = this.state;
```
ethanshar, line 150: If the component is controlled, the value should derived directly from props and not the state.

ethanshar (changes requested):
Few comments on this PR 
- The PR doesn't change the component to be a controlled component cause it still keep the value in an internal state instead of getting it from the user props. 
- This is a breaking change to the component and we can't push it directly to master.
It means the user must pass the value prop, otherwise it won't react to changes.
It should be part of v6. 
- We need to make sure One App users migrate accordingly before we merge this.

src/components/dateTimePicker/index.js
Inbal-Tish: You're right, generally speaking, but when we want a controlled component and still need to keep an internal state that's the way to do it. See Carousel, DialogDismissibleView, PageControl, Picker, GradientSlider for example.

src/components/dateTimePicker/index.js
Inbal-Tish: I need to keep an internal state for the onChange

Inbal-Tish:
@ethanshar I fixed the branching, should be only 1 file change. I don't see a breaking change here, the user doesn't have to pass 'value' prop. The change is made for when the user does pass 'value' prop and change it with screen state change.

src/components/textField/index.js
Inbal-Tish: Done.

ethanshar (approved):

PR #990, opened 2020-10-19
Fix/text field right icon alignment

Inbal-Tish:
## Description
TextField - Fix right icon/button alignment when 'floatingPlaceholder' and 'useTopErrors' error is visible.

## Changelog
TextField - fix right icon/button alignment.


ethanshar:
@Inbal-Tish 
This fix breaks Picker icon for some reason 
Screen Shot 2020-10-20 at 7 52 02

Inbal-Tish:
@ethanshar I added a fix but it's only a patch, we'll have to take care of it in the new TextField.

ethanshar:
@Inbal-Tish 
I see now that it breaks the main screen search input. 
Notice that this input is completely custom so I understand how it can get broken.. 
What worries me how it will affect open source users... ?

Inbal-Tish:
@ethanshar Ok. I'll rethink it...

PR #991, opened 2020-10-19
migrate Switch to TS

lidord-wix:
## Changelog
Migration of Switch and SwitchScreen to typescript.


src/components/switch/index.tsx
```diff
@@ -92,7 +94,7 @@ class Switch extends BaseComponent {
     };
   }
 
-  toggle(value) {
+  toggle(value: any) {
```
M-i-k-e-l, line 97: Don't use `any`, unless you have to.

src/components/switch/index.tsx
```diff
@@ -121,8 +123,8 @@ class Switch extends BaseComponent {
   }
 
   getSwitchStyle() {
-    const {value, onColor, offColor, style: propsStyle, disabled, disabledColor} = this.getThemeProps();
-    const style = [this.styles.switch];
+    const {value, onColor, offColor, style: propsStyle, disabled, disabledColor} = this.props;
+    let style: any = [this.styles.switch];
```
M-i-k-e-l, line 127: 1. Why change from `const` to `let`?
2. Don't use `any`, unless you have to.

src/components/switch/index.tsx
```diff
@@ -153,9 +155,9 @@ class Switch extends BaseComponent {
   }
 
   render() {
-    const {...others} = this.getThemeProps();
-
+    const {...others} = this.props;
     return (
+      // @ts-ignore
```
M-i-k-e-l, line 160: Did you get an error here (I don't), not sure why (because I think we got those for the accessibility props).
Maybe the new change in typescript solved it, can you try to clean-install and check?

lidord-wix: I clean-installed and still get an error here..

src/components/switch/index.tsx
lidord-wix: I tried something with this let and forgot to return it to const :)

src/components/switch/index.tsx
```diff
@@ -1,87 +1,86 @@
...
+  componentDidUpdate(prevProps: SwitchProps) {
+    const {value} = this.props;
     if (prevProps.value !== value) {
       this.toggle(value);
     }
   }
 
   getAccessibilityProps() {
-    const {disabled, value} = this.getThemeProps();
+    const {disabled, value} = this.props;
```
lidord-wix, line 83: I changed every `this.getThemeProps()` to `this.props`, is that ok? Is there another way to handle the error on `this.getThemeProps()`?

src/components/switch/index.tsx
M-i-k-e-l: Yes, it seems I had a problem on my end (restarted VS Code and I have an error).

src/components/switch/index.tsx
M-i-k-e-l: That is correct, since the component is wrapped with `asBaseComponent` this is how it should be.

src/components/switch/index.tsx
```diff
@@ -121,8 +120,8 @@ class Switch extends BaseComponent {
   }
 
   getSwitchStyle() {
-    const {value, onColor, offColor, style: propsStyle, disabled, disabledColor} = this.getThemeProps();
-    const style = [this.styles.switch];
+    const {value, onColor, offColor, style: propsStyle, disabled, disabledColor} = this.props;
+    const style: object[] = [this.styles.switch];
```
M-i-k-e-l, line 124: Try to see how we define `Style` elsewhere in the props of other TS components.
BTW, it is more important that the actual `style` prop is fixed, but the same can be done here.

M-i-k-e-l (changes requested):
style prop has the wrong type

src/components/switch/index.tsx
M-i-k-e-l: Very nice!

M-i-k-e-l (approved):

PR #993, opened 2020-10-20
Infra/typescript page control

M-i-k-e-l:
## Description
Move PageControl to typescript

## Changelog
Move PageControl to typescript


src/components/pageControl/index.tsx
```diff
@@ -220,7 +231,7 @@ class PageControl extends PureComponent {
   }
 }
 
-export default asBaseComponent(forwardRef(PageControl));
+export default asBaseComponent<PageControlProps>(forwardRef(PageControl));
```
M-i-k-e-l, line 234: @ethanshar, I'm getting an error in this line (see below), I can fix it by making `forwardRef` more generic (see below), but that causes other errors in other files that are using `forwardRef`, the other option is changing `numOfPages` and `currentPage` to be optional, but that feels wrong.
I feel like the generic `forwardRef` is more correct but not sure how I can fix those other issues (some I can but they cause other errors etc.).
What do you think?



```
Argument of type 'typeof PageControl' is not assignable to parameter of type 'ComponentType<any>'.
  Type 'typeof PageControl' is not assignable to type 'ComponentClass<any, any>'.
    Types of property 'getDerivedStateFromProps' are incompatible.
      Type '(nextProps: PageControlProps, prevState: State) => { prevPage?: number | undefined; pagesOffset?: number | undefined; largeIndicatorsOffset?: number | undefined; } | null' is not assignable to type 'GetDerivedStateFromProps<any, any>'.
        Types of parameters 'nextProps' and 'nextProps' are incompatible.
          Type 'Readonly<any>' is missing the following properties from type 'PageControlProps': numOfPages, currentPage
```

```
export default function forwardRef<P = {}>(WrappedComponent: React.ComponentType<P>): React.ComponentType<P> {
  function forwardRef(props: P, ref: any) {
    return <WrappedComponent {...props} forwardedRef={ref}/>;
  }
...
```

ethanshar: Did you find any help in the guild? 
Maybe Dor or Nir can help..

M-i-k-e-l: Dor helped :)

ethanshar (approved):

PR #1000, opened 2020-10-25
Picker Refactor #1

ethanshar:
## Description
Step 1 - Migrate PickerItem from Class to a Function component. 

This in order to start using Context for communicating between the Picker parent component and its children (PickerItems) 
Once PickerItem is a Function component I'll be able to use hooks and specifically the `useContext` hook.

## Changelog
Refactor PicketItem component


src/components/picker/PickerItem.js
```diff
@@ -16,153 +16,138 @@ import Text from '../text';
...
+  const _onPress = useCallback(() => {
+    // onPress(_.isObject(value) ? value : {value, label});
+    onPress(value);
+  }, [value, onPress]);
 
-  getLabel() {
-    const {value, label} = this.props;
+  const onSelectedLayout = useCallback((...args) => {
+    _.invoke(props, 'onSelectedLayout', ...args);
+  }, []);
```
mendyEdri, line 46: shouldn't we pass `props` as a dependency?

src/components/picker/PickerItem.js
```diff
@@ -16,153 +16,138 @@ import Text from '../text';
...
       console.warn('UILib Picker.Item will stop supporting passing object as value & label (e.g {value, label}) in the next major version. Please pass separate label and value props');
     }
-  }
-  
+  }, []);
 
-  generateStyles() {
-    this.styles = createStyles(this.props);
-  }
+  // TODO: deprecate the check for object
```
mendyEdri, line 38: TODO hints for `isObject` check, but in `onPress` it's commented out

src/components/picker/PickerItem.js
```diff
@@ -16,153 +16,138 @@ import Text from '../text';
...
+    testID
+  } = props;
+
+  useEffect(() => {
+    if (_.isPlainObject(value)) {
       console.warn('UILib Picker.Item will stop supporting passing object as value & label (e.g {value, label}) in the next major version. Please pass separate label and value props');
     }
-  }
-  
+  }, []);
```
mendyEdri, line 36: There is a lint error,
We should either pass `value` as a dependency or no square brackets to trigger on each render

mendyEdri (changes requested):

src/components/picker/PickerItem.js
ethanshar: this is internal usage. 
`onSelectedLayout` is being sent from the Picker and we don't really need to pass on the props.

src/components/picker/PickerItem.js
ethanshar: I actually removed it in a following PR

src/components/picker/PickerItem.js
```diff
@@ -16,153 +16,138 @@ import Text from '../text';
...
     if (isSelected) {
       return <Image source={selectedIcon} tintColor={disabled ? Colors.dark60 : selectedIconColor}/>;
     }
-  }
-
-  renderItem() {
-    const {disabled} = this.props;
+  };
 
+  const _renderItem = () => {
```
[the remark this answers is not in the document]
ethanshar: I don't want to overkill the component with memoization.
I read in a blog post about when to useMemo and it says it's not recommended to overuse memoization cause it the it can also affect performance. 
From what I read, the thumb rule is to start simple and to consider memoizing when we face performance issues. 

memoizing this will require wrapping `getLabel` with `useCallback` and I prefer not to complicate it at the moment. 

Anyway, Personally I still not certain what is best practice with how often to use `useMemo` so again, for now I prefer to go easy on it..

src/components/picker/PickerItem.js
```diff
@@ -16,153 +16,138 @@ import Text from '../text';
...
+   * Callback for onPress action
+   */
+  onPress: PropTypes.func,
+  /**
+   * Callback for onLayout event
+   */
+  onSelectedLayout: PropTypes.func
+};
 
 export default PickerItem;
```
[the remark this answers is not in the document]
ethanshar: Due tho some of the of the props in PickerItem (like `renderItem` callback prop), it's very possible that memoizing the whole component might cause some side effects and break behaviors. 
For now I think it will be better to keep parity with the previous implementation (where PickerItem was simple Component and not a pure one) and later on consider optimizations if necessary.

src/components/picker/PickerItem.js
```diff
@@ -16,153 +16,138 @@ import Text from '../text';
...
-  renderSelectedIndicator() {
-    const {
-      isSelected,
-      disabled,
-      selectedIcon = Assets.icons.check,
-      selectedIconColor = Colors.primary
-    } = this.props;
+  };
 
+  const renderSelectedIndicator = () => {
```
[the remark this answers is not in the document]
ethanshar: done

src/components/picker/PickerItem.js
ethanshar: done

src/components/picker/PickerItem.js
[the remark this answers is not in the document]
mendyEdri: Gotcha. confusing topic indeed :)

src/components/picker/PickerItem.js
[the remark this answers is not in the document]
mendyEdri: Understood :)

mendyEdri (approved):
I've learned few things from your comments 👏

PR #1001, opened 2020-10-25
Infra/picker refactor #2

ethanshar:
## Description
(This PR depends on #1000)
- Use context to pass Picker value to Picker Items. 
This way we don't need to clone children and inject the props. 
Also this support rendering children that are not picker items (like section separators)

- Fixed issue with search functionality (broke after recent API migration) 
- Fixed issue with multi select mode with new API 

## Changelog
Refactor Picker component


src/components/picker/PickerItem.js
```diff
@@ -18,57 +20,58 @@ import Text from '../text';
...
-  };
-
   const selectedIndicator = useMemo(() => {
     if (isSelected) {
       return <Image source={selectedIcon} tintColor={disabled ? Colors.dark60 : selectedIconColor}/>;
     }
   }, [isSelected, disabled, selectedIcon, selectedIconColor]);
 
+  const _onPress = useCallback(() => {
+    context.onPress(value);
```
Inbal-Tish, line 53: don't we need to safely invoke 'onPress' here?

src/components/picker/index.js
```diff
@@ -337,7 +339,13 @@ class Picker extends PureComponent {
         onCancel: this.cancelSelect
       };
 
-      return renderCustomModal(modalProps);
+      return (
+        <>
```
Inbal-Tish, line 343: What is the use of the empty tag?

src/components/picker/PickerItem.js
```diff
@@ -18,57 +20,58 @@ import Text from '../text';
...
+  const itemLabel = getItemLabel(label, value, props.getItemLabel || context.getItemLabel);
+  const accessibilityProps = {
+    accessibilityState: isSelected ? {selected: true} : undefined,
+    accessibilityHint: 'Double click to select this suggestion',
+    ...Modifiers.extractAccessibilityProps(props)
+  };
 
   useEffect(() => {
     if (_.isPlainObject(value)) {
       console.warn('UILib Picker.Item will stop supporting passing object as value & label (e.g {value, label}) in the next major version. Please pass separate label and value props');
```
Inbal-Tish, line 42: Why not use LogService instead of console?

src/components/picker/PickerItem.js
ethanshar: nope. 
We own the context therefore we know we pass an `onPress` handler.

src/components/picker/index.js
ethanshar: It is shorter way to wrap JSX with React.Fragment, so instead of doing
```
<React.Fragment>
....
</React.Fragment>
```
you can just do 
```
<>
....
</>
```

src/components/picker/PickerItem.js
ethanshar: Fixed

Inbal-Tish (approved):

PR #1002, opened 2020-10-26
Commons - exporting props

Inbal-Tish:
#863  Description
Exporting BaseComponentInjectedProps, ForwardRefInjectedProps and ModifiersProps (to extract sub-modifiers props, like MarginModifiers).

## Changelog
Commons - exporting props.

### DIFF
--- a/src/commons/index.js
+++ b/src/commons/index.js
@@ -27,5 +27,14 @@ module.exports = {
   },
   get withScrollReached() {
     return require('./withScrollReached').default;
+  },
+  get BaseComponentInjectedProps() {
+    return require('./asBaseComponent').BaseComponentInjectedProps;
+  },
+  get ForwardRefInjectedProps() {
+    return require('./forwardRef').ForwardRefInjectedProps;
+  },
+  get ModifiersProps() {
+    return require('./modifiers').default;
   }
 };
--- a/src/index.js
+++ b/src/index.js
@@ -226,6 +226,15 @@ export default {
   get withScrollReached() {
     return require('./commons').withScrollReached;
   },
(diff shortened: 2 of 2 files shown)


src/commons/index.js
```diff
@@ -27,5 +27,14 @@ module.exports = {
   },
   get withScrollReached() {
     return require('./withScrollReached').default;
+  },
```
ethanshar, line 30: This one is not good.. 
There's a new temp file `commons/new.ts` (that will replace commons/index.js eventually) 
The typings are already exported there..

src/index.js
```diff
@@ -226,6 +226,15 @@ export default {
...
   },
+  get BaseComponentInjectedProps() {
+    return require('./commons').BaseComponentInjectedProps;
+  },
+  get ForwardRefInjectedProps() {
+    return require('./commons').ForwardRefInjectedProps;
+  },
+  get ModifiersProps() {
+    return require('./commons').default;
+  },
```
ethanshar, line 237: You need to export the typings in the ts files
`generatedTypes/index.d.ts`
`src/index.ts`

Inbal-Tish: Sorry, I don't understand what is what with all the fake files... Why they should be in `generatedTypes/index` if these are TS fils? And where all the other commons in `src/index.ts`?

src/commons/index.js
Inbal-Tish: Where is the call for the `commons/new.ts` file? I see only `commons/index.js`...

ethanshar: We're using it in the components that need these typings.. 
View, Text, Button.. almost all of them..

PR #1004, opened 2020-10-26
Export all modifiers and other common typings

ethanshar:
## Description
Export all common infra typings

## Changelog
Export all common infra typings

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,16 @@
 
 export * from './style';
 export * from './services';
+export {
+  BaseComponentInjectedProps,
+  ForwardRefInjectedProps,
+  ContainerModifiers,
+  MarginModifiers,
+  PaddingModifiers,
+  TypographyModifiers,
+  ColorsModifiers,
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -6,6 +6,16 @@
 
 export * from './style';
 export * from './services';
+export {
+  BaseComponentInjectedProps,
+  ForwardRefInjectedProps,
+  ContainerModifiers,
+  MarginModifiers,
```
Inbal-Tish, line 13: MarginModifiers and PaddingModifiers are part of ContainerModifiers, is there a reason to import them separately?

Inbal-Tish:
Thanks!

src/index.ts
ethanshar: Yes, some components only support margins, while some support the whole package

Inbal-Tish: Ok

Inbal-Tish (approved):

PR #1006, opened 2020-10-27
Fixes TouchableOpacity mandatory props error

Dor256:
## Description
`TouchableOpacity` had `RNTouchableOpacity` props mandatory, this PR makes it optional

## Changelog
Fixes TS errors in `TouchableOpacity`


typings/components/TouchableOpacity.d.ts
```diff
@@ -1,7 +1,7 @@
 
 import {TouchableOpacityProps as RNTouchableOpacityProps} from 'react-native';
 import {PureBaseComponent} from '../commons';
-import {AlignmentModifiers, FlexModifiers} from '../modifiers';
+import {AlignmentModifiers, FlexModifiers} from '../../src/commons/modifiers';
 import {ColorValue} from '../style/colors';
```
ethanshar, line 6: If I not mistaken, I think this file should be removed.. 
not sure why we still have it
We're using the auto generated file 
https://github.com/wix/react-native-ui-lib/blob/96e6f1e37a6df633744f9571fb8efbea58b4b25f/generatedTypes/components/touchableOpacity/index.d.ts#L1

Maybe that's the reason we had issues with the component typings?

Dor256: Could be, let me check this locally before we remove it

Dor256:
I deleted the old typings file, I'm not entirely sure yet that it will fix the problem but let's try

ethanshar (approved):

PR #1013, opened 2020-11-01
V6 - change - Renaming 'ComponentPropTypes' to 'ComponentProps'

Inbal-Tish:
## Description
Rename 'ComponentPropTypes' to 'ComponentProps'.

## Changelog
Rename 'ComponentPropTypes' to 'ComponentProps'.


ethanshar (commented):
I think this might create lots of annoying conflicts till we finally merge v6 (and it might take some time) 
So I suggest that for now to add new exports for the typings with the right names together with the current exports and eventually we can just remove them and fix them internally.. 
WDYT?

Inbal-Tish:
Sure

PR #1017, opened 2020-11-03
feat/tracking view allow safeArea

mendyEdri:
## Description
KeyboardTrackingView now allow users to disable the default safe-area space, default is true to avoid breaking changes 

## Changelog
KeyboardTrackingView now allow users to disable the default safe-area space by using `useSafeArea={false}`


M-i-k-e-l (commented):
Did you manage to see why the `addBottomView` is not working? Or is it something else?

mendyEdri:
> Did you manage to see why the `addBottomView` is not working? Or is it something else?

@M-i-k-e-l 
addBottomView is part of CustomKeyboardView and we are using here KeyboardTrackingView. 
(the addBottomView is the view between the closed keyboard to the bottom of the screen)

M-i-k-e-l (approved):

PR #1019, opened 2020-11-03
feat/useToggleValue

mendyEdri:
## Description
Introducing new utility hook to easily control any toggled values

## Changelog
Introducing `useToggleValue` to toggle between any tow values with ease!


src/commons/useToggleValue/index.ts
```diff
@@ -0,0 +1,23 @@
+import {useState, useRef} from 'react';
```
ethanshar, line 1: There's a little mess with the "commons" path and what we export there, also after TS it got a little messier due the `commons/new` file we had to add. 
So till we sort that mess I think it will be better to move our hooks into a different hierarchy, so what about keeping it under `src/hooks` ? 

Also, I'm considering if we should export each hook as a standalone or export them under "Hooks" namespace? 
So instead of the user import like this
```
import {useToggleValue} from 'react-native-ui-lib`
```

They will 
```
import {Hooks} from 'react-native-ui-lib`
const {useToggleValue} from Hooks;
```

WDYT?

mendyEdri: I like it! 
On it

ethanshar (approved):

PR #1021, opened 2020-11-04
fix firstLoad logic in floatingButton

lidord-wix:
## Description
Bug: floatingButton component shows briefly when visible=false.
Fix: update the firstLoad logic to keep `this.firstLoad` as true, until the first time that `visible` is set to true

## Changelog
Fix floatingButton visibility bug when visible is initialised to false

### DIFF
--- a/src/components/floatingButton/index.tsx
+++ b/src/components/floatingButton/index.tsx
@@ -64,10 +64,6 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
     this.firstLoad = true;
   }
 
-  componentDidMount() {
-    this.firstLoad = false;
-  }
-
   onAnimationEnd = () => {
     this.setState({animating: false});
   };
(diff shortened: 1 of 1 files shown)


src/components/floatingButton/index.tsx
```diff
@@ -117,6 +113,9 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
     const {withoutAnimation, secondaryButton, visible} = this.props;
     const Container = !withoutAnimation ? AnimatableView : View;
 
+    // NOTE: keep this.firstLoad as true as long as the visibility changed to true
+    this.firstLoad && !visible ? this.firstLoad = true : this.firstLoad = false;
```
Inbal-Tish, line 117: What if it visible=true, then visible=false, and then visible=false again?

lidord-wix: It won't cause a re-render since the props are the same..
The problem was only when the user sends visible=false at the beginning, and we changed the firstLoad to false even though the component isn't loaded yet

Inbal-Tish (approved):

PR #1024, opened 2020-11-08
Fix docs

### LINKED ISSUE #1022: Missed Button component from documentation site
**Describe the bug**
I cannot find description of Button component on the doc site
https://wix.github.io/react-native-ui-lib/getting-started/setup

**Component?**
Button
no_button_in_doc

ethanshar:
## Description
We encounter (again) with issues with the doc generation. 
Some components were missing from the docs site. 

After some investigation, I found out that react-docgen failed to parsing typescript files of components with **getter** methods. 
I tried to find to culprit, but unfortunately didn't find which package cause that. 

Anyway, I decided to make some changes in the way we generate our docs
- I excluded TS files 
- I added the deploy docs phase to our build (so now docs will be generated from the JS files after the TS build) 
- ~~The docs will now be created during CI build (something we wanted to do for a while)~~
- Fixed usage of _.chain, similar to what we did in private docs 

## Changelog
Fix issue with missing component in our docs site (Fix #1022)


uilib-docs/package.json
```diff
@@ -6,7 +6,7 @@
   "author": "Kyle Mathews <mathews.kyle@gmail.com>",
   "dependencies": {
     "classnames": "^2.2.6",
-    "gatsby": "2.23.22-static-query-template.8",
```
M-i-k-e-l, line 9: Did you test this on production?
We have this open bug: https://github.com/gatsbyjs/gatsby/issues/26038
(They want us to validate something but I haven't gotten around to it...)

ethanshar: It seems to work now. 
The current docs were generated from this branch.

package.json
```diff
@@ -32,7 +32,7 @@
     "log": "react-native log-ios | grep 'ethan -'",
     "docs:install": "(cd ./uilib-docs && rm -rf node_modules && rm -rf package-lock.json && npm install)",
     "docs:deploy": "(cd ./uilib-docs && gatsby build --prefix-paths && gh-pages -d public --branch gh-pages)",
-    "docs:build": "(cd ./uilib-docs && gatsby build --prefix-paths)",
+    "docs:build": "(cd ./uilib-docs && npm install && gatsby build --prefix-paths && gh-pages -d public --branch gh-pages)",
```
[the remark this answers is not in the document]
ethanshar: As you can see I scratch that in the PR description.
At first I called `docs:build` from the `build` script. 
But it seems to not work and require more investigation.. so I dropped it for now. 
`docs:build` is not being used at the moment.. but that should be the script we will want to run in CI once we'll make it work.

uilib-docs/package.json
M-i-k-e-l: I'm not sure it has been, locally I had `Button` and I don't see it in production.

ethanshar: I believe someone had deployed the docs from master and "broke" it :/ 
I just triggered docs again from this branch and now you can see the Button

M-i-k-e-l (approved):

PR #1027, opened 2020-11-10
add dateFormatter prop to dateTimePicker

lidord-wix:
## Description
Add dateFormatter prop to dateTimePicker

## Changelog
Add dateFormatter prop to dateTimePicker

### DIFF
--- a/src/components/dateTimePicker/index.js
+++ b/src/components/dateTimePicker/index.js
@@ -150,21 +150,18 @@ class DateTimePicker extends BaseComponent {
     const {value} = this.state;
     const {mode, dateFormat, timeFormat, dateFormatter, timeFormatter} = this.getThemeProps();
     if (value) {
-      if (dateFormatter && mode === MODES.DATE) {
-        return dateFormatter(value);
+      switch (mode) {
+        case MODES.DATE:
+          if (dateFormatter) {
+            return dateFormatter(value);
+          }
+          return dateFormat ? moment(value).format(dateFormat) : value.toLocaleDateString();
+        case MODES.TIME:
+          if (timeFormatter) {
+            return timeFormatter(value);
+          }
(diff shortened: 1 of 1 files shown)


src/components/dateTimePicker/index.js
```diff
@@ -140,8 +148,14 @@ class DateTimePicker extends BaseComponent {
 
   getStringValue = () => {
     const {value} = this.state;
-    const {mode, dateFormat, timeFormat} = this.getThemeProps();
+    const {mode, dateFormat, timeFormat, dateFormatter, timeFormatter} = this.getThemeProps();
     if (value) {
+      if (dateFormatter && mode === MODES.DATE) {
```
Inbal-Tish, line 153: I would use a switch case here and include all logic in each case to make the code more readable. What do you think?

lidord-wix: I changed to switch case - not sure what is more readable..
Let me know if you have an idea for a better logic here :)

src/components/dateTimePicker/index.js
```diff
@@ -144,20 +148,20 @@ class DateTimePicker extends BaseComponent {
...
+    const {mode, dateFormat, timeFormat, dateFormatter, timeFormatter} = this.getThemeProps();
     if (value) {
-      if (dateFormatter) {
-        return dateFormatter(value);
+      switch (mode) {
+        case MODES.DATE:
+          if (dateFormatter) {
+            return dateFormatter(value);
+          }
+          return dateFormat ? moment(value).format(dateFormat) : value.toLocaleDateString();
```
Inbal-Tish, line 158: You can have all in one trinery: 
return dateFormatter ?  dateFormatter(value) : dateFormat ? moment(value).format(dateFormat) : value.toLocaleDateString();

Inbal-Tish (approved):

PR #1034, opened 2020-11-15
LogService - fix typings

M-i-k-e-l:
## Description
LogService - fix typings

## Changelog
LogService - fix typings

### DIFF
--- a/generatedTypes/services/LogService.d.ts
+++ b/generatedTypes/services/LogService.d.ts
@@ -1,9 +1,9 @@
 declare function warn(message: string): void;
-declare function deprecationWarn({ component, oldProp, newProp }: ({
+declare function deprecationWarn({ component, oldProp, newProp }: {
     component: string;
     oldProp: string;
-    newProp: string;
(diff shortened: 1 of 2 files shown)


generatedTypes/services/LogService.d.ts
```diff
@@ -1,8 +1,8 @@
 declare function warn(message: string): void;
-declare function deprecationWarn({ component, oldProp, newProp }?: ({
-    component?: string;
-    oldProp?: string;
-    newProp?: string;
+declare function deprecationWarn({ component, oldProp, newProp }: ({
+    component: string;
+    oldProp: string;
+    newProp: string;
```
Inbal-Tish, line 5: 'newProp' parameter should stay optional as we sometimes deprecate a prop with no replacement. Which means we also need to edit the massage accordingly...

M-i-k-e-l: Added an option

Inbal-Tish (approved):

PR #1035, opened 2020-11-15
Calc modifiers in render methods so we can access context

ethanshar:
## Description
Stop using `getDerivedStateFromProp` in asBaseComponent to calculate modifiers. 
Move modifiers calculation to the render method where we can access the component's context

## Changelog
Fix issue with modifiers support based on context

### DIFF
--- a/src/commons/asBaseComponent.tsx
+++ b/src/commons/asBaseComponent.tsx
@@ -21,20 +21,9 @@ type ThemeComponent = {
 
 function asBaseComponent<PROPS, STATICS = {}>(WrappedComponent: React.ComponentType<any>): React.ComponentClass<PROPS & ThemeComponent> & STATICS {
   class BaseComponent extends UIComponent {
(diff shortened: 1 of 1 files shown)


src/commons/asBaseComponent.tsx
```diff
@@ -21,31 +21,21 @@ type ThemeComponent = {
...
-      return null;
-    }
-
     static getThemeProps = (props: any, context: any) => {
       return Modifiers.getThemeProps.call(WrappedComponent, props, context);
     }
 
     render() {
       const themeProps = BaseComponent.getThemeProps(this.props, this.context);
+      const modifiers = Modifiers.generateModifiersStyle(undefined, themeProps);
```
M-i-k-e-l, line 34: Is it worth it to memoize the result?

ethanshar: I thought about it in the past. 
But I think memoizing it will be too much since there can so many variations of props.

M-i-k-e-l (approved):

PR #1038, opened 2020-11-18
Feat/basic ui

mendyEdri:
## Description
UI changes and `active row` logic


demo/src/screens/incubatorScreens/WheelPickerScreen.tsx
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
-import {View, Text, Incubator} from 'react-native-ui-lib';
+import React, {useCallback} from 'react';
+import {View, Text, Incubator, Typography, Colors} from 'react-native-ui-lib';
 import _ from 'lodash';
+import { ItemProps } from 'src/incubator/WheelPicker/Item';
```
ethanshar, line 4: make sure to format the code with prettier (the curly braces spacings are not according to our standards) 
I think eslint should have given you error for that

src/incubator/WheelPicker/Item.tsx
```diff
@@ -30,56 +31,28 @@ export default ({
   // value,
   itemHeight,
   onSelect,
-  offset
+  activeIndex
```
ethanshar, line 34: Not using to offset makes the transition of active/inactive look bad. 
I believe you'll have to use it in order to transition nicely the colors

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,90 +2,164 @@
...
+  const scrollView = useRef<Animated.ScrollView>();
   const [offset] = useValues([0], []);
-  const onScroll = onScrollEvent({y: offset});
+  const activeIndex = useRef<number>(0);
+
+  const wrapItems = (items?: ItemProps[]): WrappedItem[] => {
+    return _.map(items, (item, index: number) => {
+      return {...item, isActive: activeIndex.current === index};
+    });
+  };
```
ethanshar, line 55: what's the purpose of `wrapItems`? 
I see you only use it for the keyExtractor, is that really necessary? 
What does it solve?

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,90 +2,164 @@
...
+    const calculatedIndex = Math.round(offset / itemHeight);
+    return valueInRange(calculatedIndex, 0, items!.length - 1);
+  };
+
+  const onScroll = useCallback((event: NativeSyntheticEvent<NativeScrollEvent>) => {
+    const currentIndex = calculateCurrentIndex(event.nativeEvent.contentOffset.y);
+    if (activeIndex.current != currentIndex) {
+      activeIndex.current = currentIndex;
+      updateItems();
+    }
```
ethanshar, line 97: This will set all of the items whenever the index change. 
Which happens a lot in wheel pickers and in a very rapid pace. 
I think it can be very problematic for performance.

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,90 +2,164 @@
...
-  const scrollview = useRef<Animated.ScrollView>();
+type WrappedItem = {
+  isActive: boolean;
+} & ItemProps;
+
+const WheelPicker = ({items, itemHeight = 48, activeItemTextStyle, inactiveItemTextStyle, onChange: onChangeEvent}: WheelPickerProps) => {
+  const height = itemHeight * 5;
+  const scrollView = useRef<Animated.ScrollView>();
   const [offset] = useValues([0], []);
-  const onScroll = onScrollEvent({y: offset});
```
ethanshar, line 28: using onScrollEvent is actually very helpful for us.
It maps the scroll event on the native side and set whatever value we want. 
This is helpful cause it can save us from updating our state. 

In general the whole implementation of this component, using Reanimated, was planned to fix performance issue. 
Using Reanimate, or animated values in general is a great way to store values over the bridge and access them using interpolation to create various transitions and animation directly on the native thread. 

It's a little different approach from what we usually do

src/incubator/WheelPicker/Item.tsx
```diff
@@ -30,56 +31,28 @@ export default ({
...
-      // outputRange: [0.8, 1, 0.8], // with scale change
-      outputRange: [1, 1, 1],
-    });
-  }, [itemHeight]);
+  
+  const isActive = activeIndex === index;
 
   const selectItem = useCallback(() => onSelect(index), [index]);
 
+  const color = isActive ? Colors.primary : Colors.grey20;
```
ethanshar, line 41: we will need to find a way to transition the color using animation. it's look bulky at the moment.

src/incubator/WheelPicker/Item.tsx
```diff
@@ -30,56 +31,28 @@ export default ({
...
+  const font = isActive ? Typography.text60BO : Typography.text60R;
```
ethanshar, line 42: Unfortunately, there's no way (I know of) to transition between font weights **with an animation**. 
I personally think we should not have that at the moment, but we can discuss it with the UX.

ethanshar (commented):
I wrote in one of the comments about how I approached this component. 
Unlike many other components, this one requires a very delicate handling when it comes to performance because the interaction here is very important. 
That's the reason I used animated values in many cases. 

Animated values are a great way to communicate with the native side without going over the bridge and without updating the state too often. 

Go over my comments, we can sync about it if you like.

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,90 +2,164 @@
...
-          <View
-            style={{
-              borderTopWidth: 1,
-              borderBottomWidth: 1,
-              height: itemHeight
-            }}
-          />
-        </View>
+        {renderFader(FaderPosition.BOTTOM)}
+        {renderFader(FaderPosition.TOP)}
```
ethanshar, line 156: These are great alternative for the fading animation. Nice 👍

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,55 +2,105 @@
...
-  itemTextStyle?: StyleProp<TextStyle>;
-  onChange: (item: ItemProps, index: number) => void;
+  /**
+   * TextStyle for the focused row
+   */
+  selectedTextColor?: string;
+  /**
+   * TextStyle for other, non-focused rows
+   */
+  unselectedTextStyle?: string;
```
ethanshar, line 34: same comment here for the names

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,55 +2,105 @@
...
+  const listSize = items?.length || 0;
+  const middleIndex = useMiddleIndex({itemHeight, listSize});
 
   const selectItem = useCallback(
     index => {
-      if (scrollview.current) {
-        scrollview.current.getNode().scrollTo({y: index * itemHeight, animated: true});
+      if (scrollView.current?.getNode()) {
+        //@ts-ignore for some reason scrollToOffset isn't recognized
+        scrollView.current.getNode().scrollToOffset({offset: index * itemHeight, animated: true});
```
ethanshar, line 58: Would that work?
```suggestion
        //@ts-ignore for some reason scrollToOffset isn't recognized
        scrollView.current?.getNode()?.scrollToOffset({offset: index * itemHeight, animated: true});
```

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,55 +2,105 @@
...
+  }, [itemHeight]);
+
+  const renderFader = useMemo(
+    () => (position: FaderPosition) => {
+      return <Fader visible position={position} size={60} />;
+    },
+    []
+  );
+
+  const renderSeparators = useMemo(() => {
```
ethanshar, line 91: As discussed in last dev sync, I think it's more readable to name `useMemo` values as the object it returns 
```suggestion
  const separators = useMemo(() => {
```

ethanshar (commented):
Looks great! 
wrote few comments on naming..

src/incubator/WheelPicker/Item.tsx
```diff
@@ -19,67 +19,49 @@ interface InternalProps extends ItemProps {
   index: number;
   offset: any;
   itemHeight: number;
-  textStyle?: TextStyle;
+  selectedColor?: string;
+  unselectedColor?: string;
```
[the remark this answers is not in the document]
mendyEdri: Agree!

ethanshar (approved):

PR #1043, opened 2020-11-23
Modifiers - expose extractAccessibilityProps & extractComponentProps

M-i-k-e-l:
## Description
Modifiers - expose extractAccessibilityProps & extractComponentProps

## Changelog
Modifiers - expose extractAccessibilityProps & extractComponentProps

### DIFF
--- a/src/commons/index.js
+++ b/src/commons/index.js
@@ -28,10 +28,7 @@ module.exports = {
   get withScrollReached() {
     return require('./withScrollReached').default;
   },
-  get extractAccessibilityProps() {
-    return require('./modifiers').extractAccessibilityProps;
-  },
-  get extractComponentProps() {
-    return require('./modifiers').extractComponentProps;
+  get modifiers() {
+    return require('./modifiers');
   }
 };
--- a/src/commons/index.js
+++ b/src/commons/index.js
@@ -27,5 +27,11 @@ module.exports = {
   },
   get withScrollReached() {
     return require('./withScrollReached').default;
+  },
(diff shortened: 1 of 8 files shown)


src/commons/index.js
```diff
@@ -27,5 +27,11 @@ module.exports = {
   },
   get withScrollReached() {
     return require('./withScrollReached').default;
+  },
```
ethanshar, line 30: Let's extract the whole Modifiers file..

M-i-k-e-l: Fixed

src/index.ts
```diff
@@ -18,6 +18,7 @@ export {
 } from './commons/new';
 export * as Incubator from './incubator';
 export * as Hooks from './hooks';
+export * as Modifiers from './commons/modifiers';
```
ethanshar, line 21: I think this is incorrect. 
This will export all Modifiers functions separately, while we want to export it as a whole class.

ethanshar (changes requested):

src/index.ts
M-i-k-e-l: You can see the usage in the private lib's PR, it is as follows:
```
import {Modifiers} from 'react-native-ui-lib';
{...Modifiers.extractAccessibilityProps(this.props)}
```
Is that not what you meant?

ethanshar: yes.. it's actually correct. 
Not sure why it look weird before :/
merging..

ethanshar (approved):

PR #1045, opened 2020-11-25
ColorPicker - 'value' and 'dialogProps' props deprecations

Inbal-Tish:
## Description
deprecating 'value' in favor of 'selectedValue' prop.
deprecating 'dialogProps' prop.

## Changelog
ColorPicker - 'value' (pass 'selectedValue' instead) and 'dialogProps' deprecation (pass Dialog props without object wrapper).

### DIFF
--- a/src/components/colorPicker/ColorPickerDialog.js
+++ b/src/components/colorPicker/ColorPickerDialog.js
@@ -4,6 +4,7 @@ import React, {PureComponent} from 'react';
 import {LayoutAnimation, StyleSheet, Keyboard, TextInput, PixelRatio, I18nManager} from 'react-native';
 import {Constants} from '../../helpers';
 import {asBaseComponent} from '../../commons';
+import {extractComponentProps} from '../../commons/modifiers';
 import Assets from '../../assets';
 import {Colors, Typography} from '../../style';
 import View from '../view';
@@ -36,7 +37,7 @@ class ColorPickerDialog extends PureComponent {
      */
     onSubmit: PropTypes.func,
     /**
-     * Props to pass the Dialog component // TODO: deprecate 'dialogProps' prop
+     * Props to pass the Dialog component // TODO: remove 'dialogProps' prop after deprecation
      */
     dialogProps: PropTypes.object,
     /**
@@ -294,11 +295,11 @@ class ColorPickerDialog extends PureComponent {
   }
 
   renderDialog() {
-    const {visible, dialogProps, testID} = this.props;
+    const {dialogProps, testID} = this.props;
+    const dialogProp = extractComponentProps(Dialog, this.props);
 
     return (
       <Dialog
-        visible={visible} //TODO: pass all Dialog props instead
         width="100%"
         height={null}
         bottom
@@ -309,6 +310,7 @@ class ColorPickerDialog extends PureComponent {
         testID={`${testID}.dialog`}
         supportedOrientations={['portrait', 'landscape', 'landscape-left', 'landscape-right']} // iOS only
         {...dialogProps}
+        {...dialogProp}
       >
         {this.renderHeader()}
         {this.renderPreview()}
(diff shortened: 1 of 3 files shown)


src/components/colorPicker/ColorPickerDialog.js
```diff
@@ -294,11 +295,11 @@ class ColorPickerDialog extends PureComponent {
   }
 
   renderDialog() {
-    const {visible, dialogProps, testID} = this.props;
+    const {dialogProps, testID} = this.props;
+    const dialogProp = extractComponentProps(Dialog, this.props);
```
ethanshar, line 299: What's the reason for that? 
Why would you want to deprecate a dedicated prop for dialog? 

I don't think flatting dialog props into ColorPicker props is a good call. 
ColorPicker doesn't really inherit a Dialog, it is more compose of a Dialog.

src/components/colorPicker/index.js
```diff
@@ -34,9 +35,10 @@ class ColorPicker extends PureComponent {
      */
     colors: PropTypes.arrayOf(PropTypes.string),
     /**
-     * The value of the selected swatch // TODO: rename prop 'selectedValue'
+     * The value of the selected swatch // TODO: remove after migration to 'selectedValue'
      */
     value: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
+    selectedValue: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
ethanshar, line 41: What's the reason for the rename? 
personally, `value` and `selectedValue` feel the same. 
Do you want change it for readability purpose? cause that will just add another migration for us..

src/components/colorPicker/ColorPickerDialog.js
Inbal-Tish: ColorPicker doesn't inherit Dialog but ColorPickerDialog is kinda is and 'dialogProps' prop is its prop to deprecate...

src/components/colorPicker/index.js
Inbal-Tish: It is for readability. Ok, I can leave it if it doesn't feel necessary.

src/components/colorPicker/ColorPickerDialog.js
ethanshar: but we still want to have a `dialogProps` prop so we can control how the dialog of ColorPicker looks.. no?

Inbal-Tish: I don't think we need a prop for that. The user should pass Dialog's props, not inside the 'dialogProps' object

ethanshar: Ok, I got a little confused...
I see you want to remove `dialogProps` from `ColorPickerDialog` - that makes sense!
But we should still have a `dialogProps` prop in the main ColorPicker component for channeling these prop to `ColorPickerDialog`.

In the previous refactor you spread `ColorPickerDialog` propsTypes into ColorPicker so it was included. 
Now, removing it will remove it completely - you should declare it in the main component. 

Let's talk if you still think otherwise, ill explain in details why do we need it..

Inbal-Tish: I want the user to pass Dialog props to ColorPicker as he would to the Dialog and I will extract these props and pass them to ColorPickerDialog. I don't think there is a need for the object wrapper that is 'dialogProps' prop. What do you think?

ethanshar: I think that there's a need for it. 
Let's say you have a prop called `containerStyle` in both Dialog and ColorPicker component. 
It can creates conflicts. 
I believe the separation here is important and having a dedicate prop as `dialogProps` is crucial. 

I think the thumb rule in this case is whether the internal component is also a standalone component (in this case, dialog is also a standalone component) 

If we'll take another case, for instance ListItem and ListItemLeftPart. 
ListItemLeftPart is just an internal implementation of ListItem, it doesn't live on its own, therefore it make sense to spread ListItemLeftPart props in ListItem props.

Does it make it sense?

Inbal-Tish: Ok. I'm closing this PR

PR #1046, opened 2020-11-25
Use Box object to wrap usage of CustomKeyboardLayout

savelichalex:
## Description
Currently there is a bug with KeyboardAccessoryView, when a custom keyboard is used on android it can't be closed (height is not set to 0) after reloading JS bundle.

The problem is that on Android, when JS bundle is reloaded, a `KeyboardInputPackage` is also reloaded, and two methods are called, first is [init](https://github.com/wix/react-native-ui-lib/blob/master/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/KeyboardInputPackage.java#L40) and second one is [createNativeModules](https://github.com/wix/react-native-ui-lib/blob/master/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/KeyboardInputPackage.java#L24). In `init` method a `CustomKeyboardLayout` is recreated with a new react context. But the [createViewManager](https://github.com/wix/react-native-ui-lib/blob/master/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/KeyboardInputPackage.java#L30) is not called during reload, therefore `CustomKeyboardRootViewManager` is using old pointer to `CustomKeyboardLayout`, because of that, when we trying to close custom keyboard it can't access shadow node to set height to 0.

In this PR I wrapped `CustomKeyboardLayout` with a `Box` object to have ability to use one pointer across modules and change instance dynamically.

## Changelog
Fix a bug when KeyboardAccessoryView didn't work after JS bundle reload

### DIFF
--- a/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardLayout.java
+++ b/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardLayout.java
@@ -186,4 +186,19 @@ public class CustomKeyboardLayout implements ReactSoftKeyboardMonitor.Listener, 
         }
     }
 
+    public static class Box {
+        private CustomKeyboardLayout instance;
+
+        Box() {
+            this.instance = null;
+        }
+
+        public void setInstance(CustomKeyboardLayout instance) {
+            this.instance = instance;
+        }
+
+        public CustomKeyboardLayout getInstance() {
+            return instance;
+        }
+    }
 }
--- a/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardRootView.java
+++ b/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardRootView.java
@@ -8,9 +8,9 @@ import androidx.annotation.NonNull;
 
 public class CustomKeyboardRootView extends FrameLayout {
 
-    private final CustomKeyboardLayout mLayout;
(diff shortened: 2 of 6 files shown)


M-i-k-e-l:
Hi @savelichalex,
Sorry for the late response.

Two questions:
1. Does this bug occur only in debug, or did I misunderstand? - If so then I'm not sure we would like to add code for it.
2. Can this bug be fixed with a new `key` to the component (on your JS code)? - It feels like that's a simpler solution with no code change.

savelichalex:
Hi @M-i-k-e-l 

1. Mostly - yes. But I could imagine a situation with services like CodePush, when bridge instance is reloaded. Anyway even in debug this is a huge pain. I spent whole day trying to understand why it's not working on Android, until realised that it works until reload and then spent another few hours with debugger in AS.
2. How a `key` could help? It will get a wrong layout anyway, as with every reload it just get a wrong pointer from the first bundle initialisation.

M-i-k-e-l:
Hi @savelichalex,
I've never worked with CodePush, but I think that it only applies changes when a restart of the app occurs.
To be sure, is the reload we're talking about is when you've changed your JS code?

savelichalex:
> To be sure, is the reload we're talking about is when you've changed your JS code?

Yes.

M-i-k-e-l:
I'm sorry for the trouble you've faced, the time you've lost, and thank you for the effort you've made to fix this issue, however I'm not sure we want to add code that is only useful in debug version. @ethanshar, what do you think?

ethanshar:
@M-i-k-e-l 
I agree.

PR #1047, opened 2020-11-25
fix/chip_ts_issue

lidord-wix:
## Description
Fix TS issue in chip component by using Partial typings.

### DIFF
--- a/src/components/chip/index.tsx
+++ b/src/components/chip/index.tsx
@@ -1,15 +1,12 @@
 import _ from 'lodash';
 import React, {useCallback} from 'react';
-import {StyleSheet, StyleProp, ViewStyle, ViewProps, ImageStyle, ImageProps, TextStyle, ImageSourcePropType} from 'react-native';
-// @ts-ignore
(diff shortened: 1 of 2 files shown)


src/components/chip/index.tsx
```diff
@@ -76,7 +76,7 @@ export type ChipProps = ViewProps & TouchableOpacityProps & {
   /**
    * Additional icon props
    */
-  iconProps?: ImageProps
+  iconProps?: Partial<ImageProps>;
```
M-i-k-e-l, line 79: 1. This means that **any** prop from `ImageProps` is optional, although I'm not sure there is any other prop that is mandatory, there might be in the future, so maybe another solution is better - look in `CardImage`.

2. There's still an error in the private lib with this change.

lidord-wix: Actually that was my second option, I thought the one I chose is more generic..

M-i-k-e-l: I don't think we want generic here, we want it specific.
Not that I like the option I suggested (not very pretty), but I think it's better than this one.

M-i-k-e-l (approved):

PR #1059, opened 2020-12-03
Fix/picker get label

mendyEdri:
## Description
Fixed `getLabel` property issue on `multi-select` mode.

## Changelog
Fixed `getLabel` property issue on `Picker`'s `multi-select` mode.

### DIFF
--- a/src/components/picker/index.js
+++ b/src/components/picker/index.js
@@ -243,12 +243,12 @@ class Picker extends PureComponent {
   getLabel(value) {
     const {getLabel} = this.props;
 
+    if (_.isFunction(getLabel) && !_.isUndefined(getLabel(value))) {
+      return getLabel(value);
+    }
+    
     if (_.isArray(value)) {
       return this.getLabelsFromArray(value);
-    }
-
-    if (_.isFunction(getLabel)) {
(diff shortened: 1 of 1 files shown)


src/components/picker/index.js
```diff
@@ -224,19 +224,9 @@ class Picker extends PureComponent {
...
-    const {value: propsValue} = this.props;
-    const {value: stateValue} = this.props;
-
-    if (this.shouldNotChangePickerLabelWhileSelecting()) {
-      return this.getLabel(propsValue);
-    }
-    return this.getLabel(stateValue);
+    const {value} = this.props;
+    return this.getLabel(value);
   };
```
ethanshar, line 230: I think that now getLabelValueText is redundant.. 
it's just a wrapper of `this.getLabel`. 
Can we remove it?

mendyEdri: Right 😁

ethanshar: I think it was valid before when you had some logic there.. 
But now we use it only in one place and all it does it extract the value prop. Which seems a little redundant, no?

ethanshar (approved):

PR #1063, opened 2020-12-07
replace @react-native-community/picker with @react-native-picker/picker

ethanshar:
## Description
**Safely** migrate from @react-native-community/picker to @react-native-picker/picker

## Changelog
Migrate from @react-native-community/picker to @react-native-picker/picker


lib/components/WheelPicker/index.js
```diff
@@ -2,11 +2,26 @@ import _ from 'lodash';
...
+/* Safely require picker component */
+let Picker = () => null;
+try {
+  Picker = require('@react-native-picker/picker').Picker; // New package
+} catch (e) {
+  try {
+    Picker = require('@react-native-community/picker').Picker; // Deprecated package
+    console.warn(`RNUILib Picker will soon migrate to use "@react-native-picker/picker" package instead of '@react-native-community/picker'`);
+  } catch (e) {
+    if (Constants.isIOS) {
```
M-i-k-e-l, line 19: This `if` can surround the whole section (the `try` `catch`).

lib/components/WheelPicker/index.js
```diff
@@ -2,11 +2,26 @@ import _ from 'lodash';
...
-import {Picker} from '@react-native-community/picker';
+
 import WheelPickerItem from './WheelPickerItem';
 import {Constants} from '../../../src/helpers';
 import {Typography, Colors} from '../../../src/style';
 
+/* Safely require picker component */
+let Picker = () => null;
+try {
+  Picker = require('@react-native-picker/picker').Picker; // New package
```
M-i-k-e-l, line 13: It seems that the metro (I assume) knows to take `@react-native-community/picker` when requiring `@react-native-picker/picker` (I did not clean-install and it works for me), do we want to keep this just in case?

M-i-k-e-l (commented):
Expo `package.json` still has the old lib, it appears that they have [already migrated](https://docs.expo.io/versions/latest/sdk/picker/), so I think we can move that as well.

lib/components/WheelPicker/index.js
ethanshar: Yes I noticed that as well, But I kept it just in case.

lib/components/WheelPicker/index.js
ethanshar: Fixed

M-i-k-e-l (commented):
I think you've missed the `Expo package.json` comment (it was a general comment since there is no file to associate).

ethanshar:
> I think you've missed the `Expo package.json` comment (it was a general comment since there is no file to associate).

right, fixed!

M-i-k-e-l (approved):

PR #1067, opened 2020-12-08
checkbox variant - outline

mackbrowne:
## Description

Alternative style for checkbox that wasn't possible with basic styling and theming alone.

Screenshot 2020-12-08 at 19 11 57
Screenshot 2020-12-08 at 19 11 52

## Changelog
Added optional boolean outline prop.  The checkbox acts normal when false and by default.

### DIFF
--- a/src/components/checkbox/index.tsx
+++ b/src/components/checkbox/index.tsx
@@ -166,11 +166,20 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     }
   };
 
-  getColor = () => this.props.disabled ? DEFAULT_DISABLED_COLOR : this.props.color || DEFAULT_COLOR;
+  getColor = () => (this.props.disabled ? DEFAULT_DISABLED_COLOR : this.props.color || DEFAULT_COLOR);
 
-  getBackgroundColor = () => this.props.outline ? 'transparent' : this.getColor();
+  getBackgroundColor = () => (this.props.outline ? 'transparent' : this.getColor());
 
-  getTintColor = () => this.props.outline ? this.getColor() : DEFAULT_ICON_COLOR;
+  getTintColor = () => {
+    const {outline, disabled, iconColor} = this.props;
+    if (outline) {
+      if (disabled) return DEFAULT_DISABLED_COLOR;
+      else return iconColor || DEFAULT_COLOR;
+    } else {
+      if (disabled) return Colors.white;
+      else return iconColor || Colors.white;
+    }
+  };
 
   getBorderStyle() {
     const borderColor = {borderColor: this.getColor()};
@@ -231,7 +240,12 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
 }
 
 function createStyles(props: CheckboxProps) {
-  const {color = DEFAULT_COLOR, iconColor = DEFAULT_ICON_COLOR, size = DEFAULT_SIZE, borderRadius = DEFAULT_BORDER_RADIUS} = props;
+  const {
+    color = DEFAULT_COLOR,
+    iconColor = DEFAULT_ICON_COLOR,
+    size = DEFAULT_SIZE,
+    borderRadius = DEFAULT_BORDER_RADIUS
+  } = props;
 
   return StyleSheet.create({
     container: {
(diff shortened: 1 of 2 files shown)


src/components/checkbox/index.tsx
```diff
@@ -186,14 +201,18 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
+              this.styles.container,
+              {backgroundColor: outline ? 'transparent' : disabled ? DEFAULT_DISABLED_COLOR : this.getColor()},
+              {opacity: this.animationStyle.opacity}
+            ]}
           >
             <Animated.Image
               style={[
                 this.styles.selectedIcon,
-                color && {tintColor: iconColor},
+                color && {tintColor: outline ? this.getColor() : DEFAULT_ICON_COLOR},
```
ethanshar, line 213: There's an issue here. 
You ignore the `iconColor` prop. 
If a checkbox starts disabled and then become enabled in runtime. 
The checkmark icon color will stay grey 

Also, with when passing the `outline` flag, `iconColor` is being ignored. 

I suggest moving the logic of the icon color to a separate  function (getIconColor) and apply in runtime. 
setting it in `createStyles` will not take effect in runtime..

ethanshar (commented):
Hi @mackbrowne 
Great feature. Thanks for the PR. 
See my comment, I found two issues.

src/components/checkbox/index.tsx
mackbrowne: @ethanshar thanks for the quick feedback!  I'll add this soon. 

I also noticed a bug with 'outline - no color - disabled` seen in the screenshot above which I think is caused by the runtime scenario you explained.   I'll do another round of cleaning it up today and get back to you.

mackbrowne: @ethanshar this should be fixed now with the new commit!  

I also took the liberty of creating DEFAULT_BORDER_WIDTH and DEFAULT_BORDER_RADIUS constants and included them with the rest of the constants for this page.  I hope this is ok!

mackbrowne:
Updated Screenshots 

Simulator Screen Shot - iPhone 11 - 2020-12-09 at 13 55 02
Simulator Screen Shot - iPhone 11 - 2020-12-09 at 13 55 06

src/components/checkbox/index.tsx
mackbrowne: @ethanshar  what is the process here?  Should I resolve this or leave it for you to review then close?

src/components/checkbox/index.tsx
```diff
@@ -159,14 +166,15 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
 
-  getColor() {
-    const {color, disabled} = this.props;
-    return disabled ? DEFAULT_DISABLED_COLOR : color || DEFAULT_COLOR;
-  }
+  getColor = () => this.props.disabled ? DEFAULT_DISABLED_COLOR : this.props.color || DEFAULT_COLOR;
+
+  getBackgroundColor = () => this.props.outline ? 'transparent' : this.getColor();
+
+  getTintColor = () => this.props.outline ? this.getColor() : DEFAULT_ICON_COLOR;
```
ethanshar, line 173: Hi, 
This still has issue. 
It ignore the `iconColor` prop and use the `color` prop instead. 
See my code suggestion below
```suggestion
  getTintColor = () => {
    const {outline, disabled, iconColor} = this.props;
    let tintColor = iconColor;
    if (outline) {
      tintColor = disabled ? DEFAULT_DISABLED_COLOR : tintColor || DEFAULT_COLOR;
    } else {
      tintColor = disabled ? Colors.white : tintColor || Colors.white;
    }
    return tintColor;
  }
```

src/components/checkbox/index.tsx
ethanshar: Hi,
Sorry for the late response. 
There's still small issue.
See my comment.

src/components/checkbox/index.tsx
mackbrowne: I understand now.  I changed this suggestion slightly in my commit to use more if/else instead of terniary operators.  I find it more readable now.  Hope that's ok!

ethanshar (approved):

PR #1069, opened 2020-12-09
Fix/drawer types source

Inbal-Tish:
## Description
Removing Drawer's manual typings and fixing generated types.
Adding 'customValue' prop to pass along from user to component and back on component's events.
Declaring props.

## Changelog
Drawer - Removing Drawer's manual typings and fixing generated types. Adding 'customValue' prop to pass along from user to component and back on component's events. Declaring missing props.


generatedTypes/index.d.ts
```diff
@@ -35,6 +35,7 @@ export {default as TouchableOpacity, TouchableOpacityProps} from './components/t
 export {default as Button, ButtonPropTypes, ButtonProps} from './components/button';
 export {default as Checkbox, CheckboxPropTypes, CheckboxProps} from './components/checkbox';
 export {default as Chip, ChipPropTypes, ChipProps} from './components/chip';
+export {default as Drawer, DrawerProps, ItemProps} from './components/drawer';
```
ethanshar, line 38: We can't export a typing called `ItemProps` it's too vague. 
Better rename it to DrawerItemProps

generatedTypes/index.d.ts
```diff
@@ -67,7 +68,6 @@ export {default as Carousel, CarouselProps} from './components/carousel';
 export {
   ActionSheet,
   ConnectionStatusBar,
-  Drawer,
```
ethanshar, line 70: Make sure to update the exports also in `src/index.ts` file

generatedTypes/index.d.ts
Inbal-Tish: Done

generatedTypes/index.d.ts
Inbal-Tish: Done

ethanshar (approved):

PR #1071, opened 2020-12-10
Disable panning in native picker dialogs

ethanshar:
## Description
Disable panning in native picker dialogs since they contains a wheel picker, scrolling thru the picker conflicts with the dialog panning. So it's best to just disable the panning in this case. 

## Changelog
Fix wheePicker swipe conflicts with dialog panning by disabling the dialog panning


src/components/picker/PickerDialog.android.js
```diff
@@ -83,7 +83,7 @@ class PickerDialog extends BaseComponent {
   render() {
     const dialogProps = extractComponentProps(Dialog, this.props);
```
mendyEdri, line 84: `this.props` already contains `panDirection` as null, passed from `NativePicker`.
For some reason `extractComponentProps` not populating it when spreading passed props

ethanshar: You are right, I missed that. 
It should have already worked. 
Checking...

ethanshar: Fixed (kinda)

mendyEdri (approved):

PR #1078, opened 2020-12-15
ScrollBar - TS migration

Inbal-Tish:
## Description
ScrollBar - migrate to TS.

## Changelog
ScrollBar - migrate to TS.

### DIFF
--- a/src/components/scrollBar/index.tsx
+++ b/src/components/scrollBar/index.tsx
@@ -17,7 +17,7 @@ import View from '../view';
 import Image from '../image';
 
 
-interface Props extends FlatListProps<any> {
+interface Props extends FlatListProps<any>, ForwardRefInjectedProps {
   /**
      * Whether to use a FlatList. NOTE: you must pass 'data' and 'renderItem' props as well
      */
@@ -61,7 +61,7 @@ interface Props extends FlatListProps<any> {
     focusIndex?: number
 };
 
-export type ScrollBarProps = Props & ForwardRefInjectedProps;
+export type ScrollBarProps = Props;
 
 export type State = {
   gradientOpacity: Animated.Value,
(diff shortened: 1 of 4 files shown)


ethanshar (commented):
You need to export it also in src/index.ts and in generateTypes/index.d.ts 
It's not exported as a stub

ethanshar:
@Inbal-Tish 
Im getting TS errors locally when I run `npm run build:dev` 
It's weird that the CI still pass green. 
Anyway, can you check..

Inbal-Tish:
I know there are errors. I told you I have an issue with 'forwardedRef', among others...

src/components/scrollBar/index.tsx
```diff
@@ -17,7 +17,7 @@ import View from '../view';
 import Image from '../image';
 
 
-interface Props extends FlatListProps<any> {
+interface Props extends FlatListProps<any>, ForwardRefInjectedProps {
```
ethanshar: The props you're using for export should not include internal typings like `ForwardRefInjectedProps`.

`ForwardRefInjectedProps` typings are internal, we use them because we wrap our component with the `forwardRef` HOC. 

The way we do that is like this: 

```
interface Props {}

type InternalProps = Props & ForwardRefInjectedProps; 


// Here you use "InternalProps" cause you also expecting to get internal props like "forwardedRed", etc.. 
class YourComponent extends Component<InternalProps, State> {}

// Here you use "Props" - this is the API we expose for our users. 
export default asBaseComponent<Props, typeof YourComponent>(forwardRef<Props>(YourComponent));
```

Inbal-Tish: So what do I export? If I do this `export type ScrollBarProps = Props & ForwardRefInjectedProps;` I also export the internal prop, no?

ethanshar: Yes, 
See my code snippet. 
I break the typings into two `Props` & `InternalProps`. Eventually I export `Props` only

Inbal-Tish: Done

ethanshar (approved):

PR #1079, opened 2020-12-16
Typescript/tab controller move to typescript

M-i-k-e-l:
## Description
TabController - move to typescript
This is step one from 3 (step 2 is move to functional component and 3 is to add a presenter for the centering).
There are a few TODOs for typescript, some will be removed once the move to functional component is done, but I'm not sure all will be removed, especially the ones that are related to reanimated, they don't export all of their types.

## Changelog
TabController - move to typescript


src/index.ts
```diff
@@ -36,6 +36,7 @@ export {default as Overlay, OverlayTypes} from './components/overlay';
 export {default as RadioButton, RadioButtonPropTypes, RadioButtonProps} from './components/radioButton/RadioButton';
 export {default as RadioGroup, RadioGroupPropTypes, RadioGroupProps} from './components/radioButton/RadioGroup';
 export {default as Switch, SwitchProps} from './components/switch';
+export {default as TabController, TabControllerProps, TabBarItemProps} from './components/tabController';
```
ethanshar, line 39: `TabBarItemProps` might conflict in the future with our TabBar component which will probably have also TabBatItemProps :/ 
Maybe we can name this one TabControllerItemProps, WDYT?

src/components/tabController/PageCarousel.tsx
```diff
@@ -23,16 +23,16 @@ class PageCarousel extends PureComponent {
...
   }
 
-  onTabChange = ([index]) => {
-    this.scrollToPage(index, true);
+  onTabChange = ([index]: readonly number[]) => {
+    this.scrollToPage(index);
   };
 
-  scrollToPage = (pageIndex, animated) => {
+  scrollToPage = (pageIndex: number) => {
```
ethanshar, line 35: It seems like you changes the functionality here. 
Why did you remove the `animated` parameter?

src/components/tabController/TabBar.tsx
```diff
@@ -241,20 +284,21 @@ class TabBar extends PureComponent {
...
         if (Constants.isRTL && Constants.isAndroid) {
           const scrollingWidth = Math.max(0, this.contentWidth - this.containerWidth);
           targetOffset = scrollingWidth - targetOffset;
         }
 
-        this.tabBar.current.scrollTo({x: targetOffset, animated});
+        if (this.tabBar && this.tabBar.current) {
+          this.tabBar.current.scrollTo({x: targetOffset, animated});
+        }
       }
```
ethanshar, line 297: Consider using optional chaining here (:

src/components/tabController/TabBarItem.tsx
```diff
@@ -0,0 +1,325 @@
+// TODO: support commented props
```
ethanshar, line 1: Do you think it's possible to fix this file push so it won't be consider as a new file and will replace the original JS file.
If it's too complicated then never mind.. But it'll be nice not to lost git history for this file.

src/components/tabController/TabPage.tsx
```diff
@@ -68,7 +72,7 @@ export default class TabPage extends PureComponent {
     const {targetPage} = this.context;
     const {index, lazy} = this.props;
     return block([
-      cond(and(eq(targetPage, index), lazy, eq(this._loaded, 0)), [set(this._loaded, 1), call([], this.lazyLoad)]),
+      cond(and(eq(targetPage, index), Number(lazy), eq(this._loaded, 0)), [set(this._loaded, 1), call([], this.lazyLoad)]),
```
ethanshar, line 75: Did you find a bug here? Does the number conversion here actually fix something?

src/components/tabController/index.tsx
```diff
@@ -204,4 +237,13 @@ TabController.TabBar = TabBar;
...
 TabController.PageCarousel = PageCarousel;
-export default asBaseComponent(TabController);
+export default asBaseComponent<
+  TabControllerProps,
+  {
+    TabBar: typeof TabBar;
+    TabBarItem: typeof TabBarItem;
+    TabPage: typeof TabPage;
+    PageCarousel: typeof PageCarousel;
+  }
```
ethanshar, line 247: What about doing `typeof TabController` like we usually do. 
Wouldn't that work?

ethanshar (commented):
Overall look good. 
Wrote some comments. 

Anyway, i Noticed some performance issues with the component, but I think it has nothing to do with your PR. 
Recently, we update reanimated version to v13.1 and I noticed some weird issues in TabController. 
Do you experience them as well?

src/index.ts
M-i-k-e-l: I think it might get confusing since all the sub components there are (for TabController), and I'm not sure people won't need their props as well.
Maybe we can add them statically? Either on the component or create another "MegaProp" that has them all?
So either TabController.Props, TabController.TabBar.Props... or TabControllerProps.Main, TabControllerProps.TabBar...
Not sure how easy it is to do or understand, WDYT?

src/components/tabController/PageCarousel.tsx
M-i-k-e-l: It was not used, `false` was always sent (and I had an error about it).

M-i-k-e-l:
> Overall look good.
> Wrote some comments.
> 
> Anyway, i Noticed some performance issues with the component, but I think it has nothing to do with your PR.
> Recently, we update reanimated version to v13.1 and I noticed some weird issues in TabController.
> Do you experience them as well?

Yes, I think I know what you mean, it also sometimes does not go to the correct page. I was thinking that it might be solved with the new logic 🤞

src/components/tabController/TabBar.tsx
M-i-k-e-l: Fixed :)

src/components/tabController/TabBarItem.tsx
M-i-k-e-l: Looking in VSCode I have the history, where are you looking?

src/components/tabController/TabPage.tsx
M-i-k-e-l: I think it was for typescript, but you can see it's the same on line 53.

src/components/tabController/index.tsx
M-i-k-e-l: Yes, my mistake, fixed

ethanshar:
> Yes, I think I know what you mean, it also sometimes does not go to the correct page. I was thinking that it might be solved with the new logic 🤞
> Edit: looking at it again, not sure the new logic will fix it since it happens not only in centered mode.

Ok.. so I just verified.
On react-native-reanimated@1.9.0 (the previous version we had), everything works great
After updating to react-native-reanimated@1.13.1 (which was required for something else) things started to go south.. 

So I suggest that after we merge this PR, We'll investigate the issue first before going forward.
Usually when there are issues because of version upgrade, it's probably something small we need to tweak. 🤞

M-i-k-e-l:
> > Yes, I think I know what you mean, it also sometimes does not go to the correct page. I was thinking that it might be solved with the new logic 🤞
> > Edit: looking at it again, not sure the new logic will fix it since it happens not only in centered mode.
> 
> Ok.. so I just verified.
> On react-native-reanimated@1.9.0 (the previous version we had), everything works great
> After updating to react-native-reanimated@1.13.1 (which was required for something else) things started to go south..
> 
> So I suggest that after we merge this PR, We'll investigate the issue first before going forward.
> Usually when there are issues because of version upgrade, it's probably something small we need to tweak. 🤞

ok

src/components/tabController/TabPage.tsx
ethanshar: 👍

src/components/tabController/TabBarItem.tsx
ethanshar: In github? 
This file is considered new, so it's a little difficult to spot what has changed.. 
It's not big deal, but if you can easily fix it, that'll be great..

src/components/tabController/PageCarousel.tsx
ethanshar: ok.

src/index.ts
ethanshar: I'm not sure you can create this type of nesting in typings... 
If it's possible then let's do it.. 

I guess that's the reason other libraries' typings have weird long names..

src/components/tabController/TabBarItem.tsx
M-i-k-e-l: Weird, I did the same thing I did with other files (`git mv <old_file> <new_file>`)...
And I do see the history in VSCode, I think we'll 🤞 for it to be ok?

src/components/tabController/index.tsx
```diff
@@ -204,4 +237,8 @@ TabController.TabBar = TabBar;
 TabController.TabBarItem = TabBarItem;
 TabController.TabPage = TabPage;
 TabController.PageCarousel = PageCarousel;
-export default asBaseComponent(TabController);
+export default asBaseComponent<
+  TabControllerProps,
+  typeof TabController
+  //@ts-ignore typescript - will be fixed when moved to functional component
+>(TabController);
```
ethanshar, line 244: no need for asBaseComponent

M-i-k-e-l: It seems it was `asBaseComponent` to begin with, I see we're using its `config` in private for `carouselPageWidth`.

ethanshar: you are completely right, I accidentally looked at the old TabController file that sits under Incubator 🤦‍♂️

src/index.ts
M-i-k-e-l: I went with `TabControllerProps` namespace, let me know if that's ok

src/components/tabController/index.tsx
M-i-k-e-l: Yes, tell me about it, I spent some time trying to figure about why my changes are not applied in VSCode (why are my files still in javascript)... :)

src/index.ts
ethanshar: The namespace is nice, but now users will need to do  this?

`TabControllerProps.TabControllerProps`? it's a little weird don't you think?

src/components/tabController/index.tsx
ethanshar: yes, we plan to remove these files in V6. 
My bad, I should have warn you about it..

M-i-k-e-l: No worries :)

src/index.ts
M-i-k-e-l: :s

Any suggestions? `TabControllerProps.Props` or maybe create the `namespace` as `TabController` and everything will be inside it (but the it'll be `TabController.TabController` and worse `TabController.TabController.TabBar`).

ethanshar: TBH, Im starting to think we should drop the namespace thing, it's a little messy and too nested. 
Im not sure people are used to import typings this way.
I prefer to have 
TabControllerProps
TabControllerBarProps
TabControllerItemProps
TabControllerPageProps, etc...

It's not the best, but feels simpler.. WDYT?

M-i-k-e-l: I agree, changed as you recommended

ethanshar (approved):

PR #1081, opened 2020-12-16
Counter Icon badge

lidord-wix:
## Description
Add prop `withCounterIcon` to support rendering badge with icon and label 

## Changelog
Support rendering badge with icon and label


src/components/badge/index.tsx
```diff
@@ -222,13 +236,27 @@ class Badge extends PureComponent<BadgeProps> {
...
           ...iconStyle
         }}
       />
     );
   }
 
+  renderContent() {
+    const {withCounterIcon, icon, label} = this.props;
+
+    if (withCounterIcon && icon && label) {
```
ethanshar, line 249: I think it's best to allow render both icon and label always (without an extra prop). 
It will be the user responsibility to pass either an icon, label or both.

TBH, not sure why we blocked it to begin with

src/components/badge/index.tsx
```diff
@@ -143,6 +147,15 @@ class Badge extends PureComponent<BadgeProps> {
       height: badgeHeight,
       minWidth: badgeHeight
     };
+    if (withCounterIcon) {
```
ethanshar, line 150: I think it's more accurate to check if the user has passed both icon and label together, no?

src/components/badge/index.tsx
lidord-wix: No problem, I tried to preserve the original logic 👍🏽

src/components/badge/index.tsx
lidord-wix: Yup

ethanshar (approved):

PR #1083, opened 2020-12-17
Feat/new wheel picker integration

mendyEdri:
## Description
Integrate the new `WheelPicker` into our Picker when passing `useNativePicker` prop.

WIP:
- <s>Unwanted extra space in the `WheelPicker` when it's under the `safeArea`.</s>
- <s>FlatList's `initialScrollIndex` doesn't work, hacked with timer for now (could be a permanent fix)</s>.
- <s>WheelPicker is not fully controlled; when passing a hard-coded data it still change when scrolling the WheelPicker, 
example: `<WheelPicker selectedValue={‘Java’} />` when scrolling, it's not scrolling back to the 'Java' row.</s>
- <s>Implement and check behaviour on Android.</s>
- <s>Extract finding index to presenter/other type.</s>
- <s>add unit test for WheelPicker presenter.</s>

## Changelog
Integrate the new customizable `WheelPicker` into our Picker when passing `useNativePicker` prop.


src/components/picker/PickerDialog.ios.js
```diff
@@ -56,7 +56,7 @@ class PickerDialog extends BaseComponent {
     // TODO: should be taken from dialogProps but there's an issue with "babel-plugin-typescript-to-proptypes" plugin
     const {panDirection} = this.props;
     return (
-      <Dialog {...dialogProps} height={250} width="100%" migrate bottom animationConfig={{duration: 300}} panDirection={panDirection}>
+      <Dialog {...dialogProps} width="100%" migrate bottom animationConfig={{duration: 300}} panDirection={panDirection}>
```
ethanshar, line 59: @mendyEdri 
Regarding 
> Unwanted extra space in the WheelPicker when it's under the safeArea.

Consider doing this 

```
      <Dialog {...dialogProps} width="100%" migrate bottom animationConfig={{duration: 300}} panDirection={panDirection}>
        <View flex bg-white>
          {this.renderHeader()}
          <View centerV flex>
            {this.renderPicker()}
          </View>
          <View useSafeArea/>
        </View>
      </Dialog>
```

I added `<View useSafeArea/>` under the wrapper View of the picker.

ethanshar:
> * Unwanted extra space in the `WheelPicker` when it's under the `safeArea`.

See my comment in the code

> * FlatList's `initialScrollIndex` doesn't work, hacked with timer for now (could be a permanent fix).

Instead of hacking with setTimeout. Try scrolling to initial index on FlatList's `onLayout` prop. 

> * WheelPicker is not fully controlled; when passing a hard-coded data it still change when scrolling the WheelPicker,
>   example: `<WheelPicker selectedValue={‘Java’} />` when scrolling, it's not scrolling back to the 'Java' row.

I don't think WheelPicker can ever be fully controlled because of the way users interact with it. 
We can't have the wheel's value to wait for prop change, the interaction is too fast - so it must be uncontrolled in that way. 
Having said that, the dialog picker can be controlled. cause the final value is determined only when pressing 'done'.

mendyEdri:
> the dialog picker can be
@ethanshar 
Agree, user interaction can constantly change the `value`.
though, I tried to mimic the current behaviour of WheelPicker, that was re-scrolling to `value` if it was controlled, so is it ok that the behaviour will differ?

ethanshar:
> > the dialog picker can be
> 
> Agree, user interaction can constantly change the `value`.
> though, I tried to mimic the current behaviour of WheelPicker, that was re-scrolling to `value` if it was controlled, so is it ok that the behaviour will differ?
> 2nd: when w'll have a few pickers, wi'll need a way to control the value.
> (ex: min to max, max can't be lower then min)

I see what you mean.. 
I believe you want to treat the value prop as an initialValue and if the user change it in runtime it will reset to that value, right?

mendyEdri:
> and if the user change it in runtime it will reset to that value,

right. we probably going to need it

src/components/picker/NativePicker.js
```diff
@@ -59,9 +64,11 @@ class NativePicker extends BaseComponent {
 
   renderPickerDialog = () => {
     const {selectedValue, showDialog} = this.state;
-
+    
     return (
       <PickerDialog
+        height={this.PICKER_HEIGHT + this.MENU_HEIGHT}
+        pickerStyle={{width: '100%'}}
```
ethanshar, line 71: Not sure this is required.. 
usually Views are stretched horizontally (unless you change their flexDirection to be row)
So the width actually determined by the Dialog width

src/incubator/WheelPicker/index.tsx
```diff
@@ -2,16 +2,17 @@
...
+import {TextStyle, ViewStyle, FlatList, NativeSyntheticEvent, NativeScrollEvent} from 'react-native';
 import Animated from 'react-native-reanimated';
 import {onScrollEvent, useValues} from 'react-native-redash';
 import {Colors} from '../../../src/style';
 import View from '../../components/view';
 import Fader, {FaderPosition} from '../../components/fader';
 import {Constants} from '../../helpers';
-import useMiddleIndex from './helpers/useListMiddleIndex';
 import Item, {ItemProps} from './Item';
+import {Spacings} from 'react-native-ui-lib';
```
ethanshar, line 14: you "can't" import from `react-native-ui-lib` in source files. 
You can get `Spacings` from this import 
`import {Colors} from '../../../src/style';`

src/incubator/WheelPicker/index.tsx
```diff
@@ -39,47 +46,104 @@ export interface WheelPickerProps {
   /**
    * Event, on active row change
    */
-  onChange: (index: number, item?: ItemProps) => void;
```
ethanshar, line 42: Don't you prefer to keep it synced with the Picker's `onChange` API?

src/incubator/WheelPicker/index.tsx
```diff
@@ -39,47 +46,104 @@ export interface WheelPickerProps {
...
+}: WheelPickerProps) => {
   const scrollView = useRef<Animated.ScrollView>();
   const [offset] = useValues([0], []);
+  let scrollOffset = useRef<number>(0).current;
   const onScroll = onScrollEvent({y: offset});
-  
-  const listSize = items?.length || 0;
-  const middleIndex = useMiddleIndex({itemHeight, listSize});
+
+  const {height, items, shouldControlComponent, index: currentIndex, getRowItemAtOffset} = usePresenter({
```
ethanshar, line 81: Is `currentIndex` actually represents the initial index?

src/incubator/WheelPicker/index.tsx
```diff
@@ -39,47 +46,104 @@ export interface WheelPickerProps {
...
+          style={textStyle}
+          {...item}
+          onSelect={selectItem}
+        />
+      );
+    },
+    [itemHeight]
+  );
 
   const fader = useMemo(
```
ethanshar, line 148: Is there any benefit from using `useMemo` here (for `fader`)? 
Why not just render `<Fader....` twice at the bottom

src/incubator/WheelPicker/usePresenter.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const middleIndex = useMiddleIndex({itemHeight, listSize: items.length});
+
+  const getSelectedValueIndex = (): number => {
+    if (_.isString(selectedValue) || _.isNumber(selectedValue)) {
+      return _.findIndex(items, {value: selectedValue});
+    }
+    return _.findIndex(items, {value: selectedValue?.value});
+  }
+
+  const shouldControlComponent = (offset: number): boolean => {
```
ethanshar, line 50: at any point when swiping the wheel picker and `selectedValue` stays the same this will return true

src/incubator/WheelPicker/index.tsx
```diff
@@ -39,47 +46,104 @@ export interface WheelPickerProps {
...
+    controlComponent();
+  });
+
+  /**
+   * The picker is a controlled component. This means we expect the
+   * to relay on `selectedValue` prop to be our
+   * source of truth - not the picker current value.
+   * This way, you can control disallow or mutate selection of some values.
+   */
+  const controlComponent = () => {
```
ethanshar, line 99: Verify that initial value works.. right now it always takes the first one.

src/incubator/WheelPicker/index.tsx
```diff
@@ -39,47 +46,104 @@ export interface WheelPickerProps {
...
+  children?: JSX.Element | JSX.Element[];
+  /**
+   * WheelPicker initial value, can be ItemProps.value, number as index
+   */
+  selectedValue: ItemProps | number | string;
 }
 
-const WheelPicker = ({items, itemHeight = 48, activeTextColor, inactiveTextColor, textStyle, onChange: onChangeEvent}: WheelPickerProps) => {
-  const height = itemHeight * 5;
+const WheelPicker = ({
```
ethanshar, line 64: use React.Memo to have the component pure

src/incubator/WheelPicker/index.tsx
mendyEdri: Right, auto-import

src/incubator/WheelPicker/index.tsx
mendyEdri: Do you mean to `onChange: PropTypes.func` without specified values / index?

src/incubator/WheelPicker/index.tsx
mendyEdri: Not the initial, the current index (`onChange` updates the presenter with the current offset, and the presenter calculates it based on the `selectedValue` )

src/incubator/WheelPicker/index.tsx
mendyEdri: So it works, it based on the value given as `items` or `children`, I fixed the demo so it be more clear it's not an index..
`_.map(months, value => ({label: value, value: value}));`

src/incubator/WheelPicker/index.tsx
mendyEdri: I think Android could slow down UI thread sometime, while render fade / gradient views (also in native), so I thought to memo it even though it's not logically necessary.
but I removed from `separators` since it's just a simple view.

WDYT?

mendyEdri:
# Please DO NOT merge this until engine's reanimated release
Depend on engine PR [https://github.com/wix-private/wix-one-app-engine/pull/2123](url)

ethanshar:
@mendyEdri 
The reason this is pending is because we wanted to check it on a real one app version, right? 
Did we tested on one app after the onChange fixes?

ethanshar (approved):
Approving, don't merge yet. 
Let's figure how we want to rollout this feature carefully.

mendyEdri:
> > @mendyEdri
> > The reason this is pending is because we wanted to check it on a real one app version, right?
> > Did we tested on one app after the onChange fixes?
> 
> @ethanshar, exactly. i've added in the jira-ticket the relevant one-app version with this fix :)


> Approving, don't merge yet.
> Let's figure how we want to rollout this feature carefully.

@ethanshar, We have UX approval for including it without haptic (yet). 
Can I merge?

PR #1085, opened 2020-12-17
fix padding/margin modifiers: boolean -> number

doronpr:
## Description
fix padding/margin modifiers: boolean -> number
e.g. <Button padding="5"/> should be legal (currently <Button padding={true}/> is required)
e.g. <Button marginR="5"/> should be legal (currently <Button marginR={true}/> is required)

## Changelog
fix padding/margin modifiers: boolean -> number


ethanshar:
@doronpr 
Not sure I understand what this fix exactly?

doronpr:
> @doronpr
> Not sure I understand what this fix exactly?

it fixes this: https://jira.wixpress.com/browse/WOAUILIB-1395

ethanshar:
@doronpr 
This fix covers for React Native issue with their typings, and this is something I don't won't to include in our code. 
It'll be better to PR react-native code.

`marginRight` prop (or `marginTop`, `..Left`, etc..) comes from a react-native's View and will throw you a TS error also.

doronpr:
> @doronpr
> This fix covers for React Native issue with their typings, and this is something I don't won't to include in our code.
> It'll be better to PR react-native code.
> 
> `marginRight` prop (or `marginTop`, `..Left`, etc..) comes from a react-native's View and will throw you a TS error also.

@ethanshar why do you say the source of the problem is in RN? Can you link some info?
Regardless, this fixes an *existing* typing declaration in ui-lib that is incorrect. If you don't want to fix it it should be removed.
It's better to have a no fix to an external problem then an incorrect fix.

ethanshar:
> @ethanshar why do you say the source of the problem is in RN? Can you link some info?
> Regardless, this fixes an _existing_ typing declaration in ui-lib that is incorrect. If you don't want to fix it it should be removed IMHO.
> It's better to have no fix to an external problem than an incorrect fix.

you can just import View from react-native and see for yourself. 
```
import {View} from 'react-native';



render() {
  return <View marginRight={100}>
.....
  </View>
}

```
You will see the prop's working also for RN View (it's not a uilib feature). 
And you will get the TS error here as well.. RN doesn't declare these typings apparently.

doronpr:
Closing but plz take A look at the outstanding issue: https://jira.wixpress.com/browse/WOAUILIB-1395

PR #1086, opened 2020-12-17
icon prop - fix type to be both number and object.

Inbal-Tish:
## Description
icon prop - fix type to be both number and object.

## Changelog
icon prop - fix type to be both number and object.


src/components/tabController/TabBarItem.js
```diff
@@ -47,7 +47,7 @@ export default class TabBarItem extends PureComponent {
     /**
      * icon of the tab
      */
-    icon: PropTypes.number,
+    icon: PropTypes.oneOfType([PropTypes.object, PropTypes.number]),
     /**
```
ethanshar, line 51: Miki's currently working on migrating TabController to TS. 
So let's remove this one so it won't creates him conflicts..

ethanshar (changes requested):

src/components/tabController/TabBarItem.js
Effanuel: When could we expect this PR's changes to be merged?

ethanshar: I believe next version (in two weeks or so..)

ethanshar (approved):

PR #1089, opened 2020-12-18
Fix/tab controller transition

ethanshar:
## Description
Fix animation issues in TabController, there are two commits here
**Notice this PR branched from your PR (with all the TS changes, so there won't be any conflicts) - so merging it, will update your PR.** 

The First:
Fix the issue with the first time you press a tab bar item - there's no response

The second:
Fix a bigger issue. 
Apparently they fixed something in reanimated@13.1 that expose a bug on our end. 
It seems the `<Code>....</Code>` blocks kept re-rendering while they should have been monetized. 
Re-redner the `Code` block on state change cause the animation declarations to reset - which caused a huge problem. 

## Changelog
Fix TabController animation issues with reanimated@13.1

### DIFF
--- a/src/components/tabController/PageCarousel.tsx
+++ b/src/components/tabController/PageCarousel.tsx
@@ -40,13 +40,20 @@ class PageCarousel extends PureComponent {
     }
   };
 
-  renderCodeBlock = () => {
+  renderCodeBlock = _.memoize(() => {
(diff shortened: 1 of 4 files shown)


M-i-k-e-l:
Cool, how did you find this? Release notes?

PR #1092, opened 2020-12-20
Feat/radio button customize styling

ckuijjer:
## Description
The `style` property on the _RadioButton_ component was applied to both the `<Container />` component wrapping the radio button and label, and to the radio button itself. This made customizing the styling of the component less possible.

I simply added `style` to the destructuring in the `render` function, removing it from `...others` and not being applied to the `<Container />`. An option would be to also allow styling the container, by adding a `containerStyle` prop, just like in the code for the CheckBox.

Two commits, one solely for the prettier format and the 2nd containing the change

## Changelog
* The `style` prop can be used to apply customized styling to the radio button.

### DIFF
--- a/src/components/radioButton/RadioButton.tsx
+++ b/src/components/radioButton/RadioButton.tsx
@@ -77,6 +77,10 @@ export type RadioButtonProps = RadioGroupContextProps &
      * Should the content be rendered right to the button
      */
     contentOnRight?: boolean;
+    /**
+     * Additional styling for the container
+     */
+    containerStyle?: StyleProp<ViewStyle>;
   };
 export type RadioButtonPropTypes = RadioButtonProps; //TODO: remove after ComponentPropTypes deprecation;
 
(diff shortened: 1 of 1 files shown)


ethanshar:
Hi @ckuijjer 
Thanks for submitting a PR. 
Regarding your change, I suggest adding a dedicated `containerStyle` prop for those who still want to style the whole component container.

src/components/radioButton/RadioButton.tsx
```diff
@@ -244,19 +245,12 @@ class RadioButton extends PureComponent<Props, RadioButtonState> {
...
       // @ts-ignore
-      <Container
-        row
-        centerV
-        activeOpacity={1}
-        {...others}
-        onPress={this.onPress}
-        {...this.getAccessibilityProps()}
-      >
+      <Container row centerV activeOpacity={1} {...others} onPress={this.onPress} {...this.getAccessibilityProps()}>
```
ethanshar, line 253: Pass a new prop (`containerStyle`) to the Container's `style` 
It will fix your issue and still allow styling the button container view.

ckuijjer: Thanks for reviewing! It's indeed a good idea to allow styling of both elements. I've added a `containerStyle` prop. Would this be good?

ethanshar (approved):

PR #1093, opened 2020-12-21
Fix/dialog swiping does not trigger on dialog dismissed

M-i-k-e-l:
## Description
Dialog - fix swiping to dismiss does not trigger onDialogDismissed

## Changelog
Dialog - fix swiping to dismiss does not trigger onDialogDismissed

### DIFF
--- a/src/components/dialog/index.tsx
+++ b/src/components/dialog/index.tsx
@@ -87,6 +87,7 @@ interface DialogState {
   orientationKey: orientations;
   modalVisibility?: boolean;
   dialogVisibility?: boolean;
+  fadeOut?: boolean;
 }
 
 const DEFAULT_OVERLAY_BACKGROUND_COLOR = Colors.rgba(Colors.dark10, 0.6);
@@ -171,7 +172,7 @@ class Dialog extends Component<DialogProps, DialogState> {
     }
   }
 
-  onDismiss = () => {
+  _onDismiss = () => {
     this.setState({modalVisibility: false}, () => {
       const props = this.props;
       if (props.visible) {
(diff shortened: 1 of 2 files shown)


src/components/dialog/index.tsx
```diff
@@ -182,6 +183,19 @@ class Dialog extends Component<DialogProps, DialogState> {
         _.invoke(props, 'onDialogDismissed', props);
       }
     });
+  }
+
+  onDismiss = () => {
+    let fadeOut = false;
+    if (Constants.isIOS && this.props.visible) {
+      fadeOut = true;
+    }
```
ethanshar, line 192: Try that.. 
```suggestion
    const fadeOut = Constants.isIOS && this.props.visible;
```

src/components/dialog/index.tsx
```diff
@@ -171,8 +172,8 @@ class Dialog extends Component<DialogProps, DialogState> {
     }
   }
 
-  onDismiss = () => {
-    this.setState({modalVisibility: false}, () => {
+  _onDismiss = () => {
+    this.setState({modalVisibility: false, fadeOut: false}, () => {
```
ethanshar, line 176: I didn't completely understand the `fadeOut` fix.. 
Also, It seem the dialog works as expected when dismissing it using swipe (on master) 
Maybe we should sync about it..

M-i-k-e-l: Did you test it in public or private?

ethanshar: hmm wait.. 
Wasn't the issue that after dismissing using swipe the Dialog doesn't show the next time?

M-i-k-e-l: ActionSheet (private)

ethanshar: So shouldn't the fix be part of ActionSheet code?

M-i-k-e-l: No, it's the same as the previous PR to the Dialog for RN63, it fixes the `onDialogDismissed` API

src/components/dialog/index.tsx
M-i-k-e-l: 10x :)

ethanshar (approved):

PR #1094, opened 2020-12-21
prop value shape deprecation lint rule

lidord-wix:
## Description
I tried to make the rule readable as possible,
I also adjusted the schema, not sure it's the right way because when I set `additionalProps: false` the test is failing.

## Changelog
Set new lint rule for deprecation of prop value shape


eslint-rules/lib/rules/prop-value-shape-deprecation.js
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    function checkAttributeProperties(attributeProperties, attributeName, deprecation, nodeName, node, context) {
+      for (let i = 0; i <= attributeProperties.length; i++) {
+        const propertyName = _.get(attributeProperties[i], 'key.name');
+        const origin = propertyName && _.find(deprecation.shape, ['prop', propertyName]);
+        if (origin && origin.prop && propertyName === origin.prop) {
+          const destination = _.get(origin, 'fix.propName');
+          const message = `The shape of '${attributeName}' prop of '${nodeName}' doesn't contain '${origin.prop}' anymore. Please use '${destination}' instead (fix is available).`;
+          context.report({
```
M-i-k-e-l, line 124: I'm not sure if generally it's expected the report to be first (in a separate function), but it seems that's the convention we have in our rules.
Please look at our other rules and apply the same structure so it'll be easier to compare.

eslint-rules/lib/rules/prop-value-shape-deprecation.js
```diff
@@ -0,0 +1,141 @@
...
+      recommended: true
+    },
+    messages: {
+      uiLib: "This prop's value shape is deprecated."
+    },
+    fixable: 'code',
+    schema: [MAP_SCHEMA]
+  },
+  create(context) {
+    function testJSXAttributes(node) {
```
M-i-k-e-l, line 56: This function is (way) too long, please divide into smaller functions.
I did not look at all the logic, I'll look closer after you'll refactor it.

eslint-rules/lib/rules/prop-value-shape-deprecation.js
```diff
@@ -0,0 +1,141 @@
...
+      try {
+        const {deprecations} = _.get(context, 'options[0]');
+        const nodeName = _.get(node, 'name.name');
+        _.forEach(deprecations, deprecation => {
```
M-i-k-e-l, line 60: In other rules we check if we have imports that are relevant to this rule.
I'm not sure that our previous way is better since it doesn't seem to take any time to do this on a large example (yes, I've tried on a somewhat large screen :) ).
Well done!

eslint-rules/tests/lib/rules/prop-value-shape-deprecation.js
```diff
@@ -0,0 +1,116 @@
+const RuleTester = require('eslint').RuleTester;
```
M-i-k-e-l, line 1: Some general comments on the tests:
1. Add more, including more valid ones (I don't mind adding as well if you want me to).
2. It was not obvious what's the difference between the code and the output, consider moving `imageSource` and `source` to a couple of `const`s.

M-i-k-e-l (commented):
A few comments, but generally nice work!
I think it's a hard place to get in to.

eslint-rules/tests/lib/rules/prop-value-shape-deprecation.js
lidord-wix: 1. I've added 2 more valid examples, feel free to suggest more :) 
2. Moved it to `const`s.

eslint-rules/lib/rules/prop-value-shape-deprecation.js
M-i-k-e-l: I see now that you don't make sure it's the correct import in another way (I assumed you did).
Consider a case where someone is using a component from `react-native` and not `wix-react-native-ui-lib` and it has the same prop name (we have at least a few components with the same name), we will notify a deprecation on a source we don't control.
If you look at `component_deprecation` you'll see `"source": "module-with-deprecations",`, I'm not sure you'll need to add support for that as well, but it might be needed to send in the import name.

M-i-k-e-l: @lidord-wix, I'm working on fixing a bug in the `component-deprecation`.
I'm basically creating a new rule for component-prop-deprecation (later we will remove this part from the old rule and fix that rule as well).
I am creating new functions in the utils, based on the idea (you started 👍 ) to get all the data from the node (without creating the imports list etc).
You might want to wait for it to be ready before continuing this PR.

lidord-wix: No problem!
About the import - I consulted with Ethan when I worked on this ticket and we decided not to check the import in this rule, we can consider it again if you think it is important.

M-i-k-e-l (commented):
Good job!
I've added a few tests, only one of which failed (it's commented out; I suggest that you'll fix it once we'll merge my utils change, we might consider adding a source then, since it'll be easier and perhaps necessary).

M-i-k-e-l:
Oh, almost forgot...
You'll need to change the version in the `package.json`, but please do it after [this PR](https://github.com/wix/react-native-ui-lib/pull/1137) is merged so there are no conflicts (I'm guessing it'll be merged by the time you look at it since you are not working tomorrow as far as I know :) ).

lidord-wix:
I fixed the issue with the last test case.
As part of the solution, I added a `getComponentName` function to the utils.

eslint-rules/lib/rules/prop-value-shape-deprecation.js
```diff
@@ -51,16 +51,16 @@ module.exports = {
     schema: [MAP_SCHEMA]
   },
   create(context) {
-    function reportPropValueShapeDeprecation(attributePropertyKey, attributeName, origin, node) {
-      const nodeName = _.get(node, 'name.name');
+    function reportPropValueShapeDeprecation(propKey, attribute, origin, node) {
+      const componentName = utils.getComponentName(node);
       const destination = _.get(origin, 'fix.propName');
-      const message = `The shape of '${attributeName}' prop of '${nodeName}' doesn't contain '${origin.prop}' anymore. Please use '${destination}' instead (fix is available).`;
+      const message = `The shape of '${attribute}' prop of '${componentName}' doesn't contain '${origin.prop}' anymore. Please use '${destination}' instead (fix is available).`;
```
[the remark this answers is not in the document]
lidord-wix: 1. Done
2. What do you mean? isn't the fix required here?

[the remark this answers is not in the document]
M-i-k-e-l: Is there no scenario were the prop is deprecated (not replaced)?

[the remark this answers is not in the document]
lidord-wix: Ok, I added a message when there's no fix and one more test (for deprecation without fix)
Let me know if this is what you meant for

eslint-rules/lib/rules/prop-value-shape-deprecation.js
```diff
@@ -54,7 +54,10 @@ module.exports = {
     function reportPropValueShapeDeprecation(propKey, prop, deprecation, node) {
       const componentName = utils.getComponentName(node);
       const newProp = _.get(deprecation, 'fix.propName');
-      const message = `The shape of '${prop}' prop of '${componentName}' doesn't contain '${deprecation.prop}' anymore. Please use '${newProp}' instead (fix is available).`;
```
M-i-k-e-l: OK, sorry, I was wrong about adding a fix message (not about the possibility of no fix), but the message should (partially) coming from the `deprecation.message` (see the components' rule for an example).

M-i-k-e-l: Nice! Thanks and sorry again.

M-i-k-e-l (approved):

PR #1098, opened 2020-12-27
Typescript/tab controller move to function component

M-i-k-e-l:
## Description
1. Move `TabBar` to function component.
2. Move logic for fading to `FadedScrollView` (should we make this more generic, it has some RTL logic).
3. Fix `Fader`'s RTL logic - please take a look at the `FaderScreen`, it has some logic, I'm not sure if that's ok.

## Changelog
Move TabBar to function component & add support for Fader in RTL


src/components/tabController/TabBar.tsx
```diff
@@ -128,364 +128,289 @@ interface Props extends TabControllerBarProps, BaseComponentInjectedProps, Forwa
...
+  const onScroll = useCallback(({nativeEvent: {contentOffset}}: NativeSyntheticEvent<NativeScrollEvent>) => {
+    tabBarScrollOffset.current = contentOffset.x;
+    if (Constants.isRTL && Constants.isAndroid) {
+      const scrollingWidth = Math.max(0, contentWidth.current - containerWidth);
+      tabBarScrollOffset.current = scrollingWidth - tabBarScrollOffset.current;
+    }
+  },
+  [containerWidth]);
+
+  const getSnapBreakpoints = useMemo(() => {
```
ethanshar, line 227:
```suggestion
  const snapBreakpoints = useMemo(() => {
```

src/components/tabController/TabBar.tsx
```diff
@@ -128,364 +128,289 @@ interface Props extends TabControllerBarProps, BaseComponentInjectedProps, Forwa
...
-  }
+  const renderSelectedIndicator =
+    itemsWidths && itemsWidths.length > 0 ? (
+      <Reanimated.View style={[styles.selectedIndicator, indicatorStyle, _indicatorTransitionStyle]}/>
+    ) : undefined;
 
-  renderCodeBlock = _.memoize(() => {
-    const {currentPage, targetPage} = this.context;
-    const {itemsWidths, itemsOffsets} = this.state;
+  const renderCodeBlock = _.memoize(() => {
```
ethanshar, line 360: _.memoize is a lodash memo util.. 
Not sure how it works in a react function component.
Maybe we would use useMemo here?

src/components/tabController/TabBar.tsx
```diff
@@ -128,364 +128,289 @@ interface Props extends TabControllerBarProps, BaseComponentInjectedProps, Forwa
...
-          scrollEventThrottle={16}
-          testID={testID}
-          snapToOffsets={this.getSnapBreakpoints()}
-          decelerationRate={'fast'}
+  const shadowStyle = enableShadow ? propsShadowStyle || styles.containerShadow : undefined;
+  return (
+    <View style={[styles.container, shadowStyle, {width: containerWidth}, containerStyle]}>
+      <FadedScrollView
+        /**
+         // @ts-ignore TODO: typescript */
```
ethanshar, line 381:
```suggestion
         // @ts-ignore TODO: typescript
```

src/components/tabController/TabBar.tsx
```diff
@@ -128,364 +128,289 @@ interface Props extends TabControllerBarProps, BaseComponentInjectedProps, Forwa
...
+    if (propsChildren) {
+      LogService.warn('uilib: Please pass the "items" prop to TabController.TabBar instead of children');
+    }
+
+    if ((propsItems || children.current) && !contextItems) {
+      _registerTabItems();
+    }
+  }, []);
+
+  const [itemsWidths, setItemsWidths] = useState<number[]>([]);
```
ethanshar, line 200: consider moving some parts of the code to a separate custom hook (it does't have to be every part)
This will cleanup some of TabBar's code and create better separation. 
For instance, maybe we can move the whole logic that calculates itemWidth and itemOffset to a separate hook.

src/components/tabController/FadedScrollView.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const FadedScrollView = (props: Props) => {
+  const {scrollEnablerProps, scrollReachedProps, children, onScroll: propsOnScroll, ...other} = props;
+  const showLeft =
+    scrollEnablerProps.scrollEnabled &&
+    (Constants.isRTL ? !scrollReachedProps.isScrollAtEnd : !scrollReachedProps.isScrollAtStart);
+  const leftPosition = Constants.isRTL ? Fader.position.RIGHT : Fader.position.LEFT;
+  const showRight =
+    scrollEnablerProps.scrollEnabled &&
+    (Constants.isRTL ? !scrollReachedProps.isScrollAtStart : !scrollReachedProps.isScrollAtEnd);
```
ethanshar, line 28: if Fader will support RTL out of the box, does that mean we don't to check RTL here?

src/components/tabController/FadedScrollView.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <Fader visible={showRight} position={rightPosition} size={FADER_SIZE} supportRTL/>
+      </>
+    );
+  }
+
+  return null;
+};
+
+// TODO: fix withScrollEnabler props (add <>)
+export default withScrollReached<FadedScrollViewProps>(withScrollEnabler(forwardRef<Props>(FadedScrollView)), {
```
ethanshar, line 64: This is a little out of scope. 
But what's nice about hooks (as long you work with function components) is that they are great alternative for HOC. 
For example in this case, you can create hooks for useScrollReached and useScrollEnabler that do calculation and return the relevant info which you can pass on to your ScrollView. 
Maybe it's a good opportunity to create an equivalent hooks for these HOCs

ethanshar (commented):
Looks Great! 
Anyway, I wish we can maybe clean up some more the code in TabBar. 
(Not saying it was super clean before..)

But see in one of my comments, I suggested using custom hooks for different parts of the code, no need to over do it, but it can make the TabBar's code more readable. 
What do you think?

src/components/tabController/TabBar.tsx
M-i-k-e-l: This is done in the second PR ✅

src/components/tabController/FadedScrollView.tsx
M-i-k-e-l: Sounds great, however I'd rather do it in the second PR, so I won't have too many conflicts there (please add this comment there, so I don't forget 📝 ).

src/components/fader/index.tsx
```diff
@@ -80,6 +80,7 @@ function Fader(props: FaderProps) {
     <View pointerEvents={'none'} style={styles.containerStyle}>
       {(props.visible || _.isUndefined(props.visible)) && (
         <Image
+          supportRTL={props.supportRTL}
           source={styles.imageSource}
```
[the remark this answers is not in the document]
M-i-k-e-l: I guess so 🙇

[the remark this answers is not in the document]
M-i-k-e-l: True :)

demo/src/screens/componentScreens/FaderScreen.tsx
```diff
@@ -34,7 +35,7 @@ class FaderScreen extends Component<WithScrollReachedProps> {
     const {scrollReachedProps} = this.props;
     const visible =
       faderPosition === Fader.position.BOTTOM ||
-      faderPosition === Fader.position.RIGHT
+      (faderPosition === Fader.position.RIGHT && !Constants.isRTL)
```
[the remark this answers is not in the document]
M-i-k-e-l: Turns out: names matter - deprecated LEFT\RIGHT into START\END - made the code much simpler

src/components/tabController/FadedScrollView.tsx
M-i-k-e-l: Turns out: names matter - deprecated LEFT\RIGHT into START\END - made the code much simpler.
Android is still bugged, will be handled in the relevant ticket.

src/components/tabController/TabBar.tsx
M-i-k-e-l: Weird, pretty sure it gave me an error previously...

src/components/tabController/TabBar.tsx
M-i-k-e-l: Done

src/components/tabController/TabBar.tsx
M-i-k-e-l: I removed it and did not see unnecessary calls or performance problem, WDYT?

ethanshar: ok, cool.. 
Let's keep it like that.

src/components/tabController/FadedScrollView.tsx
ethanshar: coolio 👍

ethanshar (approved):

PR #1103, opened 2020-12-29
Feat/change tab controller center selected logic

M-i-k-e-l:
## Depends on
https://github.com/wix/react-native-ui-lib/pull/1098

## Description
Change TabController center selected logic (and move all the focusItems logic to a helper class)

## Changelog
Change TabController center selected logic (and move all the focusItems logic to a helper class)


ethanshar:
@M-i-k-e-l 
I think there are conflicts because of the changes we did in #1098

M-i-k-e-l:
Reminder:
>This is a little out of scope.
But what's nice about hooks (as long you work with function components) is that they are great alternative for HOC.
For example in this case, you can create hooks for useScrollReached and useScrollEnabler that do calculation and return the relevant info which you can pass on to your ScrollView.
Maybe it's a good opportunity to create an equivalent hooks for these HOCs

M-i-k-e-l:
> Reminder:
> 
> > This is a little out of scope.
> > But what's nice about hooks (as long you work with function components) is that they are great alternative for HOC.
> > For example in this case, you can create hooks for useScrollReached and useScrollEnabler that do calculation and return the relevant info which you can pass on to your ScrollView.
> > Maybe it's a good opportunity to create an equivalent hooks for these HOCs

Done

M-i-k-e-l:
> @M-i-k-e-l
> I think there are conflicts because of the changes we did in #1098

Fixed and added the new hooks

demo/src/screens/componentScreens/FaderScreen.tsx
```diff
@@ -10,6 +11,7 @@ const itemWidth = 100;
 const itemHeight = 100;
 const tintColor = undefined;
 
+// @ts-ignore
```
ethanshar, line 14: did you read the TS error here? 
it's actually right (: the condition will have return false...

src/commons/index.js
```diff
@@ -25,9 +25,15 @@ module.exports = {
...
   },
+  get useScrollEnabler() {
+    return require('./useScrollEnabler').default;
+  },
   get withScrollReached() {
     return require('./withScrollReached').default;
   },
+  get useScrollReached() {
+    return require('./useScrollReached').default;
+  },
```
ethanshar, line 36: we recently added a new Hooks folder, I suggest moving the hooks there..

src/helpers/useFocusItemHelper.ts
```diff
@@ -0,0 +1,171 @@
+import _ from 'lodash';
```
ethanshar, line 1: Few comments on this.. 
- Is this hook really a common one or mainly for TabController purpose? 
I assume you want to use it in FilterChips as well, in that case, consider moving it to Hooks folder. 
- The name of the hook needs a little work in my opinion `..Helper` doesn't really say much, clearly it's a hook that helps doing something (: Let's try to find a better name.

src/helpers/useFocusItemHelper.ts
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Add a margin to the offset (default to true)
+   * This gives a better UX
+   * Not relevant to OffsetType.CENTER
+   */
+  addOffsetMargin?: boolean;
+  /**
+   * How much space (padding \ margin) is there on he left\right of the items
+   */
+  sideSpacing?: number;
```
ethanshar, line 39: I liked innerSpacing, maybe this one can be outerSpacing? 
```suggestion
  outerSpacing?: number;
```

src/helpers/useFocusItemHelper.ts
```diff
@@ -0,0 +1,171 @@
...
+   */
+  offsetType?: OffsetType;
+  /**
+   * Add a margin to the offset (default to true)
+   * This gives a better UX
+   * Not relevant to OffsetType.CENTER
+   */
+  addOffsetMargin?: boolean;
+  /**
+   * How much space (padding \ margin) is there on he left\right of the items
```
ethanshar, line 37: typo
```suggestion
   * How much space (padding \ margin) is there on the left\right of the items
```

src/helpers/useFocusItemHelper.ts
```diff
@@ -0,0 +1,171 @@
...
+    itemsCount,
+    selectedIndex,
+    offsetType = OffsetType.CENTER,
+    addOffsetMargin = true,
+    sideSpacing = 0,
+    innerSpacing = 0
+  } = props;
+  const itemsWidths = useRef<(number | null)[]>(_.times(itemsCount, () => null));
+  const currentIndex = useRef<number>(selectedIndex || 0);
+  const [offsets, setOffsets] = useState<Offsets>({CENTER: [], LEFT: [], RIGHT: []});
```
ethanshar, line 80: Do we need to keep the offsets in state?
They don't **really** affect the render..

src/helpers/useFocusItemHelper.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    if (addOffsetMargin) {
+      index = 1;
+      while (index < itemsCount - 1) {
+        leftOffsets[index] -= itemsWidths[index - 1];
+        rightOffsets[index] += itemsWidths[index + 1];
+        ++index;
+      }
+    }
```
ethanshar, line 118: Can you explain a little what this does?
Is it for fixing the focusing so the items won't be too near to the edges?

src/helpers/useFocusItemHelper.ts
```diff
@@ -0,0 +1,171 @@
...
+  const onItemLayout = useCallback((event: LayoutChangeEvent, index: number) => {
+    const {width} = event.nativeEvent.layout;
+    itemsWidths.current[index] = width;
+    if (!_.includes(itemsWidths.current, null)) {
+      setSnapBreakpoints(itemsWidths.current as number[]);
+    }
+  },
+  [setSnapBreakpoints]);
+
+  const scroll = (scrollTo: number, animated: boolean) => {
```
ethanshar, line 133: Is this in order to support ScrollView, FlatList and SectionList? 
I think I have a similar code somewhere.. (: 

Maybe we should add a specific hook for that (:
Anyway, not for this scope - just FYI

src/index.js
```diff
@@ -223,9 +223,15 @@ export default {
...
+  get useScrollEnabler() {
+    return require('./commons').useScrollEnabler;
+  },
   get withScrollReached() {
     return require('./commons').withScrollReached;
   },
+  get useScrollReached() {
+    return require('./commons').useScrollReached;
+  },
   get Modifiers() {
```
ethanshar, line 235: should be exported as part of `Hooks`..

src/components/tabController/TabBar.tsx
```diff
@@ -357,44 +309,54 @@ const TabBar = (props: Props) => {
...
-          style={[
-            styles.tabBar,
-            !_.isUndefined(height) && {height},
-            {paddingHorizontal: centerOffset, backgroundColor}
-          ]}
-        >
-          {renderTabBarItems}
-        </View>
+        <View style={indicatorContainerStyle}>{renderTabBarItems}</View>
         {renderSelectedIndicator}
```
ethanshar, line 360: this should be `selectedIndicator` since it's an element and not a function

ethanshar (commented):
Looks great. Good job! 
I wrote a few comments..

demo/src/screens/componentScreens/FaderScreen.tsx
M-i-k-e-l: I know, it depends on the "screen configuration" above (numberOfItems ... tintColor), we wanted to keep the screen simple (no buttons etc) so I've put the configuration in code only.
But I did want the screen to adjust properly to the configuration change, so we get an error because of this "constant configuration".

src/commons/index.js
M-i-k-e-l: Done

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: I'll move it to the hooks

useFocusItem
useScrollToItem

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: Done

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: Thanks

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: Actually it does (verified it now), `focusIndex` needs to change when the `offsets` have been calculated.

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: Yes, it is for fixing the margin, it gives a margin of the previous' item's width.

src/index.js
M-i-k-e-l: Done

src/components/tabController/TabBar.tsx
M-i-k-e-l: Done

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: No time like the present.
I'm just not sure it's a hook, maybe a helper function, since it doesn't need `useState`, `useRef` etc, WDYT?

src/helpers/useFocusItemHelper.ts
ethanshar: I think useScrollIToItem hints better what the hook does..

src/helpers/useFocusItemHelper.ts
ethanshar: Technically you can have this hook to create the scrollView ref and provide (and also use it internally for the scrollTo)
something like that.. 
```
const [scrollRef, scrollTo] = useScrollTo()
```
and maybe if someone already has the ref we can support passing it 
```
const myOwnRef = useRef();
const [scrollRef, scrollTo] = useScrollTo(myOwnRef)
```

WDYT?

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: Done

src/helpers/useFocusItemHelper.ts
M-i-k-e-l: Done

ethanshar (approved):

PR #1106, opened 2020-12-30
bug fixes for TabBar refactor

Inbal-Tish:
## Description
Fix refactor bugs.

## Changelog
TabBar - fix refactor bugs

### DIFF
--- a/src/components/scrollBar/index.tsx
+++ b/src/components/scrollBar/index.tsx
@@ -25,7 +25,7 @@ export interface ScrollBarProps extends FlatListProps<any> {
     /**
      * The element to use as a container, instead of a View
      */
-    containerView?: object,
+    containerView?: React.ComponentClass,
     /**
      * The props to pass the container
      */
--- a/src/components/scrollBar/index.tsx
+++ b/src/components/scrollBar/index.tsx
@@ -25,7 +25,7 @@ export interface ScrollBarProps extends FlatListProps<any> {
     /**
      * The element to use as a container, instead of a View
      */
-    containerView?: object,
+    containerView?: JSX.ElementClass,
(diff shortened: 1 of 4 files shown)


src/components/scrollBar/index.tsx
```diff
@@ -25,7 +25,7 @@ export interface ScrollBarProps extends FlatListProps<any> {
     /**
      * The element to use as a container, instead of a View
      */
-    containerView?: JSX.Element,
+    containerView?: object,
```
ethanshar, line 28: Is that right? 
Shouldn't `containerView` be an element?

ethanshar (commented):
I noticed you don't push changes in generateTypes. 
Make sure to always run `npm run build:dev`. It verifies there are no TS errors and also create the updated generatedTypes

src/components/scrollBar/index.tsx
Inbal-Tish: No. Since we're not passing <TabletView/>, we're passing TabletView.

Inbal-Tish:
> I noticed you don't push changes in generateTypes.
> Make sure to always run `npm run build:dev`. It verifies there are no TS errors and also create the updated generatedTypes

Done

src/components/scrollBar/index.tsx
ethanshar: Still it's not a object. 
I think there's a dedicate type for it.. 

Try searching for the right prop.

Inbal-Tish: I couldn't find it, the 'object' type is the only one that doesn't error...

ethanshar (approved):

PR #1111, opened 2020-12-31
Feat/incubator text field preset

ethanshar:
## Description
Support `preset` prop in the new TextField component and add the first "default" preset. 
This should keep parity with the old TextField UI but still allows to keep the UI code separated 

## Changelog
Support "default" preset to our Incubator.TextField component


src/incubator/TextField/index.tsx
```diff
@@ -83,9 +84,13 @@ type TextFieldProps = MarginModifiers &
      * Container style of the whole component
      */
     containerStyle?: ViewStyle;
+    /**
+     * Predefined preset to use for styling the field
+     */
+    preset?: 'default' | undefined;
```
mendyEdri, line 90: Is the type could be optional boolean? 
(`useDefaultPreset` for example)

mendyEdri (commented):
small comment regarding a type :)

src/incubator/TextField/index.tsx
ethanshar: Actually the idea was to add more presets in the future, so the `preset` prop was meant to be used as an enum. 
For now I just added the `default` preset so it feels redundant..

mendyEdri: Yeah that's what I thought it meant to be, merging

mendyEdri (approved):

PR #1112, opened 2020-12-31
Infra/useDidUpdate

mendyEdri:
## Description
Added new hook tool - useDidUpdate to avoid getting side effect calls on initial value.

## Changelog
Introduce new useDidUpdate hook, that will skip update on initial value.


src/hooks/useDidUpdate/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * This hook avoid calling useEffect on the initial value of his dependency array
+ */
+const useDidUpdate = (callback: () => void, dep: [any]) => {
+  const _dep = useRef<[any] | undefined>(dep);
+
+  const isDataChanged = useCallback(() => {
+    return !_.isEqual(_dep.current, dep);
+  }, [dep]);
```
ethanshar, line 12: Not sure there's a need for useCallback here, you don't pass the function reference anywhere..

src/hooks/useDidUpdate/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  useEffect(() => {
```
ethanshar, line 14: Maybe im missing something.. 
shouldn't there be a code that prevents from this effect to trigger on initial mount?

src/hooks/useDidUpdate/index.ts
mendyEdri: Right, I should change it to useMemo to avoid re do the `_. isEqual` work

src/hooks/useDidUpdate/index.ts
mendyEdri: Since the inner dependancies set to be as the passed prop dependancies, it will actually start as the same value..

ethanshar (approved):

PR #1113, opened 2020-12-31
Create a middleware TextField migrator

ethanshar:
## Description
Create TextFieldMigrator to migrate users from the old TextField implementation to the new. 

## Changelog
Start migrating to our new TextField implementation by passing the `migrate` prop.


Inbal-Tish:
@ethanshar Don't you want to set up the lint here as well so users can enjoy the auto fix?

Inbal-Tish (approved):

ethanshar:
@Inbal-Tish 
But we configure lint in the private lib..

Inbal-Tish:
> @Inbal-Tish
> But we configure lint in the private lib..

I know. I thought we can configure it here as well.

ethanshar:
No one is using it in the community..

PR #1114, opened 2021-01-03
Fix/card section removing image source deprecation

Inbal-Tish:
## Description
Fix/CardSection removing imageSource deprecation

## Changelog
CardSection - removing 'imageSource' deprecation

### DIFF
--- a/src/components/card/CardSection.tsx
+++ b/src/components/card/CardSection.tsx
@@ -1,7 +1,6 @@
 import _ from 'lodash';
 import React, {PureComponent} from 'react';
 import {StyleSheet, ViewStyle, ImageStyle, ImageSourcePropType, StyleProp} from 'react-native';
-import {LogService} from '../../services';
 import {asBaseComponent} from '../../commons/new';
 import View, {ViewProps} from '../view';
 import Text, {TextProps} from '../text';
@@ -36,7 +35,6 @@ export type CardSectionProps = ViewProps & {
    * Will be used for the background when provided
    */
   imageSource?: ImageSourcePropType;
-  source?: ImageSourcePropType;
   /**
    * The style for the background image
    */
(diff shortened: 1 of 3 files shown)


src/components/card/CardSection.tsx
```diff
@@ -56,14 +54,6 @@ type Props = CardSectionProps & asCardChildProps;
...
   static displayName = 'Card.Section';
 
-  constructor(props: Props) {
-    super(props);
-
-    if (props.imageSource) {
-      LogService.deprecationWarn({component: 'CardSection', oldProp: 'imageSource', newProp: 'source'});
-    }
-  }
-
```
ethanshar, line 66: Unfortunately, I think now we need to have the opposite warning.. 
It's been for a few months, no?

ethanshar (commented):
@Inbal-Tish 
See my comment. 
Are we sure we can just remove the `source` prop? 
If we've been instructing people to use it for some time, we might need to deprecate it with a warning...

Inbal-Tish:
@ethanshar I think we should leave it as 'source', same as in StateScreen. This is the Image prop's name and the description explains that this is the source for the image. I don't think we need to have our own prop name for it....

ethanshar:
> @ethanshar I think we should leave it as 'source', same as in StateScreen. This is the Image prop's name and the description explains that this is the source for the image. I don't think we need to have our own prop name for it....

I disagree. 
When a component is image orientated then I think it make sense to keep the prop name as `source` (Avatar, Thumbnail..)
Where the component is not necessarily based on an image, I think it's important to explicitly call the prop `imageSource`. 

This is what we agreed on When Miki created this component, I don't want to change it because we accidentally migrated the prop.

Inbal-Tish:
We did accidentally migrated it. In this case, StateScreen should be un-migrated as well, no?

PR #1116, opened 2021-01-04
Adding 'defaultSource' for image source in case of an error

Inbal-Tish:
## Description
Image - Adding 'defaultSource' prop for image source in case of an error.

## Changelog
Image - Adding 'defaultSource' prop for image source in case of an error.


src/components/image/index.tsx
```diff
@@ -51,16 +60,25 @@ export type ImageProps = RNImageProps & MarginModifiers & {
    * Render an overlay with custom content
    */
   customOverlayContent?: JSX.Element;
+  /**
+   * Default image source in case of an error
+   */
+  defaultSource?: ImageSourcePropType
```
ethanshar, line 66: notice that `defaultSource` is already exist (react-native's Image has this prop) 
And it's purpose is set a default source as placeholder for images. 

Our new prop's purpose is only for broken sources, It needs to have a more explicit name. (defaultFailingSource, or something like that..)

ethanshar (changes requested):

src/components/image/index.tsx
Inbal-Tish: Done

ethanshar (approved):

PR #1123, opened 2021-01-07
Remove Blur (and fix CardsScreen)

M-i-k-e-l:
## Description
Remove Blur (and fix CardsScreen)

## Changelog
Remove Blur (and fix CardsScreen)

### DIFF
--- a/demo/src/screens/componentScreens/CardsScreen.tsx
+++ b/demo/src/screens/componentScreens/CardsScreen.tsx
@@ -99,7 +99,6 @@ export default class CardsScreen extends Component<CardsScreenProps, CardsScreen
         onPress={() => {}}
         borderRadius={borderRadius}
         useNative
-        backgroundColor={Colors.white}
         activeOpacity={1}
         activeScale={isImageOnLeft ? 0.96 : 1.04}
       >
@@ -124,7 +123,6 @@ export default class CardsScreen extends Component<CardsScreenProps, CardsScreen
     return (
       <Card style={{marginBottom: 10}} onPress={() => {}}>
         <Card.Section
-          bg-white
           content={[
             {text: 'You’re Invited!', text70: true, grey10: true},
             {text: 'join now', text90: true, grey50: true}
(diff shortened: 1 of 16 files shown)


demo/src/screens/componentScreens/CardsScreen.tsx
```diff
@@ -125,6 +124,7 @@ export default class CardsScreen extends Component<CardsScreenProps, CardsScreen
     return (
       <Card style={{marginBottom: 10}} onPress={() => {}}>
         <Card.Section
+          bg-white
```
Inbal-Tish, line 127: Don't we want to have Card and Card.Section bg color white by default?

src/components/modal/index.tsx
```diff
@@ -73,16 +64,14 @@ class Modal extends Component<ModalProps> {
...
   render() {
-    const {blurView, enableModalBlur, visible, ...others} = this.props;
-    const defaultContainer = enableModalBlur && Constants.isIOS ? BlurView : View;
-    const Container: any = blurView ? blurView : defaultContainer;
+    const {visible, ...others} = this.props;
 
     return (
       <RNModal visible={Boolean(visible)} {...others}>
-        <Container style={{flex: 1}} blurType="light">
+        <View style={{flex: 1}}>
```
Inbal-Tish, line 71: Do we really need this View? Is the touchableOverlay can't be the container view?

demo/src/screens/componentScreens/CardsScreen.tsx
M-i-k-e-l: Good question, I'm not sure.
@ethanshar, should we add `defaultProps` for `Card` and `CardSection` with `backgroundColor: Colors.white`?

src/components/modal/index.tsx
M-i-k-e-l: I think it can simply be removed, tested on the places I could think of and it looked ok.

demo/src/screens/componentScreens/CardsScreen.tsx
ethanshar: Yes, backgroundColor should be white by default

M-i-k-e-l: Fixed

Inbal-Tish (approved):

PR #1125, opened 2021-01-08
Stop using AnimatableManager in Hint

ethanshar:
## Description
- Remove usage of AniamtableManger from our Hint component
- Use Colors.primary instead of Colors.blue30 

## Changelog
* Replace usage of `react-native-animatable` from our Hint component with our own implementation

### DIFF
--- a/src/components/hint/index.js
+++ b/src/components/hint/index.js
@@ -2,9 +2,8 @@
 import _ from 'lodash';
 import PropTypes from 'prop-types';
 import React from 'react';
-import {StyleSheet, AccessibilityInfo, findNodeHandle} from 'react-native';
-import {View as AnimatableView} from 'react-native-animatable';
-import {Typography, Spacings, Colors, BorderRadiuses, AnimatableManager} from '../../style';
+import {Animated, StyleSheet, AccessibilityInfo, findNodeHandle} from 'react-native';
+import {Typography, Spacings, Colors, BorderRadiuses} from '../../style';
 import {Constants} from '../../helpers';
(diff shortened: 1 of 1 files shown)


src/components/hint/index.js
```diff
@@ -343,19 +353,18 @@ class Hint extends BaseComponent {
...
-    const shownUp = position === HINT_POSITIONS.TOP;
+    const {message, messageStyle, icon, iconStyle, borderRadius, color, testID} = this.getThemeProps();
 
     if (this.showHint) {
       return (
-        <AnimatableView
-          animation={shownUp ? AnimatableManager.animations.hintAppearUp : AnimatableManager.animations.hintAppearDown}
-          duration={200}
+        <View
+          animated
```
Inbal-Tish, line 361: how do you control the type of animation to apply on the View?

ethanshar: What do you mean?

Inbal-Tish: With Animatable you can load the animations, like 'hintApearUp'. How do you do it with the 'animated' prop of View. What animation does the Hint will have now?

Inbal-Tish: Got it. getHintAnimatedStyle

Inbal-Tish (approved):

src/components/hint/index.js
ethanshar: The default one. 
Users won't be able to configure it now. 
I don't thing people were aware of that and I think it's pretty redundant..
Of course if someone will request we can reconsider it

PR #1137, opened 2021-01-14
Fix Typography lint catching non truthy props

M-i-k-e-l:
## Description
Fix Typography lint catching non truthy props

## Changelog
Fix Typography lint catching non truthy props


eslint-rules/lib/rules/typography-deprecation.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
     }
 
     function testJSXAttribute(node) {
-      if (node && node.name) {
+      if (node && node.name && !node.value) {
```
ethanshar, line 66: I assume this check is because people use typography modifiers like this? 
```
<Text bodyBold />
```
And the prop doesn't have an actual value (it behaves as boolean true) so you assume it's a modifier? 

If that's right I think you might be missing something. 
What if someone renders this
```
const useBodyBold = <some condition>;
<Text bodyBold={useBodyBold} />
```
In this case the value will exist and can be true or false. 

Maybe the check should be more explicit and ignore boolean values

M-i-k-e-l: Yes, it was the `title` case (which can be an actual `title` in some components).
That's a good point, I'll add a test and fix accordingly.

M-i-k-e-l: From what I can tell there's only the `Identifier` type (no `boolean`, `string` etc), so we can't tell what the value is unless we start to dig in deeper into the tree, I don't think it's worth it at this stage, WDYT?

For reference I'll leave two example `node.value`:
title (string):
{"type":"JSXExpressionContainer","start":155,"end":162,"loc":{"start":{"line":5,"column":27},"end":{"line":5,"column":34}},"expression":{"type":"Identifier","start":156,"end":161,"loc":{"start":{"line":5,"column":28},"end":{"line":5,"column":33},"identifierName":"title"},"name":"title","range":{"0":156,"1":161},"_babelType":"Identifier"},"range":{"0":155,"1":162},"_babelType":"JSXExpressionContainer"}
isItDeprecated (true):
{"type":"JSXExpressionContainer","start":103,"end":119,"loc":{"start":{"line":3,"column":29},"end":{"line":3,"column":45}},"expression":{"type":"Identifier","start":104,"end":118,"loc":{"start":{"line":3,"column":30},"end":{"line":3,"column":44},"identifierName":"isItDeprecated"},"name":"isItDeprecated","range":{"0":104,"1":118},"_babelType":"Identifier"},"range":{"0":103,"1":119},"_babelType":"JSXExpressionContainer"}

ethanshar: I checked in https://astexplorer.net
And inside `JSXExpressionContainer` you can find an `expression` that holds the value. 
You can check `typeof value` to get the type.

M-i-k-e-l: Nice site, thanks :)
Not sure how I can get the value though (again, without traversal).
Do you have a way to do that without traversal?
If not, I suggest we open a task and do that after the refactor.

Code examples:
1.
import {Typography} from 'our-source';
import {List} from 'another-source';
export default class OurList extends Component {
  render() {
	const title = 'bla';
  	return (
	  <List.Item title={title}/>
	)
  }
}

2. Same but with:
const title = true;

ethanshar: You mean we need a way to go back in the tree and the find the origin value of the identifier (`title`) ?
We have in our utils the `findValueNodeOfIdentifier`. I think we used it before for the same purpose..

M-i-k-e-l: I see, I've tried a few cases and I think we cannot catch all of them (what if the value depends on `props`), and it will be more complicated than we want (what if it's a const that depends on a previous const etc).
We can go into a recursive solution (for the above case) which will also have many branches, examples, each one will required a different handling:
`const titleVal = true`                         (type = 'Literal', boolean)
`const titleVal = 'text'`                       (type = 'Literal', string)
`const titleVal = this.props.title`   (type = 'MemberExpression')
`const titleVal = !this.props.isTitle`  (type = 'UnaryExpression')
... 🤯 

Before I dig deeper (if you like me to), maybe a better solution is to tie the prop to specific components (i.e. `Text` etc), WDYT?
Or am I missing a simple solution?

ethanshar: How tying to specific components help? 
People can still pass this props in the ways you mention for `Text` component. 
Tying it to specific components can reduce the chance we'll report on a wrong use case.. so anyway this is something we probably wanna do. 

But.. I don't want us to go crazy. 
Let's focus on the common use cases for now that are possible for us to check.. 
It's ok if it doesn't cover all use cases.. we'll deal with them one at a time..

M-i-k-e-l: It'll help because generally components that have `Props` that extend `TextProps` or `TypographyModifiers` do not have the `title` prop.
So, if they have `title` as a prop, it'll be the typography, and we don't need to check for a value.
I think that if a component will have two props with the same name (i.e. a prop and a modifier that have the same name), it'll be a bigger problem than the lint rules, so it should not happen.

ethanshar: Ok sounds good..
But we can still have this, no? 

```
const useTitle = ....;
<Text title={useTitle}>My Text</Text>
```

M-i-k-e-l: Yes, but if the user has `title` it's already "bad", i.e. deprecated, no matter what the value is

ethanshar: ohhhh right right. 
ok sounds good. 
Let's do that - check on specific components

M-i-k-e-l: Done.
I've added some commented tests for stuff we don't yet support  (renamed components - should be easily supported after the refactor, props of components - might be supported with Lidor's PR).

eslint-rules/tests/lib/rules/typography-deprecation.js
```diff
@@ -439,6 +519,194 @@ ruleTester.run('typography-deprecation', rule, {
...
+            const titleVal = isTitle;
+            const subtitleVal = !isTitle;
+            return (
+              <TextField title={titleVal} subtitle={subtitleVal}/>
+            )
+          }
+        }`,
+        errors: [{message: `'Typography.title' is deprecated. Please use 'Typography.heading' instead (fix is available).`}]
+    },
+    // {
```
ethanshar, line 621: Why are we keeping these? 
These are the use cases we don't support, right? 
Maybe write a comment that explain these tests and that we can't/don't support them

ethanshar (commented):
Looks good!
just couple of comments..

eslint-rules/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-uilib",
-  "version": "1.0.27",
+  "version": "1.0.28",
```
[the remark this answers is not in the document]
M-i-k-e-l: I think so (does hoping really hard helps? 🤞 ).
Do you have any reservations?

eslint-rules/tests/lib/rules/typography-deprecation.js
M-i-k-e-l: Added comments (TODOs)

ethanshar (approved):

PR #1138, opened 2021-01-18
Fix/text field migrator/set display name

ethanshar:
## Description
Set proper displayName on our TextFieldMigrator

## Changelog
Set proper displayName on our TextFieldMigrator

### DIFF
--- a/src/components/textField/TextFieldMigrator.tsx
+++ b/src/components/textField/TextFieldMigrator.tsx
@@ -52,7 +52,7 @@ function migrateProps(props: any) {
   return fixedProps;
 }
 
-export default forwardRef(({migrate = false, ...props}: any, ref) => {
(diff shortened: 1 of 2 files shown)


src/components/textField/TextFieldMigrator.tsx
```diff
@@ -62,3 +62,6 @@ export default forwardRef(({migrate = false, ...props}: any, ref) => {
     return <OldTextField {...props} ref={ref}/>;
   }
 });
+
+TextFieldMigrator.displayName = 'TextField';
```
M-i-k-e-l, line 66: Could there be a problem when two components have the same display name (`TextField` still has its `displayName`)?

ethanshar: There might.. 
but it's temporary...

M-i-k-e-l (approved):

PR #1139, opened 2021-01-19
fix prop types for ChipsInput props

Dor256:
## Description
Fix typing of `ChipsInputProps` by extending `TextInputProps`

## Changelog
*Add a quick message for our users about this change*
Fix `ChipsInput` prop typings

### DIFF
--- a/typings/components/ChipsInput.d.ts
+++ b/typings/components/ChipsInput.d.ts
@@ -2,6 +2,7 @@ import {ReactElement} from 'react';
 import {StyleProp, TextStyle, ViewStyle, AccessibilityProps, ColorValue} from 'react-native';
 import {BaseComponent} from '../commons';
 import {AccessibilityProps} from '../modifiers';
+import {TextFieldProps} from './Inputs';
 
 export type Tag =
   | string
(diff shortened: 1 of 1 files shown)


typings/components/ChipsInput.d.ts
```diff
@@ -1,7 +1,9 @@
+import { TypographyModifiers } from 'generatedTypes';
```
M-i-k-e-l, line 1: Looks like you've left this be mistake

typings/components/ChipsInput.d.ts
```diff
@@ -1,7 +1,9 @@
...
 import {ReactElement} from 'react';
 import {StyleProp, TextStyle, ViewStyle, AccessibilityProps, ColorValue} from 'react-native';
 import {BaseComponent} from '../commons';
 import {AccessibilityProps} from '../modifiers';
+import { TextInputProps } from './Inputs';
```
M-i-k-e-l, line 6: Shouldn't this be `TextFieldProps`?

M-i-k-e-l (changes requested):

typings/components/ChipsInput.d.ts
Dor256: woops - fixing

typings/components/ChipsInput.d.ts
Dor256: Hmm, I don't really know as I've only seen this used in our module as `TextInput` you know the use cases better than me :)

Dor256: Fixed

M-i-k-e-l: > you know the use cases better than me :)
Not really, it's just using a `TextField` with `{...others}` :)

M-i-k-e-l (approved):

PR #1147, opened 2021-01-23
POC for dark mode support

ethanshar:
## Description
Add new Colors api for loading schemes for dark mode
for example: 
```
Colors.loadSchemes({light: {...}, dark: {...}})
```
Also added a listener in our asBaseComponent that for appearance change so our components can react to dark mode change

## Changelog
Support dark mode using `Colors.loadSchemes` method.


AbdeenM:
Hello All,

Was wondering if someone can review and approve, in need of using loadSchemes on our application.

Cheers!

src/commons/asBaseComponent.tsx
```diff
@@ -23,11 +24,22 @@ function asBaseComponent<PROPS, STATICS = {}>(WrappedComponent: React.ComponentT
     static displayName: string | undefined;
     static propTypes: any;
     static defaultProps: any;
+    appearanceListener: any;
```
M-i-k-e-l, line 27:
```suggestion
    appearanceListener: appearanceListener?: Appearance.AppearanceListener;
```
Once this is changed you'll see the need for the other two changes in this file.

src/commons/asBaseComponent.tsx
```diff
@@ -23,11 +24,22 @@ function asBaseComponent<PROPS, STATICS = {}>(WrappedComponent: React.ComponentT
...
 
     state = {
       error: false
     };
 
+    componentDidMount() {
+      this.appearanceListener = Appearance.addChangeListener(({colorScheme}: {colorScheme: ColorSchemeName}) => {
+        this.setState({colorScheme});
+      });
```
M-i-k-e-l, line 36:
```suggestion
      this.appearanceListener = ({colorScheme}: {colorScheme: ColorSchemeName}) => {
        this.setState({colorScheme});
      };

      Appearance.addChangeListener(this.appearanceListener);
```

src/commons/asBaseComponent.tsx
```diff
@@ -23,11 +24,22 @@ function asBaseComponent<PROPS, STATICS = {}>(WrappedComponent: React.ComponentT
...
+    }
+
+    componentWillUnmount() {
+      Appearance.removeChangeListener(this.appearanceListener);
```
M-i-k-e-l, line 40:
```suggestion
      if (this.appearanceListener) {
        Appearance.removeChangeListener(this.appearanceListener);
      }
```

src/commons/modifiers.ts
```diff
@@ -96,26 +96,28 @@ export type ContainerModifiers =
   BorderRadiusModifiers &
   BackgroundColorModifier;
 
-
 export function extractColorValue(props: Dictionary<any>) {
-  // const props = this.getThemeProps();
-  const allColorsKeys: Array<keyof typeof Colors> = _.keys(Colors);
+  const scheme = Appearance.getColorScheme();
+  const schemeColors = Colors.schemes[scheme!];
```
M-i-k-e-l, line 101: How do you know `scheme` in not `null` or `undefined` (here and below)?

demo/src/screens/foundationScreens/DarkModeScreen.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+class DarkModeScreen extends Component {
+  state = {};
+  render() {
+    return (
+      <View flex padding-page bg-screenBG>
+        <Text h1 textColor>
+          Dark Mode
+        </Text>
+        <Text marginT-s2 body textColor>
```
M-i-k-e-l, line 13: Consider showing this only on iOS [simulator](https://stackoverflow.com/a/35529545/4759619).

src/commons/asBaseComponent.tsx
M-i-k-e-l: After thinking about it, this is better:
```
appearanceListener = ({colorScheme}: {colorScheme: ColorSchemeName}) => {
  this.setState({colorScheme});
};
```
This will remove the need for the `if` in `componentWillUnmount`

src/commons/modifiers.ts
ethanshar: I don't but technically if it's null or undefined then schemeColors will be undefined and it'll ignore them. 

Im not sure why `Appearance.getColorScheme()` can return undefined or null, maybe the user can set it in the device preferences? 
In this case, I guess we should ignore the schemes, no?

demo/src/screens/foundationScreens/DarkModeScreen.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <View flex padding-page bg-screenBG>
+        <Text h1 textColor>
+          Dark Mode
+        </Text>
+        {Constants.isIOS ? (
+          <Text marginT-s2 body textColor>
+            Change to dark mode in simulator by pressing Cmd+Shift+A
+          </Text>
+        ) : (
+          <Text>Change to dark mode</Text>
```
M-i-k-e-l, line 18:
```suggestion
          <Tex marginT-s2 body textColort>Change to dark mode</Text>
```

src/commons/modifiers.ts
M-i-k-e-l: I'm not sure why but this is the definition in react-native:
```
type ColorSchemeName = 'light' | 'dark' | null | undefined;
```
Actually it's more weird because the [original code](https://github.com/facebook/react-native/blob/93377ff5089d981e56094329b6f6893b3124be12/Libraries/Utilities/NativeAppearance.js#L14) does not have them (does not seem to have changed in the last two years)...

BTW, you can now remove the `!` since it has no purpose now.

ethanshar: fixed

M-i-k-e-l (approved):

PR #1151, opened 2021-01-25
ChispInput backspase fix for Android

vidmantas0010:
## Description

There is an issue in react-native lib.  https://github.com/facebook/react-native/issues/29689 
onChangeText is called before onKeyPress on Android - state.input are empty when deleting last character. Debounce fixes execution order.

onKeyPress is called properly for backspace button. No need additional event listener.
`DeviceEventEmitter.addListener('onBackspacePress', this.onKeyPress);	` 
Found out this caused double onKeyPress events.

## Changelog

Fixed ChipsInput backspace tag removal issue on Android.


M-i-k-e-l (approved):
Very nice ❗ 
Maybe share this in the ticket you've mentioned?

PR #1155, opened 2021-01-26
Add accessibility props to avatar component

Dor256:
## Description
Add missing accessibility typing to `Avatar` component. 

## Changelog
Fix accessibility typings to `Avatar` component

### DIFF
--- a/src/components/avatar/index.tsx
+++ b/src/components/avatar/index.tsx
@@ -8,7 +8,8 @@ import {
   TouchableOpacity,
   ImagePropsBase,
   ImageStyle,
-  TextStyle
+  TextStyle,
+  AccessibilityProps
 } from 'react-native';
 import {Colors} from '../../style';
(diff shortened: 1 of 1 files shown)


src/components/avatar/index.tsx
```diff
@@ -136,7 +137,7 @@ export type AvatarProps = {
    * Used as a testing identifier
    */
   testID?: string;
-};
+} & AccessibilityProps;
```
M-i-k-e-l, line 140: 1. Do you need them all? Perhaps you can [Pick the way we do in Dialog](https://github.com/wix/react-native-ui-lib/blob/master/src/components/dialog/index.tsx#L22-L24)?
2. Either way, it should be at the start (`export type AvatarProps = AccessibilityProps & {`).

ethanshar:
@Dor256 
Add description and changelog :(

M-i-k-e-l (approved):

PR #1170, opened 2021-02-01
DialogDismissibleView - refactor to function

M-i-k-e-l:
## Description
DialogDismissibleView - refactor to function

## Changelog
DialogDismissibleView - refactor to function


src/components/dialog/DialogDismissibleView.tsx
```diff
@@ -54,216 +55,200 @@ interface LocationProps {
...
+const DialogDismissibleView = (props: Props) => {
+  const {direction = DEFAULT_DIRECTION, visible: propsVisible, containerStyle, style, children, onDismiss} = props;
+  // @ts-expect-error
+  const {isPanning, dragDeltas, swipeDirections} = useContext<PanContextProps>(PanningContext);
+
+  const width = useRef<number>(Constants.screenWidth);
+  const height = useRef<number>(Constants.screenHeight);
+  const thresholdX = useRef<number>(0);
+  const thresholdY = useRef<number>(0);
+  const counter = useRef<number>(0);
```
ethanshar, line 70: `counter` is a little vague, consider changing it to something that explains it better..

src/components/dialog/DialogDismissibleView.tsx
```diff
@@ -54,216 +55,200 @@ interface LocationProps {
...
+  const containerRef = useRef<typeof View>();
+  const animatedValue = useRef<Animated.AnimatedValue>(new Animated.Value(0));
+  const swipe = useRef<PanDirectionsProps>({});
```
ethanshar, line 73: Same here with `swipe`, consider renaming to something more readable

src/components/dialog/DialogDismissibleView.tsx
```diff
@@ -54,216 +55,200 @@ interface LocationProps {
...
+  const prevDragDeltas = useRef<PanAmountsProps>();
+  const prevSwipeDirections = useRef<PanDirectionsProps>();
+  const visible = useRef<boolean>(Boolean(propsVisible));
+
+  const getHiddenLocation = useCallback((left: number, top: number): LocationProps => {
```
ethanshar, line 78: Why are you using `useCallback` here? You're not passing it as a prop

src/components/dialog/DialogDismissibleView.tsx
```diff
@@ -54,216 +55,200 @@ interface LocationProps {
...
-  private hiddenLocation: LocationProps;
-  private animatedValue = new Animated.Value(0);
-  private width = Constants.screenWidth;
-  private height = Constants.screenHeight;
-  private counter = 0;
-  private swipe: PanDirectionsProps = {};
-  private thresholdX = 0;
-  private thresholdY = 0;
-  private ref = React.createRef<any>();
+  const animateTo = useCallback((toValue: number, animationEndCallback?: Animated.EndCallback) => {
```
ethanshar, line 102: No need for useCallback

src/components/dialog/DialogDismissibleView.tsx
```diff
@@ -54,216 +55,200 @@ interface LocationProps {
...
+      toValue,
+      duration: 300,
+      easing: Easing.bezier(0.2, 0, 0.35, 1),
+      useNativeDriver: true
+    }).start(animationEndCallback);
+  }, []);
 
-  constructor(props: Props) {
-    super(props);
+  const isSwiping = useCallback((): boolean => {
```
ethanshar, line 111: Also here.. useCallback seem redundant.
Even the function content only use reference values

ethanshar: Also, verify you really need to declare a return type. 
Typescript usually handles it automaticaly

ethanshar (commented):
@M-i-k-e-l 
I think there's is a misusage of the useCallback hook. 
I wrote a few comments, but overall I think it most cases you don't to use `useCallback` 

The main purpose for using it for a function is when you want to pass a callback as a prop for another component. 
And since here it's not the case, most function can just be a regular functions.

src/components/dialog/DialogDismissibleView.tsx
M-i-k-e-l: Changed to `dragsCounter`

src/components/dialog/DialogDismissibleView.tsx
M-i-k-e-l: Changed to mutableSwipeDirections

src/components/dialog/DialogDismissibleView.tsx
M-i-k-e-l: But it's being used in `onLayout` which is passed as a prop.

src/components/dialog/DialogDismissibleView.tsx
M-i-k-e-l: Used in `onLayout` as well.

src/components/dialog/DialogDismissibleView.tsx
M-i-k-e-l: `isSwiping` is used in `onDrag` which is used in `useEffect`.
Removed a couple of return values.

src/components/dialog/DialogDismissibleView.tsx
ethanshar: Then I think it's enough that only onLayout will be wrapped with useCallback, no? 
`getHiddenLocation` can be a pure function that has no dependencies

M-i-k-e-l: Are we leaving this as is (following our meeting today)?

ethanshar (approved):
Approved. 
There's is a conflict needed to resolve

PR #1172, opened 2021-02-03
Changed `useSafeArea` default value

mendyEdri:
## Description
CustomKeyboard had a fix some time ago, to remove the safe area (annoying grey part at the bottom).
In order to avoid migration back at the time, I've set `useSafeArea` default value to `true` although other UILib components set it to `false` by default. 
In this PR it fixed, and moved to be `false` by default

## Changelog
When using `KeyboardAccessoryView`, `useSafeArea` default value has changed from `true` to `false`, to match all other UILib's api that by default it is `false`.

### DIFF
--- a/lib/components/Keyboard/KeyboardTracking/KeyboardTrackingView.ios.js
+++ b/lib/components/Keyboard/KeyboardTracking/KeyboardTrackingView.ios.js
@@ -36,7 +36,6 @@ class KeyboardTrackingView extends PureComponent {
     useSafeArea: PropTypes.bool
   };
 
-  /** V6 should be change to default false */
   static defaultProps = {
     useSafeArea: true
   };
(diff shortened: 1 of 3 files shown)


lib/components/Keyboard/KeyboardTracking/KeyboardTrackingView.ios.js
```diff
@@ -36,7 +36,6 @@ class KeyboardTrackingView extends PureComponent {
     useSafeArea: PropTypes.bool
   };
 
-  /** V6 should be change to default false */
   static defaultProps = {
     useSafeArea: true
```
Inbal-Tish, line 40: Why here it's not 'false'?

mendyEdri: It's a different feature, that we reverted eventually.
(The row I deleted in the spreadsheet)

Inbal-Tish:
@mendyEdri Please check why the build is faling

lib/components/Keyboard/KeyboardTracking/KeyboardTrackingView.ios.js
Inbal-Tish: ok

Inbal-Tish (approved):

PR #1178, opened 2021-02-04
Fix/tab controller center android with rtl

M-i-k-e-l:
## Description
TabController - center tabs - Android with RTL - fix scroll to
Fix offset for Android with RTL (i.e. fix Fader in TabController)

Unsolved bugs (I don't think they are regressions):
1. Using `asCarousel` is reversed in Android RTL.
2. Fader in `ScrollBar` - I think this component should be re-written anyway (function with hooks + new `Fader` etc).

## Changelog
TabController - Android with RTL - fix scroll to and Fader


ethanshar:
- I don't think we have usages of asCarousel in the app
- I think ScrollBar should be deprecated in favor of the HOCs/Hooks you created for the fader effect

ethanshar (approved):

PR #1179, opened 2021-02-05
Safe require blur-view dependency in Card

ethanshar:
## Description
**Don't Merge, Just Review**
I have recently came across this proposal (which was released last weekend)
https://github.com/react-native-community/discussions-and-proposals/issues/120

This is an amazing feature we've been waiting for a long time. 
It allows us to conditionally require dependencies and use them safely. 
It looks something like this 
```
let a;
try {
  a = require('optional-module');
} catch (e) {
  // Handle failure from loading the module
}
```

For instance, In our **Card** component we have a dependency on **BlurView**. 
Today, if the user is missing this dependency the whole component will crash the app. 
With this new feature, we can safely require BlurView and handle it accordingly (see PR code to understand it better).

This will allow us to use dependencies conditionally and have the users decide what and when to use specific features. 
For instance, in the case of the whole haptic feedback effort. 
We can add the haptic support directly in public code and condition it by checking if the dependency exists. 

## Changelog
Card's enableBlur is won't crash the app if missing @react-native-community/blur dependency


M-i-k-e-l:
Sounds really nice, what's the next step? Should this be done in the V6 branch (to avoid conflicts)?

ethanshar:
> Sounds really nice, what's the next step? Should this be done in the V6 branch (to avoid conflicts)?

Technically yes, 
But I think we can also add this in v5, it's not a breaking change in any way. 
I didn't want to merge yet, to avoid conflicts.. 
Let's have @Inbal-Tish review and see what she things.

M-i-k-e-l:
This will need to be applied to the `Modal` as well

Inbal-Tish:
@ethanshar @M-i-k-e-l Looks great! Very long-awaited feature!

Inbal-Tish (approved):

ethanshar:
> This will need to be applied to the `Modal` as well

Done!

PR #1182, opened 2021-02-08
TabController - remove from incubator and remove support for children in TabBar

M-i-k-e-l:
## Description
TabController - remove from incubator and remove support for children in TabBar
+ some minor fixes.

## Changelog
TabController - remove from incubator and remove support for children in TabBar


generatedTypes/components/badge/index.d.ts
```diff
@@ -344,7 +344,7 @@ declare const _default: React.ComponentClass<ViewProps & TouchableOpacityProps &
     /**
      * the badge size (default, small)
      */
-    size: number;
+    size?: number | undefined;
```
Inbal-Tish, line 347: Why do you need to specify undefined type for an optional prop?

src/components/tabController/TabBar.tsx
```diff
@@ -144,47 +139,27 @@ const TabBar = (props: Props) => {
...
   useEffect(() => {
-    if (propsChildren) {
-      LogService.warn('uilib: Please pass the "items" prop to TabController.TabBar instead of children');
-    }
-
-    if ((propsItems || children.current) && !contextItems) {
+    if (propsItems && !contextItems) {
       _registerTabItems();
     }
   }, []);
```
Inbal-Tish, line 165: Why no dependencies here, like 'items'? Won't it cause unnecessary updates?

generatedTypes/components/badge/index.d.ts
M-i-k-e-l: I don't, this is the automatically generated `.d.ts` file (when running `npm run build:dev`).

src/components/tabController/TabBar.tsx
M-i-k-e-l: On the contrary, no dependencies means no re-renders (this is like `componentDidMount`).

Inbal-Tish (approved):

PR #1183, opened 2021-02-09
no-direct-import eslint rule - adding `applyAutofix` flag

Inbal-Tish:
## Description
no-direct-import eslint rule - adding `applyAutofix` flag to control autofix option

## Changelog
no-direct-import eslint rule - adding `applyAutofix` flag to control autofix option


ethanshar (commented):
@Inbal-Tish 
I suggest bumping the version so we can publish it after we merge

Inbal-Tish:
> @Inbal-Tish
> I suggest bumping the version so we can publish it after we merge

Done

eslint-rules/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-uilib",
-  "version": "1.0.28",
+  "version": "1.1.0",
```
ethanshar, line 3: Shono pacakge which brings us a depdency is locked using `~`
That means it gets only patches.. 
So if we want this new version to be taken we need to update the patch version and not minor. 

```suggestion
  "version": "1.0.29",
```

ethanshar (changes requested):

ethanshar (approved):

PR #1184, opened 2021-02-09
Infra - change Dialog to use reanimated 2

M-i-k-e-l:
## Description
Infra - change Dialog to use reanimated 2.
Note that this has `Infra/reanimated-2-part-1-refactor-dialog` set as base and not `master`.
Known issues:
Android - gestures do not work (only the animation works), this is probably because we need to add `gestureHandlerRootHOC` or `GestureHandlerRootView` somewhere (I did not find the correct way to do it).
~~Android & iOS (more visible on Android): the view is shown for a short time in the "visible location" before animating in from the "hidden location"~~ fixed with `persistentAnimationValue`, not sure if that's the best solution.
I'm not sure I've solved *all* the errors from migrating to v2.

## Changelog
Infra - change Dialog to use reanimated 2

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -4,7 +4,7 @@ import com.android.build.OutputFile
 
 project.ext.react = [
     entryFile: "index.js",
-    enableHermes: false,  // clean and rebuild if changing
+    enableHermes: true,  // clean and rebuild if changing
 ]
 
 apply from: "../../node_modules/react-native/react.gradle"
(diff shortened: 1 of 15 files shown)


ethanshar:
@M-i-k-e-l 
I split your PR into two 
#1223  - Only migrating the demo project intro reanimated v2, the is now enough for migrating since reanimated handle gracefully the migration. But it does throw warnings. 

#1224 - Migrate all of the component that use Reanimated to use the new API with a fallback to the old one. This way our users can gradually migrate when they are ready. And if someone install reanimated v2 they won't get the annoying warnings.

M-i-k-e-l:
Keeping this for the Dialog part

PR #1197, opened 2021-02-19
Updated Typescript definitions inside TouchableOpacity

vikasg603:
I compared the props for TouchableOpacity and also for View.

What I noticed was that View was exporting PropType, but TouchableOpacity was not.

And also, TouchableOpacity was using Type instead of Interface.

I am not sure it will resolve the issue properly or not but check it once.

### DIFF
--- a/src/components/touchableOpacity/index.tsx
+++ b/src/components/touchableOpacity/index.tsx
@@ -16,7 +16,7 @@ import {
 } from '../../commons/new';
 import IncubatorTouchableOpacity from '../../incubator/TouchableOpacity';
 
-export interface TouchableOpacityProps extends Omit<RNViewProps, 'style' | 'onPress'>, ContainerModifiers {
+export interface TouchableOpacityProps extends Omit<RNTouchableOpacityProps, 'style' | 'onPress'>, ContainerModifiers {
     /**
      * background color for TouchableOpacity
      */
--- a/src/components/touchableOpacity/index.tsx
+++ b/src/components/touchableOpacity/index.tsx
@@ -16,9 +16,7 @@ import {
 } from '../../commons/new';
 import IncubatorTouchableOpacity from '../../incubator/TouchableOpacity';
 
-
-export type TouchableOpacityProps = Omit<RNTouchableOpacityProps, 'style' | 'onPress'> &
-  ContainerModifiers & {
+export interface TouchableOpacityProps extends Omit<RNViewProps, 'style' | 'onPress'>, ContainerModifiers {
     /**
      * background color for TouchableOpacity
      */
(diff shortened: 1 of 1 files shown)


vikasg603 (commented):
Resolved Build time error.

src/components/touchableOpacity/index.tsx
```diff
@@ -44,8 +42,10 @@ export type TouchableOpacityProps = Omit<RNTouchableOpacityProps, 'style' | 'onP
      */
     customValue?: any;
     style?: StyleProp<ViewStyle> | Animated.AnimatedProps<StyleProp<ViewStyle>>;
-    onPress?: (props: TouchableOpacityProps) => void;
-  };
+    onPress?: () => void;
```
ethanshar, line 45: Why did u change the signature of onPress? 
Is that relevant for the fix? 
It just we have a custom signature for onPress because we pass the props back to the user

vikasg603: No, not required, but I am pretty much sure that TouchableOpacity doesn't take any arguments when used onPress.
And even in UILib, we are not passing anything as an argument in onPress.

vikasg603: But actually, React Native Touchable Opacity accepts 1 argument as GestureResponderEvent, but what you had passed was not GestureResponderEvent but rather complete GestureResponderEvent.

Let me update the pull request with GestureResponderEvent

ethanshar: But it's not suppose to be an event. 
If you'll look at the Button component implementation, you'l see we pass back the prop to the user. 
This is helpful with avoiding passing inline callbacks to `onPress`

vikasg603: Oh, my bad, I thought it might be an event.
Okay, let me update the code again.

vikasg603:
Updated the code again.

src/components/touchableOpacity/index.tsx
```diff
@@ -45,7 +43,9 @@ export type TouchableOpacityProps = Omit<RNTouchableOpacityProps, 'style' | 'onP
     customValue?: any;
     style?: StyleProp<ViewStyle> | Animated.AnimatedProps<StyleProp<ViewStyle>>;
     onPress?: (props: TouchableOpacityProps) => void;
-  };
+}
+
+export type TouchableOpacityPropTypes = TouchableOpacityProps; //TODO: remove after ComponentPropTypes deprecation;
```
ethanshar, line 48: @vikasg603 
One last comment, what is that for? (:

vikasg603: It was there in your View component.
I just copied the same code with some small changes :)

ethanshar: It's not required here (:

src/components/touchableOpacity/index.tsx
```diff
@@ -16,9 +16,7 @@ import {
 } from '../../commons/new';
 import IncubatorTouchableOpacity from '../../incubator/TouchableOpacity';
 
-
-export type TouchableOpacityProps = Omit<RNTouchableOpacityProps, 'style' | 'onPress'> &
-  ContainerModifiers & {
+export interface TouchableOpacityProps extends Omit<RNTouchableOpacityProps, 'style' | 'onPress'>, ContainerModifiers {
```
ethanshar, line 19: btw, exporting as interface fixes the issue?

ethanshar:
Hi @vikasg603 
Sorry for the long ping-pong. 
I verified ur fix and pushed a commit. 
I'm closing this, the fix will be available in the next version

PR #1209, opened 2021-03-01
DialogDismissibleView - fix bug with delayed insets

M-i-k-e-l:
## Description
DialogDismissibleView - fix bug with delayed insets (we have await in our private repo)
BTW, I think that code can probably be moved to public.

## Changelog
DialogDismissibleView - fix bug with delayed insets

### DIFF
--- a/src/components/dialog/DialogDismissibleView.tsx
+++ b/src/components/dialog/DialogDismissibleView.tsx
@@ -55,8 +55,6 @@ interface LocationProps {
 }
 
 const DEFAULT_DIRECTION = PanningProvider.Directions.DOWN;
-const TOP_INSET = Constants.isIphoneX ? Constants.getSafeAreaInsets().top : Constants.isIOS ? 20 : 0;
-const BOTTOM_INSET = Constants.isIphoneX ? Constants.getSafeAreaInsets().bottom : Constants.isIOS ? 20 : 0;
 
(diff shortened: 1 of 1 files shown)


src/components/dialog/DialogDismissibleView.tsx
```diff
@@ -65,6 +63,8 @@ const DialogDismissibleView = (props: Props) => {
 
   const width = useRef<number>(Constants.screenWidth);
   const height = useRef<number>(Constants.screenHeight);
+  const TOP_INSET = useRef<number>(Constants.isIphoneX ? Constants.getSafeAreaInsets().top : Constants.isIOS ? 20 : 0);
+  const BOTTOM_INSET = useRef<number>(Constants.isIphoneX ? Constants.getSafeAreaInsets().bottom : Constants.isIOS ? 20 : 0);
```
ethanshar, line 67: Looking at it now, it feels this kind of logic should be elsewhere, I guess in Constants. WDYT? 
I can approve and we can refactor later of course

ethanshar (approved):

src/components/dialog/DialogDismissibleView.tsx
M-i-k-e-l: I'm adding a ticket for this, it'll require some adjusting in both public and private, so I'd rather not mix it in this fix.

PR #1219, opened 2021-03-11
fix #1218 where carousel is not calculating children correctly

### LINKED ISSUE #1218: Carousel tries to render empty children
### Describe the bug

The carousel thinks that `null` or `[]` is a child, even though they're not.

### To Reproduce
```javascript
<Carousel pageControlPosition={Carousel.pageControlPositions.UNDER}>
  {null}
  <Text>Foobar</Text>
</Carousel>
```

causes an issue where it thinks the length of the children is two. The reason you might want to do this, for instance in my example, is to render a list of images and then add a final item allowing you to add more images. On mount, my array is actually null so you'd do something like `` and then render the 

```javascript
<Carousel pageControlPosition={Carousel.pageControlPositions.UNDER}>
  {fields?.map(({ image }) => <MyComponent image={image} />)
  <AddAnotherImage />
</Carousel>
```

### Expected behavior

Only the actual children should be rendered

### Screenshots

image

iOS 14
The latest release of this package

scripter-co:
## Description
A fix for #1218

## Changelog

Replaced `children.length` with Reacts children count helper.

### DIFF
--- a/src/components/carousel/CarouselPresenter.ts
+++ b/src/components/carousel/CarouselPresenter.ts
@@ -1,9 +1,11 @@
+import React, {ReactNode} from 'react';
 import _ from 'lodash';
 import {CarouselProps, CarouselState} from './types';
 
-export function getChildrenLength(props: CarouselProps): number {
-  const length = _.get(props, 'children.length') || 0;
-  return length;
+type PropsWithChildren = CarouselProps & { children?: ReactNode }
+
+export function getChildrenLength(props: PropsWithChildren): number {
+  return React.Children.count(props.children);
 }
 
 export function calcOffset(props: CarouselProps, state: Omit<CarouselState, 'initialOffset' | 'prevProps'>) {
--- a/src/components/carousel/__tests__/CarouselPresenter.spec.js
+++ b/src/components/carousel/__tests__/CarouselPresenter.spec.js
@@ -4,7 +4,6 @@ import * as uut from '../CarouselPresenter';
 describe('Carousel presenter', () => {
   it('should getChildrenLength', () => {
     expect(uut.getChildrenLength({children: [<></>, <></>, <></>]})).toBe(3);
-    expect(uut.getChildrenLength({children: [<></>, <></>, <></>]})).toBe(3);
     expect(uut.getChildrenLength({children: [<></>]})).toBe(1);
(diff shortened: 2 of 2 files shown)


src/components/carousel/__tests__/CarouselPresenter.spec.js
```diff
@@ -1,50 +1,53 @@
+import React from 'react';
 import * as uut from '../CarouselPresenter';
 
 describe('Carousel presenter', () => {
   it('should getChildrenLength', () => {
```
scripter-co, line 5: you could argue this test is pointless now and can be deleted. It's just testing that `React.Children.count` works - which doesn't need to be tested.

Inbal-Tish: Why are you changing the tests from objects to tags?

src/components/carousel/CarouselPresenter.ts
```diff
@@ -1,9 +1,11 @@
+import React, {ReactNode} from 'react';
 import _ from 'lodash';
 import {CarouselProps, CarouselState} from './types';
 
-export function getChildrenLength(props: CarouselProps): number {
-  const length = _.get(props, 'children.length') || 0;
-  return length;
+type PropsWithChildren = CarouselProps & { children?: ReactNode }
```
Inbal-Tish, line 5: 'children' are not already part of CarouselProps?

src/components/carousel/__tests__/CarouselPresenter.spec.js
scripter-co: because `{}` is not a valid component in react. If you try and call `React.children.count()` with `{}` you'll get `Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.`

src/components/carousel/CarouselPresenter.ts
scripter-co: [No](https://github.com/wix/react-native-ui-lib/blob/16afa97a1c6d511d398ee7c0b6a3195b9cca4d9b/src/components/carousel/types.tsx#L10) - It is fine in the component because the class uses a generic which looks like `readonly props: Readonly<P> & Readonly<{ children?: ReactNode }>;`

Inbal-Tish (approved):

PR #1225, opened 2021-03-14
Add extension to generate code snippets

jonathanmos:
## Description
Added extension that allows generation of code snippets for vscode. The extension code is in the snippetGenerator subproject

To compile and test the extension:
1) go to the snippetGenerator root
2) npm install
3) npm install -g vsce
4) vsce package
5) you get a file in the folder -> snippetgenerator-[version#frompackagejson].vsix that you can install in vscode with ‘extensions: install from vsix’ 

## Changelog
SnippetGenerator can be packaged with vsce and then installed on vscode. Allows interactive generation of snippets, or semiautomatic generation by copying jsx that we want to create a snippet from


ethanshar:
@jonathanmos 
In the 4th step -> `vsce package`, I get the following error: 
```
      TS2304: Cannot find name 'expect'.
```

ethanshar (commented):
I forgot to catch you after your presentation. 
While the extension is pretty useful I had something else in mind. 
I wanted to have an extension that offers predefined uilib components snippets. 
So people can have quick snippet for components when start working on a new code. 

I think this specific extension is not so relevant to uilib, maybe you want to move it to a different repo?

PR #1226, opened 2021-03-14
Support SVG image

M-i-k-e-l:
## Description
Support SVG image
I'm not sure about the `isSvg` check (in `Image`).
Two typings problems: (1) how do we import the lib's typings (while using `try` `catch`) (2) I had to use `any` as `data`'s type, it's not critical, it's internal for now.

## Changelog
Support SVG image


metro.config.js
```diff
@@ -1,17 +1,28 @@
...
+        }
+      }),
+      babelTransformerPath: require.resolve('react-native-svg-transformer')
+    },
+    resolver: {
+      assetExts: assetExts.filter(ext => ext !== 'svg'),
+      sourceExts: [...sourceExts, 'svg']
+    }
+  };
+})();
```
ethanshar, line 28: can you explain a little about this change? 
Is it something our users need to do to make this feature work?

package.json
```diff
@@ -109,6 +111,8 @@
     "react-native": ">=0.63.4",
     "react-native-gesture-handler": ">=1.9.0",
     "react-native-reanimated": ">=1.13.2",
+    "react-native-svg": "^12.1.0",
+    "react-native-svg-transformer": "^0.14.3",
```
ethanshar, line 115: We should reconsider adding optional dependencies in our peer deps. 
If the dependency is optional we maybe don't want our users to get warnings about it (not until the use the relevant features). WDYT?

src/components/image/SvgImage.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  if (typeof data === 'string') {
+    return <SvgXml xml={data} {...others}/>;
+  } else if (data) {
+    const File = data; // Must be with capital letter
+    return <File {...others}/>;
+  } else {
+    return null;
+  }
```
ethanshar, line 26: Is it necessary? 
if we know we're getting here only when passing string/function

demo/src/screens/incubatorScreens/index.js
```diff
@@ -5,4 +5,5 @@ export function registerScreens(registrar) {
     gestureHandlerRootHOC(require('./TouchableOpacityScreen').default));
   registrar('unicorn.components.IncubatorTextFieldScreen', () => require('./IncubatorTextFieldScreen').default);
   registrar('unicorn.incubator.WheelPickerScreen', () => gestureHandlerRootHOC(require('./WheelPickerScreen').default));
+  registrar('unicorn.incubator.SvgImageScreen', () => require('./SvgImageScreen').default);
```
ethanshar, line 8: Why this screen is under Incubator? 
The feature is part of our source code, don't we want to show it in the Image example screen?

demo/src/screens/incubatorScreens/SvgImageScreen.tsx
```diff
@@ -0,0 +1,59 @@
+import React, {Component} from 'react';
+import {Image, View, Text} from 'react-native-ui-lib';
+// @ts-ignore
+import {renderBooleanOption, renderRadioGroup} from '../ExampleScreenPresenter';
+const file = require('../../../../uilib-docs/src/images/newDesign/headerLogo.svg').default;
```
ethanshar, line 5: Consider loading this asset as any other image asset here
`./demo/src/configurations.js`

metro.config.js
M-i-k-e-l: This is part of the configuration of `react-native-svg-transformer`, this part is needed when loading files.
Yes, they will need to add this if they want to load SVG files.

package.json
M-i-k-e-l: I agree, removed (only the ones added in this PR)

src/components/image/SvgImage.tsx
M-i-k-e-l: It is probably not needed, but I think it's better safe than sorry.

demo/src/screens/incubatorScreens/index.js
M-i-k-e-l: I thought we're on a POC ATM, should I move it?

demo/src/screens/incubatorScreens/SvgImageScreen.tsx
M-i-k-e-l: Done

metro.config.js
ethanshar: This is part of the SVG library instructions? 
Maybe we can mention it somewhere in the component docs..

demo/src/screens/incubatorScreens/index.js
ethanshar: I would normally agree, but since it's part of the Image component I think maybe it should be part of the Image example screen?

src/components/image/SvgImage.tsx
ethanshar: Then consider returning at the end null regardless of the condition, wdyt?

metro.config.js
M-i-k-e-l: Yes, added docs.

demo/src/screens/incubatorScreens/index.js
M-i-k-e-l: Done

src/components/image/SvgImage.tsx
M-i-k-e-l: Done

ethanshar (approved):

PR #1236, opened 2021-03-22
Infra/delete utils_old file

lidord-wix:
## Description
Remove the old utils file
Move the relevant function inside prop-value-shape-deprecation
Update findValueNodeOfIdentifier() to be recursive

## Changelog
Remove the old utils file


eslint-rules/lib/rules/prop-value-shape-deprecation.js
```diff
@@ -130,6 +130,22 @@ module.exports = {
       }
     }
 
+    function getComponentName(node) {
```
M-i-k-e-l, line 133: Why not move to the new `imports` + `getComponentName(componentLocalName, imports)`?

lidord-wix: I changed to `getComponentLocalName(node)`
It seems like `imports` + `getComponentName` doesn't work here

M-i-k-e-l (approved):
Nice 👍 
Generally you should increase the version (in `package.json`), but I have a fix to eslint so I'll update there.

PR #1238, opened 2021-03-24
Feat/segmentetControl new component

lidord-wix:
## Description
Segmented control component

## Changelog
New component - SegmentedControl


ethanshar (commented):
The code looks great! 
Wrote few comments..

src/components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+    activeBackgroundColor = Colors.white,
+    unActiveColor = Colors.grey20,
+    outlineColor = Colors.grey60,
+    outlineWidth = BORDER_WIDTH
+  } = props;
+  const [selectedSegment, setSelectedSegment] = useState(-1);
+
+  const segments = useRef([] as {x: number; width: number}[]);
+  const segmentsCounter = useRef(0);
+  const animatedValue = new Reanimated.Value(selectedSegment);
```
ethanshar, line 74: Same as the counter, the animated value is not stored in the state, so on every render it creates a new animated value which breaks your animation continuousness

ethanshar (commented):
@lidord-wix 
See my recent comment regarding the animation issue

src/components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+   */
+  segments?: SegmentItemProps[];
+  /**
+   * The color of the active segment (label and outline).
+   */
+  activeColor?: string;
+  /**
+   * The color of the inactive segments (label).
+   */
+  inActiveColor?: string;
```
ethanshar, line 27:
```suggestion
  inactiveColor?: string;
```

src/components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    return (
+      <View row center style={[styles.container, style]}>
+        {animatedStyle && <Reanimated.View style={[styles.selectedSegment, animatedStyle, {borderColor: color}]}/>}
+        {_.map(labels, (_value, index) => {
+          return (
+            <Segment
+              segmentOnLayout={this.onLayout}
+              index={index}
+              onPress={index => this.onSegmentPress(index)}
```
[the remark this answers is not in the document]
ethanshar: If I'm adding a console.log to `Segment` component it still prints all of the segments when switching the active segment. 
I suggest investigating what's causing the extra renders.

src/components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+   */
+  backgroundColor?: string;
+  /**
+   * The background color of the active segment
+   */
+  activeBackgroundColor?: string;
+  /**
+   * The color of the segmentedControl outline
+   */
+  outlineColor?: string;
```
ethanshar, line 51: I would think this prop is meant for the active indicator and not the whole segment control. 
For the whole SegmentControl the user can pass the containerStyle. WDYT?

src/components/segmentedControl/segment.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const renderIcon = () => {
+    return iconSource && <Image source={iconSource} style={[{tintColor: segmentedColor}, iconStyle]}/>;
+  };
+
+  return (
+    <TouchableOpacity
+      onLayout={event => segmentOnLayout?.(index, event)}
+      style={segmentStyle}
+      onPress={() => onPress(index)}
```
ethanshar, line 83: When and if possible, try to avoid inline callbacks.

ethanshar (approved):

PR #1239, opened 2021-03-24
Avatar - add name with initials and backgroundColor specific to that name

M-i-k-e-l:
## Description
Avatar - add name with initials and backgroundColor specific to that name (i.e. hash)

## Changelog
Avatar - add name with initials and backgroundColor specific to that name (i.e. hash)


src/components/avatar/index.tsx
```diff
@@ -93,6 +124,20 @@ export type AvatarProps = Pick<AccessibilityProps, 'accessibilityLabel'> & {
...
+  /**
+   * The name of the avatar user.
+   * Text initials will be generated from the name.
+   * Has lower priority than label.
+   */
+  name?: string;
+  /**
+   * Override the logic for creating initials of the user name (default from AvatarHelper)
+   */
+  getInitials?: (name?: string, limit?: number) => string;
```
ethanshar, line 136: Avoid adding props that the user can have on their end. 
If the user wants to implement a custom logic for initials, they can do it and pass the `label` prop.

src/components/avatar/index.tsx
```diff
@@ -93,6 +124,20 @@ export type AvatarProps = Pick<AccessibilityProps, 'accessibilityLabel'> & {
...
+  /**
+   * Send this to use the name to infer a backgroundColor
+   */
+  nameToColorProps?: NameToColorProps;
```
ethanshar, line 140: Maybe a better name can be colorHashConfig? autoColorsConfig? dyanmicColorsConfog? (:

src/components/avatar/index.tsx
```diff
@@ -44,6 +46,35 @@ export enum BadgePosition {
 
 const DEFAULT_BADGE_SIZE = 'pimpleBig';
 
+export type NameToColorProps = {
+  /**
+   * Hash the name (or label) to get a color, so each name will have a specific color.
+   * Default is false.
+   */
+  useHashBackgroundColor: boolean;
```
ethanshar, line 54: missing `?`
```suggestion
  useHashBackgroundColor?: boolean;
```

Also, I think the flag prop should not be part of the config but alongside it. 

```
<Avatar useHashBackgroundColor hashColorConfig={....} name={...} />
```

src/components/avatar/index.tsx
```diff
@@ -44,6 +46,35 @@ export enum BadgePosition {
...
+   */
+  hashFunction?: (name?: string) => number;
+  /**
+   * Background color in cases where the getBackgroundColor returns undefined.
+   */
+  defaultColor?: string;
+  /**
+   * Replace the default logic.
+   */
+  getBackgroundColor?: (
```
ethanshar, line 70: same as `getInitials`, if the user wants to have a custom logic for the background color they can implement on their end and pass thw `backgroundColor` prop

src/components/avatar/index.tsx
```diff
@@ -44,6 +46,35 @@ export enum BadgePosition {
...
+   */
+  avatarColors?: string[];
+  /**
+   * Replace the default hashing function (name -> number)
+   */
+  hashFunction?: (name?: string) => number;
+  /**
+   * Background color in cases where the getBackgroundColor returns undefined.
+   */
+  defaultColor?: string;
```
ethanshar, line 66: See my comment on `getBackgroundColor`, not sure it's still relevant

src/components/avatar/index.tsx
```diff
@@ -318,14 +362,61 @@ class Avatar extends PureComponent<AvatarProps> {
...
+    return text;
+  });
+
+  get text() {
+    const {label, name, getInitials = AvatarHelper.getInitials} = this.props;
+    return this.getText(label, name, getInitials);
+  }
+
+  getBackgroundColor = memoize((text,
+    useHashBackgroundColor,
```
ethanshar, line 380: Why go into this method if the user pass `useHashBackgroundColor: false`?

src/helpers/AvatarHelper.ts
```diff
@@ -27,13 +27,13 @@ export function getColorById(id: string, avatarColors = getAvatarColors()) {
   return avatarColors[colorIndex];
 }
 
-export function getInitials(name: string) {
+export function getInitials(name?: string, limit = 2) {
```
ethanshar, line 30: Consider adding proper unit test for the new limit argument.

src/helpers/AvatarHelper.ts
```diff
@@ -43,6 +43,19 @@ export function getInitials(name: string) {
   return _.toUpper(initials);
 }
 
+export function getBackgroundColor(name?: string,
```
ethanshar, line 46: consider adding few simple tests to cover the new functionalities

src/components/avatar/index.tsx
M-i-k-e-l: But we want the coloring logic to only apply when `name` is sent, and `name` is ignored when `label` is sent.

src/components/avatar/index.tsx
M-i-k-e-l: `autoColorsConfig` with `AutoColorsProps` 10x :)

src/components/avatar/index.tsx
M-i-k-e-l: Moved (renamed to `useAutoColorsConfig`)

src/components/avatar/index.tsx
M-i-k-e-l: Removed

src/components/avatar/index.tsx
M-i-k-e-l: Removed

src/components/avatar/index.tsx
M-i-k-e-l: I think we need this as a backup (errors etc)

src/components/avatar/index.tsx
M-i-k-e-l: Fixed

src/helpers/AvatarHelper.ts
M-i-k-e-l: Added

src/helpers/AvatarHelper.ts
M-i-k-e-l: Done (found a bug!)

generatedTypes/components/avatar/index.d.ts
```diff
@@ -63,6 +77,21 @@ export declare type AvatarProps = Pick<AccessibilityProps, 'accessibilityLabel'>
...
+     * The name of the avatar user.
+     * If no label is provided, the initials will be generated from the name.
+     * autoColorsConfig will use the name to create the background color of the Avatar.
+     */
+    name?: string;
+    /**
+     * Hash the name (or label) to get a color, so each name will have a specific color.
+     * Default is false.
+     */
+    useAutoColorsConfig?: boolean;
```
ethanshar, line 90: `useAutoColorsConfig` sounds like the user must pass it. I think `useAutoColors` is better
```suggestion
    useAutoColors?: boolean;
```

src/components/avatar/index.tsx
```diff
@@ -318,14 +349,48 @@ class Avatar extends PureComponent<AvatarProps> {
...
+  get backgroundColor() {
+    const {backgroundColor, useAutoColorsConfig, autoColorsConfig, name} = this.props;
+    if (backgroundColor) {
+      return backgroundColor;
+    }
+
+    const {
+      avatarColors = AvatarHelper.getAvatarColors(),
+      hashFunction = AvatarHelper.hashStringToNumber,
+      defaultColor = Colors.dark80
```
ethanshar, line 379: dark preset should be deprecated eventually...
```suggestion
      defaultColor = Colors.grey80
```

src/helpers/AvatarHelper.ts
ethanshar: The power of testings (:

ethanshar:
Looks great!
Last small fixes (:

generatedTypes/components/avatar/index.d.ts
M-i-k-e-l: Done

ethanshar (approved):

PR #1245, opened 2021-03-31
Infra/set haptic infra as an optional dependency

lidord-wix:
## Description
Set haptic infrastructure as an optional dependency

## Changelog
Set haptic infrastructure as an optional dependency


package.json
```diff
@@ -54,6 +54,7 @@
     "prop-types": "^15.5.10",
     "react-native-animatable": "^1.1.0",
     "react-native-color": "0.0.10",
+    "react-native-haptic-feedback": "^1.11.0",
```
ethanshar, line 57: The whole idea of the optional dependencies is that we don't force to install those dependencies for the user. 
This is to avoid huge bundle size with packages our consumer don't necessarily use/need. 

If the user decide to use a component that has a dependency on one of the optionals deps, they can install it.

src/services/index.ts
```diff
@@ -1 +1,2 @@
 export {default as LogService} from './LogService';
+export {HapticMethods, triggerHaptic} from './HapticService';
```
ethanshar, line 2: Export the service as a whole, otherwise `services/index` fill will export a bunch of unrelated methods from various services.

demo/src/screens/componentScreens/HapticScreen.tsx
```diff
@@ -0,0 +1,27 @@
+import _ from 'lodash';
+import React, {Component} from 'react';
+import {ScrollView} from 'react-native';
+import {View, Text, Button} from 'react-native-ui-lib';
+import {HapticMethods, triggerHaptic} from '../../../../src/services';
```
ethanshar, line 5: See my comment on how you export haptic service.

demo/src/screens/componentScreens/HapticScreen.tsx
```diff
@@ -2,11 +2,11 @@ import _ from 'lodash';
 import React, {Component} from 'react';
 import {ScrollView} from 'react-native';
 import {View, Text, Button} from 'react-native-ui-lib';
-import {HapticMethods, triggerHaptic} from '../../../../src/services';
+import {HapticService} from '../../../../src/services';
```
ethanshar, line 5: I missed something 😬 
You should also export the HapticService in the main src/index and use it from there

ethanshar (approved):

PR #1246, opened 2021-04-01
Configure module-resolver plugin

ethanshar:
## Description
POC for using `babel-plugin-module-resolver` package. 
It allows us to create aliases to long relative paths (eg, `../../style`, `../../helpers`, etc..) 
Instead we can just import from our predefined aliases (eg. `style`, `helpers`, etc...)

In this PR I just made couple of changes in Button code to test the usage of this plugin. 
I'd like to give it some time and see that it integrates well in private uilib and in One App. 
If we won't have any complains or issues I'll fully integrate this plugin 

## Changelog
Use internally `babel-plugin-module-resolver` for resolving relative paths


tsconfig.dev.json
```diff
@@ -1,28 +1,5 @@
 {
-  "compilerOptions": {
```
lidord-wix, line 2: Don't we need the compilerOptoins anymore?

ethanshar: Nope, 
You can see that I now extends our base tsconfig.json file that already contains the relevant compilerOptions
`"extends": "./tsconfig",`

lidord-wix (approved):

PR #1250, opened 2021-04-03
Fix callback props typings in various components

ethanshar:
## Description
Fix callback props typings in components like: TouchableOpacity, Card and Modal.TopBar

## Changelog
Fix callback props typings in components like: TouchableOpacity, Card and Modal.TopBar

### DIFF
--- a/generatedTypes/components/card/index.d.ts
+++ b/generatedTypes/components/card/index.d.ts
@@ -25,7 +25,7 @@ export declare type CardProps = ViewProps & TouchableOpacityProps & {
     /**
      * action for when pressing the card
      */
-    onPress?: () => void;
+    onPress?: TouchableOpacityProps['onPress'];
     /**
      * whether the card should have shadow or not
      */
@@ -83,7 +83,7 @@ declare const _default: React.ComponentClass<ViewProps & TouchableOpacityProps &
     /**
      * action for when pressing the card
      */
-    onPress?: (() => void) | undefined;
+    onPress?: ((props?: any) => void) | undefined;
     /**
      * whether the card should have shadow or not
      */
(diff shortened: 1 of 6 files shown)


generatedTypes/components/touchableOpacity/index.d.ts
```diff
@@ -30,7 +30,7 @@ export interface TouchableOpacityProps extends Omit<RNTouchableOpacityProps, 'st
      */
     customValue?: any;
     style?: StyleProp<ViewStyle> | Animated.AnimatedProps<StyleProp<ViewStyle>>;
-    onPress?: (props: TouchableOpacityProps | any) => void;
+    onPress?: (props?: TouchableOpacityProps | any) => void;
```
mendyEdri, line 33: What the reason we need `| any` here?

generatedTypes/components/card/index.d.ts
```diff
@@ -83,7 +83,7 @@ declare const _default: React.ComponentClass<ViewProps & TouchableOpacityProps &
     /**
      * action for when pressing the card
      */
-    onPress?: (() => void) | undefined;
+    onPress?: ((props?: any) => void) | undefined;
```
mendyEdri, line 86: Is typescript still make us use `| undefined` although `onPress` is an optional?

generatedTypes/components/touchableOpacity/index.d.ts
ethanshar: Hmm good point. 
It was here before, but it might be irrelevant.. 
ill try to remove it and see if it affect us

ethanshar: Ok, there was a reason for the `any` type.
In TouchableOpacity we allow to pass a prop called `customValue` which can be anything you want to pass and get back in the `onPress` callback. 
Because `customValue` can be anything, we have to support `any` type in the onPress callback, cause we can't tell what it'll be

generatedTypes/components/card/index.d.ts
ethanshar: You are looking at the generatedTypes file, this is how it generates the d.ts files apparently. 
The source code doesn't have it

generatedTypes/components/touchableOpacity/index.d.ts
mendyEdri: Yeah I see now. make sense :)

mendyEdri (approved):

PR #1253, opened 2021-04-06
ChipScreen - adding example for dynamic label

Inbal-Tish:
ChipScreen - adding example for onPress to open Dialog and dynamic label.
Chip - moving default values from static.
WheelPickerDialog - container self-alignment and code cleanup.

## Description
ChipScreen - adding example for dynamic label

## Changelog
ChipScreen - adding example for onPress to open Dialog and dynamic label.


demo/src/screens/componentScreens/ChipScreen.tsx
```diff
@@ -1,13 +1,90 @@
...
+        {color.label}
+      </Text>
+    );
+  };
+
+  renderContent = () => {
+    const {selectedValue} = this.state;
+
+    return (
+      <WheelPickerDialog
```
ethanshar, line 60: Why did you add a picker to the chip example screen? 
Don't you think it's making the example screen a bit more complicated than needed? 

What do you want to demonstrate here?

Inbal-Tish: That was a request by ux

ethanshar: ohh ok :/

ethanshar (approved):

PR #1254, opened 2021-04-06
Picker - filter children (for long lists with search)

### LINKED ISSUE #1244: Search in Picker component
### Describe the bug

When we has very much data in Picker component, the Children will be optimized in FlatList which is used in Picker  and searching will not work correctly.  How to resolve this problem?

M-i-k-e-l:
## Description
Picker - filter children (for long lists with search).
I think this solved the issue, but it has some code duplication with the `Picker.Item`, do you think it can be removed from the item?
Should resolve #1244

## Changelog
Picker - filter children (for long lists with search)


src/components/picker/index.js
```diff
@@ -271,6 +273,25 @@ class Picker extends Component {
...
+  get children() {
+    const {searchValue} = this.state;
+    const {children, showSearch} = this.props;
+    if (showSearch && !_.isEmpty(searchValue)) {
+      return this.getFilteredChildren(children, searchValue);
+    }
+
+    return children;
+  }
+
```
ethanshar, line 294: Looks good! we should remove the search logic we have in PickerItem code since this one covers it.

M-i-k-e-l: Done

src/components/picker/PickerItem.js
```diff
@@ -80,10 +80,6 @@ const PickerItem = props => {
     );
   };
 
-  if (context.showSearch && shouldFilterOut(context.searchValue, itemLabel)) {
```
ethanshar, line 83: You can also stop passing `showSearch` and `searchValue` in the context provided by the Picker

M-i-k-e-l: Done

ethanshar (approved):

PR #1259, opened 2021-04-11
Feat/add label to wheelPicker

lidord-wix:
## Description
Add right label to the wheelPicker selected value

## Changelog
Add right label to the wheelPicker selected value

### DIFF
--- a/generatedTypes/incubator/WheelPicker/index.d.ts
+++ b/generatedTypes/incubator/WheelPicker/index.d.ts
@@ -1,6 +1,7 @@
 import React from 'react';
 import { TextStyle, ViewStyle } from 'react-native';
 import { ItemProps } from './Item';
+import { TextProps } from '../../components/text';
 export interface WheelPickerProps {
     /**
      * Data source for WheelPicker
@@ -29,6 +30,18 @@ export interface WheelPickerProps {
      */
     textStyle?: TextStyle;
     /**
+     * Additional label on the right of the item text
+     */
+    rightLabel?: string;
+    /**
+     * The Additional right label's style
+     */
+    rightLabelStyle?: TextStyle;
+    /**
+     * The Additional right label's props
+     */
+    rightLabelProps?: TextProps;
+    /**
      * Event, on active row change
      */
     onChange?: (item: string | number, index: number) => void;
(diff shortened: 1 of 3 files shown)


src/incubator/WheelPicker/index.tsx
```diff
@@ -153,7 +169,12 @@ const WheelPicker = React.memo(
               height: Spacings.s9,
               borderColor: Colors.grey60
             }}
-          />
+            center
+          >
+            <Text marginL-80 text80M {...rightLabelProps} color={activeTextColor} style={rightLabelStyle}>
```
mendyEdri, line 174: Maybe we should change the function name / move the label out, since now it's not rendering only the `separators`

src/incubator/WheelPicker/index.tsx
```diff
@@ -39,6 +40,18 @@ export interface WheelPickerProps {
    * Row text style
    */
   textStyle?: TextStyle;
+  /**
+   * Additional label on the right of the item text
+   */
+  rightLabel?: string;
```
mendyEdri, line 46: Is there going to be another label? if not, maybe we should call it `label`

generatedTypes/incubator/WheelPicker/index.d.ts
```diff
@@ -28,6 +29,18 @@ export interface WheelPickerProps {
      * Row text style
      */
     textStyle?: TextStyle;
+    /**
```
mendyEdri, line 32: I'm not sure we supposed to do any changes in this files (it's under `generatedType`)

mendyEdri (changes requested):

generatedTypes/incubator/WheelPicker/index.d.ts
lidord-wix: I didn't, It's automatic :)

src/incubator/WheelPicker/index.tsx
```diff
@@ -143,7 +159,7 @@ const WheelPicker = React.memo(
       [itemHeight]
     );
 
-    const separators = () => {
+    const renderCentralFrame = () => {
```
mendyEdri, line 162: Sorry to nudge you 😬
`renderSeparatorsAndLabel` or something, make more sense - at least to me, didn't get at first what a central frame is..

lidord-wix: Done.
I also changed the screen a bit to make it fit for small screens.
verified on Android as well

mendyEdri (approved):

PR #1260, opened 2021-04-12
Feat/SectionsWheelPicker new component

lidord-wix:
## Description
Create a new SectionsWheelPicker component to present a set of wheelPickers
Jira 155

## Changelog
Create a new SectionsWheelPicker component to present a set of wheelPickers


demo/src/screens/componentScreens/SectionsWheelPickerScreen.tsx
```diff
@@ -0,0 +1,108 @@
+import _ from 'lodash';
+import React, {useState, useCallback} from 'react';
+import {Alert} from 'react-native';
+import {Text, View, SectionsWheelPicker, SegmentedControl, Button} from 'react-native-ui-lib';
+import {WheelPickerProps} from '../../../../src/incubator/WheelPicker';
```
ethanshar, line 5: We're not exporting Incubator WheelPickerProps in our src/index?

ethanshar:
Looks great! I wrote a few comments..

demo/src/screens/componentScreens/SectionsWheelPickerScreen.tsx
lidord-wix: No, adding it

src/components/sectionsWheelPicker/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+export type SectionsWheelPickerProps = {
+  /**
+   * Array of sections.
+   */
+  sections?: WheelPickerProps[];
+  /**
+   * Describe the width of each section in the SectionsWheelPicker
+   * default value: '33%'
+   */
+  sectionWidth?: string | number;
```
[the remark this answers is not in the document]
lidord-wix: I'm not sure this is what you meant, but I changed it to be affected by the style prop of each section.
In addition, I updated the default width to be equal for each section, no matter how many sections there are.
Let me know if you thought about a better solution.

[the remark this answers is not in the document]
ethanshar: The width shouldn't be fixed or equal by default. 
Usually each section will have a different width based on its content (values and label)
I think the default behavior should be a width that is based on each section content.

src/components/sectionsWheelPicker/index.tsx
```diff
@@ -56,7 +54,7 @@ const SectionsWheelPicker = (props: SectionsWheelPickerProps) => {
   const renderSections = () =>
     _.map(sections, (section, index) => {
       return (
-        <View height={1} width={section.style?.width || defaultWidth} key={index} testID={testID}>
+        <View height={1} width={section.style?.width} key={index} testID={testID}>
```
ethanshar, line 57: Passing the `width` here is redundant cause the width should come from the wheel picker content and the style is passed there

src/components/sectionsWheelPicker/index.tsx
```diff
@@ -56,7 +54,7 @@ const SectionsWheelPicker = (props: SectionsWheelPickerProps) => {
...
           <WheelPicker testID={`${testID}.${index}`} {...wheelPickerProps} {...section} style={{width: '100%'}}/>
```
ethanshar, line 58: Same here, no reason to pass the width, the content of the wheel picker should determine its size. 
Also, regardless, passing the style like that overrides the user style

src/incubator/WheelPicker/Item.tsx
```diff
@@ -35,31 +34,30 @@ export default ({
...
-  
+
   return (
     <AnimatedTouchableOpacity
       activeOpacity={1}
       style={{
-        height: itemHeight
+        height: itemHeight,
+        minWidth: Spacings.s10
       }}
```
ethanshar, line 56: Use StyleSheet for static styles. 
And I suggest memoizing the style we pass here, It can save performance.

src/index.ts
```diff
@@ -40,7 +40,7 @@ export {default as Overlay, OverlayTypes} from './components/overlay';
 export {default as RadioButton, RadioButtonPropTypes, RadioButtonProps} from './components/radioButton/RadioButton';
 export {default as RadioGroup, RadioGroupPropTypes, RadioGroupProps} from './components/radioButton/RadioGroup';
 export {default as SectionsWheelPicker, SectionsWheelPickerProps} from './components/sectionsWheelPicker';
-export {default as SegmentedControl, SegmentedControlProps} from './components/segmentedControl';
+export {default as SegmentedControl, SegmentedControlProps, SegmentedControlItemProps} from './components/segmentedControl';
 export {default as Switch, SwitchProps} from './components/switch';
```
ethanshar: I don't think this one is part of the PR (: 
Merging master will probably fix that

ethanshar (commented):
Wrote some last few comments 👍

src/index.ts
lidord-wix: There was a conflict here after merging master

src/incubator/WheelPicker/Item.tsx
```diff
@@ -2,7 +2,7 @@ import React, {useCallback, useMemo} from 'react';
 import Animated, {interpolateColors} from 'react-native-reanimated';
 import Text from '../../components/text';
 import TouchableOpacity from '../../components/touchableOpacity';
-import {TextStyle} from 'react-native';
+import {TextStyle, StyleSheet} from 'react-native';
 import {Colors, Spacings} from '../../../src/style';
```
ethanshar: keep module imports at the top, relative imports after

ethanshar (approved):
Looks great, I approved. 
I wrote a small comment after the imports order

PR #1266, opened 2021-04-18
Feat/add haptics to Drawer

lidord-wix:
## Description
Add haptic feedback to the Drawer component
Jira 1201

Depending on https://github.com/wix/react-native-ui-lib/pull/1277 

## Changelog
Add haptic feedback to the Drawer component


src/components/drawer/Swipeable.tsx
```diff
@@ -123,7 +125,8 @@ export default class Swipeable extends Component<Props, StateType> {
       if (!this.dragThresholdReached && x >= threshold && x < threshold + 10) {
         // move item right
         this.dragThresholdReached = true;
-        onToggleSwipeLeft({rowWidth, leftWidth, dragX: x, triggerHaptic: true});
+        HapticService.triggerHaptic(HAPTIC_METHOD, 'Drawer');
+        onToggleSwipeLeft({rowWidth, leftWidth, dragX: x});
```
Inbal-Tish, line 129: I think we should keep the haptic optional, not all users will want to use it.

src/components/drawer/index.tsx
```diff
@@ -97,10 +97,6 @@ interface Props {
    * Callback for just before right item full swipe
    */
   onWillFullSwipeRight?: Function;
-  /**
-   * Haptic trigger function to use onToggleSwipeLeft
-   */
-  leftToggleHapticTrigger?: Function;
```
Inbal-Tish, line 103: This is a breaking change. You can't just delete props, you need to deprecate them. Give the user a chance to migrate.

lidord-wix: Right, adding it to the V7 spreadsheet

src/components/drawer/Swipeable.tsx
Inbal-Tish: Why not get the haptic method from the HapticService, i.e. `HapticMethods.impactMedium`?

lidord-wix: I'll be able to do that only when I'll fix the `HapticMethods` in the `HapticService` as we talked about in our dev sync.
I'm waiting for this PR to be merged to do so

Inbal-Tish: Ok. Add the other PR as a dependency and we'll merge this after it.

src/components/drawer/Swipeable.tsx
```diff
@@ -123,7 +126,8 @@ export default class Swipeable extends Component<Props, StateType> {
       if (!this.dragThresholdReached && x >= threshold && x < threshold + 10) {
         // move item right
         this.dragThresholdReached = true;
-        onToggleSwipeLeft({rowWidth, leftWidth, dragX: x, triggerHaptic: true});
+        !disableHaptic && HapticService.triggerHaptic(HAPTIC_METHOD, 'Drawer');
```
Inbal-Tish, line 129: this line repeats 3 times. How about moving it to a method?

src/components/drawer/Swipeable.tsx
```diff
@@ -11,10 +11,12 @@ import React, {Component} from 'react';
 import {Animated, StyleSheet, View, I18nManager} from 'react-native';
 import {PanGestureHandler, TapGestureHandler, State} from 'react-native-gesture-handler';
 import {Constants} from '../../helpers';
+import {HapticService, HapticType} from '../../services';
 
 
 const DRAG_TOSS = 0.05;
 const LEFT_TOGGLE_THRESHOLD = 0.6;
+const HAPTIC_METHOD = HapticType.impactMedium;
```
Inbal-Tish, line 19: Is this working? The other PR merged?

lidord-wix: It is not merged yet - I added a link in the description.
And yes it works, I checked it..

Inbal-Tish: If you're using HAPTIC_METHOD only once, why store it in a const?

src/components/drawer/Swipeable.tsx
```diff
@@ -249,7 +257,7 @@ export default class Swipeable extends Component<Props, StateType> {
       rightThreshold = rightWidth / 2,
       fullLeftThreshold,
       fullRightThreshold,
-      onToggleSwipeLeft
+      onToggleSwipeLeft,
```
Inbal-Tish, line 260: Why is there a comma here? Doesn't it fail on lint tests?

src/components/drawer/Swipeable.tsx
Inbal-Tish: Also, why `impactMedium` is the default?

src/components/drawer/Swipeable.tsx
lidord-wix: No, but I removed it.

src/components/drawer/Swipeable.tsx
lidord-wix: These are the UX guidelines.
We can open it to a discussion in the future if any user will ask to change it.

Inbal-Tish: I'm asking because our private config (Drawer.config) is setting it to `impactLight`, so it should be the default or it should be open for user's change.

src/components/drawer/Swipeable.tsx
Inbal-Tish: Is the dependency merged?

lidord-wix: Yes :)

src/components/drawer/Swipeable.tsx
lidord-wix: It has updated to `impactMedium`, which is the default now.

Inbal-Tish (approved):

Issue #1262, opened 2021-04-14
Is it possible to prevent a Dialog from being dismissed after a tap on the background?


stanica:
I'd like to make the dialog box dismissable only when pressing a button

ethanshar:
Released in v5.23.0

PR #1267, opened 2021-04-19
Feat/add IgnoreBackgroundPress prop

lidord-wix:
## Description
Add IgnoreBackgroundPress prop to Dialog component to support not dismissing the dialog on background press 
Add example in DialogScreen
relevant issue - https://github.com/wix/react-native-ui-lib/issues/1262

## Changelog
Add IgnoreBackgroundPress prop to Dialog component

### DIFF
--- a/src/components/dialog/index.tsx
+++ b/src/components/dialog/index.tsx
@@ -289,7 +289,7 @@ class Dialog extends Component<DialogProps, DialogState> {
         visible={modalVisibility}
         animationType={'none'}
         onBackgroundPress={onBackgroundPress}
-        onRequestClose={this.hideDialogView}
+        onRequestClose={onBackgroundPress}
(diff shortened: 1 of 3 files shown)


src/components/dialog/index.tsx
```diff
@@ -283,7 +288,7 @@ class Dialog extends Component<DialogProps, DialogState> {
         transparent
         visible={modalVisibility}
         animationType={'none'}
-        onBackgroundPress={this.hideDialogView}
+        onBackgroundPress={onBackgroundPress}
         onRequestClose={this.hideDialogView}
```
M-i-k-e-l, line 292: It feels like this should be affected as well (`onRequestClose` is the back button in Android)

M-i-k-e-l (approved):

PR #1269, opened 2021-04-21
Update dateTimePicker to support Xcode 12 and iOS 14

yedidyak:
See https://github.com/react-native-datetimepicker/datetimepicker/issues/285 for more details

## Description
Set the display mode of the DateTimePicker to 'inline' on iOS to fit the previous native style for date and time pickers, when built with Xcode 12 for iOS 14.

## Changelog
Fixes DateTimePicker component style for apps built with Xcode 12 running on iOS 14.

### DIFF
--- a/src/components/dateTimePicker/index.js
+++ b/src/components/dateTimePicker/index.js
@@ -249,7 +249,7 @@ class DateTimePicker extends Component {
           is24Hour={is24Hour}
           minuteInterval={minuteInterval}
           timeZoneOffsetInMinutes={timeZoneOffsetInMinutes}
-          display={Constants.isIOS ? "inline" : "default"}
+          display={Constants.isIOS ? 'inline' : 'default'}
         />
       );
     }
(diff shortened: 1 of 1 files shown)


src/components/dateTimePicker/index.js
```diff
@@ -249,6 +249,7 @@ class DateTimePicker extends Component {
           is24Hour={is24Hour}
           minuteInterval={minuteInterval}
           timeZoneOffsetInMinutes={timeZoneOffsetInMinutes}
+          display={Platform.OS === "ios" ? "inline" : "default"}
```
ethanshar, line 252: You don't have Platform imported. 
Anyway, better use our Constants 
```suggestion
          display={Constants.isIOS ? "inline" : "default"}
```

Inbal-Tish, line 252: You're not importing Platform. You can use Constants.isIOS.

ethanshar (commented):
@yedidyak 
Can you add a short description/changelog

ethanshar (approved):

PR #1270, opened 2021-04-21
Feat/Picker selectionLimit prop

lidord-wix:
## Description
Add selectionLimit prop to the Picker component to support limiting the number of the selected items.
Jira 1702

## Changelog
Add selectionLimit prop to the Picker component to support limiting the number of the selected items.

### DIFF
--- a/src/components/picker/PickerItem.js
+++ b/src/components/picker/PickerItem.js
@@ -34,12 +34,17 @@ const PickerItem = props => {
   const itemValue = !migrate && _.isPlainObject(value) ? value?.value : value;
   const isSelected = isItemSelected(itemValue, context.value);
   const itemLabel = getItemLabel(label, value, props.getItemLabel || context.getItemLabel);
+  const selectedCounter = context.selectionLimit && context.value?.length;
   const accessibilityProps = {
     accessibilityState: isSelected ? {selected: true} : undefined,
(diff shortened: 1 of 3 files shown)


src/components/picker/PickerItem.js
```diff
@@ -34,12 +34,17 @@ const PickerItem = props => {
...
+  const selectedCounter = context.selectionLimit && context.value?.length;
   const accessibilityProps = {
     accessibilityState: isSelected ? {selected: true} : undefined,
     accessibilityHint: 'Double click to select this suggestion',
     ...Modifiers.extractAccessibilityProps(props)
   };
 
+  const isItemDisabled = useMemo(() => {
+    return disabled || (!isSelected && context.selectionLimit && context.selectionLimit === selectedCounter);
+  }, [selectedCounter]);
```
ethanshar, line 46: Notice you use a simple variable as a dependency in this useMemo. 
Each time the component will render,  `selectedCounter` will be recreated and will invoke the useMemo
Overall it's not that dramatic in this case, but better do it right. 

I'd move `selectedCounter` into the useMemo method and include in the dependencies array values from props/context/state.

ethanshar (changes requested):
Looks good! 
wrote you one thing worth fixing

ethanshar (approved):

PR #1273, opened 2021-04-26
Feat/slider orientation support

mendyEdri:
## Description
Supporting Slider react and recalculate when orientation is changed

## Changelog
Add support to Slider component to react to orientation changes

### DIFF
--- a/demo/src/screens/componentScreens/SliderScreen.js
+++ b/demo/src/screens/componentScreens/SliderScreen.js
@@ -41,7 +41,7 @@ export default class SliderScreen extends Component {
             <Image assetName={'megaphone'} style={styles.image}/>
             <Slider 
               onValueChange={this.onSliderValueChange} 
-              value={INITIAL_VALUE} 
+              value={this.state.sliderValue} 
               minimumValue={0} 
               maximumValue={100} 
               step={1} 
--- a/demo/src/screens/componentScreens/SliderScreen.js
+++ b/demo/src/screens/componentScreens/SliderScreen.js
@@ -41,7 +41,7 @@ export default class SliderScreen extends Component {
             <Image assetName={'megaphone'} style={styles.image}/>
             <Slider 
               onValueChange={this.onSliderValueChange} 
-              value={this.state.sliderValue} 
+              value={INITIAL_VALUE}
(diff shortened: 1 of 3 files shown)


src/components/slider/index.js
```diff
@@ -401,10 +414,10 @@ export default class Slider extends PureBaseComponent {
     if (currentSize && width === currentSize.width && height === currentSize.height) {
       return;
     }
-
     this[layoutName] = size;
 
     if (this.containerSize && this.thumbSize && this.trackSize) {
+      // console.warn('post return');
```
Inbal-Tish, line 420: You can delete this console

demo/src/screens/componentScreens/SliderScreen.js
```diff
@@ -41,7 +41,7 @@ export default class SliderScreen extends Component {
             <Image assetName={'megaphone'} style={styles.image}/>
             <Slider 
               onValueChange={this.onSliderValueChange} 
-              value={INITIAL_VALUE} 
+              value={this.state.sliderValue}
```
Inbal-Tish, line 44: Why do we need to keep updating the value by state here? The 'value' prop is for the **initial** value, once you update it the Slider takes care of the value and uses 'onValueChange' to update the user with the new value. When we update it back to the Slider we'll have an extra render on any value change which can affect performance.

src/components/slider/index.js
```diff
@@ -291,9 +301,9 @@ export default class Slider extends PureBaseComponent {
   }
 
   getXForValue(v) {
-    const {minimumValue, maximumValue} = this.props;
+    const {minimumValue} = this.props;
     const range = this.getRange();
-    const relativeValue = maximumValue > 0 ? minimumValue - v : maximumValue - v; // for negatives in min value
+    const relativeValue = minimumValue - v;
```
Inbal-Tish, line 306: Is there a way to test this part. In unit tests or e2e?

Inbal-Tish (approved):

Inbal-Tish:
@mendyEdri I approved the PR. Merge when you want. I just wondering if it's not better to add tests first to make sure we didn't break any one...

PR #1277, opened 2021-04-27
Add HapticType

lidord-wix:
## Description
Export enum HapticType instead of hapticMethods

## Changelog
Export enum HapticType instead of hapticMethods

### DIFF
--- a/src/services/HapticService.ts
+++ b/src/services/HapticService.ts
@@ -5,17 +5,17 @@ const options = {
   ignoreAndroidSystemSettings: false
 };
 
-const HapticMethods = [
-  'selection',
-  'impactLight',
-  'impactMedium',
-  'impactHeavy',
-  'notificationSuccess',
-  'notificationWarning',
-  'notificationError'
-];
+export enum HapticType {
+  selection = 'selection',
+  impactLight = 'impactLight',
+  impactMedium = 'impactMedium',
+  impactHeavy = 'impactHeavy',
+  notificationSuccess = 'notificationSuccess',
+  notificationWarning = 'notificationWarning',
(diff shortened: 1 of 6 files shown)


src/services/HapticService.ts
```diff
@@ -24,6 +24,6 @@ function triggerHaptic(hapticMethod: string, componentName: string) {
 }
 
 export default {
-  HapticMethods,
+  HapticType,
```
mendyEdri, line 27: No migration needed?

src/services/HapticService.ts
```diff
@@ -5,17 +5,17 @@ const options = {
...
+  impactLight = 'impactLight',
+  impactMedium = 'impactMedium',
+  impactHeavy = 'impactHeavy',
+  notificationSuccess = 'notificationSuccess',
+  notificationWarning = 'notificationWarning',
+  notificationError = 'notificationError'
+}
 
-function triggerHaptic(hapticMethod: string, componentName: string) {
+function triggerHaptic(hapticMethod: HapticType, componentName: string) {
```
mendyEdri, line 18: It's intentionally kept its name `method` where all other places has changed to `type`?

mendyEdri (commented):
Minor changes :)

src/services/HapticService.ts
lidord-wix: Nope, it's a new code so it's not in use yet :)

src/services/HapticService.ts
lidord-wix: It was intentional, but on second thought it is better to be aligned with our 'old' code in the private repo.

mendyEdri (approved):

PR #1279, opened 2021-04-28
Fix issue with Card's backgroundColor not working

ethanshar:
## Description
Fix issue with Card's backgroundColor prop not working

## Changelog
Fix issue with Card's backgroundColor prop not working

### DIFF
--- a/src/components/card/index.tsx
+++ b/src/components/card/index.tsx
@@ -194,13 +194,13 @@ class Card extends PureComponent<PropTypes, State> {
     }
   }
 
-  get blurBgStyle() {
-    const {enableBlur} = this.props;
+  get backgroundStyle() {
+    const {enableBlur, backgroundColor = Colors.white} = this.props;
 
     if (Constants.isIOS && enableBlur) {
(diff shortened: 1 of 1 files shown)


src/components/card/index.tsx
```diff
@@ -194,13 +194,13 @@ class Card extends PureComponent<PropTypes, State> {
     }
   }
 
-  get blurBgStyle() {
-    const {enableBlur} = this.props;
+  get backgroundStyle() {
+    const {enableBlur, backgroundColor = Colors.white} = this.props;
 
     if (Constants.isIOS && enableBlur) {
       return {backgroundColor: Colors.rgba(Colors.white, 0.85)};
```
mendyEdri, line 201: when the blur is enabled, aren't we supposed to pass the `backgroundColor` to `rgba` function as well? 
`return {backgroundColor: Colors.rgba(backgroundColor, 0.85)};`

ethanshar: Completely right! Fixed ✅

mendyEdri (approved):

PR #1281, opened 2021-04-29
Migrate Hint component to TypeScript

saulve:
## Description
This PR migrates Hint component and its example screen to TypeScript.

## Changelog
Removes manual typings in `Hint.d.ts`


src/components/hint/index.tsx
```diff
@@ -172,7 +217,7 @@ class Hint extends BaseComponent {
   }
 
   get containerWidth() {
-    const {containerWidth} = this.getThemeProps();
+    const {containerWidth} = getThemeProps(this.props, this.context);
```
Inbal-Tish, line 220: When wrapping the component with `asBaseComponent` you can simply extract from `this.props`, as the themeProps already passed to the WrappedComponent (see asBaseComponent line #54).

src/components/hint/index.tsx
```diff
@@ -195,13 +240,13 @@ class Hint extends BaseComponent {
   }
 
   get hintOffset() {
-    const {offset} = this.getThemeProps();
-    return offset || DEFAULT_HINT_OFFSET;
+    const {offset} = getThemeProps(this.props, this.context);
```
Inbal-Tish, line 243: same here

src/components/hint/index.tsx
```diff
@@ -195,13 +240,13 @@ class Hint extends BaseComponent {
...
+    return (offset || DEFAULT_HINT_OFFSET) as number;
   }
 
   get edgeMargins() {
-    const {edgeMargins} = this.getThemeProps();
-    return edgeMargins || DEFAULT_EDGE_MARGINS;
+    const {edgeMargins} = getThemeProps(this.props, this.context);
+    return (edgeMargins || DEFAULT_EDGE_MARGINS) as number;
```
Inbal-Tish, line 249: same here

src/components/hint/index.tsx
```diff
@@ -275,36 +326,39 @@ class Hint extends BaseComponent {
   }
 
   getTipPosition() {
-    const {position} = this.getThemeProps();
-    const tipPositionStyle = {};
+    const {position} = getThemeProps(this.props, this.context);
```
Inbal-Tish, line 329: same here

src/components/hint/index.tsx
```diff
@@ -195,13 +240,13 @@ class Hint extends BaseComponent {
   }
 
   get hintOffset() {
-    const {offset} = this.getThemeProps();
-    return offset || DEFAULT_HINT_OFFSET;
+    const {offset} = getThemeProps(this.props, this.context);
+    return (offset || DEFAULT_HINT_OFFSET) as number;
```
Inbal-Tish, line 244: Why the `as Number`?

src/components/hint/index.tsx
```diff
@@ -232,11 +279,15 @@ class Hint extends BaseComponent {
...
+
+    if (this.targetLayout?.x) {
+      hintPositionStyle.left = -this.targetLayout.x;
+    }
 
-    if (position === HINT_POSITIONS.TOP) {
+    if (position === HintPositions.TOP) {
       hintPositionStyle.bottom = 0;
-    } else {
+    } else if (this.targetLayout?.height) {
```
Inbal-Tish, line 290: Are you sure this `if` will not break anything? Cause you can set it `undefined`...

src/components/hint/index.tsx
```diff
@@ -251,12 +302,12 @@ class Hint extends BaseComponent {
   }
 
   getHintPadding() {
-    const paddings = {paddingVertical: this.hintOffset, paddingHorizontal: this.edgeMargins};
-    if (this.useSideTip) {
+    const paddings: Paddings = {paddingVertical: this.hintOffset, paddingHorizontal: this.edgeMargins};
+    if (this.useSideTip && this.targetLayout?.x) {
```
Inbal-Tish, line 306: Are you sure this does not break anything, when not setting the `paddings.paddingLeft` to undefined?

src/components/hint/index.tsx
```diff
@@ -335,9 +389,9 @@ class Hint extends BaseComponent {
   // }
 
   renderHintTip() {
-    const {position, color} = this.getThemeProps();
+    const {position, color} = getThemeProps(this.props, this.context);
```
Inbal-Tish, line 392: same here - you can extract from `this.props`

src/components/hint/index.tsx
```diff
@@ -353,7 +407,7 @@ class Hint extends BaseComponent {
   }
 
   renderHint() {
-    const {message, messageStyle, icon, iconStyle, borderRadius, color, testID} = this.getThemeProps();
+    const {message, messageStyle, icon, iconStyle, borderRadius, color, testID} = getThemeProps(this.props, this.context);
```
Inbal-Tish, line 410: Same here

src/components/hint/index.tsx
```diff
@@ -418,12 +472,11 @@ class Hint extends BaseComponent {
       <React.Fragment>
         {onBackgroundPress ? (
           <Modal
-            pointerEvents="box-none"
```
Inbal-Tish, line 421: Why are you removing the `pointerEvents`?

src/components/hint/index.tsx
```diff
@@ -418,12 +472,11 @@ class Hint extends BaseComponent {
...
             visible={this.showHint}
             animationType="none"
             transparent
             onBackgroundPress={onBackgroundPress}
-            onRequestClose={onBackgroundPress}
+            onRequestClose={onBackgroundPress as () => void}
```
Inbal-Tish, line 479: Why the change of type to `() => void`?

src/components/hint/index.tsx
saulve: I see, will add the change 👍🏼

src/components/hint/index.tsx
saulve: `getThemeProps` returns `any`, however I'll remove this, as it's not needed if we use `this.props`

src/components/hint/index.tsx
saulve: Not sure I understand the issue. `this.targetLayout.x` is necessary when setting both `paddings.paddingLeft` and `paddings.paddingRight`. This prevents a runtime error if `this.targetLayout` is undefined. It seems to work fine as much as I tested. Am I missing something?

src/components/hint/index.tsx
saulve: `pointerEvents` is a `View`, not `Modal` prop.
https://reactnative.dev/docs/view#pointerevents

src/components/hint/index.tsx
saulve: because `onRequestClose` expects a function with `() => void` signature, however `onBackgroundPress` has `(event: GestureResponderEvent) => void`. Therefore I'm adding a type assertion here.

M-i-k-e-l: This looks like a bug, because the user that sends the `onBackgroundPress` can expect to receive an `event` and may rely on it. I think you need to change the prop to:
```
onBackgroundPress?: (event?: GestureResponderEvent) => void;
```
(and remove the ` as () => void`)

src/components/hint/index.tsx
```diff
@@ -476,4 +528,4 @@ const styles = StyleSheet.create({
   }
 });
 
-export default Hint;
+export default asBaseComponent<HintProps>(Hint);
```
M-i-k-e-l, line 531:
```suggestion
export default asBaseComponent<HintProps, typeof Hint>(Hint);
```
This will solve the error you have in the screen in:
```
position={showBottomHint ? Hint.positions.BOTTOM : Hint.positions.TOP}
```

M-i-k-e-l (commented):
Nice work and thank you!

I've left some comments, you also need to update `generatedTypes` (it's a little duplication of `src/index` :/ ).
There are some things that I'd like to see changed, but I'm not sure how easy they are, feel free to try:
1. Changing the `refs` to something other than `any`.
2. Reducing the number of `interfaces`, I've suggested one, you can open a ticket on this if you don't see other improvements - this might require some infra work.

demo/src/screens/componentScreens/HintsScreen.tsx
```diff
@@ -1,11 +1,23 @@
 import React, {Component} from 'react';
-import {View, Text, Hint, Button, RadioGroup, RadioButton, Switch} from 'react-native-ui-lib'; //eslint-disable-line
```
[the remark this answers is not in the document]
saulve: makes sense, but I wonder how do you quickly test the changes with demo files? E.g. in this case it requires publishing `Hint` component changes to npm in order to test if it actually works. Makes it a bit of a hindrance :/

M-i-k-e-l:
Oh, another thing, the branch name should be something a little more specific (maybe `hint-migraion-to-ts`), for the next PR :)

demo/src/screens/componentScreens/HintsScreen.tsx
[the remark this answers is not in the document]
M-i-k-e-l: TBH I do not know how, but Ethan made it work (I don't think it has to do with npm, maybe it's magic :) )

src/components/hint/index.tsx
```diff
@@ -32,127 +34,169 @@ const TARGET_POSITIONS = {
...
+  bottom?: number;
+  left?: number;
+  right?: number;
+}
+
+interface HintPositionStyle extends Position {
+  alignItems?: string;
+}
+
+interface Paddings {
```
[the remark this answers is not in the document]
saulve: Would make sense, but afaik we'd still need an extra type if we wanted to reuse `NativePaddingKeyType` since it doesn't map to numbers:
```
type NumericValue<T> = {
  [Property in T]: number;
}

type Paddings = NumericValue<NativePaddingKeyType>
```
So seems like the number of types in the file would remain more or less the same 😅. Also, not all of the `NativePaddingKeyType` keys are used here. So I'd prefer keeping things stupid and simple. Unless there's already a mapped type I could reuse (but I couldn't find one)?

demo/src/screens/componentScreens/HintsScreen.tsx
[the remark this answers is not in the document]
saulve: Will move it back how it was, to keep consistency with other demo files, but I'm curious to find out how @ethanshar made it work :)

src/components/hint/index.tsx
```diff
@@ -32,127 +34,169 @@ const TARGET_POSITIONS = {
...
   static defaultProps = {
-    position: HINT_POSITIONS.BOTTOM
+    position: HintPositions.BOTTOM
   };
 
-  static positions = HINT_POSITIONS;
+  static positions = HintPositions;
+
+  private targetRef?: any;
+  private hintRef?: any;
```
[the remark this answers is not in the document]
saulve: I don't like `any` as well, but setting the ref type to `RefObject`, causes all sorts of issues when passing it to `findNodeHandle` func. Couldn't find a better alternative.

src/components/hint/index.tsx
```diff
@@ -172,7 +216,7 @@ class Hint extends BaseComponent {
   }
 
   get containerWidth() {
-    const {containerWidth} = this.getThemeProps();
+    const {containerWidth} = this.props;
     return containerWidth || Constants.screenWidth;
```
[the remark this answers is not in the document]
saulve: I think in the case of a falsy `containerWidth` value `Constants.screenWidth` will be returned anyway (because of logical `OR`)

src/components/hint/index.tsx
saulve: I think it's actually the other way around.

`ModalProps` ( in `src/components/modal/index.tsx`) is defined:
```
/**
  * allow dismissing a modal when clicking on its background
  */
onBackgroundPress?: (event: GestureResponderEvent) => void;
```
Whereas `onRequestClose` is defined:
```
/**
  * The `onRequestClose` prop allows passing a function that will be called once the modal has been dismissed.
  * _On the Android platform, this is a required function._
  */
onRequestClose?: () => void;
```

IMO it's important to keep the `onBackgroundPress` definition as is because if the user uses `onBackgroundPress` prop then `event` should always be defined (as you've described). `onRequestClose`, however does not expect a function argument, so we can override it. I think the only reason we use `onRequestClose` is because it's a required function for android (`onBackgroundPress` is what should actually dismiss the modal) . Alternatively, maybe we could even have something like `onRequestClose={this.noop}` (`noop = () => {}`)?

src/components/hint/index.tsx
```diff
@@ -397,7 +450,7 @@ class Hint extends BaseComponent {
   renderChildren() {
     const {targetFrame} = this.props;
 
-    if (!targetFrame) {
+    if (!targetFrame && React.isValidElement(this.props.children)) {
```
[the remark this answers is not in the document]
saulve: This method is used to ensure that `this.props.children` can be passed to `React.cloneElement()`.

https://stackoverflow.com/questions/42261783/how-to-assign-the-correct-typing-to-react-cloneelement-when-giving-properties-to

src/components/hint/index.tsx
```diff
@@ -32,127 +34,169 @@ const TARGET_POSITIONS = {
...
 
   focusAccessibilityOnHint = () => {
     const {message} = this.props;
-    const targetRefTag = findNodeHandle(this.targetRef);
-    const hintRefTag = findNodeHandle(this.hintRef);
-    if (targetRefTag && _.isString(message)) {
-      AccessibilityInfo.setAccessibilityFocus(targetRefTag);
-    } else if (hintRefTag) {
-      AccessibilityInfo.setAccessibilityFocus(hintRefTag);
+    if (this.targetRef && this.hintRef) {
```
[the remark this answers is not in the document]
saulve: `targetRef` and `hintRef` are undefined at the start by definition, but you're not getting errors because the type is `any`, which is not great, I know. I found a potential solution for this problem though. Will push the changes soon.

src/components/hint/index.tsx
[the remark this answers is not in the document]
M-i-k-e-l: Nice work :)

src/components/hint/index.tsx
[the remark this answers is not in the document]
M-i-k-e-l: Because it's a default value it will behave differently for `undefined` and `0`, but it's ok if you prefer not to change the logic.

src/components/hint/index.tsx
```diff
@@ -12,7 +12,8 @@ import {
   StyleProp,
   TextStyle,
   ViewStyle,
-  LayoutChangeEvent
+  LayoutChangeEvent,
+  View as ViewRN
```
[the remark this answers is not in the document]
saulve: No worries will add the change 👍🏼

saulve:
@M-i-k-e-l thanks for your time reviewing this. Is there anything else left for me to address?

src/components/hint/index.tsx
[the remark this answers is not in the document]
M-i-k-e-l: I see, my mistake :/

src/components/hint/index.tsx
M-i-k-e-l: Maybe we should take this offline.

M-i-k-e-l (commented):
Nice work, I think there are only two things left, maybe we should talk about the `onBackgroundPress`

M-i-k-e-l (approved):

PR #1290, opened 2021-05-05
FeatureHighlight  - migrate to new api, remove unsafe method

mendyEdri:
## Description
Replaced UNSAFE `componentWillReceiveProps` method with `getDerivedStateFromProps`

## Changelog
Infra: FeatureHighlight migrated to newer React api, replaced `UNSAFE` method with `getDerivedStateFromProps`


src/components/featureHighlight/index.js
```diff
@@ -163,7 +171,7 @@ class FeatureHighlight extends BaseComponent {
     AccessibilityInfo.setAccessibilityFocus(reactTag);
   }
 
-  findTargetNode(target) {
+  static findTargetNode(target) {
```
Inbal-Tish, line 174: Why static?

src/components/featureHighlight/index.js
```diff
@@ -148,11 +148,19 @@ class FeatureHighlight extends BaseComponent {
     this.setTargetPosition();
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    this.setTargetPosition(nextProps);
+  static getDerivedStateFromProps(nextProps, prevState) {
+    const target = nextProps?.getTarget?.();
+    const node = FeatureHighlight.findTargetNode(target);
+    if (node && node !== prevState?.node) {
+      return {...prevState, node};
```
Inbal-Tish, line 155: Why setting all '...prevState' and not only 'node'? What else has changed?

src/components/featureHighlight/index.js
```diff
@@ -148,11 +148,19 @@ class FeatureHighlight extends BaseComponent {
...
+    }
+    return null;
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(nextProps) {
+    if (!_.isEqual(nextProps, this.props)) {
```
Inbal-Tish, line 161: Is it really necessary to check all the props? All of them are relevant for setting the target position?

src/components/featureHighlight/index.js
mendyEdri: Because we need to use it in `static getDerivedStateFromProps`

src/components/featureHighlight/index.js
mendyEdri: `contentPosition`, it works without setting it, but any future state change will be reset in prop change, no?

src/components/featureHighlight/index.js
mendyEdri: Right, fixed

src/components/featureHighlight/index.js
Inbal-Tish: oh ok

src/components/featureHighlight/index.js
```diff
@@ -148,11 +148,19 @@ class FeatureHighlight extends BaseComponent {
     this.setTargetPosition();
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    this.setTargetPosition(nextProps);
+  static getDerivedStateFromProps(nextProps, prevState) {
+    const target = nextProps?.getTarget?.();
```
Inbal-Tish, line 152: Don't you want to check if 'getTarget' was changed before running this all method?

src/components/featureHighlight/index.js
Inbal-Tish: Once a prop changed and `getDerivedStateFromProps` is called you'll get the last state in 'prevState' parameter. Think of this function as `setState()`, you don't need to set the whole state object again, only the part you want to change as a result of a prop change. When you don't return anything from this function, the default is 'null'. Returning null doesn't reset the state, it doesn't touch it...

src/components/featureHighlight/index.js
mendyEdri: I think this is practically what happens, node is null and return false, wdyt?

src/components/featureHighlight/index.js
```diff
@@ -148,11 +148,27 @@ class FeatureHighlight extends BaseComponent {
...
+    const node = FeatureHighlight.findTargetNode(target);
+    if (node && node !== prevState?.node) {
+      return {...prevState, node};
+    }
+    return null;
   }
 
-  componentDidUpdate() {
+  shouldSetTargetPosition = (nextProps) => {
```
Inbal-Tish, line 160: Looks like 'shouldComponentUpdate'... If it's not a pure component you can just implement that method instead and the component will not be updated if it doesn't meet the condition.

src/components/featureHighlight/index.js
Inbal-Tish: Why node will return null for the target if it did not change?

mendyEdri: How do you check if `getTarget` has changed? we have access to `nextProps` and since it's a static method we don't have access to `this.props`

src/components/featureHighlight/index.js
mendyEdri: I've tried it before and now again, it's not quite working.
the position of the `FeatureHighlight` view is getting the previous data, that's why I went with this solution.
(I even tried `!_.isEqual(nextProps, this.props)`)

src/components/featureHighlight/index.js
Inbal-Tish: You can put the target on state and compare prevState.target with nextPros.target

src/components/featureHighlight/index.js
Inbal-Tish: Wierd. Ok...

src/components/featureHighlight/index.js
mendyEdri: Got it, thanks!

Inbal-Tish:
@mendyEdri Where do we stand with this PR?

mendyEdri:
> @mendyEdri Where do we stand with this PR?

Fixed what needed, you can review it :)

Inbal-Tish (approved):

PR #1291, opened 2021-05-06
Fix various typing issues

ethanshar:
## Description
Fix various typing issues

## Changelog
Fix Button iconSource type, Card.Section ImageProps to not require source and Text to support animated text style


src/components/text/index.tsx
```diff
@@ -39,6 +39,7 @@ export type TextProps = RNTextProps & TypographyModifiers & ColorsModifiers & Ma
    */
   animated?: boolean;
   textAlign?: string;
+  style?: StyleProp<TextStyle | Animated.AnimatedProps<TextStyle>>;
```
M-i-k-e-l, line 42: You need to `Omit<RNTextProps, 'style'>` in `TextProps`

src/components/card/CardSection.tsx
```diff
@@ -44,7 +44,7 @@ export type CardSectionProps = ViewProps & {
   /**
    * Other image props that will be passed to the image
    */
-  imageProps?: ImageProps;
+  imageProps?: Partial<ImageProps>;
```
M-i-k-e-l, line 47: This: `imageProps?: Omit<ImageProps, 'source' | 'style'>;` feels a bit more accurate, WDYT?

src/components/text/index.tsx
ethanshar: Yes, weirdly it didn't throw an error on that, anyway fixed it.

M-i-k-e-l: We should move more screens to tsx (found it there)

src/components/card/CardSection.tsx
M-i-k-e-l: I know, but I'm not sure it's intentional, approving

PR #1295, opened 2021-05-10
ChipsInput - clear input text when new tag added from props

nadavlac:
## Description
clear input text when new tag was added from props (same behaviour as internal add tag has)

## Changelog
I moved the input ref to be in state in order for it to be available in the getDerivedStateFromProps. Another solution would be to do it in the componentDidMount method, but I think it's good to have all the logic in one place and it will happen a bit faster because this way it happens before rendering.


nadavlac:
Question:
in this line (https://github.com/wix/react-native-ui-lib/blob/dc79399828b198f4d0416775b4233815c96ba974/src/components/chipsInput/index.js#L113)
We compare two refs, so it will always be true. Is this intentional?

src/components/chipsInput/index.js
```diff
@@ -96,27 +96,32 @@ class ChipsInput extends Component {
...
-        initialTags: nextProps.tags,
-        tags: nextProps.tags
-      };
+      nextState.initialTags = nextProps.tags;
+      nextState.tags = nextProps.tags;
     }
-    return null;
+    if (nextProps.tags.length > prevState.initialTags.length) {
+      nextState.value = ''
+      prevState.inputRef && prevState.inputRef.clear()
```
Inbal-Tish, line 121: This is not the place to clear the input. Use `componentDidUpdate`.

nadavlac: so I should clear the state's value in the derivedProps and clear the input in didMount?

src/components/chipsInput/index.js
```diff
@@ -292,7 +297,7 @@ class ChipsInput extends Component {
     return (
       <View style={styles.inputWrapper}>
         <TextField
-          ref={r => (this.input = r)}
+          ref={r => (this.state.inputRef = r)}
```
Inbal-Tish, line 300: You can use `React.createRef()` instead

src/components/chipsInput/index.js
```diff
@@ -96,27 +96,32 @@ class ChipsInput extends Component {
...
-      initialTags: props.tags
+      initialTags: props.tags,
+      inputRef: undefined
     };
   }
 
   componentDidMount() {
     if (Constants.isAndroid) {
-      const textInputHandle = ReactNative.findNodeHandle(this.input);
+      const textInputHandle = ReactNative.findNodeHandle(this.state.inputRef);
```
Inbal-Tish, line 106: The ref should not be on the state. See my comment in `getDerivedStateFromProps`.

src/components/chipsInput/index.js
nadavlac: If I move it do didMount, it can stay the was it was

src/components/chipsInput/index.js
```diff
@@ -96,27 +96,32 @@ class ChipsInput extends Component {
...
     if (nextProps.tags !== prevState.initialTags) {
-      return {
-        initialTags: nextProps.tags,
-        tags: nextProps.tags
-      };
+      nextState.initialTags = nextProps.tags;
+      nextState.tags = nextProps.tags;
     }
-    return null;
+    if (nextProps.tags.length > prevState.initialTags.length) {
```
Inbal-Tish, line 119: What if tags.length is smaller than the initialTags's?

nadavlac: I guess the text should still be there. It means that the user just removed a tag.
In out scenario we show a list based on the text, so there is not reason to reset it. I think that having a new tag is the only case

src/components/chipsInput/index.js
nadavlac: @Inbal-Tish

Inbal-Tish: In `ComponentDidUpdate` when meeting your condition.

src/components/chipsInput/index.js
Inbal-Tish: Move what to `DidMount`?

nadavlac: it's my bad! I meant `DidUpdate`

src/components/chipsInput/index.js
nadavlac: So I'll move only the ref clean to `DidUpdate`. okay?

src/components/chipsInput/index.js
Inbal-Tish: You don't need to update the ref. You create it once in the c'tor: `this.input = React.createRef();`.

src/components/chipsInput/index.js
Inbal-Tish: Sure. But I still don't understand what you're trying to solve and why clear the value will help you do that? Can you please add it to the description. Screenshots/videos will help.

src/components/chipsInput/index.js
nadavlac: @Inbal-Tish I edited the description.
Please review the new commit

PR #1301, opened 2021-05-13
Fix/ WheelPicker onChange in Android

lidord-wix:
## Description
`onMomentumScrollEnd` not being called in Android - https://github.com/facebook/react-native/issues/26661
Add a call for the `onChange()` callback manually.

## Changelog
Fix inconsistent between iOS and Android by calling the `onChange()` function manually

### DIFF
--- a/src/incubator/WheelPicker/index.tsx
+++ b/src/incubator/WheelPicker/index.tsx
@@ -123,15 +123,14 @@ const WheelPicker = React.memo(({
   };
 
   const scrollToIndex = (index: number, animated: boolean) => {
-      //@ts-ignore for some reason scrollToOffset isn't recognized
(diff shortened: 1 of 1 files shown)


ethanshar (commented):
@lidord-wix 
I didn't dive into the code yet, but I noticed an issue with this fix. 
See that now on Android, when selecting a different value using a click, it doesn't scroll to it, but immediately change the value.

lidord-wix:
> @lidord-wix
> I didn't dive into the code yet, but I noticed an issue with this fix.
> See that now on Android, when selecting a different value using a click, it doesn't scroll to it, but immediately change the value.

Ohh I see..
Solved it with setTimeout - I'm not sure it's the best solution but I couldn't find another :)

ethanshar (approved):

Issue #895, opened 2020-08-15
Support icon as component instead of only image


thebergamo:
Hi,

First of all, after getting know this lib I fall in love, so cheers guys!

Would you be open to support passing icon in those components as component instead of supporting only images.

For example when we've the Chip component we would use it like this:

```javascript
import { Chip } from 'react-native-ui-lib';
import Icon from 'react-native-vector-icons/MaterialIcons';

export default (props) => 
<Chip label="3km" icon={<Icon name="map" />} />
```

I believe it's quite standard using `react-native-vector-icons`, so we could have a more connected ecosystem.

If you're open to have it as an option into those components, I would create a PR for it :)

dhcmega:
Hi, is there any change to have this? thanks!

ethanshar:
Added to our board

lidord-wix:
Hey @thebergamo and @dhcmega,
We can add it to our `Chip` component, there are other components you would like to have this support in?

thebergamo:
I believe all components that have option to pass a image could benefit from being able to pass a component to it :)

dhcmega:
@lidord-wix hi! Out of the top of my head, I think I would use it as transparent back button for the header, regular buttons, lists and custom touchable buttons for the "home/desktop" of my app. Thanks!

richardrobberse:
@lidord-wix @ethanshar I think I'm facing a similar issue - trying to pass an icon coming from `@expo/vector-icons` to the `rightIconSource` prop of the `Picker` component. Kind of expected the render function to work no matter where the icon is coming from? Any ideas?

lidord-wix:
Hey @richardrobberse,
We're working on a new TextField component with `trailingAccessory` and `leadingAccessory` props that doing the functionality you described, this TextField will be used by our Picker component so you'll be able to use it there soon :)

mlabrum:
TabBar.Item is another candidate for this.

I've even tried using the react-native-vector-icons getImageSource to pass the image directly to it
```
<TabBar.Item key={label} label={label} onPress={onPress} icon={Icon.getImageSourceSync("account-arrow-left", 20, "red")}/>
```

lidord-wix:
@thebergamo, @dhcmega, @richardrobberse, @mlabrum,
I'm closing this issue since we added the required support to most of our main components.
If there's a specific need, please open an issue for that and we'll address it.

ConnorLanglois:
[...]

PR #1302, opened 2021-05-13
Feat/ Add custom content to Hint

lidord-wix:
## Description
Support passing custom content to the Hint component.
Related issue: #895 

## Changelog
Hint component now supports passing customContent prop.

### DIFF
--- a/demo/src/screens/componentScreens/HintsScreen.tsx
+++ b/demo/src/screens/componentScreens/HintsScreen.tsx
@@ -1,17 +1,19 @@
 import React, {Component} from 'react';
+import {Alert} from 'react-native';
 import {View, Text, Hint, Button, RadioGroup, RadioButton, Switch} from 'react-native-ui-lib'; //eslint-disable-line
 
 const settingsIcon = require('../../assets/icons/settings.png');
 
 type HintScreenProps = {};
 type HintScreenState = {
-  showHint: boolean,
-  useShortMessage: boolean,
-  showBottomHint: boolean,
-  showIcon: boolean,
-  targetPosition: string,
-  useTargetFrame?: boolean,
-  useSideTip?: boolean
+  showHint: boolean;
+  useShortMessage: boolean;
+  showBottomHint: boolean;
+  showIcon: boolean;
+  targetPosition: string;
+  useTargetFrame?: boolean;
+  useSideTip?: boolean;
+  showCustomContent?: boolean;
 };
 
 export default class HintsScreen extends Component<HintScreenProps, HintScreenState> {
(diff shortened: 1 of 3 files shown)


src/components/hint/index.tsx
```diff
@@ -435,8 +439,9 @@ class Hint extends Component<HintProps, HintState> {
             style={[styles.hint, color && {backgroundColor: color}, !_.isUndefined(borderRadius) && {borderRadius}]}
             ref={this.setHintRef}
           >
-            {icon && <Image source={icon} style={[styles.icon, iconStyle]}/>}
-            <Text style={[styles.hintMessage, messageStyle]}>{message}</Text>
+            {customContent}
```
M-i-k-e-l, line 442: What do you think about adding a `renderContent` method?

demo/src/screens/componentScreens/HintsScreen.tsx
```diff
@@ -85,6 +89,17 @@ export default class HintsScreen extends Component<HintScreenProps, HintScreenSt
             // borderRadius={BorderRadiuses.br40}
             // edgeMargins={30}
             // onBackgroundPress={() => this.setState({showHint: !showHint})}
+            customContent={
+              showCustomContent ? (
+                <Text text70 white>
+                  Click
+                  <Text onPress={() => Alert.alert('custom content :)')} text70BO violet50>
```
M-i-k-e-l, line 96: Not sure why this is not working on Android (does not look like it's [this](https://github.com/facebook/react-native/issues/22747)), consider changing the example to a button so this works properly.

M-i-k-e-l: Removing the code you've mentioned (below) makes me **very** uncomfortable, adding the `collapsable={false}` usually comes in to fix bugs, the `zIndex` most likely fixes a bug as well, we can look it up (in the tickets or Slack) if you'd like.
And the comment **mentions** it fixes a bug in Android.
```
collapsable={false}
---
zIndex: 100,
// This is a hack to make hint render correctly on Android
borderWidth: 1,
borderColor: 'transparent'
```

lidord-wix: Right, but it looks like a little bit old code and Ethan said it might be solved by `react-native` since then. 
Anyway, I did check it in Android and in iOS, and we can also make a version for QA before releasing it.
What do you think?

M-i-k-e-l: OK, let's create a version and test it.

lidord-wix: Tested by Yulia and it looks as expected :)

M-i-k-e-l: Cool 👍

M-i-k-e-l (approved):

PR #1311, opened 2021-05-18
Feat/ update gradient for TOP and BOTTOM overlays

lidord-wix:
## Description
Update gradient for TOP and BOTTOM Image overlays
Ticket 1743

## Changelog
Update gradient for TOP and BOTTOM Image overlays


M-i-k-e-l:
It does not look like there's a difference between the old and the new gradient, add the text ("Show Overlay Content") and you'll see it's not similar to the design.

lidord-wix:
> It does not look like there's a difference between the old and the new gradient, add the text ("Show Overlay Content") and you'll see it's not similar to the design.

There's a difference, you can see it when you toggle from vertical to top (or to bottom).
And for me, it looks like in the guidelines - you can find here 2 examples where I changed the text: 
image
image

M-i-k-e-l (approved):

PR #1312, opened 2021-05-18
Add hint onPress support

mendyEdri:
## Description
Added `onPress` function to Hint component props.

## Changelog
Press on Hint component can be detected now by passing `onPress` function.


src/components/hint/index.tsx
```diff
@@ -436,25 +441,27 @@ class Hint extends Component<HintProps, HintState> {
...
-            this.getHintPadding(),
-            this.getHintAnimatedStyle()
-          ]}
-          pointerEvents="box-none"
-          testID={testID}
-        >
-          {this.renderHintTip()}
-          {this.renderContent()}
-        </View>
+        <TouchableOpacity activeOpacity={opacity} onPress={onPress}>
```
ethanshar, line 448: Instead of wrapping with another component consider swapping between the wrapping container, we do it often in such cases.. 
For instance,
```

<Container .....>
```

ethanshar (commented):
@mendyEdri 
I just checked the new feature on Android and it doesn't work :( 
Can you take a look

ethanshar:
> @mendyEdri
> I just checked the new feature on Android and it doesn't work :(
> Can you take a look

Ignore that, I didn't pull!

ethanshar (approved):

Issue #1156, opened 2021-01-27
[TabBarItem / TabController.TabBarItem] Option to disable tint, set icon size


bneigher:
### Which component?
TabBarItem / TabController.TabBarItem

### What is the new feature?
Two very simple changes to support my (and probs common use case of using color icons as the tab icon)

<img width="401" alt="before" src="https://user-images.githubusercontent.com/4627728/105915313-66fd7180-5fe4-11eb-9f42-98444b1038d0.png">
<img width="401" alt="after" src="https://user-images.githubusercontent.com/4627728/105949229-37208f00-6021-11eb-950a-84f4d792ebc0.png">

### Code snippet
https://github.com/wix/react-native-ui-lib/blob/master/src/components/tabController/TabBarItem.tsx#L228-L243

disableIconTint prop (or item setting)
```javascript
getIconStyle() {
    const {index, currentPage, iconColor, selectedIconColor, labelColor, selectedLabelColor, disableIconTint, ignore} = this.props;

    /* new stuff here */
    if (disableIconTint) {
       return {}
    }
    /* end new stuff */

    const activeColor = selectedIconColor || selectedLabelColor || DEFAULT_SELECTED_LABEL_COLOR;
    const inactiveColor = iconColor || labelColor || DEFAULT_LABEL_COLOR;

    const tintColor = cond(eq(currentPage, index),
      // TODO: using processColor here broke functionality,
      // not using it seem to not be very performant
      activeColor,
      ignore ? activeColor : inactiveColor);

    return {
      tintColor
    };
  }
```

```javascript
render() {
    const {
      label,
      icon,
      badge,
      state,
      uppercase,
      activeOpacity,
      activeBackgroundColor,
      testID,
      disableIconTint = false, <-- new (bool) default: false
      iconStyle = {} <-- new (object) default: {}
    } = this.props;
    return <TouchableOpacity ref={this.itemRef} pressState={state} style={[styles.tabItem, this.getItemStyle()]} onLayout={this.onLayout} feedbackColor={activeBackgroundColor} activeOpacity={activeOpacity} onPress={this.onPress} testID={testID}>
        {icon && <Reanimated.Image source={icon} style={[iconStyle, !_.isUndefined(label) && styles.tabItemIconWithLabel, !disableIconTint && this.getIconStyle()]} />}
        {!_.isEmpty(label) && <Reanimated.Text style={[styles.tabItemLabel, this.getLabelStyle()]}>
[...]

PR #1314, opened 2021-05-20
added `disableIconTintColor` prop to TabController.TabBarItem

nedjulius:
## Description
Added prop to disable icon's tint color on `TabController.TabBarItem` as per #1156

## Changelog
- added `disableIconTintColor` prop to TabController.TabBarItem

### DIFF
--- a/src/components/tabController/TabBarItem.tsx
+++ b/src/components/tabController/TabBarItem.tsx
@@ -92,6 +92,10 @@ export interface TabControllerItemProps {
    * Used as a testing identifier
    */
   testID?: string;
+  /**
+   * disables icon's tint color
+   */
+  disableIconTintColor?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


src/components/tabController/TabBarItem.tsx
```diff
@@ -226,7 +230,20 @@ export default class TabBarItem extends PureComponent<Props> {
...
+      iconColor,
+      selectedIconColor,
+      labelColor,
+      selectedLabelColor,
+      ignore,
+      disableIconTintColor
+    } = this.props;
+
+    if (disableIconTintColor) {
+      return {};
```
ethanshar, line 245: suggest return undefined here and not an empty object. 
new object with a new ref might trigger a render, while undefined won't

ethanshar (approved):

PR #1315, opened 2021-05-20
Add testID to wheel picker list

ethanshar:
## Description
Add testID to wheel picker list

## Changelog
Add testID to wheel picker list (`${testID}.list`)


lidord-wix (commented):
I think we should modify the testID of the items to be:
`testID={'${testID}.item'}`

ethanshar:
> I think we should modify the testID of the items to be:
> `testID={'${testID}.item'}`

Done ✅

lidord-wix (approved):

PR #1320, opened 2021-05-24
Support checking for required props in a component

ethanshar:
## Description
Add support to our `component-prop-deprecation` rule to verify required props

## Changelog
Add support to our `component-prop-deprecation` rule to verify required props

### DIFF
--- a/eslint-rules/lib/rules/component-prop-deprecation.js
+++ b/eslint-rules/lib/rules/component-prop-deprecation.js
@@ -30,6 +30,9 @@ const MAP_SCHEMA = {
           message: {
             type: 'string'
           },
+          isRequired: {
+            type: 'boolean'
+          },
           fix: {
             type: 'object',
             required: ['propName'],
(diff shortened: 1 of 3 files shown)


lidord-wix (commented):
The code looks great, but honestly, I think it should be a different rule since it's not really a prop deprecation.
Ideally, the component-prop-deprecation.json (in private) shouldn't contain anything, but the JSON file of this rule should contain props permanently..

ethanshar:
> The code looks great, but honestly, I think it should be a different rule since it's not really a prop deprecation.
> Ideally, the component-prop-deprecation.json (in private) shouldn't contain anything, but the JSON file of this rule should contain props permanently..

After what we discussed you still think we need to have this rule separated?

lidord-wix:
> > The code looks great, but honestly, I think it should be a different rule since it's not really a prop deprecation.
> > Ideally, the component-prop-deprecation.json (in private) shouldn't contain anything, but the JSON file of this rule should contain props permanently..
> 
> After what we discussed you still think we need to have this rule separated?

Nope, approving :)

lidord-wix (approved):

PR #1327, opened 2021-06-01
Feat/segmented control new style

lidord-wix:
## Description
Update SegmentedControl style
Add a fullWidth mode

## Changelog
Update SegmentedControl style and add a fullWidth mode

### DIFF
--- a/src/components/segmentedControl/index.tsx
+++ b/src/components/segmentedControl/index.tsx
@@ -11,6 +11,7 @@ const {interpolate: _interpolate, interpolateNode} = Reanimated;
 const interpolate = interpolateNode || _interpolate;
 const Easing = EasingNode || _Easing;
 const BORDER_WIDTH = 1;
+const SEGMENT_HEIGHT = 30;
 
 export type SegmentedControlItemProps = SegmentProps;
 export type SegmentedControlProps = {
@@ -79,12 +80,12 @@ const SegmentedControl = (props: SegmentedControlProps) => {
     initialIndex = 0,
     containerStyle,
     segments,
-    activeColor = Colors.primary,
+    activeColor = Colors.white,
     borderRadius = BorderRadiuses.br100,
     backgroundColor = Colors.grey80,
-    activeBackgroundColor = Colors.white,
+    activeBackgroundColor = Colors.grey10,
     inactiveColor = Colors.grey20,
-    outlineColor = activeColor,
+    outlineColor = Colors.grey20,
     outlineWidth = BORDER_WIDTH,
     fullWidth
   } = props;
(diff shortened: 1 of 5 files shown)


src/components/segmentedControl/index.tsx
```diff
@@ -144,35 +150,38 @@ const SegmentedControl = (props: SegmentedControlProps) => {
...
-    borderColor: Colors.grey60
+    borderColor: Colors.grey60,
+    paddingVertical: Spacings.s2,
+    borderWidth: BORDER_WIDTH
   },
   selectedSegment: {
-    height: Spacings.s7 - 2 * BORDER_WIDTH,
-    position: 'absolute',
-    backgroundColor: Colors.white
+    height: SEGMENT_HEIGHT,
```
ethanshar, line 183: Seeing it again, I don't think that passing a fixed height is correct. 
It limits the component height and in case the content is bigger (bigger font for example) the component doesn't grow nicely.. the font will be cut

src/components/segmentedControl/index.tsx
```diff
@@ -62,6 +63,10 @@ export type SegmentedControlProps = {
    * Additional spacing styles for the container
    */
   containerStyle?: StyleProp<ViewStyle>;
+  /**
+   * Whether the segmentedControl should be in a full width mode 
+   */
+  fullWidth?: boolean,
```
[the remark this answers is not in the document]
lidord-wix: Fixed :)

src/components/segmentedControl/index.tsx
lidord-wix: Fixed :)

khunghang16:
How I can custom style label in Segment? Tks

src/components/segmentedControl/index.tsx
```diff
@@ -150,29 +153,30 @@ const SegmentedControl = (props: SegmentedControlProps) => {
...
-      <Reanimated.View
-        style={[
-          styles.selectedSegment,
-          animatedStyle,
-          {borderColor: outlineColor, borderRadius, backgroundColor: activeBackgroundColor, borderWidth: outlineWidth}
-        ]}
-      />
-      {renderSegments()}
+    <View>
+      <View row center style={[styles.container, containerStyle, {borderRadius, backgroundColor}]}>
```
ethanshar, line 157: I think it's important to pass the `containerStyle` to the actual container view. 
This way the user can pass things like `alignSelf: flex-start` for alignment and have overall have better control of the container style. 
Maybe we can have a `style` prop for the segment control instead. 
WDYT? 

Usually, I would have done a migration for it, but I think since it's a new component, it's safe enough to make the change.

src/components/segmentedControl/index.tsx
```diff
@@ -75,17 +76,18 @@ const SegmentedControl = (props: SegmentedControlProps) => {
...
-    activeColor = Colors.primary,
+    activeColor = Colors.white,
     borderRadius = BorderRadiuses.br100,
     backgroundColor = Colors.grey80,
-    activeBackgroundColor = Colors.white,
+    activeBackgroundColor = Colors.grey10,
     inactiveColor = Colors.grey20,
-    outlineColor = activeColor,
+    outlineColor = Colors.grey20,
     outlineWidth = BORDER_WIDTH
```
ethanshar, line 85: As discussed, consider making UI changes in our private config.

ethanshar (approved):

lidord-wix:
> How I can custom style label in Segment? Tks

@khunghang16 
We don't support it at the moment, please open an issue if necessary :)

Issue #734, opened 2020-04-03
[Slider] slider thumb not exactly centered


bneigher:
It looks like the thumb for the slider isn't exactly positioned correctly, seems like the rightmost part of the thumb component is seeked to the correct position. Examples in the demo app and my own are attached.  Reproduce = set value to 50 on a 0 to 100 slider

In each screenshot, the lines were drawn and then copied and pasted to ensure they are equal in length

<img width="437" alt="Screen Shot 2020-04-03 at 3 11 42 PM" src="https://user-images.githubusercontent.com/4627728/78409056-9ce8e980-75bd-11ea-9ffd-a14a5a79437b.png">

<img width="430" alt="Screen Shot 2020-04-03 at 3 05 59 PM" src="https://user-images.githubusercontent.com/4627728/78409062-9eb2ad00-75bd-11ea-9862-d79d2f1edcaf.png">

I'm using version 4.2.0
React Native 0.62.0
iOS simulator

PR #1334, opened 2021-06-02
Fix/ Slider's thumb position

lidord-wix:
## Description
Fix slider's thumb position
Fix https://github.com/wix/react-native-ui-lib/issues/734

## Changelog
Fix slider's thumb position

### DIFF
--- a/src/components/slider/index.js
+++ b/src/components/slider/index.js
@@ -306,7 +306,7 @@ export default class Slider extends PureBaseComponent {
     const relativeValue = minimumValue - v;
     const value = minimumValue < 0 ? Math.abs(relativeValue) : v - minimumValue; // for negatives
     const ratio = value / range;
-    const x = ratio * (this.state.trackSize.width - this.initialThumbSize.width / 2);
+    const x = ratio * this.state.trackSize.width;
     return x;
   }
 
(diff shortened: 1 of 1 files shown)


src/components/slider/index.js
```diff
@@ -250,7 +250,7 @@ export default class Slider extends PureBaseComponent {
     }
 
     if (this.minTrack) {
-      this._minTrackStyles.width = x;
+      this._minTrackStyles.width = Math.min(this.state.trackSize.width, x);
```
mendyEdri, line 253: I think `x` will forever be less than `width` (since the width it's the track width and the `x` is the current position of the thumb).
anyway, is this needed for some reason?

mendyEdri (commented):
Small comment :)

src/components/slider/index.js
lidord-wix: There is an issue when you scroll from the track itself, see the attached video: 

https://user-images.githubusercontent.com/66782556/120628489-6a5ea900-c46d-11eb-8b5f-ec65079b164f.mov

mendyEdri: Got it, great catch :)

mendyEdri (approved):

PR #1338, opened 2021-06-03
Typescript/feature highlight

mendyEdri:
## Description
Migrate `FeatureHighlight` component into typescript.

## Changelog
Migrate `FeatureHighlight` component into typescript.


src/components/featureHighlight/index.tsx
```diff
@@ -35,96 +151,15 @@ const contentViewHeight = Constants.isAndroid ? 268 : 282;
...
-    /**
-     * The padding of the highlight frame around the highlighted element's frame (only when passing ref in 'getTarget')
-     */
-    innerPadding: PropTypes.number,
-    /**
-     * PageControl component's props
-     */
-    pageControlProps: PropTypes.shape(PageControl.propTypes)
-  };
+  static propTypes = {};
```
Inbal-Tish, line 156: Why did you leave the `propTypes` static object?

src/components/featureHighlight/index.tsx
```diff
@@ -298,7 +332,7 @@ class FeatureHighlight extends BaseComponent {
               titleStyle
             ]}
             numberOfLines={titleNumberOfLines}
-            pointerEvents="none"
+            pointerEvents
```
Inbal-Tish, line 335: I don't understand the meaning of this change. `pointerEvets` prop is not a boolean so why did you drop the value ("none")?

src/components/featureHighlight/index.tsx
```diff
@@ -219,8 +252,8 @@ class FeatureHighlight extends BaseComponent {
   }
 
   getContentPosition() {
-    const {highlightFrame, minimumRectSize, innerPadding} = this.props;
-    const {top, height} = this.targetPosition;
+    const {highlightFrame, minimumRectSize = {height: 0}, innerPadding = 0} = this.props;
```
Inbal-Tish, line 255: `minimumRectSize` is of type `ReactSize`, is setting only the height as default enough?

mendyEdri: Yes, since it's only used the height in line 262

src/components/featureHighlight/index.tsx
mendyEdri: It becomes boolean from out modifiers, and some how react-native haven't export this prop with a type, 
so I changed it back and added `@ts-expect-error`

src/components/featureHighlight/index.tsx
mendyEdri: removed

src/components/featureHighlight/index.tsx
Inbal-Tish: Yes, I know. I just thought it might give you a type error...

Inbal-Tish (approved):

Inbal-Tish:
@mendyEdri Resolve conflicts and merge please.

PR #1339, opened 2021-06-03
Typescript/stack aggregator

mendyEdri:
## Description
Migrate `StackAggregator` component into typescript.

## Changelog
Migrate `StackAggregator` component into typescript.


src/components/stackAggregator/index.tsx
```diff
@@ -1,107 +1,115 @@
...
     /**
      * A callback for item press
      */
-    onItemPress: PropTypes.func,
+    onItemPress?: (index: number) => void;
     /**
      * A callback for collapse state will change (value is future collapsed state)
      */
-    onCollapseWillChange: PropTypes.func,
+    onCollapseWillChange?: (change: boolean) => void;
```
Inbal-Tish, line 50: 'changed' instead of 'change' (as 'change' refers to the change itself and not its state)

src/components/stackAggregator/index.tsx
```diff
@@ -1,107 +1,115 @@
...
     /**
-    * A callback for collapse state change (value is collapsed state)
-    */
-    onCollapseChanged: PropTypes.func,
+     * A callback for collapse state change (value is collapsed state)
+     */
+    onCollapseChanged?: (change: boolean) => void;
```
Inbal-Tish, line 54: same here

src/components/stackAggregator/index.tsx
```diff
@@ -1,107 +1,115 @@
...
+
+/**
+ * @description: Stack aggregator component
+ * @modifiers: margin, padding
+ * @example: https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/StackAggregatorScreen.js
+ */
+class StackAggregator extends PureComponent<StackAggregatorProps, State> {
+  static displayName = 'StackAggregator';
+
+  animatedScale: any;
```
Inbal-Tish, line 75: why type 'any' and not 'Animated.Value' on all the animated variables?

Inbal-Tish (approved):

PR #1345, opened 2021-06-06
Feat/swipe to dismiss

M-i-k-e-l:
## Description
Add SwipeToDismissView - component that allows for easy swipe to dismiss (locks the swipe direction once a swipe has started).
We have a few components for swiping, not sure about the naming, but I think this makes a good component that simplifies the lower level APIs.

## Changelog
Add SwipeToDismissView - component that allows for easy swipe to dismiss


ethanshar:
@M-i-k-e-l 
We have PanDismissibleView component and now SwipeToDismissView component. 
Also we have PanGestureView, PanListenerView and PanResponderView. 

It's getting messy, don't you think?

ethanshar:
Maybe we should consider keeping some components internal.. like: PanGestureView, PanListenerView and PanResponderView

Regarding this new component, why is it different than `PanDismissibleView`? 
Name wise it sound the same..

M-i-k-e-l:
I'm not sure, perhaps we should bind them under something to remove some of the exports (like we did in `Keyboard`).

1. You do not have to use the lower-level components.
2. It locks the drag\swipe direction once a drag has started.
3. Fixes some "jumps" when dragging in two directions (up + left etc).

ethanshar:
1. I prefer not to have 5 components that sound the same, in the docs it's very confusing. Let's remove them from docs (using IGNORE) and make them internal
2. Sounds like a feature, why not add it the PanDismissibleView
3. Sound like a bug, why not fix it in the PanDismissibleView

M-i-k-e-l:
1. I feel that they can be used for more complex implementations (like we're doing in the `DialogDismissibleView`), but even if we want to remove them, shouldn't we do that in a major version?
2. `SwipeToDismissView` is using `PanDismissibleView`, but it's only the coupled version (everything is dragable).
3. This bug happens only when locking the drag.

I've looked at it and I think that merging them would either lose features or will require more props than the current design.
Do you want to schedule a meeting about this?

PR #1346, opened 2021-06-06
Support passing custom onPress to Picker.Item

ethanshar:
## Description
We have a requirement to have a custom onPress handling for picker item. 
So first, I noticed we don't even trigger the item's `onPress` prop - so I fixed that. 
Also, I excluded the custom onPress from the disabled state according to the new requirement

related ticket WOAUILIB-1790

## Changelog
Support passing custom `onPress` callback for Picker.Item


src/components/picker/PickerPresenter.js
```diff
@@ -35,3 +35,8 @@ export function getItemLabel(label, value, getItemLabel) {
 export function shouldFilterOut(searchValue, itemLabel) {
   return !_.includes(_.lowerCase(itemLabel), _.lowerCase(searchValue));
 }
+
+export function shouldDisableItem(itemProps, isSelected, selectionLimit, selectedCounter) {
+  const isUnselectable = !isSelected && selectionLimit && selectionLimit === selectedCounter;
+  return Boolean((itemProps?.disabled && !itemProps?.onPress) || isUnselectable);
```
M-i-k-e-l, line 41: Is it intentional that the item does not have the disabled look even when the user sent the `disabled` props? - I don't think that the `onPress` should override that, it should only not close the `Picker`, or am I wrong?

ethanshar: We don't apply the disabled UI, just the functionality. 
And we needed a way to disable an item but also provide a custom functionality to its press. 
So this is how I chose to do it. 

If the user just pass the disabled prop (the more common case) the item won't be clickable
If the user needs a custom onPress they can pass it and do what they want - the users has a full control when they can pass the onPress themselves.

M-i-k-e-l: I still feel like the default UI should be disabled, and the user will be able to change that if that's he's choice. I think that this will feel like a bug if left this way.

I understand a user wants to have the option to press the item when it is disabled, but it is still disabled (at list as a PickerItem) and thus should be shown as such.

PR #1355, opened 2021-06-09
added ts-ignore for file that is currently js and can't be imported i…

mendyEdri:
## Description
Fixed lint issue importing js file from ts-file

## Changelog
Fix eslint import error, importing js file from ts file, added `// @ts-expect-error`


mendyEdri:
already fixed

PR #1357, opened 2021-06-09
DateTimePicker - Implement themeVariant prop

boris-etingof-flarehr:
## Description
`DateTimePicker` renders white on white text when Dark theme is enabled on iOS device. `react-native-datetimepicker` has an override prop for this called `themeVariant`. This change is to Implement this prop in `react-native-ui-lib` `DateTimePicker`.

## Changelog
Add `themeVariant` prop to `DateTimePicker`.

### DIFF
--- a/src/components/dateTimePicker/index.js
+++ b/src/components/dateTimePicker/index.js
@@ -19,6 +19,13 @@ const MODES = {
   TIME: 'time'
 };
 
+
+const THEME_VARIANTS = {
+  LIGHT: 'light',
+  DARK: 'dark'
+};
+
+
 /*eslint-disable*/
 /**
  * @description: Date and Time Picker Component that wraps RNDateTimePicker for date and time modes.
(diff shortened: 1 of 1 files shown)


src/components/dateTimePicker/index.js
```diff
@@ -98,7 +105,11 @@ class DateTimePicker extends Component {
     /**
      * Render custom input
      */
-    renderInput: PropTypes.elementType
+    renderInput: PropTypes.elementType,
+    /**
+     * Override system theme variant (dark or light mode) used by the date picker.
+     */
+    themeVariant: PropTypes.oneOf(Object.values(THEME_VARIANTS))
```
VivekRajagopal, line 112: This is a required prop? Can it be an optional one?

boris-etingof-flarehr: > This is a required prop? Can it be an optional one?

It is optional. Required would be `themeVariant: PropTypes.oneOf(Object.values(THEME_VARIANTS)).isRequired`

VivekRajagopal (approved):
LGTM

ethanshar (approved):

PR #1363, opened 2021-06-14
fix hint animation

mendyEdri:
## Description
Fix missing Hint bubble fade-out animation

## Changelog
Added missing hide animation for `Hint` bubble

### DIFF
--- a/src/components/hint/index.tsx
+++ b/src/components/hint/index.tsx
@@ -146,7 +146,7 @@ export interface HintProps {
 interface HintState {
   targetLayout?: HintTargetFrame;
   targetLayoutInWindow?: HintTargetFrame;
-  animationEnded: boolean;
+  hintUnmounted: boolean;
 }
 
 /**
@@ -171,7 +171,7 @@ class Hint extends Component<HintProps, HintState> {
   state = {
     targetLayoutInWindow: undefined,
     targetLayout: this.props.targetFrame,
-    animationEnded: false
+    hintUnmounted: false
   };
 
   visibleAnimated = new Animated.Value(Number(!!this.props.visible));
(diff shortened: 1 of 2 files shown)


ethanshar (commented):
@mendyEdri 
Android throws an error

src/components/hint/index.tsx
```diff
@@ -496,11 +496,7 @@ class Hint extends Component<HintProps, HintState> {
   }
 
   render() {
-    const {visible, onBackgroundPress, testID} = this.props;
-
-    if (!visible) {
-      return this.props.children;
-    }
```
ethanshar, line 503: This is intention and not a bug (: 
When the Hint is not visible we want to return the original item as it is.

ethanshar (commented):
@mendyEdri 
Android doesn't throw an error now, but also doesn't work. 
See video

Also, check my comment, I don't think that's the right fix (: 

https://user-images.githubusercontent.com/1780255/122710224-88e3e300-d268-11eb-876a-ae62a6434cce.mov

src/components/hint/index.tsx
```diff
@@ -496,9 +512,9 @@ class Hint extends Component<HintProps, HintState> {
   }
 
   render() {
-    const {visible, onBackgroundPress, testID} = this.props;
-
-    if (!visible) {
+    const {onBackgroundPress, testID} = this.props;
+    if (!this.props.visible && this.state.animationEnded) {
+      console.log('removed');
```
ethanshar, line 517: You forgot a console.log (:

src/components/hint/index.tsx
```diff
@@ -165,24 +166,38 @@ class Hint extends Component<HintProps, HintState> {
...
+    }).start(this.toggleAnimationEndedToRemoveHint);
+  };
+
+  toggleAnimationEndedToRemoveHint = () => {
+    this.setState({animationEnded: true}, () => {
+      setTimeout(() => {
+        this.setState({animationEnded: false});
+      }, this.animationDuration);
+    });
+  };
```
ethanshar, line 199: This forces us to trigger two setState calls every time we trigger the animation. twice on enter twice on exit. 
Maybe we can reduce that.. 
If you think about it, while the Hint is visible it doesn't matter really what's the value of `animationEnded` state. 

What if instead of keeping `animationEnded` you'll keep something else, like `actuallyVisible`, `hasUnmounted` (or anything with a better name (: )
Then you'll update it once on enter and once on exit...

mendyEdri: Right, fixed

ethanshar (approved):

PR #1375, opened 2021-06-24
Fix/segmentedControl performance

lidord-wix:
## Description
Fix segmentedControl performance
Jira WOAUILIB-1853

## Changelog
Fix segmentedControl performance

### DIFF
--- a/demo/src/screens/componentScreens/SegmentedControlScreen.tsx
+++ b/demo/src/screens/componentScreens/SegmentedControlScreen.tsx
@@ -54,12 +54,12 @@ const SegmentedControlScreen = () => {
         </View>
         <SegmentedControl
           containerStyle={styles.container}
-          onChangeIndex={(index: number) => onChangeIndex('second', index)}
+          onChangeIndex={(index: number) => onChangeIndex('fifth', index)}
           segments={segments.fifth}
         />
         <SegmentedControl
           containerStyle={styles.container}
-          onChangeIndex={(index: number) => onChangeIndex('second', index)}
+          onChangeIndex={(index: number) => onChangeIndex('sixth', index)}
           segments={segments.sixth}
         />
       </View>
--- a/demo/src/screens/componentScreens/SegmentedControlScreen.tsx
+++ b/demo/src/screens/componentScreens/SegmentedControlScreen.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useCallback} from 'react';
 import {StyleSheet} from 'react-native';
 import {Text, View, Colors, SegmentedControl, Assets, Spacings, BorderRadiuses} from 'react-native-ui-lib';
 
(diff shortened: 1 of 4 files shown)


src/components/segmentedControl/index.tsx
```diff
@@ -110,7 +110,7 @@ const SegmentedControl = (props: SegmentedControlProps) => {
       updateSelectedSegment(index);
     }
   },
-  [onChangeIndex, selectedSegment, updateSelectedSegment]);
+  [onChangeIndex, updateSelectedSegment]);
```
ethanshar, line 113: Are you sure that's correct? 
You are using selectedSegment in this callback, removing it can lead to bugs

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
```diff
@@ -54,12 +54,12 @@ const SegmentedControlScreen = () => {
         </View>
         <SegmentedControl
           containerStyle={styles.container}
-          onChangeIndex={(index: number) => onChangeIndex('second', index)}
+          onChangeIndex={(index: number) => onChangeIndex('fifth', index)}
```
ethanshar, line 57: Sometimes performance issue can be caused by the example screen. 
Consider not passing and inline anonymous callback

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
```diff
@@ -54,12 +54,12 @@ const SegmentedControlScreen = () => {
...
           segments={segments.fifth}
         />
         <SegmentedControl
           containerStyle={styles.container}
-          onChangeIndex={(index: number) => onChangeIndex('second', index)}
+          onChangeIndex={(index: number) => onChangeIndex('sixth', index)}
```
ethanshar, line 62: Same here..

src/components/segmentedControl/index.tsx
lidord-wix: You are right, but actually, I think the condition is not required at all.. 
Fixed :)

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
lidord-wix: I agree, but in this specific case, the performance issues are probably from the component itself since the complaint came from Dine devs.. 
Anyway, I removed the inline callbacks from the screen :)

src/components/segmentedControl/index.tsx
ethanshar: I think the condition is required (: 
It meant to prevent calling the onChangeIndex callback in case the user press twice at the same segment. 
Now when pressing twice on the same index I still get this callback 😬

lidord-wix: Actually, IMO the user should handle this (maybe the prop name is misleading and `onSegmentPress` is better).
Because the user might want a callback for any press and we should support it.
Let me know what you think, 'cause I'm not sure what is better :)

ethanshar: We can add a separate prop for onPress though I'm not sure it's needed. 
If you think about it, product wise the user wants to be notified on index change, that's the purpose of the component, to change between different segments.
Our goal is to provide an API that is useful for the user

lidord-wix:
@ethanshar 
I've added a setTimeout for the onChangeIndex and now the animation looks much better.. 
We should migrate it to reanimated 2 once we'll assure it improves performance :)

src/components/segmentedControl/index.tsx
```diff
@@ -106,8 +106,8 @@ const SegmentedControl = (props: SegmentedControlProps) => {
 
   const onSegmentPress = useCallback((index: number) => {
     if (selectedSegment !== index) {
-      onChangeIndex?.(index);
       updateSelectedSegment(index);
+      setTimeout(() => onChangeIndex?.(index), 400);
```
ethanshar, line 110: Other than delaying the callback, I'd consider also to throttle the calls. 
For instance, if the user press quickly on multiple different segments, maybe we can throttles all of these calls and trigger only the last one. 

Where are you checking the performance of this component? It's mostly noticeable in our SectionWheelPicker example screen. 

Another thing worth checking (can be in another PR) is to move to reanimated2, other than simplifying the code it can also help with performance.

lidord-wix: I tried to wrap the callback with lodash throttle but couldn't see any change.. also, I couldn't see the throttle effect in other places we use it (PickerModal for example), let me know if it's only me

I'm checking it in the SectionWheelPickerScreen as well :)

I have a draft PR on moving this component to reanimated2, I'll publish it after this PR :)

ethanshar (approved):

PR #1380, opened 2021-07-01
Reimplement Incubator.TouchableOpacity with reanimatedv2

ethanshar:
## Description
Reimplement Incubator.TouchableOpacity with Reanimated v2
The component is currently exported as `Incubator.TouchableOpacity2`
eventually, we'll replace it with `Incubator.TouchableOpacity` (without a migration, since the API is the same)

## Changelog
Reimplement Incubator.TouchableOpacity with Reanimated v2 as Incubator.TouchableOpacity2

### DIFF
--- a/src/incubator/TouchableOpacity2.tsx
+++ b/src/incubator/TouchableOpacity2.tsx
@@ -78,6 +78,9 @@ function TouchableOpacity(props: Props) {
   const {borderRadius, paddings, margins, alignments, flexStyle} = modifiers;
 
   const isActive = useSharedValue(0);
+  /* This flag is for fixing an issue with long press triggering twice
+  TODO: Consider revisiting this issue to see if it still occurs */
+  const isLongPressed = useSharedValue(false);
 
   const backgroundColor = useMemo(() => {
     return props.backgroundColor || modifiers.backgroundColor;
@@ -98,13 +101,22 @@ function TouchableOpacity(props: Props) {
     onEnd: () => {
       isActive.value = withTiming(0, {duration: 200});
       runOnJS(onPress)();
+    },
+    onFail: () => {
+      isActive.value = withTiming(0, {duration: 200});
     }
   });
 
   const longPressGestureHandler = useAnimatedGestureHandler({
-    onEnd: () => {
-      isActive.value = withTiming(0, {duration: 200});
-      runOnJS(onLongPress)();
+    onActive: () => {
+      if (!isLongPressed.value) {
+        isLongPressed.value = true;
+        isActive.value = withTiming(0, {duration: 200});
+        runOnJS(onLongPress)();
+      }
+    },
+    onFinish: () => {
+      isLongPressed.value = false;
     }
   });
 
@@ -129,7 +141,7 @@ function TouchableOpacity(props: Props) {
     >
       <Reanimated.View>
         {/* @ts-expect-error */}
-        <LongPressGestureHandler onGestureEvent={longPressGestureHandler}>
+        <LongPressGestureHandler onGestureEvent={longPressGestureHandler} shouldCancelWhenOutside>
           <Reanimated.View
             {...others}
             ref={forwardedRef}
--- a/src/incubator/TouchableOpacity2.tsx
+++ b/src/incubator/TouchableOpacity2.tsx
@@ -50,6 +50,10 @@ export type TouchableOpacityProps = {
    * Pass custom style
    */
   style?: ViewProps['style'];
+  /**
+   * Custom value of any type to pass on to TouchableOpacity and receive back in onPress callback
+   */
+  customValue?: any;
   onLayout?: (event: LayoutChangeEvent) => void;
   testID?: string;
 };
@@ -86,11 +90,11 @@ function TouchableOpacity(props: Props) {
 
   const onPress = useCallback(() => {
     props.onPress?.(props);
-  }, [props]);
+  }, [props.onPress, props.customValue]);
 
   const onLongPress = useCallback(() => {
     props.onLongPress?.(props);
-  }, [props]);
+  }, [props.onPress, props.customValue]);
 
   const toggleActive = (value: number) => {
     'worklet';
--- a/src/incubator/TouchableOpacity2.tsx
+++ b/src/incubator/TouchableOpacity2.tsx
@@ -1,0 +1,157 @@
+import React, {PropsWithChildren, useCallback, useMemo} from 'react';
+import {LayoutChangeEvent} from 'react-native';
+import Reanimated, {
+  useAnimatedGestureHandler,
+  useAnimatedStyle,
+  useSharedValue,
+  withTiming,
+  interpolate,
+  interpolateColor,
+  runOnJS
+} from 'react-native-reanimated';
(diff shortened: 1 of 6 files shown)


lidord-wix:
Hey, few things:
1. onLongPress acts on release instead of after a constant time.
2. The onCancel doesn't work.
3. The tabController doesn't work while using this touchableOpacity.

ethanshar:
@lidord-wix 
I fixed both 1. and 2. 
And about 3. The current TabController is not using this TouchableOpacity so there shouldn't be any problem

lidord-wix:
Will the TabController not use this TouchableOpacity in the future?

ethanshar:
> @ethanshar
> Will the TabController not use this TouchableOpacity in the future?

I'm planning to use it in TabController2

src/incubator/TouchableOpacity2.tsx
```diff
@@ -0,0 +1,169 @@
...
+  TODO: Consider revisiting this issue to see if it still occurs */
+  const isLongPressed = useSharedValue(false);
+
+  const backgroundColor = useMemo(() => {
+    return props.backgroundColor || modifiers.backgroundColor;
+  }, [props.backgroundColor, modifiers.backgroundColor]);
+
+  const onPress = useCallback(() => {
+    props.onPress?.(props);
+  }, [props]);
```
lidord-wix, line 91: Why does the onPress need all the props as an argument?
This will update the callback for every prop change

lidord-wix, line 91: Why does the onPress need all the props as an argument?
This will update the callback for every prop change

src/incubator/TouchableOpacity2.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const onLongPress = useCallback(() => {
+    props.onLongPress?.(props);
+  }, [props]);
```
lidord-wix, line 95: Same here :)

src/incubator/TouchableOpacity2.tsx
```diff
@@ -0,0 +1,169 @@
...
+ * @modifiers: flex, margin, padding, background
+ * @example: https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/incubatorScreens/TouchableOpacityScreen.js
+ */
+function TouchableOpacity(props: Props) {
+  const {
+    children,
+    modifiers,
+    style,
+    // onPress = _.noop,
+    // onLongPress,
```
lidord-wix, line 70: I guess we can remove these two props from here

src/incubator/TouchableOpacity2.tsx
```diff
@@ -0,0 +1,169 @@
...
+    onFail: () => {
+      isActive.value = withTiming(0, {duration: 200});
+    }
+  });
+
+  const longPressGestureHandler = useAnimatedGestureHandler({
+    onActive: () => {
+      if (!isLongPressed.value) {
+        isLongPressed.value = true;
+        isActive.value = withTiming(0, {duration: 200});
```
lidord-wix, line 114: What is the meaning of the duration here? I can't see any difference between `{duration: 0}` and `{duration: 2000}`

src/incubator/TouchableOpacity2.tsx
```diff
@@ -0,0 +1,169 @@
...
+   */
+  activeScale?: number;
+  /**
+   * Callback for when tapping the touchable
+   */
+  onPress?: (props: any) => void;
+  /**
+   * Callback for when long pressing the touchable
+   */
+  onLongPress?: (props: any) => void;
```
lidord-wix, line 40: Why does the type of `props` here is `any`?

ethanshar: It's because of a behavior/feature we added in uilib to our touchable components. 
We pass back the props the user originally passed to the component. 

Why? 
It solve a common issue when rendering multiple repeating items (like in the case of lists) and other use-cases.. 

For instance, You can have the following `renderItem` method for a FlatList
```
renderItem = ({item, index}) => {
  return <TouchableOpacity onPress={() => doSomething(index)}  />
}
```
You see we have to pass an inline callback here (which is bad for performance) cause we need to pass on the specific index of the item, for each item. 

With the feature we added, the user can do the following
```
doSomething = (touchableProps) => {
  const {index} = touchableProps;
  console.log('item index is:`, index)
}

renderItem = ({item, index}) => {
  return <TouchableOpacity onPress={doSomething} index={index}  />
}
```
Here we didn't have to pass an inline callback, we passed the callback by ref and another custom prop, the `index`. 

That's why `props` in the onPress callback type is any, cause the user can basically pass/accept anything here. 

Let me know if it's not clear, we can sync about It quickly..

src/incubator/TouchableOpacity2.tsx
ethanshar: Removed (:

src/incubator/TouchableOpacity2.tsx
ethanshar: Well technically you are right, we might consider an optimization for it. 
But see first my big comment on the props feature we pass back during press callbacks.


.....That's the reason we pass all props, and that's why we all the props are in the deps arrays. 

Anyway, I agree it's not optimal. 
Maybe we can make an assumption that the "custom" prop the user passed is a static prop that won't change, then we can ignore all the props in the deps array. It means the user will get back the initial value they passed with a custom prop. 
WDYT?

src/incubator/TouchableOpacity2.tsx
ethanshar: Same answer (:

src/incubator/TouchableOpacity2.tsx
ethanshar: There's definitely a difference (: 
Try again. Try refreshing the screen after making the change in the code. 
It dictates the time it takes for the activeness transition.
I did a little refactor here cause these lines repeat itself so pull the new code

src/incubator/TouchableOpacity2.tsx
lidord-wix: Is that a reasonable assumption? 
Maybe the user wants to know about props change, no?

ethanshar: I wanna say that yes, it's reasonable. But technically if our assumption is wrong then it might introduce a bug. 
Anyway, I have another solution. 


Not so long ago, we added a dedicate prop called `customValue` to the original TouchableOpacity component. 
See here -> https://github.com/wix/react-native-ui-lib/blob/master/src/components/touchableOpacity/index.tsx#L45

It's purpose was to be an explicit prop to pass on a custom value as I described in the previous comment. 
Also the reason we had to declare an explicit type for it is due to typings. 
When people use `<TouchableOpacity someValue={...}/>` it threw a TS error that `someValue` is not a valid prop. 
So we have `customValue` and now they can just `<TouchableOpacity customValue={...}/>`

So back to our concern, see my recent code change. 
I only condition the onPress callbacks on `props.onPress` and on `props.customValue` - which is the value that the user expect to pass on.. 

Let me know it all makes sense? (:

lidord-wix (approved):

PR #1399, opened 2021-07-11
Feat/hint add props

Inbal-Tish:
## Description
Hint - adding 'senableShadow' (for white hints only) and 'removePaddings' props.

## Changelog
Hint - adding 'senableShadow' (for white hints only) and 'removePaddings' props.

### DIFF
--- a/demo/src/screens/componentScreens/HintsScreen.tsx
+++ b/demo/src/screens/componentScreens/HintsScreen.tsx
@@ -1,9 +1,11 @@
+import _ from 'lodash';
 import React, {Component} from 'react';
 import {Alert} from 'react-native';
 import {Colors, View, Text, Hint, Button, RadioGroup, RadioButton, Switch} from 'react-native-ui-lib';
 
 
 const settingsIcon = require('../../assets/icons/settings.png');
+const reactions = ['❤️', '😮', '😔', '😂', '😡'];
 
 type HintScreenProps = {};
 type HintScreenState = {
@@ -45,6 +47,10 @@ export default class HintsScreen extends Component<HintScreenProps, HintScreenSt
     Alert.alert('Hint Pressed');
   }
 
+  onReactionPress = () => {
+    Alert.alert('Reaction button pressed');
+  }
+
   renderCustomContent() {
     return (
       <Text text70 white>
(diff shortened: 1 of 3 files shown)


src/components/hint/index.tsx
```diff
@@ -392,6 +407,11 @@ class Hint extends Component<HintProps, HintState> {
     return tipPositionStyle;
   }
 
+  shouldEnableShadow() {
+    const {color, enableShadow} = this.props;
+    return enableShadow && color === Colors.white;
```
mendyEdri, line 412: Why `enableShadow` works only when the color is white?
If so, it's counterintuitive to pass `color={white}` + `enableShadow` to make the shadow work

demo/src/screens/componentScreens/HintsScreen.tsx
```diff
@@ -95,16 +126,12 @@ export default class HintsScreen extends Component<HintScreenProps, HintScreenSt
...
-                  </Text>
-                  for more information
-                </Text>
-              ) : undefined
+              showCustomContent ? 
+                this.renderCustomContent() : showReactionStrip ? this.renderReactionStrip() : undefined
             }
+            color={!showCustomContent && showReactionStrip ? Colors.white : undefined}
+            enableShadow={showReactionStrip}
+            removePaddings={!showCustomContent && showReactionStrip}
```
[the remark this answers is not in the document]
Inbal-Tish: Yes. It was requested that the user will be able to set its own padding. For that it is better to allow it to remove our padding and add then to his custom view.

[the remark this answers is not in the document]
mendyEdri: Got it 💪🏽

src/components/hint/index.tsx
Inbal-Tish: Shadow works for a white container. If we want to support shadow in all cases we need to wrap the hint with another container. Since the request was for a shadow on a white bg hint I "saved" us a view on the hierarchy. I think we can add it if we'll have the requirement.

mendyEdri: No sure if we need to wrap the hint
If you just remove the `&& color === Colors.white`, (and `enableShadow={true}`) it will work with other colors 🙂

Inbal-Tish: Fixed ❤️

mendyEdri:
Loved the demo component! 😍

mendyEdri (approved):

PR #1403, opened 2021-07-13
WheelPicker - fix warning about getNode

M-i-k-e-l:
## Description
WheelPicker - fix warning about getNode

## Changelog
WheelPicker - fix warning about getNode


lidord-wix (commented):
Looks good, I suggest changing the all ts-ignore to ts-expect-error :)

lidord-wix (approved):

PR #1405, opened 2021-07-14
Feat/ show the beginning of a long text in textField on Android

lidord-wix:
## Description
Show the beginning of a long text in textField on Android
WOAUILIB-1892

## Changelog
Show the beginning of a long text in textField on Android

### DIFF
--- a/src/components/textField/index.js
+++ b/src/components/textField/index.js
@@ -621,7 +621,6 @@ export default class TextField extends BaseInput {
         value={value}
         placeholder={placeholderText}
         placeholderTextColor={placeholderColor}
-        selection={Constants.isAndroid && {start: 0}}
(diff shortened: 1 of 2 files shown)


src/components/textField/index.js
```diff
@@ -621,6 +621,7 @@ export default class TextField extends BaseInput {
         value={value}
         placeholder={placeholderText}
         placeholderTextColor={placeholderColor}
+        selection={Constants.isAndroid && {start: 0}}
```
ethanshar, line 624: Have you checked this on Android? 
It creates a bug when typing in a TextField

lidord-wix: I did, but I checked it only in pickers 🙈
Fixed :)

ethanshar (approved):

Issue #1274, opened 2021-04-26
Ref is not forwarding to Picker


Derron210:
Picker can not be accessed via ref.  Is it a bug or was made intentionally?

This happening because of
```
export default asBaseComponent(forwardRef(Picker));
```
in Picker/index.js

Removing of forwardRef will fix this issue.

harish0507:
Any update on this issue?

PR #1407, opened 2021-07-15
Fix forwarded ref in Picker

ethanshar:
## Description
We didn't forward the ref in the picker component but did wrap It with forwardRef HOC which resulted in a null ref. 
This fix issue #1274

## Changelog
Fix issue with Picker component ref not being forwarded

### DIFF
--- a/src/components/picker/index.js
+++ b/src/components/picker/index.js
@@ -22,12 +22,13 @@ const PICKER_MODES = {
   MULTI: 'MULTI'
 };
 const ItemType = PropTypes.oneOfType([
-  PropTypes.number, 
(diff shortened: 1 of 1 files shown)


Inbal-Tish (approved):

riamon-v:
Well, I would love to see an example of how you actually go the reference of the Picker because after trying a hundred solution I still get null or undefined

riamon-v:
> Well, I would love to see an example of how you actually go the reference of the Picker because after trying a hundred solution I still get null or undefined

Wellmy issue is that I was using the native picker component, the ref is not forwarded to it

PR #1409, opened 2021-07-15
Infra/add pre push hook

M-i-k-e-l:
## Description
Add pre-push hook that checks format, typings and runs tests.

There are 3 commits:
1. The actual change.
2. Fix some problems (so formatting will not break them).
3. Auto formatted files (_**VERY big, could cause conflicts!**_).

I'm not sure if you want to add this (and we might want to consider a good timing for it), but I think it can prevent a lot of problems in the future (+ it can help contributions' simplicity); let me know what you think.

## Changelog
Add pre-push hook


ethanshar:
@M-i-k-e-l 
Let's discuss it with the rest of the team in our next sync

ethanshar:
@M-i-k-e-l 
Btw, regarding this one maybe we can have the prehook only for our tests/lint

M-i-k-e-l:
I think the tests are the most annoying part (longest); as decided we should just rely on the pre-push and build process.
Maybe we can add a check in the build process to make sure there are no updates to the `.d.ts` files? I don't think we have that. Is there something more missing from it?

PR #1412, opened 2021-07-18
Fix Picker in multi mode not changing selection when there is a customRender

### LINKED ISSUE #1387: Picker mode='MULTI' item selection not working with custom renderPicker in 5.22.0
### Describe the bug

When I use custom `renderPicker` in the `Picker` component with `mode='MULTI'` not able to select any item from the `Picker.Item` and `props.isSelected` always false.

We can reproduce in our UI lib expo examples.

```
"react-native-ui-lib": "^5.22.0",
```

It was working fine before v5.22.  and Not working in v5.27 also.

Not Working with "MULTI" and renderPicker
```
<Picker title={t('common:batches')}
  mode="MULTI"
  value={selectedStudentBatchRef.current ?? []}
  onChange={function(option: any) {
    console.log(option);
  }}
  renderPicker={function() {
    return (
      <View>
        <Button label={t('common:addBatch')} />
      </View>
    );
  }}>

...
```

Working without renderPicker
```
<Picker title={t('common:batches')}
  mode="MULTI"
  value={selectedStudentBatchRef.current ?? []}
  onChange={function(option: any) {
    console.log(option);
  }}>

...
```

_Originally posted by @saravanakumargn in https://github.com/wix/react-native-ui-lib/issues/1258#issuecomment-873763486_

M-i-k-e-l:
## Description
Fix Picker in multi mode not changing selection when there is a customRender

## Changelog
Fix Picker in multi mode not changing selection when there is a customRender

Fixes #1387

### DIFF
--- a/src/components/picker/index.js
+++ b/src/components/picker/index.js
@@ -195,7 +195,12 @@ class Picker extends Component {
           value: nextProps.value
         };
       }
-    } else if (_.isFunction(nextProps.renderPicker) && prevState.value !== nextProps.value) {
+    } else if (
+      // Allow external change of the picker's value, only applicable to single mode
+      nextProps.mode === Picker.modes.SINGLE &&
+      _.isFunction(nextProps.renderPicker) &&
+      prevState.value !== nextProps.value
+    ) {
       return {
         prevValue: prevState.value,
         value: nextProps.value


src/components/picker/index.js
```diff
@@ -195,7 +195,12 @@ class Picker extends Component {
           value: nextProps.value
         };
       }
-    } else if (_.isFunction(nextProps.renderPicker) && prevState.value !== nextProps.value) {
+    } else if (
+      // Allow external change of the picker's value, only applicable to single mode
+      nextProps.mode === Picker.modes.SINGLE &&
+      _.isFunction(nextProps.renderPicker) &&
+      prevState.value !== nextProps.value
+    ) {
```
ethanshar, line 203: So technically an external change of the value will not work for multi-mode pickers? 
Why? 

It feels like there's a root issue we need to fix here, no?

M-i-k-e-l: I agree, however I could not think of a good example of changing the information externally in multi mode (perhaps an update from the server?), and this was the easiest fix to keep both scenarios working.
If we want a better fix for this, I suggest adding an internal ticket (maybe we want this patch until we get to that).

PR #1417, opened 2021-07-20
Fix/tint color for svg image

phibo23:
## Description
- The `Image` component supports SVGs (since https://github.com/wix/react-native-ui-lib/releases/tag/5.23.0), but the `tintColor` prop didn't work
- react-native-svg-transformer has a documented configuration to replace a specified fill color in SVG components
- in components such as `TabBarItem` where Icons are based on the `Image` component, `tintColor` is applied automatically, which also caused an error with SVGs

Bildschirmfoto 2021-07-08 um 15 06 10

## Changelog
Added `tintColor` support for SVGs in Image component, if react-native-svg-transformer is configured to replace fill color

### DIFF
--- a/src/components/image/SvgImage.tsx
+++ b/src/components/image/SvgImage.tsx
@@ -5,10 +5,11 @@ const SvgXml = SvgPackage?.SvgXml;
 
 export interface SvgImageProps {
   data: any; // TODO: I thought this should be string | React.ReactNode but it doesn't work properly
+  fill?: string; // fill works on files if SVGR is configured appropriately, see https://github.com/kristerkari/react-native-svg-transformer/blob/f10c8d579490826227d1bcfa84ba6eabc124b70a/README.md#changing-svg-fill-color-in-js-code
 }
 
 function SvgImage(props: SvgImageProps) {
-  const {data, ...others} = props;
+  const {data, fill, ...others} = props;
 
   if (!SvgXml) {
     // eslint-disable-next-line max-len
@@ -17,10 +18,11 @@ function SvgImage(props: SvgImageProps) {
   }
 
   if (typeof data === 'string') {
+    // TODO can fill be applied here too?
     return <SvgXml xml={data} {...others}/>;
   } else if (data) {
     const File = data; // Must be with capital letter
-    return <File {...others}/>;
+    return <File fill={fill} {...others}/>;
   }
 
   return null;
--- a/src/components/image/index.tsx
+++ b/src/components/image/index.tsx
@@ -158,8 +158,8 @@ class Image extends PureComponent<Props, State> {
   }
 
   renderSvg = () => {
-    const {source, ...others} = this.props;
-    return <SvgImage data={source} {...others}/>;
+    const {source, tintColor, ...others} = this.props;
+    return <SvgImage data={source} fill={tintColor} {...others}/>;
   }
 
   renderRegularImage() {
--- a/src/components/image/index.tsx
+++ b/src/components/image/index.tsx
@@ -33,7 +33,7 @@ export type ImageProps = RNImageProps & MarginModifiers & {
    */
   assetGroup?: string;
   /**
-   * the asset tint
+   * the asset tint. For SVG support see note above.
    */
   tintColor?: string;
   /**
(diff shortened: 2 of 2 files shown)


src/components/image/SvgImage.tsx
```diff
@@ -5,10 +5,11 @@ const SvgXml = SvgPackage?.SvgXml;
 
 export interface SvgImageProps {
   data: any; // TODO: I thought this should be string | React.ReactNode but it doesn't work properly
+  fill?: string; // fill works on files if SVGR is configured appropriately, see https://github.com/kristerkari/react-native-svg-transformer/blob/f10c8d579490826227d1bcfa84ba6eabc124b70a/README.md#changing-svg-fill-color-in-js-code
```
M-i-k-e-l, line 8: 1. Is `fill` always the same as `tintColor`? If yes, I'd change the name.
2. I suggest putting this comment as a doc (if we want to expose it at some point).

src/components/image/index.tsx
```diff
@@ -158,8 +160,8 @@ class Image extends PureComponent<Props, State> {
   }
 
   renderSvg = () => {
-    const {source, ...others} = this.props;
-    return <SvgImage data={source} {...others}/>;
+    const {source, tintColor, ...others} = this.props;
```
M-i-k-e-l, line 163: If `fill` is indeed changed to `tintColor`, this change can be removed.

M-i-k-e-l (commented):
Hi @phibo23,
Thanks a lot for your contribution!

1. Did you have a red screen, an error or a warning in your case?
2. Could you please provide an example (even in the PR) of how this is configured?

phibo23:
> Hi @phibo23,
> Thanks a lot for your contribution!
> 
> 1. Did you have a red screen, an error or a warning in your case?
> 2. Could you please provide an example (even in the PR) of how this is configured?

Hi, sorry for taking so long to respond as I was too busy in the last weeks. I'll try to answer yout questions in the next days.

phibo23:
> Did you have a red screen, an error or a warning in your case?

a red screen was the reason why a colleague of mine suggested the workaround of using fill instead of tintColor for svg images. It is actually an issue with the underlying react-native-svg library.

Bildschirmfoto 2021-08-15 um 18 49 51

We handled it in our project by creating a patch with https://www.npmjs.com/package/patch-package. This pull request is an adaptation of that patch.

However, I have to admit, that I'm not sure if I adapted it correctly, because in the patch we only had to change one js file of the image component and in this PR I had to add some more changes required by TypeScript (which I am not very familiar with). Here is our patch file:
```patch
diff --git a/node_modules/react-native-ui-lib/src/components/image/index.js b/node_modules/react-native-ui-lib/src/components/image/index.js
index 2516e1d..84b4790 100644
--- a/node_modules/react-native-ui-lib/src/components/image/index.js
+++ b/node_modules/react-native-ui-lib/src/components/image/index.js
@@ -154,9 +154,10 @@ class Image extends PureComponent {
   renderSvg = () => {
     const {
       source,
+      tintColor,
       ...others
     } = this.props;
-    return <SvgImage data={source} {...others} />;
+    return <SvgImage data={source} fill={tintColor} {...others} />;
   };
 
   renderRegularImage() {
```

> Could you please provide an example (even in the PR) of how this is configured?

We configured SVGR according to the documentation of react-native-svg-transformer https://github.com/kristerkari/react-native-svg-transformer/blob/f10c8d579490826227d1bcfa84ba6eabc124b70a/README.md#changing-svg-fill-color-in-js-code
I was reluctant to add anything to the react-native-ui-lib example app, because this type of configuration is only suitable if you want to apply it to **every** svg in a project.

src/components/image/SvgImage.tsx
phibo23: The problem is, that the SVG component from react-native-svg also has a `tintColor` prop, which appears to be incompatible with color strings like `#12297a`. At least on iOS that causes an error, because in the original line 23 of this file (`return <File {...others}/>;`) all props passed to the `Image` component except for those destructured explicitely for this component are passed through.

Regarding the documentation: The SvgImage component does not have its own page in your documentation. I did however add the same info to the `@notes` in the Image component documentation

M-i-k-e-l:
It's been more than a month, I'm closing it; feel free to reply if you want this reopened.

PR #1419, opened 2021-07-21
Infra/no import rule multiple

mendyEdri:
## Description
Add support for multiple rules in `no-direct-import`.

## Changelog
Support having multiple eslint rules passed as array in `no-direct-import` rule.

### DIFF
--- a/eslint-rules/package.json
+++ b/eslint-rules/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-uilib",
-  "version": "2.0.9",
+  "version": "2.1.0",
   "description": "uilib set of eslint rules",
   "keywords": [
     "eslint",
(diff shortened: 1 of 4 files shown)


eslint-rules/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-uilib",
-  "version": "2.0.9",
+  "version": "2.1.0",
```
lidord-wix, line 3: I'm not sure if we should bump from 2.0.9 to 2.1.0 or to 2.0.10.. do you know if there's any logic for that?

lidord-wix (commented):
Look really good, I left one small comment :)

eslint-rules/package.json
mendyEdri: Base on [Semantic Versioning](https://semver.org/).
("MINOR version when you add functionality in a backwards compatible manner")

lidord-wix (approved):

Issue #1393, opened 2021-07-06
Picker doesn't work in landscape mode on iOS


WJimmyCook:
### Describe the bug
When you open a picker modal in landscape orientation the modal tries to open in portrait orientation then immediately closes

### To Reproduce

1. Add picker component
2. Turn phone to landscape mode
3. Open picker

### Expected behavior

Picker to work in either orientation.

### Screenshots

https://user-images.githubusercontent.com/23268736/124596649-88367980-de30-11eb-8af3-97321135993e.mp4

useNativePicker true
https://user-images.githubusercontent.com/23268736/124596949-dc415e00-de30-11eb-8b4a-b457a564af41.mp4

### Device (please complete the following information)
 - Device: iPhone (works as expected on android)

ethanshar:
Could it be it's related to this issue? 
https://github.com/wix/react-native-ui-lib/issues/1241

WJimmyCook:
That seems to be a different issue but it did lead me down the right path.

https://github.com/wix/react-native-ui-lib/blob/50ccb660e3254ebb1d73d2c7adfc3b6ac4ca3494/src/components/picker/PickerModal.js#L83

Adding `supportedOrientations={['portrait', 'landscape', 'landscape-left', 'landscape-right']}` to the Modal component in the PickerModal.js file fixes the issues. I can submit a PR with the fix if you like, let me know.

ethanshar:
> Adding `supportedOrientations={['portrait', 'landscape', 'landscape-left', 'landscape-right']}` to the Modal component in the PickerModal.js file fixes the issues. I can submit a PR with the fix if you like, let me know.

Yes. That would be great! thanks

ethanshar:
Hi, after going over your PR I realized that this is something you can easily achieve by setting the Modal component's theme. 
Like this
```
ThemeManager.setComponentTheme('Modal', {
  supportedOrientations: ['portrait', 'portrait-upside-down', 'landscape', 'landscape-left', 'landscape-right']
});
```

Anyway I created a new PR based on your that fixes it in the base Modal component (which all dialogs and pickerDialogs are based on)
https://github.com/wix/react-native-ui-lib/pull/1431

WJimmyCook:
Thanks for the explanation!

ethanshar:
Moreover we decided to not update the default value of `supportedOrientations` cause technically It might mean a breaking change for someone else. 
[...]

PR #1422, opened 2021-07-22
Allow picker to work in landscape mode on iOS

WJimmyCook:
## Description
Currently, when the phone is in landscape mode the picker still tries to open in portrait mode. This allows the picker to work in landscape mode. 

Closes this issue: https://github.com/wix/react-native-ui-lib/issues/1393

supportedOrientations is only used by iOS (it's ignored on android).


ethanshar:
Hi @WJimmyCook 
Thank you for your contribution! 
This PR has a few conflicts, I created a different PR that solves it. 
Also see my full comment on your issue which elaborates another solution

PR #1424, opened 2021-07-24
add select scheme method

s123121:
## Description
Currently, this lib can change dark/light mode base on system setting. However, most of the time, user want to change the theme directly on the app. This PR allow app developer to change theme directly without depending on user's appearance preferences

## Changelog
Add `selectScheme` to change the color scheme manually

### DIFF
--- a/src/commons/modifiers.ts
+++ b/src/commons/modifiers.ts
@@ -97,8 +97,8 @@ export type ContainerModifiers =
   BackgroundColorModifier;
 
 export function extractColorValue(props: Dictionary<any>) {
-  const scheme = Appearance.getColorScheme() || 'light';
-  const schemeColors = Colors.schemes[scheme];
+  const scheme = Colors.currentScheme === 'default' ? Appearance.getColorScheme() : Colors.currentScheme;
+  const schemeColors = Colors.schemes[scheme || 'light'];
   const allColorsKeys: Array<keyof typeof Colors> = [..._.keys(Colors), ..._.keys(schemeColors)];
   const colorPropsKeys = _.chain(props)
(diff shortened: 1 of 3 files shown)


src/style/colors.ts
```diff
@@ -8,17 +8,21 @@ import {colorsPalette, themeColors} from './colorsPalette';
 import ColorName from './colorName';
 
 type Schemes = {light: {[key: string]: string}; dark: {[key: string]: string}};
+type CurrentScheme = 'default' | 'light' | 'dark'
```
ethanshar, line 11: I suggest renaming the type to `Scheme` or `SchemeType`. 
`CurrentScheme` is the name of the instance

src/commons/modifiers.ts
```diff
@@ -97,8 +97,8 @@ export type ContainerModifiers =
   BackgroundColorModifier;
 
 export function extractColorValue(props: Dictionary<any>) {
-  const scheme = Appearance.getColorScheme() || 'light';
-  const schemeColors = Colors.schemes[scheme];
+  const scheme = Colors.currentScheme === 'default' ? Appearance.getColorScheme() : Colors.currentScheme;
+  const schemeColors = Colors.schemes[scheme || 'light'];
```
ethanshar, line 101: I think it's better to add a `getScheme` method in `Colors` just to to avoid repeating this piece of code

ethanshar (changes requested):

ethanshar:
@s123121 
Notice the PR build is red. 
I suggest running `npm run prepush` to check for potential issues (lint, TS and tests)

PR #1430, opened 2021-07-26
Allow select scheme to change color scheme manually

s123121:
## Description
Currently, this lib can change dark/light mode base on system setting. However, most of the time, user want to change the theme directly on the app. This PR allow app developer to change theme directly without depending on user's appearance preferences

Also fix a bug when app go to background, `Apperance` module listen to the wrong color scheme 

## Changelog
Add selectScheme to change the color scheme manually


s123121:
I have change per requested and also fix a bug regarding Apperance 
More info here: https://github.com/facebook/react-native/issues/28525
Also currently, on Android, developer need to change the java code so that Appearance can listen to the colorScheme change (maybe add the warning inside doc?)
https://github.com/facebook/react-native/issues/28823

ethanshar (approved):
@s123121 Looks great! 
Thanks for contributing (:

PR #1432, opened 2021-07-27
Badge accessibility

yginton:
## Description
- Added accessibiltyLabel prop to Badge component, so user can define his own message.
- If label prop is defined we use it and not just give the not-informative 'badge' accessibilty label

## Changelog
Badge Component


demo/src/screens/componentScreens/ProgressiveImageScreen.js
```diff
@@ -1,32 +0,0 @@
-import React, {Component} from 'react';
```
ethanshar, line 1: How this screen relevant to the PR?

src/components/badge/index.tsx
```diff
@@ -104,6 +104,10 @@ export type BadgeProps = ViewProps &
      * Use to identify the badge in tests
      */
     testId?: string;
+    /**
+     * Component accessibility label
+     */
+     accessibilityLabel?: string;
```
ethanshar, line 110: No reason to declare this type. 
Badge is already has this cause its typings based on ViewProps

src/components/progressiveImage/index.js
```diff
@@ -1,54 +0,0 @@
-import React from 'react';
```
ethanshar, line 1: This file also not relevant to this PR. 
I think your environment is not synced or something.. 
These are files from a PR you submitted long time ago..

src/components/badge/index.tsx
```diff
@@ -132,10 +136,12 @@ class Badge extends PureComponent<BadgeProps> {
...
   getAccessibilityProps() {
-    const {onPress, icon, label} = this.props;
-
+    const {onPress, icon, label, accessibilityLabel} = this.props;
     return {
-      accessibilityLabel: icon ? 'badge' : label ? `${label} new items` : undefined,
+      accessibilityLabel: accessibilityLabel ? accessibilityLabel :
+        label ? `${label} new items` :
+        icon ? 'badge' :
+        undefined,
```
ethanshar, line 144: couple of things.. 
First, when you do `x ? x : y` it's clean to do `x || y`
So in this case I'd would `accessibilityLabel || <fallback logic>` 

and the other thing, we always want to fallback to `badge` as an accessibility label. 
So whether it's an icon badge or empty badge we'd like to use `badge` for its a11y label.

ethanshar (changes requested):

PR #1435, opened 2021-07-27
Fix/incubator touch2 long press

mendyEdri:
## Description
Fixed `TouchableOpacity2` long-press was released, after long-press action was called. 
This behaviour matching RN's TouchableOpacity long-press behaviour. 

## Changelog
Fixed `TouchableOpacity2` long-press issue, that the touch was released after action was called.


ethanshar (approved):
Amazing! 
It works great!!

PR #1445, opened 2021-08-03
TextField - fix text alignment in RTL

M-i-k-e-l:
## Description
TextField - fix text alignment in RTL

## Changelog
TextField - fix text alignment in RTL


src/components/textField/index.js
```diff
@@ -800,7 +800,7 @@ function createStyles({centered, multiline}, rightItemTopPadding = 0) {
     },
     errorMessage: {
       color: Colors.red30,
-      textAlign: centered ? 'center' : undefined
+      textAlign: centered ? 'center' : 'left'
```
ethanshar, line 803: We don't need a similar treatment here?

M-i-k-e-l: Not from what I've tested, seems like `Text` (and probably `RNText`) is working properly unlike `RNTextInput`.

ethanshar (approved):

PR #1446, opened 2021-08-04
Fix/ drawer RTL bug

lidord-wix:
## Description
Fix issue with swiping drawer in RTL mode when there are items only on one side
issue - WOAUILIB-1965

## Changelog
Fix RTL bug in drawer

### DIFF
--- a/src/components/drawer/index.tsx
+++ b/src/components/drawer/index.tsx
@@ -156,12 +156,12 @@ class Drawer extends PureComponent<Props> {
   constructor(props: DrawerProps) {
     super(props);
 
-    this.leftRender = props.leftItem ? (Constants.isRTL ? this.renderRightActions : this.renderLeftActions) : undefined;
-    this.rightRender = props.rightItems
-      ? Constants.isRTL
-        ? this.renderLeftActions
-        : this.renderRightActions
-      : undefined;
+    this.leftRender = Constants.isRTL
+      ? props.rightItems && this.renderRightActions
+      : props.leftItem && this.renderLeftActions;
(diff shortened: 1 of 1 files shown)


src/components/drawer/index.tsx
```diff
@@ -156,12 +156,12 @@ class Drawer extends PureComponent<Props> {
...
 
-    this.leftRender = props.leftItem ? (Constants.isRTL ? this.renderRightActions : this.renderLeftActions) : undefined;
-    this.rightRender = props.rightItems
-      ? Constants.isRTL
-        ? this.renderLeftActions
-        : this.renderRightActions
-      : undefined;
+    this.leftRender = Constants.isRTL
+      ? props.rightItems && this.renderRightActions
+      : props.leftItem && this.renderLeftActions;
```
Inbal-Tish, line 161: Why not check if `leftItem` exists in the `renderLeftActions` method? (and the same for the `rightItems`...)

lidord-wix: It actually doesn't work.. the value of `this.leftRender` should be `undefined` and not `() => undefined` if the items don't exist.
I can write the methods that way, but I'm not sure if it makes it more clear: 
```
  private renderLeftActions = this.props.leftItem
    ? (progress: Animated.Value /* , dragX: Animated.Value */) => {
      const {leftItem} = this.props;
      const leftItems = leftItem ? [leftItem] : undefined;
      return this.renderActions(leftItems, progress /* , dragX */);
    }
    : undefined;

```

ethanshar (approved):

PR #1462, opened 2021-08-09
Allow setting windowSoftInputMode

### LINKED ISSUE #1360: RNUiLib forces android's softInputMode to 'adjustResize'
### Describe the bug
https://github.com/wix/react-native-ui-lib/blob/afa371e8ea6f82deacd261114690f54ee71b82e3/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardLayout.java#L169-L181

These codes force android's softInputMode to adjustResize (when keyboard is hidden)

### To Reproduce
1. Just install RNUiLib
2. Modify your AndroidManifest.xml, android:windowSoftInputMode="adjustResize" to "adjustPan"
3. Run the codes below

### Expected behavior
CustomKeyboardLayout should save previous softInputMode and restore it.

### Code snippet
```
import React, {useEffect} from 'react';
import {View, TextInput} from 'react-native';

export default function () {
    return (
        <View style={{ flex: 1 }}>
            <TextInput
                style={{
                    backgroundColor: '#00ff00'
                }}
            />
            <View style={{
                flex: 1
            }}/>
            <View style={{
                height: 30,
                width: 30,
                backgroundColor: '#ff0000'
            }}/>
        </View>
    );
};
```

Android

M-i-k-e-l:
## Description
Currently our `CustomKeyboardLayout` is forcing the app to have `windowSoftInputMode=SOFT_INPUT_ADJUST_RESIZE` even if you did not use a `KeyboardAccessoryView` in your app.
This fix changes this behaviour to only occur if a `KeyboardAccessoryView` is loaded (even in a different tab etc).

@ethanshar, I've tested on both public and private demo apps, but I think it isn't without risk, what's your opinion?

## Changelog
Do not force SOFT_INPUT_ADJUST_RESIZE on Android apps.

Fixes #1360

### DIFF
--- a/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardLayout.java
+++ b/lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardLayout.java
@@ -24,6 +24,8 @@ public class CustomKeyboardLayout implements ReactSoftKeyboardMonitor.Listener, 
     private final InputMethodManager mInputMethodManager;
     private final ReactSoftKeyboardMonitor mKeyboardMonitor;
     private WeakReference<CustomKeyboardRootViewShadow> mShadowNode = new WeakReference<>(null);
+    private int mSoftInputMode;
+    private boolean mIsShown = false;
 
     public CustomKeyboardLayout(ReactContext reactContext, ReactSoftKeyboardMonitor keyboardMonitor, ReactScreenMonitor screenMonitor) {
         mKeyboardMonitor = keyboardMonitor;
@@ -31,6 +33,18 @@ public class CustomKeyboardLayout implements ReactSoftKeyboardMonitor.Listener, 
 
         mKeyboardMonitor.setListener(this);
         screenMonitor.addListener(this);
+    }
+
+    public void setShown(boolean isShown) {
+        mIsShown = isShown;
+        Window window = getWindow();
+        if (window != null) {
+            if (mIsShown) {
+                mSoftInputMode = window.getAttributes().softInputMode;
+            } else {
+                window.setSoftInputMode(mSoftInputMode);
+            }
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


lib/android/src/main/java/com/wix/reactnativeuilib/keyboardinput/CustomKeyboardLayout.java
```diff
@@ -167,16 +181,20 @@ private int getHeightForCustomContent() {
...
     private void setKeyboardOverlayMode() {
-        Window window = getWindow();
-        if (window != null) {
-            window.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_PAN);
+        if (mIsShown) {
+            Window window = getWindow();
+            if (window != null) {
+                window.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_PAN);
+            }
         }
```
ethanshar, line 189: Maybe extract this functionality to a separate method since it pretty much returns itself

ethanshar: Yes, they are very similar, don't you think?

ethanshar: ping

M-i-k-e-l: I do think so, I usually try to minimize changes in these files; done.
Just to be sure it's not missed, there **_is_** a small risk in this PR.

ethanshar: What's the risk?

M-i-k-e-l: We have a few `Activities` that have a different `windowSoftInputMode` than `adjustResize` they might look different if this bug has indeed overridden their selection.
I can send you the details in DM.

PR #1463, opened 2021-08-10
Modify BorderRadiusesLiterals types.

zwh1666258377:
## Description
I have modified the BorderRadiusesLiterals types. Although I would like to use the string template type, but the current typescript version does not support it. So I upgrade the typescript version an rebuild the project typing, except for some field ordering changes, everything looks fine. Is that what you expected? @ethanshar  

About ts new feat: https://github.com/microsoft/TypeScript/pull/44512

## Changelog
The parameters passed in by `loadBorders()`   are checked in `br${number}`   format


package.json
```diff
@@ -110,7 +110,7 @@
     "react-native-svg-transformer": "^0.14.3",
     "react-test-renderer": "^17.0.1",
     "shell-utils": "^1.0.10",
-    "typescript": "4.3.2"
+    "typescript": "4.5.0-dev.20210810"
```
ethanshar, line 113: Do we have to update to a dev version of typescript? 
I prefer to be in a stable version, do literals not supported in v4.3.2?

zwh1666258377: Yes it does not support

zwh1666258377: I am Chinese, because there will be a time difference, so the reply will be slower, please forgive me

ethanshar: no worries. 
If it's ok, I prefer to wait for a stable version of Typescript first that supports literals

zwh1666258377: ok

ethanshar: I see that version 4.4.2 has been published recently
Do you think it contains literals?

zwh1666258377: i was try it,it works!

zwh1666258377: I upgrade typescript version to 4.4.2. Then rebuild the typings.
And many files have been modified, but I think that is ok, please confirm

zwh1666258377: I upgrade typescript version to 4.4.2. Then fixed conflect

ethanshar (approved):

PR #1472, opened 2021-08-16
change onChangeIndex throttle from trailing to leading

RonPeleg:
## Description
change onChangeIndex throttle from trailing to leading

## Changelog
change onChangeIndex throttle from trailing to leading


ethanshar:
@RonPeleg 
Why are u changing from trailing to leading?
I think we used trailing on purpose?

RonPeleg:
waiting 400 ms for onChangeIndex to be called after the user press is a long time. It would be better to call it immediately and throttle future clicks.

ethanshar:
But then we'll have performance issues again. 
Maybe we should consider fixing the performance issues in another way. 

Just to explain, by triggering the onChangeIndex callback while the animation run, the animation is affected and looks laggy.. 
Delaying the callback solves that. 

@RonPeleg 
Do you want to go over your use case and see what we can do, maybe we can fix it another way

PR #1474, opened 2021-08-17
dean-added-auto-scroll-classifier

deanblinder:
## Description
i added a parameter "isAutoScrolled" on the "onChangePage" method to  classify if the image was auto or manually scrolled.

## Changelog
the change was in this file "src/components/carousel/index.tsx"  lines 40,47,171,264


src/components/carousel/types.tsx
```diff
@@ -105,6 +106,7 @@ export interface CarouselState {
   pageHeight: number;
   initialOffset: PointPropType;
   prevProps: CarouselProps;
+  // isAutoScrolled:boolean //dean
```
ethanshar, line 109: Is that needed? 
```suggestion
```

src/components/carousel/index.tsx
```diff
@@ -276,12 +280,13 @@ class Carousel extends Component<CarouselProps, CarouselState> {
     }
 
     this.goToPage(nextPageIndex, true);
-
+
```
ethanshar, line 283: I suggest to run document formatting to remove all redundant spaces that were added 
If you use VSCode you can run the `Format Document` on the file

src/components/carousel/types.tsx
```diff
@@ -40,7 +40,7 @@ export interface CarouselProps extends ScrollViewProps {
   /**
    * callback for when page has changed
    */
-  onChangePage?: (newPageIndex: number, oldPageIndex: number) => void;
+  onChangePage?: (newPageIndex: number, oldPageIndex: number, isAutoScrolled: boolean) => void;
```
ethanshar, line 43: Consider passing the new (last) argument as an object of "info" 
If in the future we'll have to pass more info props then instead of adding more arguments, we'll just add to this info object you created.

ethanshar (changes requested):

ethanshar (approved):

ethanshar (approved):

Issue #1471, opened 2021-08-13
Hint is not appearing in right position?


KavithaMadhu31:

ethanshar:
Hi @KavithaMadhu31 
Can you share a screenshot and also a code snippet. 
In our demo app the component works

KavithaMadhu31:
Hi @ethanshar,
i want to add Hint component inside scrollview, first time when i click hint, position is perfect, after scrolling  top/ down and then clicking the hint, it appears in same position. below added screenshots and code snippet.
![1](https://user-images.githubusercontent.com/37070871/129857200-9b3e7138-c650-4ff9-9d9c-77e5762ae556.png)
![2](https://user-images.githubusercontent.com/37070871/129857225-dbdb690c-ef3f-4fbf-a4fa-70bc4ec38a86.png)
![3](https://user-images.githubusercontent.com/37070871/129857242-5c1c4c7f-80d0-431f-8cb7-70aa7778b2b3.png)

code snippet:
const [isHintVisible, setIsHintVisible] = useState(false);
return(
 <View flex-1>
      <ScrollView style={tw('flex-1')}>
        <View bg-white style={{height: 300}}></View>
        <View bg-white centerH centerV style={{height: 300}}>
          <Hint
            style={{zIndex: 10}}
            visible={isHintVisible}
            message={'Hello'}
            messageStyle={{
              fontSize: 14,
              fontFamily: 'Nunito',
              fontWeight: '700',
              lineHeight: 18,
            }}
            onBackgroundPress={() => setIsHintVisible(false)}
            useSideTip={true}>
            <TouchableOpacity onPress={() => setIsHintVisible(true)}>
              <View padding-v1>
                <HintSvg />
              </View>
            </TouchableOpacity>
          </Hint>
        </View>
        <View bg-white style={{height: 300}}></View>
        <View bg-white style={{height: 300}}></View>
    </View>
);

Inbal-Tish:
@KavithaMadhu31 Hi. Thank you for reporting. I'm working on this bug, hopefully, soon I'll have an update.

KavithaMadhu31:
` const [isHintVisible, setIsHintVisible] = useState(false);
<View flex-1>
      <ScrollView style={tw('flex-1')}>
        <View bg-white style={{height: 300}}></View>
        <View bg-white centerH centerV style={{height: 300}}>
          <Hint
            visible={isHintVisible}
            message={'Hello'}
[...]

PR #1478, opened 2021-08-18
Hint - fix target frame update

Inbal-Tish:
## Description
Hint - fix target frame update when toggling visibility in case of target frame change (due to scroll, for example).
fix issue #1471 

## Changelog
Hint - fix target frame update when toggling visibility in case of target frame change (due to scroll, for example).

### DIFF
--- a/src/components/hint/index.tsx
+++ b/src/components/hint/index.tsx
@@ -110,9 +110,9 @@ export interface HintProps {
    * Callback for Hint press
    */
   onPress?: () => void;
-  /**
-    * Callback for the background press
-    */
+  /** 
+   * Callback for the background press
+   */
   onBackgroundPress?: (event: GestureResponderEvent) => void;
   /**
    * The hint container width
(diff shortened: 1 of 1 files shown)


src/components/hint/index.tsx
```diff
@@ -217,15 +217,13 @@ class Hint extends Component<HintProps, HintState> {
...
     }
-
-    if (!this.state.targetLayoutInWindow) {
-      setTimeout(() => {
-        this.targetRef?.measureInWindow((x: number, y: number, width: number, height: number) => {
-          const targetLayoutInWindow = {x, y, width, height};
-          this.setState({targetLayoutInWindow});
-        });
+    
+    setTimeout(() => {
```
ethanshar, line 221: Maybe it's make sense to re-measure only if the onBackgroundPress prop has been sent, cause otherwise it's not necessary

Inbal-Tish: Done

ethanshar (approved):

PR #1479, opened 2021-08-18
Feat/add transition animator

M-i-k-e-l:
## Depends on
https://github.com/wix/react-native-ui-lib/pull/1475

## Description
Add `TransitionView` - a wrapping view that adds enter\exit animation (incubator).

I know there's a few `location === 'left' || location === 'right'` (or `top` and `bottom`), but to move it to a method I'll need to have two versions (`worklet` and regular), and I'll have to add `!` in some places.
I'm unsure if we can use `useHiddenLocations` elsewhere, and about the naming of `HiddenLocation` (I've renamed it to `TransitionLocation` in `TransitionView`).

A few other things:
1. There's a question (a TODO) about how to handle `opacity` from the user.
2. We can allow the user to add his\her own animation. Edit: I probably need to select a different animation (time probably) for this, I don't want to block this PR with this though, it'll be easier to test with future PRs (Dialog?).
3. The usage of `TransitionLocation` in the screen is a bit clunky, not sure if I did something wrong or if using a `union` is less usable than an enum, WDYT?

## Changelog
Add `TransitionView` - a wrapping view that adds enter\exit animation (incubator).


src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    forwardedRef,
+    style: propsStyle,
+    onLayout: propsOnLayout,
+    ...others
+  } = props;
+  const containerRef = React.createRef<RNView>();
+  const {onLayout: hiddenLocationOnLayout, hiddenLocations} = useHiddenLocations({containerRef});
+  const visible = useSharedValue<boolean>(!enterAnimationLocation);
+
+  const getLocation = (location?: HiddenLocation) => {
```
ethanshar, line 50: One more thing. 
We should aim using hooks to handle logic. 

Here, some logic is inside a hook while some is part of the component code. 

I wouldn't go crazy moving each single `if..else` to a hook. 
But in this case, wouldn't it be nice to have a general hook that handles these locations functionalities? 

Take a another look at the code, see if you can more relevant logic code under a hook.

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+      opacity: Number(visible.value)
+    };
+  }, []);
+
+  const animationEnded = useCallback((animationType: TransitionAnimationEndType) => {
+    onAnimationEnd?.(animationType);
+  },
+  [onAnimationEnd]);
+
+  const outAnimationEnd = useCallback((isFinished: boolean) => {
```
ethanshar, line 76: Few things on the animate methods.. 
It took me some time to understand why are there so many animate methods. 

First, notice the order you have: 
outAnimationEnd, animate, animateOut, useImperativeHandler, inAnimationEnd, animateIn. 

I think we can order them in a more sensible order to make it more readable. 
the `useImperativeHandler` in the middle is not helping.. 

Second, I'd consider moving all of these methods into a separate hook. 

One last thing, some of the methods names were confusing to me.. 
For instance I read `inAnimationEnd` as a method that returns a boolean value.. 
Consider renaming some of the methods: enterAnimatedEnd, exitAnimationEnd, onEnter, onExist, etc..

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    style: propsStyle,
+    onLayout: propsOnLayout,
+    ...others
+  } = props;
+  const containerRef = React.createRef<RNView>();
+  const {onLayout: hiddenLocationOnLayout, hiddenLocations} = useHiddenLocations({containerRef});
+  const visible = useSharedValue<boolean>(!enterAnimationLocation);
+
+  const getLocation = (location?: HiddenLocation) => {
+    if (location === 'left' || location === 'right') {
```
ethanshar, line 51: I personally like to use includes in these cases (`['left', 'right'].includes(location)`)
I found it shorter and more readable. 

In general this whole method can be shorten to 

```
return {
  x: ['left', 'right'].includes(location) ? hiddenLocations[location] : 0,
  y: ['top', 'bottom'].includes(location) ? hiddenLocations[location] : 0,
}
```

ethanshar (commented):
See my comments. 
I mostly focused on readability and code structure. 

I'm curious if this API will fit to our use cases. 
(I mean the imperative API over the declarative one)

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    } else if (location === 'top' || location === 'bottom') {
+      return {x: 0, y: hiddenLocations[location]};
+    } else {
+      return {x: 0, y: 0};
+    }
+  };
+
+  // Has to start at {0, 0} with {opacity: 0} so layout can be measured
+  const translateX = useSharedValue<number>(0);
+  const translateY = useSharedValue<number>(0);
```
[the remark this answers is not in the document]
M-i-k-e-l: Done

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+      transform: [{translateX: translateX.value}, {translateY: translateY.value}],
+      // TODO: do we want to take the component's opacity here? - I think combining opacities is buggy
+      opacity: Number(visible.value)
+    };
+  }, []);
+
+  const animationEnded = useCallback((animationType: TransitionAnimationEndType) => {
+    onAnimationEnd?.(animationType);
+  },
+  [onAnimationEnd]);
```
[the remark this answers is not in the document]
M-i-k-e-l: `runOnJS` does not receive `undefined`, I think I can add a default `noop` if you'd like.

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import useHiddenLocations, {HiddenLocation} from './useHiddenLocations';
+const AnimatedView = Animated.createAnimatedComponent(View);
+export {HiddenLocation as TransitionLocation};
+
+const DEFAULT_ANIMATION_VELOCITY = 300;
+const DEFAULT_ANIMATION_CONFIG = {velocity: DEFAULT_ANIMATION_VELOCITY, damping: 18, stiffness: 300, mass: 0.4};
+
+export type TransitionAnimationEndType = 'in' | 'out';
+
+export interface TransitionAnimatorProps extends ViewProps {
```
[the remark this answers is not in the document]
M-i-k-e-l: The visibility is decided by the user.
It is working with `Dialog` when using a `Modal`; I'm not sure how exactly it will work with something like a `Toast`, but I think it's possible to change the `enterFrom`; would you like me to test it or should we do that later?

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+const TransitionAnimator = (props: Props) => {
+  const {
+    onAnimationEnd,
+    enterAnimationLocation,
+    exitAnimationLocation,
+    forwardedRef,
+    style: propsStyle,
+    onLayout: propsOnLayout,
```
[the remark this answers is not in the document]
M-i-k-e-l: I prefer for `others` to not have it; I find it cleaner and safer (refactoring etc).

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+      visible.value = true;
+    }
+  }, [hiddenLocations.isDefault]);
+
+  const onLayout = useCallback((event: LayoutChangeEvent) => {
+    hiddenLocationOnLayout(event);
+    propsOnLayout?.(event);
+  }, []);
+
+  return <AnimatedView {...others} onLayout={onLayout} style={[propsStyle, animatedStyle]} ref={containerRef}/>;
```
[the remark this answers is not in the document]
M-i-k-e-l: Same as before, let me know if you want me to change it.

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    onAnimationEnd,
+    enterAnimationLocation,
+    exitAnimationLocation,
+    forwardedRef,
+    style: propsStyle,
+    onLayout: propsOnLayout,
+    ...others
+  } = props;
+  const containerRef = React.createRef<RNView>();
+  const {onLayout: hiddenLocationOnLayout, hiddenLocations} = useHiddenLocations({containerRef});
```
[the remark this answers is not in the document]
M-i-k-e-l: `top`, `bottom`, `left` and `right`, I look at it as plural.

src/incubator/TransitionAnimator/index.tsx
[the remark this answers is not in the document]
M-i-k-e-l: I'm not happy with the naming and I agree it's confusing with only the `s` difference.
We had directions previously, but here it's source and destination, not really a direction.
Maybe `Position`, do you have any other ideas?

src/incubator/TransitionAnimator/index.tsx
M-i-k-e-l: Moved things around and rename (some differently than suggested), let me know what you think.

src/incubator/TransitionAnimator/index.tsx
M-i-k-e-l: I don't think I see what you mean\see; would you like to talk about it or perhaps delay this to a future PR (`useHiddenLocations` will be used by `PanView` as well, so I might see some re-usable code).

src/incubator/TransitionAnimator/index.tsx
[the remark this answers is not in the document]
ethanshar: No, let's keep it as it is and later on we'll see if it's really a problem

src/incubator/TransitionAnimator/index.tsx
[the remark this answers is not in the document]
ethanshar: But it's a single object of a point on the screen.. 
Plural make it sound like it's an array

src/incubator/TransitionAnimator/index.tsx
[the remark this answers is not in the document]
ethanshar: Why source/destination don't work with "directions"?
Something can come from a direction and also go to a direction..

ethanshar: Why the hook will be used in PanView? 
I don't think we need to make this hook generic and multi purpose.. 

I think that as a thumb rule we should try using hooks to move out logic code out of the component code
While working on this component, try to focus only on it and avoid connecting it to the PanView

src/incubator/TransitionAnimator/index.tsx
ethanshar: Of course it's the main functionality but it doesn't mean it can't be separated.. 
I think the big challenge we're facing with function components is their readability. 
Unlike classes they tend to become messy very easily and it's really difficult to understand the flow and logic of the code. 
With hooks the code become more organized and sectionalized. 
Usually, it's hard to tell what each useEffect does and to what part of the code it belong. 

If I'm using hooks, you can break it down like this which is way more readable.. 
```
const ... = useHiddenLocation()
const ... = useAnimatedTransition()
const ... = useBlaBla()

return <View>....</View>
```

src/incubator/TransitionAnimator/index.tsx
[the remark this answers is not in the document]
M-i-k-e-l: mmm, I get what you're saying, I need to think of the names, including this one.

src/incubator/TransitionAnimator/index.tsx
M-i-k-e-l: It is needed so the animation is more precise (i.e. shorter); currently it's not dismissing to the correct location.

src/incubator/TransitionAnimator/index.tsx
M-i-k-e-l: Ah, now I understand what you mean, I'll take another look.

M-i-k-e-l:
I've renamed and refactored a few things, let me know if anything else comes to mind.

src/incubator/TransitionAnimator/useAnimatedTranslator.ts
```diff
@@ -0,0 +1,62 @@
...
+        translateY.value = withTiming(to.y, {duration: 0}, callback);
+      }
+    }
+
+    visible.value = true;
+  },
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  []);
+
+  const animate = useCallback((to: {x: number; y: number},
```
ethanshar, line 38: The arguments order here feels a little weird. 
You have `to` the end callback and then the direction.. 
Don't you think it's more appropriate to have (to, direction, endCallback) ?

src/incubator/TransitionAnimator/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+// TODO: might need to create a file for types and create a fake component for docs
+export type TransitionAnimatorProps = AnimatedTransitionProps & ViewProps;
+
+type Props = PropsWithChildren<TransitionAnimatorProps> & ForwardRefInjectedProps;
+interface Statics {
+  animateOut: () => void;
+}
+
+const TransitionAnimator = (props: Props) => {
```
ethanshar, line 20: As discussed, consider renaming the component name to `TransitionView`

src/incubator/TransitionAnimator/useAnimatedTranslator.ts
M-i-k-e-l: Yes, but the direction can be undefined and `A required parameter cannot follow an optional parameter.`

src/incubator/index.ts
```diff
@@ -6,8 +6,8 @@ export {default as TouchableOpacity2} from './TouchableOpacity2';
 export {default as WheelPicker, WheelPickerProps} from './WheelPicker';
 export {default as PanView, PanViewProps, PanViewDirections, PanViewDismissThreshold} from './panView';
 export {
-  default as TransitionAnimator,
-  TransitionAnimatorProps,
+  default as TransitionView,
+  TransitionViewProps,
   Direction,
   TransitionAnimationEndType
```
ethanshar: Maybe export with TransitionView prefix so the types are connected to a component
Direction seem arbitrary, instead better name it `TransitionViewDirection`
And for `TransitionAnimationEndType` calling it `TransitionViewTransitionAnimationEndType` is a bit long, do you think we can shorten this type name a little?

M-i-k-e-l: Regarding the `Direction`, it comes from `useHiddenLocation`, and I really think it should be used in `PanView` as well, so I'm not sure about `TransitionViewDirection`, will we want duplicate props?

How about `TransitionEndType`?

ethanshar: So would `TransitionType` would be `TransitionViewTransitionType`? or `TransitionViewType`? 

And for `Direction`, will PanView export its own directions type?

M-i-k-e-l: I thought just `TransitionType`, but if not then I guess it can be `TransitionViewTransitionType`; `TransitionViewType` sounds like a different thing (a type of the view and not the transition).

I'm not sure TBH, do you want to postpone and set it to `TransitionViewDirection` for now?

src/incubator/TransitionAnimator/useAnimatedTranslator.ts
ethanshar: Why is `direction` argument undefined? 
In `animate` implementation, the method does nothing if direction was not passed...
Also, in all usages, you always pass the direction.

M-i-k-e-l: I guess you're right, fixed.

src/incubator/index.ts
M-i-k-e-l: Done

ethanshar (approved):

src/incubator/TransitionAnimator/useAnimatedTranslator.ts
M-i-k-e-l: @ethanshar 
I thought I fixed this, I see in master it's not fixed, I'll fix in a future PR.

src/incubator/index.ts
M-i-k-e-l: @ethanshar
I thought I've renamed to `TransitionViewDirection`, I see in master it's not fixed, I'll fix in a future PR.

PR #1480, opened 2021-08-19
Remove "hoist-non-react-statics" from global type declaration

saulve:
## Description
Remove "hoist-non-react-statics" and "url-parse" from global type declaration.

## Changelog
Changes `typings/globalTypes.d.ts`


src/commons/asBaseComponent.tsx
```diff
@@ -1,8 +1,6 @@
 import React from 'react';
 import {Appearance} from 'react-native';
-//@ts-ignore
 import hoistStatics from 'hoist-non-react-statics';
```
saulve, line 3: noticed that `//@ts-ignore` is used in a lot of places due to undeclared `hoist-non-react-statics`, so I removed those.

package.json
```diff
@@ -71,6 +71,7 @@
     "@react-native-community/netinfo": "^5.6.2",
     "@react-native-picker/picker": "^1.9.4",
     "@testing-library/react-hooks": "^3.4.2",
+    "@types/hoist-non-react-statics": "^3.3.1",
```
saulve, line 74: this is the main change (added to devDependencies)

ethanshar (approved):

PR #1486, opened 2021-08-22
Support passing a static validation message

### LINKED ISSUE #1485: Incubator.TextField  not Support “error” property

ethanshar:
## Description
Support rendering a validation message in Incubator.TextField component regardless to validate prop 
If the user just passed `validationMessage` without `validate` prop, we just the render the message with no condition

Fix #1485

## Changelog
Support passing a validation message to Incubator.TextField component regardless to validate prop


generatedTypes/src/incubator/TextField/ValidationMessage.d.ts
```diff
@@ -14,9 +15,10 @@ export interface ValidationMessageProps {
      */
     validationMessageStyle?: TextStyle;
     retainSpace?: boolean;
+    validate: FieldStateProps['validateOnChange'];
```
mendyEdri, line 18:
```suggestion
    validate?: FieldStateProps['validateOnChange'];
```

mendyEdri: Somehow this `FieldStateProps['validateOnChange']` is a `boolean` value, and it creates typing errors in the demo-screen on `validate` prop, that getting there an array of strings

mendyEdri (changes requested):

generatedTypes/src/incubator/TextField/ValidationMessage.d.ts
ethanshar: done

mendyEdri (approved):

PR #1492, opened 2021-08-23
[Lint] Fix findValueNodeOfIdentifier - take into account valueNode can be false

M-i-k-e-l:
## Description
[Lint] Fix findValueNodeOfIdentifier - take into account that valueNode initial value is false (related to isRequired)

## Changelog
[Lint] Fix findValueNodeOfIdentifier - take into account that valueNode initial value is false (related to isRequired)

### DIFF
--- a/eslint-rules/tests/lib/rules/component-prop-deprecation.js
+++ b/eslint-rules/tests/lib/rules/component-prop-deprecation.js
@@ -89,6 +89,20 @@ ruleTester.run('component-prop-deprecation', rule, {
       <Picker value="value" migrate={true}/>
       `
     },
+    {
+      options: ruleOptions,
+      code: 'import {Picker} from \'module-with-deprecations\'; <Picker t="title" s="subtitle" migrate/>',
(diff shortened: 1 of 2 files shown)


eslint-rules/tests/lib/rules/component-prop-deprecation.js
```diff
@@ -89,6 +89,20 @@ ruleTester.run('component-prop-deprecation', rule, {
       <Picker value="value" migrate={true}/>
       `
     },
+    {
+      options: ruleOptions,
+      code: 'import {Picker} from \'module-with-deprecations\'; <Picker t="title" s="subtitle" migrate/>',
+      errors: [
```
Inbal-Tish, line 95: If that's a valid example why the error? (Same for the next entry)

M-i-k-e-l: 10x, copy paste mistake, wonder why it's not failing the tests; removed.

PR #1509, opened 2021-08-29
Support combining multiple abs modifiers

ethanshar:
## Description
Support combining multiple `abs` modifiers

So far we have `abs` modifiers to quickly position views in an absolute position. We have: absT, absR, absB, absL, etc...
This PR add support for combining multiple abs modifiers, so basically we can do this
```
<View absB absR>
  <Text>xxxx </Text>
</View>
```
to position a view at the right bottom of the screen 

## Changelog
Support passing multiple abs modifiers (e.g absR and absB together)


lidord-wix (approved):
Really cool!

PR #1523, opened 2021-09-05
Feat/ TextField initial width

lidord-wix:
## Description
set TextField initial width to be as the placeholder width

## Changelog
set TextField initial width to be as the placeholder width

### DIFF
--- a/src/incubator/TextField/index.tsx
+++ b/src/incubator/TextField/index.tsx
@@ -176,7 +176,7 @@ const TextField = (props: InternalTextFieldProps) => {
         <View style={[paddings, fieldStyle]}>
           <View row centerV>
             {leadingAccessory}
-            <View flex>
+            <View flexG>
               {floatingPlaceholder && (
                 <FloatingPlaceholder
                   placeholder={placeholder}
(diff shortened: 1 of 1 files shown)


src/incubator/TextField/index.tsx
```diff
@@ -187,12 +187,13 @@ const TextField = (props: InternalTextFieldProps) => {
...
               )}
               <Input
+                placeholderTextColor={floatingPlaceholder ? 'transparent' : undefined}
                 {...others}
                 style={[typographyStyle, colorStyle, others.style]}
                 onFocus={onFocus}
                 onBlur={onBlur}
                 onChangeText={onChangeText}
-                placeholder={floatingPlaceholder ? undefined : placeholder}
+                placeholder={placeholder}
```
ethanshar, line 196: Was that required eventually?  
I see the flexG enough fix the issue..

lidord-wix: It doesn't fix it.. we have to pass the placeholder to set the actual width

ethanshar (approved):

PR #1528, opened 2021-09-09
Support passing a custom formatter to TextField

ethanshar:
## Description
Support passing a custom formatter to TextField

## Changelog
Support passing a custom formatter to TextField


src/incubator/TextField/Input.tsx
```diff
@@ -27,13 +27,18 @@ export interface InputProps
    * placeholder text color
    */
   placeholderTextColor?: ColorType;
+  /**
+   * Custom formatter for the input value (used only when input if not focused)
+   */
+  formatter?: (value?: string) => string;
```
mendyEdri, line 33: Is it possible for the `formatter` to return undefined?

ethanshar: Do you mean whether we should have the following signature? 
`(value?: string) => string | undefined;`?

mendyEdri: Yes, in second thought, when you format something and it's not correct, you return the same value + an error, right?

ethanshar: Well it's really up to the user implementation.. they can do whatever.. 
Technically, we can support returning `undefined` since it's a valid value for TextField. 
I'll change it

ethanshar: Fixed

mendyEdri (approved):

PR #1541, opened 2021-09-12
CardImage - remove deprecated props

Inbal-Tish:
## Description
CardImage - remove deprecated props

## Changelog
CardImage - remove deprecated props

### DIFF
--- a/src/components/card/CardImage.tsx
+++ b/src/components/card/CardImage.tsx
@@ -1,6 +1,5 @@
 import React, {PureComponent} from 'react';
 import {View, StyleSheet} from 'react-native';
-import {LogService} from '../../services';
 // import {BaseComponent} from '../../commons';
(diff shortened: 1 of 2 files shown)


src/components/card/CardImage.tsx
```diff
@@ -1,19 +1,13 @@
 import React, {PureComponent} from 'react';
-import {View, StyleSheet, ImageSourcePropType} from 'react-native';
+import {View, StyleSheet} from 'react-native';
 import {LogService} from '../../services';
```
M-i-k-e-l: This can be removed

M-i-k-e-l (approved):
You can merge after removing the import

src/components/card/CardImage.tsx
Inbal-Tish: Already did. Thanks

PR #1549, opened 2021-09-13
Progress bar new design

Inbal-Tish:
## Description
ProgressBar - new component design and API.

## Changelog
ProgressBar - new component design and API.


src/components/progressBar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+      ...progressStyle
+    };
+  }
+
+  renderCustomElement() {
+    const {customElement} = this.props;
+
+    if (customElement) {
+      return React.cloneElement(customElement, {
+        style: [customElement.props.style, styles.progress]
```
ethanshar, line 133: I don't customElement has `props.style` 
Are you sure you're doing this correctly?

Inbal-Tish: What do you mean? Use optional chaining?

ethanshar: Not exactly.. 
Does `customElement` has `props` on it.. after all we're passing a JSX, no?

Inbal-Tish: Well, yes. JSX extends ReactElement, which has 'props'

Inbal-Tish: What is the use case you're thinking about?

ethanshar (approved):

PR #1564, opened 2021-09-19
WheelPicker (incubator) fix alignment in RTL

M-i-k-e-l:
## Description
WheelPicker (incubator) fix alignment in RTL

## Changelog
WheelPicker (incubator) fix alignment in RTL

### DIFF
--- a/src/incubator/WheelPicker/index.tsx
+++ b/src/incubator/WheelPicker/index.tsx
@@ -260,7 +260,7 @@ const WheelPicker = ({
   return (
     <View testID={testID} bg-white style={style}>
       <View row centerH>
-        <View flexG>
+        <View flexG centerH>
           <AnimatedFlatList
             testID={`${testID}.list`}
             height={height}


lidord-wix:
This is a really sensitive component, any alignment change can cause a huge effect - take a look at the `SectionsWheelPicker` screen for example :)

M-i-k-e-l:
> This is a really sensitive component, any alignment change can cause a huge effect - take a look at the `SectionsWheelPicker` screen for example :)

Seems like you have a bug; look at that screen when RTL is enabled (iOS).

lidord-wix:
> > This is a really sensitive component, any alignment change can cause a huge effect - take a look at the `SectionsWheelPicker` screen for example :)
> 
> Seems like you have a bug; look at that screen when RTL is enabled (iOS).
> When looking at "real" use cases in our private app this seemed to solve the issue; is it possible that this is a bug in the screen?
> (I've looked for a while for other solutions.)

Yup, it might be a problem with the screen itself. I actually see that the new durationPicker component (which is based on the sectionsWheekPicker) looks good in RTL as well.

M-i-k-e-l:
> > > This is a really sensitive component, any alignment change can cause a huge effect - take a look at the `SectionsWheelPicker` screen for example :)
> > 
> > 
> > Seems like you have a bug; look at that screen when RTL is enabled (iOS).
> > When looking at "real" use cases in our private app this seemed to solve the issue; is it possible that this is a bug in the screen?
> > (I've looked for a while for other solutions.)
> 
> Yup, it might be a problem with the screen itself. I actually see that the new durationPicker component (which is based on the sectionsWheekPicker) looks good in RTL as well.

You mean with the change, right?

lidord-wix:
> > > > This is a really sensitive component, any alignment change can cause a huge effect - take a look at the `SectionsWheelPicker` screen for example :)
> > > 
> > > 
> > > Seems like you have a bug; look at that screen when RTL is enabled (iOS).
> > > When looking at "real" use cases in our private app this seemed to solve the issue; is it possible that this is a bug in the screen?
> > > (I've looked for a while for other solutions.)
> > 
> > 
> > Yup, it might be a problem with the screen itself. I actually see that the new durationPicker component (which is based on the sectionsWheekPicker) looks good in RTL as well.
> 
> You mean with the change, right?

I checked the durationPicker without the change :)

M-i-k-e-l:
> > > > > This is a really sensitive component, any alignment change can cause a huge effect - take a look at the `SectionsWheelPicker` screen for example :)
> > > > 
> > > > 
> > > > Seems like you have a bug; look at that screen when RTL is enabled (iOS).
> > > > When looking at "real" use cases in our private app this seemed to solve the issue; is it possible that this is a bug in the screen?
> > > > (I've looked for a while for other solutions.)
> > > 
> > > 
> > > Yup, it might be a problem with the screen itself. I actually see that the new durationPicker component (which is based on the sectionsWheekPicker) looks good in RTL as well.
> > 
> > 
> > You mean with the change, right?
> 
> I checked the durationPicker without the change :)

Could you please test again?
Clean installed + installed on a real device.

M-i-k-e-l:
Looks like it's the `flexG` [here](https://github.com/wix/react-native-ui-lib/blob/master/src/incubator/WheelPicker/index.tsx#L264).

lidord-wix:
> Looks like it's the `flexG` [here](https://github.com/wix/react-native-ui-lib/blob/master/src/incubator/WheelPicker/index.tsx#L264).

I see, but unfortunately the `flexG` is necessary for making the wheel scrollable from everywhere (and not only from the center)

M-i-k-e-l:
I've looked at it some more, I could not find a way to keep flex and RTL working properly together.

PR #1576, opened 2021-09-30
New Dialog using the new Pan and Transition Views

M-i-k-e-l:
## Description
New Dialog using the new Pan and Transition Views
This will have conflicts with #1574 
This is not feature complete, but not bad.
Missing features (or things I did not test yet): alignments, fix Android measureInWindow bug, header (do we need it? probably not), default style(?), onOrientationChange, useSafeArea.

## Changelog
New Dialog using the new Pan and Transition Views


ethanshar:
I think we talked about it already.. 
It'll be great if we can refactor the existing Dialog component and use the new infra without having a new component or any migration

src/incubator/Dialog/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+    if (direction === 'left') {
+      return [PanViewDirections.LEFT];
+    } else if (direction === 'right') {
+      return [PanViewDirections.RIGHT];
+    } else if (direction === 'top') {
+      return [PanViewDirections.UP];
+    } else {
+      return [PanViewDirections.DOWN];
+    }
+  })();
```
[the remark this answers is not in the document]
ethanshar: If we want to stop using enums and use union type for `Directions` shouldn't we align it to union type? 

Also when you say backward compatibility? you mean in Dialog component?

and regards 
> and so, we can use the up\down\left\right convention, although I found it not as good for the useHiddenLocation.

Why union type is not good for `useHiddenLocation`? I see `useHiddenLocation` also use union type

package.json
```diff
@@ -21,7 +21,7 @@
     "android": "react-native run-android",
     "pretest": "npm run lint",
     "test": "jest",
-    "test:watch": "jest --watch",
+    "test:watch": "jest --watch -- ./src/incubator/panView/__tests__/panningUtil.spec.js",
```
ethanshar, line 24: I think you let this by mistake

src/incubator/Dialog/index.tsx
```diff
@@ -1,11 +1,15 @@
 import React, {PropsWithChildren, useMemo, useCallback} from 'react';
 import {StyleSheet, StyleProp, ViewStyle} from 'react-native';
 import {useSharedValue, withTiming, useAnimatedStyle} from 'react-native-reanimated';
-import TransitionView, {TransitionViewDirection, TransitionViewAnimationType} from '../TransitionView/index';
-import PanView, {PanViewDirections} from '../panView/index';
+import TransitionView, {TransitionViewAnimationType} from '../TransitionView/index';
```
ethanshar, line 4:
```suggestion
import TransitionView, {TransitionViewAnimationType} from '../TransitionView';
```

src/incubator/Dialog/index.tsx
```diff
@@ -1,11 +1,15 @@
...
+import PanView from '../panView/index';
```
ethanshar, line 5:
```suggestion
import PanView from '../panView;
```

src/incubator/Dialog/index.tsx
```diff
@@ -1,11 +1,15 @@
...
+import {PanningDirections, PanningDirectionsEnum} from '../panView/panningUtil';
```
ethanshar, line 6: Such things should exported directly from PanView file so other component shouldn't be dependent on internal files of another component

src/incubator/hooks/useHiddenLocation.ts
```diff
@@ -1,14 +1,12 @@
 import {RefObject, useCallback, useState} from 'react';
 import {View, LayoutChangeEvent} from 'react-native';
 import {Constants} from 'helpers';
+import {PanningDirectionsEnum} from '../panView/panningUtil';
```
ethanshar, line 4: See my comment about accessing another component internal files

src/incubator/hooks/useHiddenLocation.ts
```diff
@@ -24,10 +22,10 @@ export default function useHiddenLocation<T extends View>(props: HiddenLocationP
     width = Constants.screenWidth,
     height = Constants.windowHeight,
     isDefault = true
-  }) => {
+  }): HiddenLocation => {
     return {
-      top: -y - height,
-      bottom: Constants.windowHeight - y,
+      up: -y - height,
+      down: Constants.windowHeight - y,
```
ethanshar, line 28: Why did u change to up/down?

M-i-k-e-l: So it will be backwards compatible with the old enum and remove some awkward "translations" between the two enums.

src/incubator/Dialog/index.tsx
M-i-k-e-l: Done, tried to use the `globals` but the enum did not work so I used your original suggestion

src/incubator/hooks/useHiddenLocation.ts
ethanshar: I see.. bummer I prefer top/bottom on up/down
If the components are in Incubator I don't mind changing it, I don't consider it a breaking change

M-i-k-e-l: It'll require a migration (public, private, apps...) because we're using up\down [here](https://github.com/wix/react-native-ui-lib/blob/master/src/components/panningViews/panningProvider.tsx#L4-L13).
Or some code I removed to translate it (and you did not like, if I recall correctly).
I figured we can use the old convention, at least for now, and a migration can be done in any time.

ethanshar: ok 👍

ethanshar (approved):

ethanshar:
Approved, build is still running you can merge once it's done

PR #1577, opened 2021-09-30
WheelPicker - fix missing items at start (Android)

M-i-k-e-l:
## Description
WheelPicker - fix missing items at start (Android)

## Changelog
WheelPicker - fix missing items at start (Android)

### DIFF
--- a/src/incubator/WheelPicker/index.tsx
+++ b/src/incubator/WheelPicker/index.tsx
@@ -280,8 +280,8 @@ const WheelPicker = ({
             snapToInterval={itemHeight}
             decelerationRate={Constants.isAndroid ? 0.98 : 'normal'}
             renderItem={renderItem}
-            getItemLayout={getItemLayout}
-            initialScrollIndex={currentIndex}
+            getItemLayout={Constants.isIOS ? getItemLayout : undefined}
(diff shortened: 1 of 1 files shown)


src/incubator/WheelPicker/index.tsx
```diff
@@ -280,8 +280,8 @@ const WheelPicker = ({
             snapToInterval={itemHeight}
             decelerationRate={Constants.isAndroid ? 0.98 : 'normal'}
             renderItem={renderItem}
-            getItemLayout={getItemLayout}
-            initialScrollIndex={currentIndex}
+            getItemLayout={Constants.isIOS ? getItemLayout : undefined}
+            initialScrollIndex={Constants.isIOS ? currentIndex : undefined}
```
ethanshar, line 284: I assume the issue we had was only on iOS? 
What was the issue this code fix?

M-i-k-e-l: The values return to the start (or some other value) even after a value was selected, it did not occur on my Android (would love a confirmation :) - you can see it in the `DurationPicker` if you remove these props.

M-i-k-e-l:
Does not work in all scenarios on Android

PR #1578, opened 2021-09-30
Fix error when passing a signle item to TabController

ethanshar:
## Description
Fix exception when passing a single item to TabController 

## Changelog
Fix exception when passing a single item to TabController

### DIFF
--- a/src/components/tabController/useScrollToItem.ts
+++ b/src/components/tabController/useScrollToItem.ts
@@ -100,8 +100,9 @@ const useScrollToItem = <T extends ScrollToSupportedViews>(props: ScrollToItemPr
     innerSpacing = 0
   } = props;
   const itemsWidths = useRef<(number | null)[]>(_.times(itemsCount, () => null));
-  const itemsWidthsAnimated = useSharedValue(_.times(itemsCount, () => 0));
-  const itemsOffsetsAnimated = useSharedValue(_.times(itemsCount, () => 0));
+  // Math.max(itemsCount, 2) is for setting minimum of 2 items required for the interpolation
+  const itemsWidthsAnimated = useSharedValue(_.times(Math.max(itemsCount, 2), () => 0));
+  const itemsOffsetsAnimated = useSharedValue(_.times(Math.max(itemsCount, 2), () => 0));
   const currentIndex = useRef<number>(selectedIndex || 0);
   const [offsets, setOffsets] = useState<Offsets>({CENTER: [], LEFT: [], RIGHT: []});
   const {scrollViewRef, scrollTo, onContentSizeChange, onLayout} = useScrollTo<T>({scrollViewRef: propsScrollViewRef});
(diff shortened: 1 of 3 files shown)


src/components/tabController/useScrollToItem.ts
```diff
@@ -100,8 +100,9 @@ const useScrollToItem = <T extends ScrollToSupportedViews>(props: ScrollToItemPr
     innerSpacing = 0
   } = props;
   const itemsWidths = useRef<(number | null)[]>(_.times(itemsCount, () => null));
-  const itemsWidthsAnimated = useSharedValue(_.times(itemsCount, () => 0));
-  const itemsOffsetsAnimated = useSharedValue(_.times(itemsCount, () => 0));
+  // Math.max(itemsCount, 2) is for setting minimum of 2 items required for the interpolation
+  const itemsWidthsAnimated = useSharedValue(_.times(Math.max(itemsCount, 2), () => 0));
+  const itemsOffsetsAnimated = useSharedValue(_.times(Math.max(itemsCount, 2), () => 0));
```
M-i-k-e-l, line 105: This looks like a simple solution, but it feels technically incorrect since we're returning an array with a length of 2 instead of 1.
However, when testing for another solution it will probably be more code change than this.
WDYT?

ethanshar: I preferred a simple fix. 
TBH, I initially thought about not supporting it all and instead throw an appropriate error to the users that they must pass at least two items. WDYT? 
I don't see a reason to use TabController for a single page..

M-i-k-e-l: Sounds like a possible solution.
Is there an option in one of our apps where there can be a single tab? If it is possible, then maybe we should handle it and not force the users to have an if\else.

ethanshar: I don't think it's relevant.. 
The person who reported the issue, just played with the component and passed a single item, but it wasn't his intention so it wasn't really a problem. 
So I think having an appropriate error message in this case will be suffice...

ethanshar: Ok, I added a console.error instead of throwing and handled the issue a little better (I think)
Let me know what u think..

M-i-k-e-l (approved):
I understood where this came from; I'd consider adding something like:
```
<View style={[styles.selectedIndicator, {left: indicatorInsets, width: itemsWidthsAnimated.value[0] - 2 * indicatorInsets}]}/>
```
so the indicator is seen, but I know you prefer less code --> merge if I'm right.

ethanshar:
> I understood where this came from; I'd consider adding something like:
> 
> ```
> <View style={[styles.selectedIndicator, {left: indicatorInsets, width: itemsWidthsAnimated.value[0] - 2 * indicatorInsets}]}/>
> ```
> 
> so the indicator is seen, but I know you prefer less code --> merge if I'm right.

Either way (with the indicator or without it) it doesn't look great. 
So I prefer keeping the the code cleaner..

PR #1580, opened 2021-10-03
Temporarily support string for Badge size

ethanshar:
## Description
Temporarily support string for Badge size
Once we'll migrate our one app users from passing strings, we'll remove this type 

## Changelog
Temporarily support string for Badge size for gradual migration after v6


Inbal-Tish:
@ethanshar We mapped the strings to numbers in a config file. Why add the support again?

ethanshar:
Cause it still fail builds on TS errors and I want to avoid that at the moment.

Inbal-Tish (approved):

PR #1581, opened 2021-10-03
Fix WheelPicker issue with RTL on iOS

ethanshar:
## Description
Fix WheelPicker issue with RTL on iOS

## Changelog
Fix WheelPicker issue with RTL on iOS


M-i-k-e-l (approved):
`centerContent` 🤦‍♂️  🎉

lidord-wix (approved):

PR #1585, opened 2021-10-04
Update Constants.ts

### LINKED ISSUE #1582: Warning EventEmitter.removeListener('change', ...) has been deprecated
### Describe the bug
Not exactly a bug but a warning is created while using buttons
```
EventEmitter.removeListener('change', ...): Method has been deprecated. Please instead use `remove()` on the subscription returned by `EventEmitter.addListener`.
```

### To Reproduce
1. Create a basic button (https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/ButtonsScreen.tsx)
2. Add to any app and click it
3. Check logs for warnings

### Expected behavior
No warnings

### Code snippet
Based on my logs the warning is being triggered due to this line: https://github.com/wix/react-native-ui-lib/blob/master/src/helpers/Constants.ts#L134

Not a device specific issue
Tested on an Android Emulator - Pixel XL (Android 10 (API level 29))

tacoelho:
Fix #1582

## Description
*Listener removal to fix bug.*

## Changelog
*Add remove() instead of removeListener to fix the error message*

### DIFF
--- a/src/helpers/Constants.ts
+++ b/src/helpers/Constants.ts
@@ -131,7 +131,7 @@ const constants = {
   },
   /* Dimensions */
   removeDimensionsEventListener: (callback: any) => {
-    Dimensions.removeEventListener('change', callback);
+    Dimensions.remove('change', callback);
   },
   /* Accessibility */
   get accessibility() {


src/helpers/Constants.ts
```diff
@@ -131,7 +131,7 @@ const constants = {
   },
   /* Dimensions */
   removeDimensionsEventListener: (callback: any) => {
-    Dimensions.removeEventListener('change', callback);
+    Dimensions.remove('change', callback);
```
ethanshar, line 134: Our demo app still uses RN64 which doesn't support this method. 
That means changing it is a breaking change. 
I'd consider supporting both RN APIs for now

ethanshar (changes requested):

src/helpers/Constants.ts
tacoelho: Updated RN version, forgot it.

ethanshar: That's actually want not my intention :( 

Having this change enforce our users to migrate to RN65, which is not a trivial change for people. 
So to avoid that, and to allow both users with `<0.65` and `>=0.65` to use rnuilib I suggest doing the following

```suggestion
    if (Dimensions.remove) {
      Dimensions.remove('change', callback);
    } else {
      Dimensions.removeEventListener('change', callback);
    }
```

package.json
```diff
@@ -99,7 +99,7 @@
     "react": "17.0.1",
     "react-autobind": "^1.0.6",
     "react-dom": "^15.4.2",
-    "react-native": "0.64.1",
+    "react-native": ">=0.65.0",
```
ethanshar, line 102: This should be reverted, migrating to higher version of RN requires more than just changing the version in package.json

ethanshar (changes requested):

package.json
tacoelho: Didn’t know that, sorry about that. 
I’ll fix.

package.json
```diff
@@ -99,7 +99,7 @@
     "react": "17.0.1",
     "react-autobind": "^1.0.6",
     "react-dom": "^15.4.2",
-    "react-native": "0.64.1",
+    "react-native:"0.64.1",
```
ethanshar, line 102: You missed a `"` 
```suggestion
    "react-native": "0.64.1",
```

ethanshar (changes requested):

ethanshar (approved):

src/helpers/Constants.ts
```diff
@@ -131,7 +131,11 @@ const constants = {
   },
   /* Dimensions */
   removeDimensionsEventListener: (callback: any) => {
-    Dimensions.removeEventListener('change', callback);
+    if (Dimensions.remove) {
+      Dimensions.remove('change', callback);
+    } else {
+      Dimensions.removeEventListener('change', callback);
+    }
```
ethanshar, line 138: @tacoelho I approved, but then noticed the build fails. 
It's probably because of TS. 
The best way to see the errors is to run `npm run prepush` 
Anyway, you can use `@ts-expect-error` for now
```suggestion
    // @ts-expect-error
    if (Dimensions.remove) {
      // @ts-expect-error
      Dimensions.remove('change', callback);
    } else {
      Dimensions.removeEventListener('change', callback);
    }
```

ethanshar (changes requested):

ethanshar (approved):
In order to include this in our upcoming release
I'll merge these changes and fix the last comments in master. 

Thank you for your contribution

PR #1587, opened 2021-10-06
Fix issue with Expo doesnt recognize uilib native modules

ethanshar:
## Description
Fix issue with Expo doesnt recognize uilib native modules (keyboard)

## Changelog
Fix issue with Expo doesnt recognize uilib native modules (keyboard)


M-i-k-e-l:
I'm assuming this only solves a red-screen and does not actually allows Expo to recognize it, right?

PR #1602, opened 2021-10-13
uilib-test-renderer - some fixes to findStyle

M-i-k-e-l:
## Description
uilib-test-renderer - some fixes to findStyle.
1. The const is renamed from `color` to `value` (more general).
2. Change `find` to `findLast` (the last style is the one in affect).
3. Protect from `undefined`.

## Changelog
uilib-test-renderer - some fixes to findStyle (testing infra)

### DIFF
--- a/src/components/hint/__tests__/index.spec.js
+++ b/src/components/hint/__tests__/index.spec.js
@@ -3,7 +3,7 @@ import {Hint} from 'react-native-ui-lib';
 import {Colors} from '../../../style';
 import {render, findStyle} from '../../../uilib-test-renderer';
 
-const HintTestComponent = ({showHint}) => {
+const HintTestComponent = ({showHint, color}) => {
   return (
     <Hint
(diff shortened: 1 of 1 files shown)


M-i-k-e-l:
@mendyEdri it looks like there's a bug in the `hint/index.spec.js`, you've specified `color={Colors.white}`, and so the background color will be `white` and not `primary`; you can test it by putting the following in the `PlaygroundScreen` (remove the commented lines as needed):

```
<Hint
  visible
  message={'Hint message to hint things'}
  position={Hint.positions.TOP}
  useSideTip
  key={'1'}
  targetFrame={{x: 10, y: 200, height: 1, width: 1}}
  onBackgroundPress={() => {}}
  // color={Colors.white}
  // removePaddings
  enableShadow
  testID={'Hint'}
/>
```

Should I remove the background color or change the expected result?

PR #1604, opened 2021-10-16
Upgrade to rn65

ethanshar:
## Description
Upgrade demo app to react-native 0.65

## Changelog
Upgrade demo app to react-native 0.65

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,4 +118,3 @@ class MyScreen extends Component {
   }
 }
 ```
-


ethanshar:
> Looks good overall.
> When I use the [upgrade-helper](https://react-native-community.github.io/upgrade-helper/?from=0.64.1&to=0.65.1) I see they also renamed the `_editorconfig` file to `.editorconfig`, but I'm not sure if it's mandatory..

Yes, it's not really relevant to us. 
Some files are just project files, .flowconfig is also not that relevant to us

PR #1605, opened 2021-10-16
Fix/event listener warning

ethanshar:
## Description
Depends on https://github.com/wix/react-native-ui-lib/pull/1604 (RN65 Upgrade)
Fix usage warnings of Dimensions.removeEventListener for RN65 

Fix #1598

## Changelog
Fix usage warnings of Dimensions.removeEventListener for RN65

PR #1625, opened 2021-10-28
Feat/use orientation hook

ethanshar:
## Description
a useOrientation hook that handles orientation change event and retrieve components the current orientation 

It will help unifying the solution in this PR (we also need it in private uilib)
https://github.com/wix/react-native-ui-lib/pull/1605

## Changelog
a useOrientation hook that handles orientation change event and retrieve components the current orientation


Inbal-Tish (approved):

Inbal-Tish:
@ethanshar That's a great hook!

PR #1630, opened 2021-10-30
Feat/hint pressable target

ethanshar:
## Description
- Fix issue with target not being pressable while Hint is visible by rendering a mock children (target)
- Added a new prop `backdropColor` to allow users to color the background which gives more emphasis on the hinted target
- I added a new API in ExampleScreenPresenter for rendering options using our SegmentedControl component
- I refactored the Hint example screen so the configuration options will display on top of a dialog and keep the actual screen cleaner 

WOAUILIB-2177

## Changelog
- Hint - Fix issue with target not being pressable while Hint is visible and added `backdropColor` prop to color overlay background


lidord-wix:
Few issues before reviewing the code - 
1. When using the `targetRef` prop with `onBackgroundPress` prop, the hint is rendered above its target.
2. When passing `onBackgroundPress` you still can't press the target button in one press. (both when the target is a child and when using `targetRef`).
3. Consider adding an example with onBackgroundPress to the example screen.

ethanshar:
> Few issues before reviewing the code -
> 
> 1. When using the `targetRef` prop with `onBackgroundPress` prop, the hint is rendered above its target.
> 2. When passing `onBackgroundPress` you still can't press the target button in one press. (both when the target is a child and when using `targetRef`).
> 3. Consider adding an example with onBackgroundPress to the example screen.

Technically there's no point to use onBackgroundPress with the new one. 
But anyway I fixed it so at least people won't get a bug and still can do what they want

lidord-wix:
> > Few issues before reviewing the code -
> > 
> > 1. When using the `targetRef` prop with `onBackgroundPress` prop, the hint is rendered above its target.
> > 2. When passing `onBackgroundPress` you still can't press the target button in one press. (both when the target is a child and when using `targetRef`).
> > 3. Consider adding an example with onBackgroundPress to the example screen.
> 
> Technically there's no point to use onBackgroundPress with the new one. But anyway I fixed it so at least people won't get a bug and still can do what they want

As discussed, the new behaviour doesn't work with the onBackgroundPress prop

ethanshar:
@lidord-wix 
Ok, Im back to my original implementation and removed the `targetRef` prop since it doesn't really solves anything. 
I played with different positions and overall it behave good (I think) 
Let me know if you find something.. 

I'll change the PR description and changelog you can see details there

demo/src/screens/componentScreens/HintsScreen.tsx
```diff
@@ -142,93 +188,39 @@ export default class HintsScreen extends Component<HintScreenProps, HintScreenSt
             {!useTargetFrame && (
               <Button
                 label={showHint ? 'Hide' : 'Show'}
-                onPress={() => this.setState({showHint: !showHint})}
+                onPress={this.toggleHint}
                 style={{alignSelf: targetPosition}}
                 testID={'Hint.button'}
+                ref={this.targetRef}
```
lidord-wix, line 194: same here

lidord-wix (approved):
Looks great now :)
Just need to clean two leftovers of the `targetRef` from the screen.
I'm approving so you can remove it and merge

demo/src/screens/componentScreens/HintsScreen.tsx
```diff
@@ -1,42 +1,35 @@
...
-      // useTargetFrame: true,
-      useSideTip: false,
-      showCustomContent: false,
-      showReactionStrip: false,
-      enableShadow: false
-    };
-  }
+
+export default class HintsScreen extends Component<HintScreenProps> {
+  targetRef = React.createRef();
```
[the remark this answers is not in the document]
ethanshar: Removed

demo/src/screens/componentScreens/HintsScreen.tsx
ethanshar: Removed. thanks

PR #1631, opened 2021-10-31
Card - fix missing types when sending useNative (send nativeProps instead)

M-i-k-e-l:
## Description
Card - fix missing types when sending nativeProps (send nativeProps instead)
Regarding `Omit<Exclude<NativeTouchableOpacityProps, TouchableOpacityProps>, 'style'>;`, you have to `Omit` the `style` because it causes a typing error (they two styles are not the same).

## Changelog
Card - add `nativeProps` prop, deprecate `nativeProps`, allows better typings for the allowed nativeProps.

### DIFF
--- a/demo/src/screens/componentScreens/CardsScreen.tsx
+++ b/demo/src/screens/componentScreens/CardsScreen.tsx
@@ -97,10 +97,8 @@ export default class CardsScreen extends Component<CardsScreenProps, CardsScreen
         style={{marginBottom: 15}}
         onPress={() => {}}
         borderRadius={borderRadius}
-        useNative
         backgroundColor={Colors.white}
-        activeOpacity={1}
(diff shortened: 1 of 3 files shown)


ethanshar:
@M-i-k-e-l 
Im not sure about this solution. 
The activeOpacity is part of RN TouchableOpacity props and activeScale is part of our own implementation of `Incubator.TouchableOpacity`
Also, this typing issue is not relevant only to Card but to everything that inherit from TouchableOpacity (Button as well) 

And last I find it kinda of weird that that some of the props we inherit from Touchable and can be passed directly and some should be pass through `nativeProps`

ethanshar:
Yes. 
There's a better solution that can be done with TS. 
It's called discriminated unions

https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions

PR #1633, opened 2021-11-01
TextField - Support passing custom preset and use it in fieldStyle callback

ethanshar:
## Description
- Support passing custom preset to TextField
- Pass on preset to `fieldStyle` callback so users can react to different presets

## Changelog
TextField - Support custom preset and pass it on to fieldStyle callback for custom preset styles

### DIFF
--- a/src/incubator/TextField/index.tsx
+++ b/src/incubator/TextField/index.tsx
@@ -84,7 +84,7 @@ export type TextFieldProps = MarginModifiers &
     /**
      * Internal style for the field container
      */
-    fieldStyle?: ViewStyle | ((context: FieldContextType) => ViewStyle);
+    fieldStyle?: ViewStyle | ((context: FieldContextType, props: {preset: TextFieldProps['preset']}) => ViewStyle);
     /**
      * Container style of the whole component
      */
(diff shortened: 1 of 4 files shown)


src/incubator/TextField/index.tsx
```diff
@@ -84,15 +84,15 @@ export type TextFieldProps = MarginModifiers &
     /**
      * Internal style for the field container
      */
-    fieldStyle?: ViewStyle | ((context: FieldContextType) => ViewStyle);
+    fieldStyle?: ViewStyle | ((context: FieldContextType, props: {preset: TextFieldProps['preset']}) => ViewStyle);
```
lidord-wix, line 87: Why should we pass the preset to the user? we're getting it from the user and we don't change it inside the component..

lidord-wix (commented):
Why should we pass the `preset` to the user? we're getting it from the user and we don't change it inside the component..

ethanshar:
> Why should we pass the `preset` to the user? we're getting it from the user and we don't change it inside the component..

It gives the user the ability to have different presets for TextField for example: `underline`, `frame`, `search`, etc.. 
And we send it back to the users so they can generate different fieldStyle with it.. 
you can see the [private PR](https://github.com/wix-private/wix-react-native-ui-lib/pull/2008) the use this feature..

lidord-wix (approved):

PR #1639, opened 2021-11-05
Use react-freeze to optimize TabController pages

ethanshar:
## Description
Use [react-freeze](https://github.com/software-mansion-labs/react-freeze) to optimize functionality of TabController.TabPage so pages that are not focused will not trigger renders

## Changelog
Use react-freeze to optimize functionality of TabController.TabPage so pages that are not focused will not trigger renders


src/components/tabController/TabPage.tsx
```diff
@@ -49,9 +51,14 @@ export default function TabPage({
   useAnimatedReaction(() => {
     return targetPage.value === index;
   },
-  isActive => {
+  (isActive, wasActive) => {
```
M-i-k-e-l, line 54: I think `wasActive` is actually `previousActive` or `previousActiveState`

src/components/tabController/TabPage.tsx
```diff
@@ -49,9 +51,14 @@ export default function TabPage({
...
     if (isActive) {
       runOnJS(lazyLoad)();
+      runOnJS(setFocused)(true);
```
M-i-k-e-l, line 57: Do we want to make sure this runs only once `lazyLoad` has finished?

src/components/tabController/TabPage.tsx
```diff
@@ -49,9 +51,14 @@ export default function TabPage({
...
+    }
+
+    if (wasActive && wasActive !== isActive) {
```
M-i-k-e-l, line 60: Why not `wasActive && !isActive ` or (to be sure we don't have a race condition):
```
if (isActive) {
  ...
} else if (wasActive) {
  runOnJS(setFocused)(false);
}
```

demo/src/screens/componentScreens/TabControllerScreen/tab3.tsx
```diff
@@ -13,7 +14,16 @@ class Tab2 extends Component {
...
+    /* Uncomment to test TabPage freeze functionality when the page lose focus */
+    // setInterval(() => {
+    //   this.setState({counter: this.state.counter + 1});
+    // }, 1000);
+  }
+
+  componentDidUpdate() {
+    if (this.state.counter % 3 === 0) {
+      console.warn('freeze counter', this.state.counter);
+    }
```
M-i-k-e-l, line 26: Not sure about the logic here, but either way: did you leave it uncommented on purpose?

src/components/tabController/TabPage.tsx
ethanshar: Not sure I understand? 
You mean the name is not good? 
what's the different between wasActive and previousActive?

src/components/tabController/TabPage.tsx
ethanshar: Don't think it's necessary 
They don't have to be coupled

src/components/tabController/TabPage.tsx
ethanshar: I change it.. now it's better. 
Apparently we got to this part of code more than once, after adding `currentPage` to dependency array of `useAnimatedReaction` it fixed it.. so we can rely completely on `wasActive`

demo/src/screens/componentScreens/TabControllerScreen/tab3.tsx
ethanshar: Yes, but I can comment it as well

ethanshar: done

ethanshar:
@M-i-k-e-l 
I fix the logic a little bit.. 
I fixed how we handle pages that are near the active page - this is relevant when using PageCarousel

src/components/tabController/TabPage.tsx
M-i-k-e-l: It was a little confusing to me, not sure why. I think it's more standard to call it `prev` or `previous`.
The API for `useAnimatedReaction` is:
```
export function useAnimatedReaction<D>(
  prepare: () => D,
  react: (prepareResult: D, preparePreviousResult: D | null) => void,
  deps?: DependencyList
): void;
```

This is also the case in other methods, but if you think it's more clear then I'm fine with it.

src/components/tabController/TabPage.tsx
```diff
@@ -49,18 +49,29 @@ export default function TabPage({
...
+    if (wasActive) {
       runOnJS(setFocused)(false);
     }
-  });
+  },
+  [currentPage]);
+
+  /* Handle freeze for pages that near the active page (relevant for carousel page) */
+  useAnimatedReaction(() => {
+    return asCarousel && currentPage.value - 1 === index || currentPage.value + 1 === index;
```
M-i-k-e-l, line 67: I think we're missing parentheses:
```
return asCarousel && (currentPage.value - 1 === index || currentPage.value + 1 === index);
```

M-i-k-e-l (commented):
As we saw, adding the following log reveals problems in the logic of the frozen tabs:
```
console.log('Ethan', index, !shouldLoad || !focused);
```

src/components/tabController/TabPage.tsx
ethanshar: @M-i-k-e-l 
I changed the implementation again, I think it's even simpler now
It has a single `useAnimatedReaction` so I think it's more bug prune 
Let me know if you find something 😬

src/components/tabController/TabPage.tsx
```diff
@@ -47,13 +49,25 @@ export default function TabPage({
...
   useAnimatedReaction(() => {
-    return targetPage.value === index;
+    return currentPage.value;
   },
-  isActive => {
+  (currentPage, previousPage) => {
+    const isActive = currentPage === index;
+    const wasActive = previousPage === index;
+    const nearActive = asCarousel && (currentPage - 1 === index || currentPage + 1 === index);
+    const wasNearActive = asCarousel && previousPage && (previousPage - 1 === index || previousPage + 1 === index);
```
M-i-k-e-l, line 58: I think `previousPage` should be `previousPage !== null` (for when it's `0`).

M-i-k-e-l (commented):
When on the Posts tab (index=1), do we want the Reviews (index=2) to load? It does not.

src/components/tabController/TabPage.tsx
ethanshar: Fixed that

src/components/tabController/TabPage.tsx
ethanshar: Right! 
Fixed

ethanshar:
> As we saw, adding the following log reveals problems in the logic of the frozen tabs:
> 
> ```
> console.log('Ethan', index, !shouldLoad || !focused);
> ```

The right thing to check is `focused` and not both IMO 
Cause you want to check the focus logic (which seems like it's working) 
The freeze also affected by shouldLoad which might give different results

src/components/tabController/TabPage.tsx
M-i-k-e-l: I think I forgot to delete this comment when I sent the review

M-i-k-e-l:
> > As we saw, adding the following log reveals problems in the logic of the frozen tabs:
> > ```
> > console.log('Ethan', index, !shouldLoad || !focused);
> > ```
> 
> The right thing to check is `focused` and not both IMO Cause you want to check the focus logic (which seems like it's working) The freeze also affected by shouldLoad which might give different results

But if we freeze when `!shouldLoad`, then why have a different logic for the carousel? Assuming it has the same logic for `shouldLoad`, it'll be frozen or not at the same time, wouldn't it?

M-i-k-e-l (approved):

Issue #136, opened 2018-04-08
Change the TextColor of a Picker Component


NikoMontana:
Hi, first of all the picker works great and it gives me a the abillity to learn react native, but how should I   change the text color, fontsize of the selected item in the TextInput Component of the Picker? 

Edit: also how would you remove the underline?

ethanshar:
`Picker` inherits `TextInput` therefore removing the underline is done by passing true to`hideUnderline` prop.
I fixed an issue, now you can pass style to the picker. try using it to change color, fontSize

NikoMontana:
Thank you very much! The problem was solved with your solution regarding the underline.

But setting the color prop in style prop, doesnt do anything. Here is my code..

```                
                      <Picker
                        hideUnderline
                        style={styles.categoryPicker}
                        value={this.state.task_posting_category}
                        enableModalBlur={false}
                        onChange={item =>
                          this.setState({
                            task_posting_category: item,
                            task_category_image:
                              task_category_images[item.value]
                          })
                        }
                        topBarProps={{ title: "Category" }}
                      >
                        {_.map(task_category, option => (
                          <Picker.Item
                            key={option.value}
                            value={option}
                            disabled={option.disabled}
                          />
                        ))}
                      </Picker>


  categoryPicker: {
    color: "white"
  }

ethanshar:
Hey @NikoMontana, 
In this example it works 
https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/FormScreen.js#L110

ethanshar:
Hey @NikoMontana, is this still relevant ?

NikoMontana:
Hi, @ethanshar sorry for the late response. Yes it is still relevant, the textcolor and fontsize are still not changeable.

Inbal-Tish:
@NikoMontana Hi. I see that in your example you are trying to pass 'color' in the style object, that should affect the color of the TextInput text. Since this is a picker, the onPress is opening a modal, thus the text color is irrelevant. Try to pass 'backgroundColor: "red"' to see if that works for you. Anyways, let me know what is that you're trying to modify and I might be able to guide you better. Btw, instead of passing 'enableModalBlur={false}' you can remove this prop altogether.

kallayyash:
> Thank you very much! The problem was solved with your solution regarding the underline.
> 
> But setting the color prop in style prop, doesnt do anything. Here is my code..
> 
> ```
>                       <Picker
[...]

PR #1640, opened 2021-11-07
Support passing max item width to grid view

ethanshar:
## Description
- Add `maxItemWidth` prop to GridView to support a responsive item width without having to know number of colums
- Minor refactor - keep itemSize on state instead on class since it affect the render cycle

WOAUILIB-2207

## Changelog
GridView now supports `maxItemWidth` prop for a responsive item size (this is instead of passing `numColumns` prop)


generatedTypes/src/components/gridView/index.d.ts
```diff
@@ -29,28 +33,32 @@ export interface GridViewProps {
...
     viewWidth: number;
     numColumns: number;
+    itemSize: number;
 }
-declare type ExistProps = GridViewProps & State;
 /**
  * @description: A auto-generated grid view that calculate item size according to given props
  * @example: https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/GridViewScreen.tsx
  */
-declare class GridView extends UIComponent<GridViewProps, State> {
```
Inbal-Tish, line 44: Why you're not passing the 'State'?

src/components/gridView/index.tsx
```diff
@@ -38,44 +43,38 @@ export interface GridViewProps {
...
+  itemSize: number;
 }
 
-
-type ExistProps = GridViewProps & State
 /**
  * @description: A auto-generated grid view that calculate item size according to given props
  * @example: https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/GridViewScreen.tsx
  */
-class GridView extends UIComponent<GridViewProps, State> {
```
Inbal-Tish, line 56: Why you're not passing 'State'?

src/components/gridView/index.tsx
```diff
@@ -38,44 +43,38 @@ export interface GridViewProps {
...
+class GridView extends UIComponent<GridViewProps> {
   static displayName = 'GridView';
 
   static defaultProps = {
     numColumns: 3,
```
Inbal-Tish, line 65: DEFAULT_NUM_COLUMNS instead of "3"

src/components/gridView/index.tsx
```diff
@@ -38,44 +43,38 @@ export interface GridViewProps {
...
     itemSpacing: Spacings.s4
```
Inbal-Tish, line 66: Better take this too to const as you're using it again as default in line #136

src/components/gridView/index.tsx
```diff
@@ -38,44 +43,38 @@ export interface GridViewProps {
   /**
    * whether to keep the items initial size when orientation changes,
    * in which case the apt number of columns will be calculated automatically.
+   * Irrelevant when passing maxItemWidth
```
Inbal-Tish, line 46:
```suggestion
   * Ignored when passing 'maxItemWidth'
```

src/components/gridView/index.tsx
```diff
@@ -1,15 +1,16 @@
 import _ from 'lodash';
 import React from 'react';
 import {StyleSheet} from 'react-native';
-// TODO: we should use asBaseComponent here instead of using UIComponent directly
 import {Colors, Spacings} from 'style';
+// TODO: we should use asBaseComponent here instead of using UIComponent directly
```
Inbal-Tish, line 5: Don't you want to include this in your refactor?

ethanshar: not at the moment..

src/components/gridView/index.tsx
ethanshar: Fixed

src/components/gridView/index.tsx
ethanshar: Fixed

src/components/gridView/index.tsx
ethanshar: Done

Inbal-Tish (approved):

PR #1661, opened 2021-11-11
Fix misuse of new event emitter API in useOrientation hook

ethanshar:
## Description
Fix misuse of new event emitter API in useOrientation hook

## Changelog
Fix event emitter warning coming from useOrientation hook

### DIFF
--- a/src/hooks/useOrientation/index.ts
+++ b/src/hooks/useOrientation/index.ts
@@ -1,4 +1,4 @@
-import {useEffect, useCallback, useState} from 'react';
+import {useEffect, useCallback, useState, useRef} from 'react';
 import {Constants} from 'helpers';
 import useDidUpdate from '../useDidUpdate';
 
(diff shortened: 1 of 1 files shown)


src/hooks/useOrientation/index.ts
```diff
@@ -8,14 +8,15 @@ interface UseOrientationProps {
...
+  const listener = useRef<any>();
 
   const orientationChangeListener = useCallback(() => {
     setOrientation(Constants.orientation);
   }, []);
 
   useEffect(() => {
-    Constants.addDimensionsEventListener(orientationChangeListener);
-    return () => Constants.removeDimensionsEventListener(orientationChangeListener);
+    listener.current = Constants.addDimensionsEventListener(orientationChangeListener);
```
M-i-k-e-l, line 18: I don't think we need the `useRef` here:
```
const listener = Constants.addDimensionsEventListener(orientationChangeListener);
```

ethanshar: Good point 👌 
Fixed

M-i-k-e-l (approved):

PR #1663, opened 2021-11-15
Feat/text add underline

M-i-k-e-l:
## Description
Text - add underline support (`underlineString` prop).
Moved logic to `textUtils` (added support for `utils`).
I'm not entirely happy with the new methods (`getUnifiedStyle` and `unifyTextPartsStyles`), they receive a lot of params and the latter is a bit long; however I did not see a simple way to improve the readability, maybe you will 🙏 
Reading each commit will probably make this easier.

## Changelog
Text - add underline support (`underlineString` prop)

### DIFF
--- a/src/components/text/index.tsx
+++ b/src/components/text/index.tsx
@@ -38,6 +38,10 @@ export type TextProps = RNTextProps &
      */
     highlightStyle?: TextStyle;
     /**
+     * Substring to underline
+     */
+    underlineString?: string | string[];
+    /**
      * Use Animated.Text as a container
      */
     animated?: boolean;
@@ -67,9 +71,9 @@ class Text extends PureComponent<PropsTypes> {
   // }
 
   renderText(children: any): any {
-    const {highlightString, highlightStyle} = this.props;
+    const {highlightString, highlightStyle, underlineString} = this.props;
 
-    if (!_.isEmpty(highlightString)) {
+    if (!_.isEmpty(highlightString) || !_.isEmpty(underlineString)) {
       if (_.isArray(children)) {
         return _.map(children, child => {
           return this.renderText(child);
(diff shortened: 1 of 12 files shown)


src/components/text/index.tsx
```diff
@@ -36,6 +37,10 @@ export type TextProps = RNTextProps &
      * Custom highlight style for highlight string
      */
     highlightStyle?: TextStyle;
+    /**
+     * Substring to underline
+     */
+    underlineString?: string | string[];
```
ethanshar, line 43: I didn't actually thought in that direction 
But I see how it helps in the Banner case 
I thought about underline as a modifier prop (similar to the uppercase prop)
But my thought of such prop was prior to the Banner requirement 

Anyway, in this case why not use the highlightStyle with highlightString to achieve the underline feature the way you did it?

ethanshar: I thought about it again.. 
And this solution is also problematic cause usually the user will want to underline the specific text but also connect it to link (with onPress handler) 
Also, what happen if the highlight string repeats itself in the a sentence but you want to underline only one

M-i-k-e-l: > But I see how it helps in the Banner case

I don't think it'll help much in this case - I think both solutions will work for the `hyperlink` and I don't think we want to "properly support" the inline "link" in the text, so I think we'd like to add the modifier anyway.

>Anyway, in this case why not use the highlightStyle with highlightString to achieve the underline feature the way you did it?

I like my solution more, because then you can use both highlight and underline, but I understand if you don't want to add another prop\logic.

> And this solution is also problematic cause usually the user will want to underline the specific text but also connect it to link (with onPress handler)

Yes, I agree that's a problem (I'm not sure how or if we can solve).

> Also, what happen if the highlight string repeats itself in the a sentence but you want to underline only one

I think that's possible, as long as it's the first one :)
We can think of a solution for that.

PR #1667, opened 2021-11-16
chipsInput add onFocus callback

vidmantas0010:
## Description
*ChipsInput add onFocus callback prop*

## Changelog
*chipsInput*


src/components/chipsInput/index.tsx
```diff
@@ -298,6 +298,7 @@ class ChipsInput extends Component<OwnProps, State> {
   }
 
   onFocus = () => {
+    _.invoke(this.props, 'onFocus');
```
Inbal-Tish, line 301: Please use optional chaining and pass the event (i.e. e: NativeSyntheticEvent<TextInputFocusEventData>)

Inbal-Tish: If you can do the same for 'onBlur' that will be much appreciated!

Inbal-Tish:
@vidmantas0010 Merge master to your branch (will fix failed build)

src/components/chipsInput/index.tsx
vidmantas0010: I can't use optional chaining, because onKeyPress above is using lodash invoke.

Inbal-Tish: The type should be `NativeSyntheticEvent<TextInputFocusEventData>` otherwise it fails on TS error

Inbal-Tish:
@vidmantas0010 Pull master again, we had tests failing that broke the builds

ethanshar:
Closing this. 
We changed our implementation, we're now using `Incubator.ChipsInput` internally. 
@vidmantas0010 This should now work for you

PR #1670, opened 2021-11-16
Feat/text add underline

M-i-k-e-l:
## Description
Text - add underline

## Changelog
Text - add underline


lidord-wix (commented):
Please update the snapshot tests and consider adding an example with underline

lidord-wix (approved):

PR #1673, opened 2021-11-17
Drawer - fix leftToggle for RTL

Inbal-Tish:
## Description
Drawer - fix leftToggle for RTL

## Changelog
Drawer - fix leftToggle for RTL

### DIFF
--- a/src/components/drawer/Swipeable.tsx
+++ b/src/components/drawer/Swipeable.tsx
@@ -61,7 +61,7 @@ type Props = {
   disableHaptic?: boolean
 };
 
-type StateType = {
+type State = {
   dragX: Animated.Value,
   rowTranslation: Animated.Value,
   leftWidth: number | typeof undefined,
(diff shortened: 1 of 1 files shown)


lidord-wix:
I see the following issues:
1. In RTL, without left item, the left toggle still responds.
2. In RTL, without left item, the full swipe doesn't work when using the gesture (not the action button).
3. in RTL, without right items, the full right swipe is dismissing the drawer instead of toggling the value.

Inbal-Tish:
@lidord-wix I fixed the first issue but I don't want to waste any more time on this component. We should rewrite it now that we have all the requirements and we know more about how the users use it.

lidord-wix:
@ethanshar 
We all came to the same conclusion, I think we should prioritize rewriting of this component..

Inbal-Tish:
> @ethanshar We all came to the same conclusion, I think we should prioritize rewriting of this component..

I agree but let's merge this to fix crashes

ethanshar:
Yes, let's merge this
And Inbal already created a ticket for rewriting this component

lidord-wix (approved):

PR #1681, opened 2021-11-19
Feat/new chips input

ethanshar:
## Description
Reimplement new ChipsInput based on Incubator.TextField (under Incubator)

## Changelog
New ChipsInput component based on our new TextField implementation


src/incubator/ChipsInput/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+      <>
+        {leadingAccessory}
+        {map(chips, (chip, index) => {
+          const isMarkedForRemoval = index === markedForRemoval;
+          return (
+            <Chip
+              key={index}
+              customValue={index}
+              // resetSpacings
+              // paddingH-s2
```
lidord-wix, line 96: Can be removed?

src/incubator/ChipsInput/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+      // value={value}
+      // onKeyPress={this.onKeyPress}
+      // enableErrors={false}
+      // onFocus={this.onFocus}
+      // onBlur={this.onBlur}
+      // hideUnderline
+      // selectionColor={isLastTagMarked ? 'transparent' : selectionColor}
+      // style={[inputStyle, styles.alignTextCenter]}
+      // containerStyle={{flexGrow: 0}}
+      // collapsable={false}
```
lidord-wix, line 137: Can be removed?

src/incubator/ChipsInput/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+});
+
+ChipsInput.changeReasons = {
+  ADDED: 'added',
+  REMOVED: 'removed'
+};
+
+export default forwardRef(ChipsInput);
+
+/* Old ChipsInput props to make sure we have parity */
```
lidord-wix, line 155: 1. Seems like we don't have parity, are we eliminating features? 
2. can we remove all those commented props?

src/incubator/ChipsInput/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+const styles = StyleSheet.create({
+  fieldStyle: {
+    flexWrap: 'wrap'
+  }
+});
+
+ChipsInput.changeReasons = {
+  ADDED: 'added',
+  REMOVED: 'removed'
+};
```
lidord-wix, line 151: Are w using it somewhere?

lidord-wix (commented):
Looks great, but seems like we remove some features, is it intended? 
Also, there're some commented lines that I think we can remove..

src/incubator/ChipsInput/index.tsx
ethanshar: I kept it for now so I'll have a quick reference in case there is something missing while we're migrating. 
I added a TODO

src/incubator/ChipsInput/index.tsx
ethanshar: Same here, added a TODO

src/incubator/ChipsInput/index.tsx
ethanshar: This is for our users so they can access the the `changeReasons` directly from the component

src/incubator/ChipsInput/index.tsx
ethanshar: True, I didn't want to add full parity cause some old props are not really needed. 
I kept them here as a comment just for the migration phase just in case there is something missing I missed out.

demo/src/screens/incubatorScreens/IncubatorChipsInputScreen.tsx
```diff
@@ -0,0 +1,43 @@
+import _ from 'lodash';
+import React, {Component} from 'react';
+import {StyleSheet} from 'react-native';
+import {View, Text, Card, TextField, Button, Colors, Incubator} from 'react-native-ui-lib'; //eslint-disable-line
```
lidord-wix, line 4: There're few redundant imports here..

lidord-wix (approved):
Cool, 
There're few redundant imports, so you can remove them and merge :)

PR #1684, opened 2021-11-21
Infra/render tests - TextField

ethanshar:
## Description
Add basic render tests to Incubator.TextField component

## Changelog
Add basic render tests to Incubator.TextField component


src/incubator/TextField/__tests__/index.spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+describe('TextField', () => {
+  describe('hint prop', () => {
+    const props = {
+      testID: 'field',
+      placeholder: 'Placeholder',
+      hint: 'Hint'
+    };
+
+    it('should hint text replace placeholder when input is focused', () => {
```
lidord-wix, line 13: Consider adding the same test but when the user doesn't pass `hint`

ethanshar: Done

lidord-wix (approved):

PR #1685, opened 2021-11-21
Checkbox - render tests

M-i-k-e-l:
## Description
Checkbox - render tests

## Changelog
Checkbox - render tests


src/components/checkbox/__tests__/index.spec.js
```diff
@@ -0,0 +1,94 @@
+import React, {useState, useCallback} from 'react';
+import {render, fireEvent} from '@testing-library/react-native';
+import Checkbox from '../index';
+
+export const testID = 'checkbox';
```
Inbal-Tish, line 5: Why are you exporting the testID?

src/components/checkbox/__tests__/index.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+  return <Checkbox {...others} onValueChange={_onValueChange} value={_value} testID={testID}/>;
+};
+
+describe('Checkbox renderer test', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('Default value is false', async () => {
```
Inbal-Tish, line 26: Why your tests are async?

src/components/checkbox/__tests__/index.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+    const {getByTestId} = render(<TestCase {...props}/>);
+
+    const checkbox = getByTestId(testID);
+    fireEvent.press(checkbox);
+    expect(onValueChange).toHaveBeenCalledTimes(1);
+    expect(onValueChange).toHaveBeenCalledWith(true);
+  });
+
+  it('Send value (false)', async () => {
```
Inbal-Tish, line 37: Why you need this test? You checked the default (false) in the previous test, are you checking react's passed prop mechanism or there's a different behavior when passing false over a default false?

src/components/checkbox/__tests__/index.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    const {getByTestId} = render(<TestCase {...props}/>);
+
+    const checkbox = getByTestId(testID);
+    fireEvent.press(checkbox);
+    expect(onValueChange).not.toHaveBeenCalled();
+  });
+
+  it('Disabled (value = true)', async () => {
+    const props = {onValueChange, value: true, disabled: true};
+    const element = <TestCase {...props}/>;
```
Inbal-Tish, line 86: You can pass on the 'element' const and extract getByTestId directly (like you did in the other tests)

src/components/checkbox/__tests__/index.spec.js
M-i-k-e-l: I think I've put the `TestCase` in a different file and it was copy-pasted, 10x

src/components/checkbox/__tests__/index.spec.js
M-i-k-e-l: I blame copy-pasted again. 10x

src/components/checkbox/__tests__/index.spec.js
M-i-k-e-l: I think I prefer to test explicitly as well as default behaviour, not sure what the best practice, do you want to ask the team?

src/components/checkbox/__tests__/index.spec.js
M-i-k-e-l: Missed that, 10x

src/components/checkbox/__tests__/index.spec.js
Inbal-Tish: Sure, we can ask them. Anyways Let's leave it in the meantime

Inbal-Tish (approved):

PR #1686, opened 2021-11-21
Skip animations (instant animations)

M-i-k-e-l:
## Description
Skip animations (instant animations)
Should make animations "run" instantly in tests.
Does remove warnings about `useNativeDriver`:
```
Animated: `useNativeDriver` is not supported because the native animated module is missing...
```

## Changelog
Skip animations (instant animations)


ethanshar:
@M-i-k-e-l 
Can you elaborate more in the description why did you added this change?

M-i-k-e-l:
> @M-i-k-e-l Can you elaborate more in the description why did you added this change?

Added a little more text, however I see the new tests we've added are not handled, so I'll need to work on it a bit.

ethanshar:
> > @M-i-k-e-l Can you elaborate more in the description why did you added this change?
> 
> Added a little more text, however I see the new tests we've added are not handled, so I'll need to work on it a bit.

Ok, reassign on me when it's ready

PR #1693, opened 2021-11-24
WheelPicker (incubator) - fix updates

Inbal-Tish:
## Description
WheelPicker (incubator) - allow 'items' change; block 'onChange' call when 'initialValue' change.

## Changelog
WheelPicker (incubator) - allow 'items' change; block 'onChange' call when 'initialValue' change.

### DIFF
--- a/src/incubator/WheelPicker/index.tsx
+++ b/src/incubator/WheelPicker/index.tsx
@@ -3,14 +3,14 @@ import {isFunction, isUndefined} from 'lodash';
 import React, {useCallback, useRef, useMemo, useEffect, useState} from 'react';
 import {TextStyle, ViewStyle, FlatList, NativeSyntheticEvent, NativeScrollEvent, StyleSheet} from 'react-native';
 import Animated, {useSharedValue, useAnimatedScrollHandler} from 'react-native-reanimated';
+import {Constants} from 'helpers';
 import {Colors, Spacings} from 'style';
+import {asBaseComponent} from '../../commons/new';
 import View from '../../components/view';
 import Fader, {FaderPosition} from '../../components/fader';
-import {Constants} from 'helpers';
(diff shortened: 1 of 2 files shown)


src/incubator/WheelPicker/index.tsx
```diff
@@ -163,7 +164,13 @@ const WheelPicker = ({
     // https://github.com/facebook/react-native/issues/26661
     if (Constants.isAndroid && prevIndex.current !== index) {
       prevIndex.current = index;
-      onChange?.(items?.[index]?.value, index);
+      
+      if (prevInitialValue.current !== initialValue) {
+        // don't invoke 'onChange' if 'initialValue' changed
+        prevInitialValue.current = initialValue;
```
ethanshar, line 170: I wouldn't set the `prevInitialValue` value here just because `scrollToIndex` method is being called in several places.
Instead I'd set in the `useEffect` that handle the initialValue update (right after triggering the scroll)
```

ethanshar (commented):
Looks great! 
Wrote a single comment

src/incubator/WheelPicker/index.tsx
Inbal-Tish: Not sure I understan. In all other cases it should not pass the condition...

ethanshar: Yes but technically the responsibility for re-setting the `prevInitialValue` is in the relevant useEffect and not in the method it calls to. 
If someone will change the implementation of `scrollToIndex` they might break the functionality of `prevInitialValue`

Inbal-Tish: I don't see how it works... Let's talk when you're available

ethanshar (approved):

PR #1696, opened 2021-11-26
Feat/new toast

ethanshar:
## Description
**Please Review - Don't Merge!**
_This contains lots of files, most of them are assets and generated types. It'll be best to review this PR by commits_ 

This is the new implementation of our Toast component (taken from the private lib)
The component's code has been refactored to a functional component 
Im using separate helper hooks for different parts of logic of the component. 

Few alterations I did
- I added support for `offline` preset for the private usage we have (in private we deepened on the `isConnected` prop)
- I added support for containerStyle, messageStyle, iconColor to give more customize capabilities to our usets

## Changelog
New Toast component (under Incubator) with better UI and lots of new features


src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+    centerMessage,
+    showLoader,
+    action,
+    swipeable,
+    backgroundColor,
+    children,
+    testID
+  } = props;
+
+  const directions = useRef([
```
Inbal-Tish, line 40: Why do you need useRef on Array if you're not modifying it anywhere?

src/incubator/toast/helpers/useToastPresets.ts
```diff
@@ -0,0 +1,75 @@
+import 'react';
+import {Colors} from 'style';
+import {ToastProps, ToastPreset} from '../types';
+
+const checkMarkIcon = require('../assets/checkmarkFlat.png');
+const exclamationIcon = require('../assets/exclamationFill.png');
+const infoIcon = require('../assets/info.png');
+const redCloudIcon = require('../assets/redCloud.png');
+
+export default ({preset, icon, iconColor, message}: Pick<ToastProps, 'preset' | 'icon' | 'message' | 'iconColor'>) => {
```
Inbal-Tish, line 10: Why do you use `Pick<ToastProps, ...>` when you extract the props?

src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+  const _renderAttachment = (positionStyle: object, zIndex?: number) => {
+    return (
+      <View style={[positionStyle, {zIndex}]} pointerEvents={'box-none'}>
+        {renderAttachmentContent()}
+      </View>
+    );
+  };
+
+  const renderToast = () => {
+    if (!visible && !isAnimating) {
```
Inbal-Tish, line 185: Why this check is here as well? Is there a scenario where it will not return on line 209 but will return here?

src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+    if (!visible) {
+      clearTimer();
+    }
+
+    toggleToast(visible);
+  }, [visible]);
+
+  const onDismiss = useCallback(() => {
+    clearTimer();
+    props.onDismiss?.();
```
Inbal-Tish, line 88: Better use 'onDismiss' directly and not as 'props.onDismiss'

src/incubator/toast/index.tsx
ethanshar: Cause we're passing the directions to PanView and we don't want to create a new ref for the array on each render.

src/incubator/toast/helpers/useToastPresets.ts
ethanshar: Because I want the hook to have an accurate typecheck. 
Instead of passing all props, I declare only the ones relevant to this hook

src/incubator/toast/index.tsx
ethanshar: You are right, It was copied like that from the private code and I see that we have the same issue there as well.

src/incubator/toast/index.tsx
ethanshar: fixed

Inbal-Tish (approved):

lidord-wix (commented):
Looks good, the hooks are making a lot of sense :) I left few specific comments. 
The following are more generic:
1. It might be a bug in the example screen, but when dismissing the toast by sliding it down or up, you should press the `Toggle toast` button 3 times to present it again (instead of 1).
2. There's a bug in the attachment feature, you can see it when sliding the toast right/left (happens on private as well).
3. I think it is a better API to have a `customContent` prop instead of getting it as children, it'll also be more clear to us in the future.
4. Consider exporting the toast presets so users will be able to use `Toast.presets.SUCCESS`.

src/incubator/toast/helpers/useToastAnimation.ts
```diff
@@ -0,0 +1,64 @@
...
+    Animated.timing(toastAnimatedValue.current, {
+      toValue: Number(show),
+      duration: 300,
+      delay,
+      easing: Easing.bezier(0.215, 0.61, 0.355, 1),
+      useNativeDriver: true
+    }).start(_onAnimationEnd);
+
+    if (preset === ToastPreset.FAILURE && show) {
+      HapticService.triggerHaptic(HapticType.impactMedium, 'Toast');
```
M-i-k-e-l, line 45: I think this is a problem, the Haptic is an optional dependency, and the toast should be more generic.
We should either test here or create a function (or add an option to the current `triggerHaptic`) to only trigger if it's available and not trigger an error.

src/incubator/toast/helpers/useToastAnimation.ts
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  const isTop = position === 'top';
+  const positionMultiplier = isTop ? -1 : 1;
+  const toastTranslateY = toastAnimatedValue.current.interpolate({
+    inputRange: [0, 1],
+    outputRange: [positionMultiplier * toastHeight, 0]
+  });
+
+  return {isAnimating, toggleToast, toastOpacity, toastTranslateY};
```
M-i-k-e-l, line 63: Wouldn't exporting `opacityStyle` and `translationStyle` will be a little nicer? I mean to export them like this:
`{opacity: toastOpacity}` and `{transform: [{translateY: toastTranslateY}]}`.
And maybe `toggleToast` can be simply `toggle`.

src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const renderToast = () => {
+    const isTop = position === 'top';
+
+    return (
+      <>
+        {!isTop && !!toastHeight && renderAttachmentContent()}
+        <View
+          animated
+          useSafeArea
+          style={[{opacity: toastOpacity}, containerStyle]}
```
M-i-k-e-l, line 194: Shouldn't this be memoized?

src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+    if (height !== toastHeight) {
+      setToastHeight(height);
+    }
+  },
+  [toastHeight]);
+
+  const renderRightElement = () => {
+    // NOTE: order does matter
+    if (showLoader) {
+      const loaderColors = [Colors.rgba(Colors.white, 0.3), Colors.rgba(Colors.grey30, 0.7)];
```
M-i-k-e-l, line 112: Is it better to keep this here or move it to a global const?

src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+    props.position === 'bottom' ? PanView.directions.DOWN : PanView.directions.UP,
+    PanView.directions.LEFT,
+    PanView.directions.RIGHT
+  ]);
+
+  const viewRef = useRef();
+  const [toastHeight, setToastHeight] = useState(500);
+
+  const {clearTimer, setDismissTimer} = useToastTimer(props);
+  const {icon, iconColor, accessibilityMessage} = useToastPresets(props);
```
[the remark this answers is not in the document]
ethanshar: Readability wise, I think it's better.. 
If I'll extract the relevant props at the top I'll have to do something like this
```
  const {icon: presetIcon, iconColor: presetIconColor, accessibilityMessage} = useToastPresets({preset, icon, iconColor});
```

I'll need to give everything new names, not sure if it's more readable, WDYT?

src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+        const reactTag = findNodeHandle(viewRef.current);
+        AccessibilityInfo.setAccessibilityFocus(reactTag!);
+      } else if (message) {
+        AccessibilityInfo.announceForAccessibility?.(accessibilityMessage);
+      }
+    }
+  };
+
+  const {isAnimating, toggleToast, toastOpacity, toastTranslateY} = useToastAnimation({
+    ...props,
```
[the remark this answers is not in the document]
ethanshar: Here's it's easier cause Im returning different stuff.. 
Fixed

src/incubator/toast/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+        >
+          {message}
+        </Text>
+      </View>
+    );
+  };
+
+  const renderIcon = () => {
+    const tintColor = iconColor;
+    return <Image source={icon} resizeMode={'contain'} style={styles.icon} tintColor={tintColor}/>;
```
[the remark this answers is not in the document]
ethanshar: Definitely! Fixed 👌

src/incubator/toast/helpers/useToastPresets.ts
```diff
@@ -0,0 +1,75 @@
+import 'react';
+import {Colors} from 'style';
+import {ToastProps, ToastPreset} from '../types';
+
+const checkMarkIcon = require('../assets/checkmarkFlat.png');
+const exclamationIcon = require('../assets/exclamationFill.png');
+const infoIcon = require('../assets/info.png');
+const redCloudIcon = require('../assets/redCloud.png');
```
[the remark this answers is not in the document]
ethanshar: Im considering these assets part of the component so I added them under the component asset folder.

src/incubator/toast/helpers/useToastPresets.ts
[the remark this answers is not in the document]
ethanshar: What do you mean? can you elaborate..

src/incubator/toast/helpers/useToastTimer.ts
```diff
@@ -0,0 +1,21 @@
+import {useCallback, useRef} from 'react';
+import {ToastProps} from '../types';
+
+export default ({autoDismiss, onDismiss}: Pick<ToastProps, 'autoDismiss' | 'onDismiss'>) => {
+  const timer = useRef<ReturnType<typeof setTimeout>>();
+
+  const clearTimer = useCallback(() => {
```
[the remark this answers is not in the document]
ethanshar: Yes, you are right. It's a leftover from the previous code. 
Fixed 👍

src/incubator/toast/helpers/useToastAnimation.ts
```diff
@@ -0,0 +1,64 @@
+import {useRef, useState} from 'react';
+import {Animated, Easing} from 'react-native';
+import {HapticService, HapticType} from 'services';
+import {ToastPreset, ToastProps} from '../types';
+
+interface UseToastAnimationProps extends Pick<ToastProps, 'visible' | 'preset' | 'position' | 'onAnimationEnd'> {
+  toastHeight: number;
+  playAccessibilityFeatures: () => void;
+  setDismissTimer: () => void;
```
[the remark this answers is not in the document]
ethanshar: We only use this type here and it's pretty simple type.
If it was used in other places then maybe.. WYDT? I don't mind either way

src/incubator/toast/helpers/useToastAnimation.ts
ethanshar: Yes, you are right. 
I added a dedicate prop for that (`enableHapticFeedback`) and use that instead. 
In private we should enable it based on our logic

src/incubator/toast/helpers/useToastAnimation.ts
ethanshar: I agree. 
I changed to opacityStyle and translateStyle. 
As for the `toggleToast`, I actually like that it's explicit, I find it more readable

src/incubator/toast/index.tsx
ethanshar: Fixed

src/incubator/toast/index.tsx
ethanshar: I actually removed it all together, it belong to the private loader anyway

src/incubator/toast/index.tsx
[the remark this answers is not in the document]
lidord-wix: I think it even makes more sense that way, cause you understand that the iconColor, for example, has been changed by this hook.. and you also know exactly which props are affecting this hook. 
But it's up to you :)

src/incubator/toast/helpers/useToastPresets.ts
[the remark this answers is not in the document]
M-i-k-e-l: Something like this (see `useBannerPresets` in our private lib):
```
const toastPresets = {};
toastPresets[ToastPreset.SUCCESS] = {
  icon: checkMarkIcon,
  iconColor: icon || iconColor ? iconColor : Colors.green40,
  presetText: 'Success'
};
...
return toastPresets;
```

src/incubator/toast/helpers/useToastAnimation.ts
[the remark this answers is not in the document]
M-i-k-e-l: `setTimer` is used in both `useToastAnimation` and `useToastTimer`, but it's not a big deal.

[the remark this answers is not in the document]
ethanshar: It's exported in `useToastTimer`, but we declare its typing once in `useToastAnimation`

src/incubator/toast/helpers/useToastPresets.ts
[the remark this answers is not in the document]
ethanshar: Ok, I did something similar

src/incubator/toast/index.tsx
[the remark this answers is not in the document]
ethanshar: Ok, I changed it, tell me what you think..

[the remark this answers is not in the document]
lidord-wix: Looks great :)

lidord-wix:
> Looks good, the hooks are making a lot of sense :) I left few specific comments. The following are more generic:
> 
> 1. It might be a bug in the example screen, but when dismissing the toast by sliding it down or up, you should press the `Toggle toast` button 3 times to present it again (instead of 1).
> 2. There's a bug in the attachment feature, you can see it when sliding the toast right/left (happens on private as well).
> 3. I think it is a better API to have a `customContent` prop instead of getting it as children, it'll also be more clear to us in the future.
> 4. Consider exporting the toast presets so users will be able to use `Toast.presets.SUCCESS`.

@ethanshar did you fix these issues?

src/incubator/toast/index.tsx
```diff
@@ -165,8 +168,8 @@ const Toast = (props: PropsWithChildren<ToastProps>) => {
   };
 
   const renderIcon = () => {
-    const tintColor = iconColor;
-    return <Icon source={icon} resizeMode={'contain'} style={styles.icon} tintColor={tintColor}/>;
+    const tintColor = toastPreset.iconColor;
+    return <Icon source={toastPreset.icon} resizeMode={'contain'} style={styles.icon} tintColor={tintColor}/>;
```
M-i-k-e-l, line 172: Are you keeping the `const tintColor` for readability?

ethanshar: Nope (: 
Forgot it when I refactored the code. Fixed!

ethanshar:
> > > Looks good, the hooks are making a lot of sense :) I left few specific comments. The following are more generic:
> > > 
> > > 1. It might be a bug in the example screen, but when dismissing the toast by sliding it down or up, you should press the `Toggle toast` button 3 times to present it again (instead of 1).
> > > 2. There's a bug in the attachment feature, you can see it when sliding the toast right/left (happens on private as well).
> > > 3. I think it is a better API to have a `customContent` prop instead of getting it as children, it'll also be more clear to us in the future.
> > > 4. Consider exporting the toast presets so users will be able to use `Toast.presets.SUCCESS`.
> > 
> > 
> > @ethanshar did you fix these issues?
> 
> Not all, I'll go over them again.. thanks

@lidord-wix 
1. The first issue is more of an example screen issue, anyway I fixed it, you can take a look. 
2. Fixed
3. I prefer not to change API, I copied the code the private and I'd like to have a smooth transition to the pubic implementation without having a big migration
4. Done!

lidord-wix (approved):

M-i-k-e-l (approved):

PR #1712, opened 2021-12-07
Image - support SVG from uri

M-i-k-e-l:
## Description
Image - support SVG from uri
Ticket number: 2328

## Changelog
Image - support SVG from uri

### DIFF
--- a/demo/src/screens/componentScreens/ImageScreen.tsx
+++ b/demo/src/screens/componentScreens/ImageScreen.tsx
@@ -9,6 +9,8 @@ const BROKEN_URL = 'file:///Desktop/website/img/cupcake.jpg';
 const DEFAULT_SIZE = 100;
 
 const file = Assets.svgs.demo.logo;
+const uri = {uri: 'http://thenewcode.com/assets/images/thumbnails/homer-simpson.svg'};
+const uriWithCss = {uri: 'http://thenewcode.com/assets/svg/accessibility.svg'};
 const xml = `
   <svg width="32" height="32" viewBox="0 0 32 32">
     <path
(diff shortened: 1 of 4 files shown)


demo/src/screens/componentScreens/ImageScreen.tsx
```diff
@@ -38,6 +40,13 @@ enum SizeType {
   Percentage = '50%'
 }
 
+enum SvgType {
```
Inbal-Tish, line 43: Feels like this enum should be exported from the 'SvgImage' component... You can also use it in the component by first determining the source type and the have a switch case to the return instead of many 'if' statements - will be way cleaner

M-i-k-e-l: I don't think this should be in the `SvgImage`, the user does not **need** to be aware of these different types, I'm not sure if it's "wrong" to put the different type here or not, perhaps I should add some tests though.
Regarding the single return, generally I agree with it, but it is a matter of taste, and I feel like it is fine the way it is currently (since the type is not coming from the component).

Inbal-Tish: It's your call

Inbal-Tish (approved):

demo/src/screens/componentScreens/ImageScreen.tsx
M-i-k-e-l: For the record, I tried to add some tests, but:
1. There is no `onLoad` \ `onError` props in the SVG components.
2. Snapshot tests have issues as well: some random things fail them (not sure how to mock these), might need to mock timers etc.
3. Lastly I tried to use the (old method) of snapshot testing using `react-test-renderer`, again it did not work for me, since the remote images need to load somehow.

PR #1715, opened 2021-12-09
fix missing dialog props (such as onShow)

mendyEdri:
## Description
Fix missing spread props `pickerModalProps` passed to `Picker`

## Changelog
Fix missing `pickerModalProps` props spread passed to `Picker`


src/components/dialog/index.tsx
```diff
@@ -247,7 +247,7 @@ class Dialog extends Component<DialogProps, DialogState> {
 
   render = () => {
     const {modalVisibility} = this.state;
-    const {testID, supportedOrientations, accessibilityLabel, ignoreBackgroundPress} = this.props;
+    const {testID, supportedOrientations, accessibilityLabel, ignoreBackgroundPress, ...others} = this.props;
```
M-i-k-e-l, line 250: This should be done with `modalProps`:

```
  /**
   * Additional props for the modal.
   */
  modalProps?: ModalProps;
```

ethanshar: @M-i-k-e-l 
We are currently declaring that Dialog props are extending Modal props. (code wise, we don't do it)
If you want to go with adding `modalProps` which I agree is safer, we should fix the component typings as well

M-i-k-e-l: We have a `Pick` from `ModalPropsIOS`, or did I miss something? Generally we should remove that if we're adding the suggested prop, but I don't think it's worth the migration since we're going to move to the new `Dialog` soonish.

mendyEdri: Fixed :)

M-i-k-e-l (commented):
Please run `npm run build:dev` 🙏 
And it has conflicts :/

mendyEdri:
> Please run `npm run build:dev` 🙏 And it has conflicts :/

Fixed the conflict. 
The `ts` errors are not related to the this code, for example `react-freeze` is missing

M-i-k-e-l (approved):

PR #1727, opened 2021-12-16
Feat/incubator dialog add header

M-i-k-e-l:
## Description
Incubator.Dialog - add support for header (and its parts) and default style
This is not compatible to our private implementation, we will either handle it in the private lib or it'll require a migration.
I've tried to make the API more clear and less cumbersome, please make look at the API and let me know if it an be improved.

Related ticket: WOAUILIB-2317

## Changelog
Incubator.Dialog - add support for header (and its parts) and default style


src/incubator/Dialog/index.tsx
```diff
@@ -24,10 +47,36 @@ const Dialog = (props: DialogProps) => {
...
+        <DialogHeader {...headerProps}/>
+        {children}
+      </View>
+    </ImperativeDialog>
+  );
 };
 
 Dialog.displayName = 'Incubator.Dialog';
 Dialog.directions = DialogDirectionsEnum;
+Dialog.Header = DialogHeader;
```
Inbal-Tish, line 66: Why do you expose these components? Aren't they private and controlled by the 'dialogHeaderProps'?

M-i-k-e-l: The idea is to give a default header but allow the user to override, so we give them (and possibly us in private) the option use our basic components to build something more complex.

Inbal-Tish (approved):

PR #1729, opened 2021-12-19
Feat/slider thumbHitSlop prop

lidord-wix:
## Description
Add thumbHitSlop prop to Slider component (required for the AudioPlayer component)  
Only the first commit is a real change, the two others are for formatting and typings.

## Changelog
Add thumbHitSlop prop to Slider component

### DIFF
--- a/src/components/slider/index.tsx
+++ b/src/components/slider/index.tsx
@@ -114,7 +114,7 @@ export type SliderProps = {
   testID?: string;
 } & typeof defaultProps;
 
-interface SliderState {
+interface State {
   containerSize: Measurements;
   trackSize: Measurements;
   thumbSize: Measurements;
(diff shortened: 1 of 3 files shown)


src/components/slider/index.tsx
```diff
@@ -501,30 +510,22 @@ export default class Slider extends PureComponent<SliderProps, SliderState> {
...
-      thumbStyle,
-      disabled,
-      thumbTintColor
-    } = this.props;
+    const {thumbStyle, disabled, thumbTintColor} = this.props;
 
     return (
       <Animated.View
-        hitSlop={this.thumbHitSlop}
+        hitSlop={this.props.thumbHitSlop}
```
Inbal-Tish, line 520: Why not extract from 'this.props' in line 516?

src/components/slider/index.tsx
```diff
@@ -32,109 +33,114 @@ export type SliderOnValueChange = (value: number) => void;
...
+   * If true the component will have accessibility features enabled
+   */
+  accessible?: boolean;
   /**
    * The slider's test identifier
    */
   testID?: string;
 } & typeof defaultProps;
 
 interface SliderState {
```
Inbal-Tish, line 117: Should be renamed "State"

lidord-wix: Renamed

Inbal-Tish (approved):

PR #1747, opened 2021-12-30
Fix/ Hint max width

lidord-wix:
## Description
Limit hint width to screenWidth 
WOAUILIB-2374

## Changelog
Fix Hint max width


src/components/hint/index.tsx
```diff
@@ -613,7 +614,7 @@ const styles = StyleSheet.create({
     position: 'absolute'
   },
   hint: {
-    maxWidth: 400,
+    maxWidth: DEFAULT_MAX_WIDTH,
```
M-i-k-e-l, line 617: Technically it's not default, the user cannot override it; I'd even consider not adding a const:
```
maxWidth: Math.min(Constants.screenWidth - 2 * Spacings.s4, 400),
```

lidord-wix: done

M-i-k-e-l (approved):

PR #1752, opened 2022-01-02
Infra/ design tokens infra

lidord-wix:
## Description
Add designTokens infra

## Changelog
Add designTokens infra


src/style/designTokens.ts
```diff
@@ -0,0 +1,73 @@
+import {colorsPalette} from './colorsPalette';
+
+export const designTokens = {
+  ...colorsPalette,
```
ethanshar, line 4: I wouldn't include the system colors palette in the tokens, It might be good for us to have a clear separation between the two objects in the future. 
In the Colors class you can load the all together

src/style/colors.ts
```diff
@@ -11,7 +12,7 @@ export class Colors {
   [key: string]: any;
 
   constructor() {
-    const colors = Object.assign(colorsPalette, themeColors);
+    const colors = Object.assign(designTokens, themeColors);
```
ethanshar, line 15: See my previous comment on separating between the tokens and colorsPalette, here you can join them together instead of include the colorsPalette in the tokens

src/style/colors.ts
```diff
@@ -245,7 +246,7 @@ function threeDigitHexToSix(value: string) {
   return value.replace(/./g, '$&$&');
 }
 
-const TypedColors = Colors as ExtendTypeWith<typeof Colors, typeof colorsPalette & typeof themeColors>;
+const TypedColors = Colors as ExtendTypeWith<typeof Colors, typeof designTokens & typeof themeColors>;
 const colorObject = new TypedColors();
-colorObject.loadColors(colorsPalette);
+colorObject.loadColors(designTokens);
```
ethanshar, line 251: same here

ethanshar (approved):

PR #1756, opened 2022-01-05
feat/Checkbox_animationDuration_Prop

woen4:
## Description
Allow customize the checkbox animation duration

## Changelog
Checkbox - New prop `animationDuration`, receive value in miliseconds


src/components/checkbox/index.tsx
```diff
@@ -82,6 +83,11 @@ export interface CheckboxProps extends TouchableOpacityProps {
    * Additional styling for checkbox and label container
    */
   containerStyle?: StyleProp<ViewStyle>;
+  /**
+   * Duration of check animation (in milliseconds)
+   */
+  animationDuration?: number;
```
ethanshar, line 89: Hi @woen4 
I suggest adding a more generic prop - `animationOptions` and allow passing both `duration` and `easing`
WDYT?

woen4: I would be even better, i will implement

PR #1770, opened 2022-01-09
Feat/ Slider RTL

lidord-wix:
## Description
Add RTL support to Slider component
Add `disableRTL` prop to keep the slider LTR even if the app is in RTL mode.

## Changelog
Add RTL support to Slider component

### DIFF
--- a/src/components/slider/index.tsx
+++ b/src/components/slider/index.tsx
@@ -105,6 +105,10 @@ export type SliderProps = {
    */
   disabled?: boolean;
   /**
+   * If true the Slider will stay in LTR mode even if the app is on RTL mode
+   */
+  disableRTL?: boolean;
+  /**
    * If true the component will have accessibility features enabled
    */
   accessible?: boolean;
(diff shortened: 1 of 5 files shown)


src/components/slider/index.tsx
```diff
@@ -583,6 +592,7 @@ export default class Slider extends PureComponent<SliderProps, State> {
                 styles.track,
                 trackStyle,
                 styles.minimumTrack,
+                Constants.isRTL && this.props.disableRTL && styles.disableRTL,
```
ethanshar, line 595: destruct the prop at the start with the rest of the props

ethanshar, line 595: I suggest renaming `styles.disableRTL` to something more specific like `styles.trackDisabledRTL` just to separate it from the actual prop and also because it's probably relevant only to the track element style

lidord-wix: Done :)

lidord-wix: Done :)

ethanshar (approved):

PR #1775, opened 2022-01-11
Support custom item render in GridView

ethanshar:
## Description
Support renderCustomItem in GridView

## Changelog
Support  renderCustomItem prop in GridView

### DIFF
--- a/src/components/gridView/index.tsx
+++ b/src/components/gridView/index.tsx
@@ -47,6 +47,10 @@ export interface GridViewProps {
    * Ignored when passing 'maxItemWidth'
    */
   keepItemSize?: boolean;
+  /**
+   * Pass to render a custom item
+   */
+  renderCustomItem?: (item: GridListItemProps) => React.ReactElement;
 }
 
 interface GridViewState {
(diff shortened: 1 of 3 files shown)


src/components/gridView/index.tsx
```diff
@@ -47,6 +47,10 @@ export interface GridViewProps {
    * Ignored when passing 'maxItemWidth'
    */
   keepItemSize?: boolean;
+  /**
+   * Pass to render a custom item
+   */
+  renderCustomItem?: (item: GridListItemProps) => React.ReactElement;
```
M-i-k-e-l, line 53: This should be added to the api.

src/components/gridView/index.tsx
```diff
@@ -177,30 +181,39 @@ class GridView extends UIComponent<GridViewProps, GridViewState> {
...
+        },
+        !isLastRow && {
+          marginBottom: itemSpacing
+        },
+        item.containerStyle
+      ],
+      children: isLastItem ? this.renderLastItemOverlay() : undefined
+    };
+
+    return renderCustomItem ? renderCustomItem(itemProps) : <GridListItem {...itemProps}/>;
```
M-i-k-e-l, line 216: We don't usually send (as far as I remember) our configuration to the custom render function, is this a new approach or a special case?
Also, I feel like the name (`item` in the definition above) should be more along "suggested props" if this is how it's called, not sure how to call it though.

ethanshar: Not sure I understand the comment. 
What you mean by configuration? 

What I did is just pass the props I usually pass to `GridListItem` to the custom render so it can use it. 
Technically, we render in private a wrapper of GridListItem (check the private PR)

src/components/gridView/index.tsx
ethanshar: Added

LucasBerce71 (approved):

src/components/gridView/index.tsx
M-i-k-e-l: I totally missed the email from private (probably archived it, thinking it's the same one).

M-i-k-e-l (approved):

PR #1782, opened 2022-01-17
Infra/ Incubator.WheelPicker selectedValue

lidord-wix:
## Description
Deprecate Incubator.WheelPicker `selectedValue` prop in favor of `initialValue`
(All the relevant changes are in the first commit, the second is only about formatting) 
WOAUILIB-2308

## Changelog
BREAKING CHANGE - Deprecate Incubator.WheelPicker `selectedValue` prop in favor of `initialValue`

### DIFF
--- a/demo/src/screens/incubatorScreens/WheelPickerScreen.tsx
+++ b/demo/src/screens/incubatorScreens/WheelPickerScreen.tsx
@@ -2,8 +2,6 @@ import React, {useCallback, useState} from 'react';
 import {View, Text, Incubator, Colors, Typography, Button, Dialog} from 'react-native-ui-lib';
 import _ from 'lodash';
 
-type WheelPickerValue = Incubator.WheelPickerProps['initialValue'];
-
 const monthItems = _.map([
   'January',
   'February',
@@ -25,10 +23,7 @@ const yearItems = _.times(2030, i => i)
   .map(item => ({label: `${item}`, value: item}));
 const dayItems = _.times(31, i => i + 1).map(day => ({label: `${day}`, value: day}));
 
-const useData = (initialMonth?: string, initialYear?: string, initialDays?: number) => {
-  const [selectedMonth, setMonth] = useState<WheelPickerValue>(initialMonth);
-  const [, setYear] = useState<WheelPickerValue>(initialYear);
-  const [selectedDays, setDays] = useState<WheelPickerValue>(initialDays);
+export default () => {
   const [showDialog, setShowDialog] = useState(false);
 
   const onPickDaysPress = useCallback(() => {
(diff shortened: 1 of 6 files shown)


demo/src/screens/incubatorScreens/WheelPickerScreen.tsx
```diff
@@ -39,75 +34,33 @@ const useData = (initialMonth?: string, initialYear?: string, initialDays?: numb
...
           activeTextColor={Colors.primary}
           inactiveTextColor={Colors.grey20}
           items={monthItems}
           textStyle={Typography.text60R}
-          selectedValue={selectedMonth}
         />
 
         <Text h3>Years</Text>
         <Text bodySmall grey30>
           (Uncontrolled, initialValue passed)
```
Inbal-Tish, line 53: Why do we need this example? Looks like the days example so we don't really need both now, no?

src/incubator/WheelPicker/usePresenter.ts
```diff
@@ -21,21 +20,19 @@ type RowItem = {
...
-  initialValue,
-  selectedValue,
+  initialValue = 0,
   children,
   items: propItems,
   itemHeight,
   preferredNumVisibleRows
 }: PropTypes): Presenter => {
-  const value = !_.isUndefined(selectedValue) ? selectedValue : initialValue;
+  const value = initialValue;
```
Inbal-Tish, line 35: seems like 'value' is unnecessary const now

demo/src/screens/incubatorScreens/WheelPickerScreen.tsx
Inbal-Tish: I would add an example for changing 'initialValue' to show its effect, like a button that rolls the wheel by changing the prop's value.

src/incubator/WheelPicker/index.tsx
```diff
@@ -21,13 +21,9 @@ export enum WheelPickerAlign {
...
 export interface WheelPickerProps {
   /**
-   * Initial value (doesn't work with selectedValue)
+   * Initial value
    */
   initialValue?: ItemProps | number | string;
-  /**
-   * The current selected value
-   */
-  selectedValue?: ItemProps | number | string;
```
Inbal-Tish, line 30: This is a breaking change. I know we're in the incubator but I'm thinking that we might want to at least explain to the uses what's wrong. Maybe add a deprecated message or error them when we detect they passed this prop... or we relay on the ts errors alone?

demo/src/screens/incubatorScreens/WheelPickerScreen.tsx
lidord-wix: Not sure why we had the months example originally, maybe for string values.
Anyway, I've added two buttons for updating the years value with the `initialValue`.

src/incubator/WheelPicker/index.tsx
lidord-wix: We don't have any internal users using this feature, regarding the community - we will add it under `Breaking Changes` in the release notes (and they also can decide not to update to this version until they fix it on their side)

Inbal-Tish: SectionsWheelPicker is using this component and it is not in the incubator. If someone passed 'selectedValue' to one of the sections (type WheelPickerProps) it will be affected by this change. How do we deal with that?

lidord-wix: You are right, but there's no internal use for `SectionsWheelPicker` in one app (we wrap it in private with different components), and for the community it's the same answer as before - we'll update it in the release notes, the ts we'll let them know as well, and they can stay with the previous version until they'll migrate to the other implementation.

Inbal-Tish: We usually don't introduce breaking changes in a patch versions. Let's discuss this with Ethan. @ethanshar

ethanshar: It's ok, that's why we have the Incubator components.
We can merge this and as Lidor mention we'll add a breaking changes section for this (also mention the SectionWheelPicker component). 

Moreover, this breaking change do offer an easy alternative and the community can also lock version till they handle the change.

Inbal-Tish (approved):

PR #1785, opened 2022-01-18
Button tests with test kit

Dan-Kovalsky:
## Description
Create the first version of test kit for some basic components using react-native/testing-library.
Use it to write tests for the Button component

## Changelog
expose testkits for: Text, Image, Button


src/index.ts
```diff
@@ -91,20 +91,25 @@ export {default as TouchableOpacity, TouchableOpacityProps} from './components/t
 export {default as View, ViewProps} from './components/view';
 export {default as Wizard, WizardProps, WizardStepProps, WizardStepStates, WizardStepConfig, WizardStepsConfig} from './components/wizard';
 
+/* ===== TestKit ===== */
+export {default as TextTestKit} from './components/text/Text.driver';
+export {default as ImageTestKit} from './components/image/Image.driver';
+export {default as ButtonTestKit} from './components/button/Button.driver';
```
ethanshar, line 97: Because we haven't completed our typescript migration we're in a state where we maintain both `index.ts` and `index.js`. 
It's important to add the exports to the latter as well..

src/components/button/__tests__/index.kit.spec.js
```diff
@@ -0,0 +1,142 @@
+import React, {useState} from 'react';
```
ethanshar, line 1: This is the spec that checks the driver? maybe it's better to name the file `index.driver.spec.js`?

ethanshar (commented):
Looks great! 
I wrote 2 small comments.. 
Anyway, Im still not sure why there's that one test that fail.. maybe we can skip it for now and get back to it later on.

src/components/button/__tests__/index.kit.spec.js
Dan-Kovalsky: I don't think it right to say that it checks the driver.
more like, it checks the button using the driver.
I still not sure what is good name for this, so I will change like you say

ethanshar (approved):

lidord-wix:
@Dan-Kovalsky 
A recommendation for next time - run `npm run prepush` before pushing to uilib, it'll generate typing files (which were missing this time) and run the test :)

PR #1798, opened 2022-01-23
Fix/expandable section animation

lidord-wix:
## Description
Animate the layout after any change of the `expanded` prop.
WOAUILIB-2435

## Changelog
ExpandableSection - fix layout animation for changing in the `expanded` prop

### DIFF
--- a/src/components/expandableSection/index.tsx
+++ b/src/components/expandableSection/index.tsx
@@ -1,7 +1,8 @@
-import React, {useEffect} from 'react';
+import React from 'react';
 import {LayoutAnimation, StyleSheet} from 'react-native';
 import View from '../view';
 import TouchableOpacity from '../touchableOpacity';
+import {useDidUpdate} from 'hooks';
 
 export type ExpandableSectionProps = {
   /**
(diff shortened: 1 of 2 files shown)


src/components/expandableSection/index.tsx
```diff
@@ -35,11 +35,19 @@ export type ExpandableSectionProps = {
 function ExpandableSection(props: ExpandableSectionProps) {
   const {expanded, sectionHeader, children, top} = props;
 
+  const animate = () => {
+    LayoutAnimation.configureNext({...LayoutAnimation.Presets.easeInEaseOut, duration: 300});
```
M-i-k-e-l, line 39: If you add a log here you'll see two issues:
1. You have an animation on load, you can solve that by using `useDidUpdate` instead of `useEffect`.
2. You have two animations when pressing the header, I'm not sure this is a usage issue (the screen) or the component.

M-i-k-e-l: I feel like this duplicity is indicating a problem with the API, we should have had either the `onPress` or the `expanded` :/ 
Regarding the new layout animation, make sure you test Android, there was an issue with it last time I've tested it (pre release of 2.3).

M-i-k-e-l (approved):
Approved, just wanted you to see the comment, you can merge

PR #1799, opened 2022-01-23
disable scroll for less than 1 pixel

D10S60948:
## Description
should prevent situations where a friction of pixel enables scroll and hence causes the scrollview-fader appear with no reason


M-i-k-e-l:
Hi @D10S60948,
Thanks for your contribution, however the 1 pixel is very arbitrary IMO, moreover I'm not sure this is the place for the fix.
Could you please provide a link to a code where it breaks or even better a minimal snippet or a repro in our demo app?

D10S60948:
It's hard to reproduce. It happened inside the ActionSheet in very specific situation and only after the first render (after component was re-rendered it was fine).
Since it doesn't happen any more I will close this issue for now. If it will happen again I'll reopen.
Thank you.

PR #1804, opened 2022-01-25
Update async storage module

ekoeryanto:


M-i-k-e-l:
Hi @ekoeryanto,
Thank you for contributing!
We have a limitation on this from what I can tell because of internal deps.
Can you please elaborate why you want this changed? I don't see how it affects a user.

PR #1805, opened 2022-01-25
replace moment for smaller bundler

ekoeryanto:
## Description

## Changelog


M-i-k-e-l:
@ethanshar we probably cannot merge this as-is, but perhaps we can create an optional-dependency for this?

ethanshar:
@M-i-k-e-l 
Why not use this light package instead?

M-i-k-e-l:
TBH I did not test it because I assumed we still need `moment` in private anyway, so it'll just add another dependency and not really help us.
Another thing to consider is users will need to migrate and this should be done in a major version.

ethanshar:
> TBH I did not test it because I assumed we still need `moment` in private anyway, so it'll just add another dependency and not really help us.
> Another thing to consider is users will need to migrate and this should be done in a major version.

I think we should consider using the light-date package. For our public users it makes sense not to have them install moment
And for our internal case it's redundant since it's a light package anyway

M-i-k-e-l:
Hi @ekoeryanto,

I've opened a PR that takes into account our optional dependencies (#2544), thank you very much for this contribution, apologies for not merging as is.

PR #1807, opened 2022-01-26
support margin modifiers in SkeletonView

lidord-wix:
## Description
Support margin modifiers in SkeletonView
WOAUILIB-2175

## Changelog
Support margin modifiers in SkeletonView Component

### DIFF
--- a/src/components/skeletonView/index.tsx
+++ b/src/components/skeletonView/index.tsx
@@ -4,7 +4,7 @@ import {StyleSheet, Animated, Easing, StyleProp, ViewStyle, AccessibilityProps} 
 import {BorderRadiuses, Colors, Dividers, Spacings} from '../../style';
 import {createShimmerPlaceholder, LinearGradientPackage} from 'optionalDeps';
 import View from '../view';
-import {Constants, asBaseComponent} from '../../commons/new';
+import {Constants, asBaseComponent, BaseComponentInjectedProps, MarginModifiers} from '../../commons/new';
 import {extractAccessibilityProps} from '../../commons/modifiers';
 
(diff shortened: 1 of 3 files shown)


src/components/skeletonView/index.tsx
```diff
@@ -407,22 +410,66 @@ class SkeletonView extends Component<SkeletonViewProps, SkeletonState> {
...
-          />
-        );
-      });
+      return (
+        <View {...others}>
+          {_.times(times, index => {
+            const key = timesKey ? `${timesKey}-${index}` : `${index}`;
+            return (
+              <SkeletonView
+                modifiers={{}}
```
ethanshar, line 461: Can we pass null here instead?

lidord-wix: It throws a TS error:
`Type 'null' is not assignable to type 'ExtractedStyle'.ts(2322)
`
I can set it to null and add a `@ts-expect-error`, wdyt?

ethanshar: Where the `modifiers` prop typing comes from? 
What about passing `undefined` or is that ignored?

lidord-wix: Same error for `undefined` as well.
The type comes from `asBaseComponent`:
` modifiers: ReturnType<typeof Modifiers.generateModifiersStyle>;
`

ethanshar (approved):

src/components/skeletonView/index.tsx
ethanshar: Got it, ok. Approving..

PR #1813, opened 2022-01-30
WheelPicker - allow pass labels alignment

Inbal-Tish:
## Description
WheelPicker - allow pass labels alignment.

## Changelog
WheelPicker - allow pass labels alignment.


lidord-wix (commented):
Looks good, 
A small thing we need to fix - 
The following props should move to the `InternalProps`: (my fault I guess)
  ```
  fakeLabel?: string;
  fakeLabelStyle?: TextStyle;
  fakeLabelProps?: TextProps;
```
And the new `align` prop should move to the exported `ItemProps`.
Also, you can use the `WheelPickerAlign` enum for the typings of the new `align` prop.

Inbal-Tish:
> WheelPickerAlign

What does this have to do with the current PR? I don't want to mix things...
Regarding 'WheelPickerAlign' enum, I can only use it if I move it out of the WheelPicker, otherwise, I'll have require cycle...

Inbal-Tish:
> Looks good, A small thing we need to fix - The following props should move to the `InternalProps`: (my fault I guess)
> 
> ```
> fakeLabel?: string;
> fakeLabelStyle?: TextStyle;
> fakeLabelProps?: TextProps;
> ```
> 
> And the new `align` prop should move to the exported `ItemProps`. Also, you can use the `WheelPickerAlign` enum for the typings of the new `align` prop.

Ok. I see it's just a minor change - I'll add it.
Moved WheelPickerAlign to types file as well

lidord-wix (approved):

PR #1815, opened 2022-02-01
Infra/lodash babel plugin

lidord-wix:
## Description
Reduce bundle size using lodash babel plugin - https://github.com/lodash/babel-plugin-lodash 
As part of it, I was needed to migrate all the usages of `_.chain` to `_.flow`, since `_.chain` brings all lodash with it. 
**I recommend reviewing this PR by commit since one of the commits is only for formatting**
WOAUILIB-2485

## Changelog
Reduce bundle size using lodash babel plugin


M-i-k-e-l (commented):
A general comment, I see a lot of places where you wrote `arr` (mainly) or `str` and not used the original name, was that intentional or some autogeneration?

lidord-wix:
> A general comment, I see a lot of places where you wrote `arr` (mainly) or `str` and not used the original name, was that intentional or some autogeneration?

Actually, I preferred to use `arr | str | obj` as a parameter to make it clear that it's a pure function that I pass to `_.flow()`,
but if it's not clear I can change it.
I can also save those functions in a different consts and pass them that way if you find it more readable

M-i-k-e-l:
> Actually, I preferred to use `arr | str | obj` as a parameter to make it clear that it's a pure function that I pass to `_.flow()`, but if it's not clear I can change it.

I think that if you have a proper name (`items`) you should use that, in other places (`keys` of `props`) it might be ok to use `arr`.
This might be a matter of taste, so maybe we need anther dev's input on this, your choice.

> I can also save those functions in a different consts and pass them that way if you find it more readable

Tough one, it might be more readable but it might be less readable with wrong name(s), it will also make the code a bit longer, I think this is generally better (unless you have really good names and \ or repeating functions in the same class).

src/assets/Assets.ts
```diff
@@ -1,13 +1,16 @@
...
   if (a) {
-    _(b).keys().forEach((key) => {
-      // @ts-ignore
-      Object.defineProperty(a, key, Object.getOwnPropertyDescriptor(b, key));
-    });
+    _.flow(_.keys, arr =>
+      _.forEach(arr, key => {
+        // @ts-ignore
+        Object.defineProperty(a, key, Object.getOwnPropertyDescriptor(b, key));
+      }))(b);
```
ethanshar, line 13: Why did you change here if there's no _.chain?

ethanshar (commented):
Just FYI, there was no reason to change usages outside of the `src` files (: 
We only deliver the src files and we babel only them. 

Anyway, Have you checked If the exported files got their lodash imports changed?

src/assets/Assets.ts
lidord-wix: It's an unexplicit chain - https://lodash.com/docs/4.17.15#prototype-chain

lidord-wix:
> Just FYI, there was no reason to change usages outside of the `src` files (: We only deliver the src files and we babel only them.
> 
> Anyway, Have you checked If the exported files got their lodash imports changed?

Yes, it does change the imports :)

src/assets/Assets.ts
ethanshar: Wouldn't it be enough to just change it to `_.keys(b).forEach`? Would that work? 
I personally find the _.flow not very readable, so whenever possible, I'd use a different approach. 

In this case we can also use vanilla JS and do `Object.keys(b).forEach`

lidord-wix: Done.
I changed to vanilla JS where I found it more readable and also renamed the parameters of the methods as Miki suggested.

ethanshar (approved):

PR #1816, opened 2022-02-02
feat/ Overlay color intensity

AmitShwarts:
## Description
1. support full overlay color.
2. add opacity when using color different from white.

## Changelog
in component Overlay:
1. Update OverlayIntensityType - add type to support full opacity.
2. Changed method getStyleByType - refactor.
3. Add method getOpacity - to compute relevant opacity.

### DIFF
--- a/src/components/overlay/index.tsx
+++ b/src/components/overlay/index.tsx
@@ -18,7 +18,8 @@ const OVERLY_TYPES = {
 export enum OverlayIntensityType {
   LOW = 'low',
   MEDIUM = 'medium',
-  HIGH = 'high'
+  HIGH = 'high',
+  FULL = 'full'
 }
 
 export type OverlayTypeType = typeof OVERLY_TYPES[keyof typeof OVERLY_TYPES];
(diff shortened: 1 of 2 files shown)


M-i-k-e-l (commented):
Hi @AmitShwarts,
Thanks for the contribution.

The `full` intensity has an inconsistency:
If you select `overlayType={Image.overlayTypes.BOTTOM}` and `overlayIntensity={Image.overlayIntensityType.HIGH}` it will have higher intensity than `overlayType={Image.overlayTypes.BOTTOM}` and `overlayIntensity={Image.overlayIntensityType.FULL}`.
Also I am not sure what to actually expect in the second case, perhaps this feature should go through our UX first?

Another thing, in cases where a code is refactored, I like to add tests before the refactor and make sure nothing got broken.

src/components/overlay/index.tsx
```diff
@@ -52,6 +53,23 @@ class Overlay extends PureComponent<OverlayTypes> {
   static overlayTypes = OVERLY_TYPES;
   static intensityTypes = OverlayIntensityType;
 
+  getOpacity(intensity = this.props.intensity, color = this.props.color) {
```
M-i-k-e-l, line 56: Why is the default value required here?

AmitShwarts:
we found a different solution for this.
we can send color with opacity as #ff00007f. will be red with opacity 50%.

PR #1818, opened 2022-02-03
Fix/ Hint renders on wrong screen

lidord-wix:
## Description
Use TouchableWithoutFeedback instead of Modal to handle the Hint background press.
Checked on Android and on iOS, in the HintsScreen and in the ButtonsScreen.
WOAUILIB-2236

## Changelog
Fix hint renders on the wrong screen


src/components/hint/index.tsx
```diff
@@ -397,31 +394,31 @@ class Hint extends Component<HintProps, HintState> {
...
-  //           <TouchableWithoutFeedback style={[StyleSheet.absoluteFillObject]} onPress={onBackgroundPress}>
-  //             <View flex />
-  //           </TouchableWithoutFeedback>
-  //         )}
-  //       </View>
-  //     );
-  //   }
-  // }
+  renderOverlay() {
+    const {targetLayoutInWindow = {y: 0, x: 0}} = this.state;
```
ethanshar, line 398: Why are you setting a default value here? 
Also, it makes the check in line 400 redundant.. maybe it was intentional?

src/components/hint/index.tsx
```diff
@@ -549,40 +546,24 @@ class Hint extends Component<HintProps, HintState> {
...
   render() {
-    const {onBackgroundPress, backdropColor, testID} = this.props;
-
     if (!this.props.visible && this.state.hintUnmounted) {
       return this.props.children || null;
     }
 
     return (
       <>
         {this.renderChildren()}
```
ethanshar, line 555: I suggest keeping the old implementation commented, till we know for sure there are no issues. 
Maybe even support both with a new prop (`useModal` for example)

src/components/hint/index.tsx
```diff
@@ -549,40 +546,24 @@ class Hint extends Component<HintProps, HintState> {
...
-        {onBackgroundPress ? (
-          <Modal
-            visible={this.showHint}
-            animationType={backdropColor ? 'fade' : 'none'}
-            overlayBackgroundColor={backdropColor}
```
ethanshar, line 565: Notice that `backdropColor` was supported when using the Modal. 
We should make sure it is still supported 
Better check for other props as well

ethanshar (commented):
@lidord-wix I wrote a few comments.. 
Anyway this a big change, so we have to make sure the usages in One App are not affected in a bad way. 

I would create a one app version with this fix and get a list of modules who use Hint (I hope they are not many) and then we probably want them to verify the functionality of the Hint is still intact

src/components/hint/index.tsx
lidord-wix: Just because of the TS error, added a `ts-expect-error` for now

src/components/hint/index.tsx
lidord-wix: Done

src/components/hint/index.tsx
lidord-wix: Fixed, also added the animation support.

src/components/hint/index.tsx
```diff
@@ -572,8 +583,10 @@ class Hint extends Component<HintProps, HintState> {
             {this.renderHintContainer()}
           </Modal>
         ) : (
-          // this.renderOverlay(),
-          this.renderHintContainer()
+          <>
+            {onBackgroundPress && this.renderOverlay()}
```
ethanshar, line 587: No need to condition this use case with `onBackgroundPress` 
We take care of it in `renderOverlay` function, where we render the actual overlay. 
Technically, the user can still have a hint with a background overlay

src/components/hint/index.tsx
```diff
@@ -397,31 +403,36 @@ class Hint extends Component<HintProps, HintState> {
...
+              left: containerPosition.left - targetLayoutInWindow.x,
+              backgroundColor: backdropColor,
+              opacity: this.visibleAnimated
+            }
+          ]}
+          pointerEvents="box-none"
+          testID={`${testID}.overlay`}
+        >
+          {onBackgroundPress && (
+            <TouchableWithoutFeedback style={[StyleSheet.absoluteFillObject]} onPress={onBackgroundPress}>
```
ethanshar, line 428:
```suggestion
            <TouchableWithoutFeedback style={StyleSheet.absoluteFillObject} onPress={onBackgroundPress}>
```

src/components/hint/index.tsx
ethanshar: In that case it's better to fix the actual issue. 
In this case `ts-expect-error` would never be solved, cause it's our code. If anything you can use `ts-ignore`
But I think that in this case we just fix it. 

You can declare the type of the `targetLayoutInWindow` in the state. 
You can use casting to keep it simple, something like that

```
  state = {
    targetLayoutInWindow: undefined as {x: number, y: number, width: number, height: number} | undefined,
    targetLayout: this.props.targetFrame,
    hintUnmounted: !this.props.visible
  };
```

ethanshar (approved):

PR #1821, opened 2022-02-06
Update version of react-native-reanimated to 2.4.0

yedidyak:
## Description
Update version of react-native-reanimated (2.4.0)

## Changelog
Update internally versions of react-native-reanimated and react-native-gesture-handler


M-i-k-e-l:
For future reference a few workarounds were aded:
1. `onMomentumScrollEnd={() => {}}` workaround for `useAnimatedScrollHandler.onMomentumEnd` not being called (Android) (https://github.com/software-mansion/react-native-reanimated/issues/2735)
2. `moduleNameMapper` in `package.json` see this: https://github.com/software-mansion/react-native-reanimated/issues/1196
3. `interpolateColor` mock see this https://github.com/software-mansion/react-native-reanimated/issues/2749

src/components/tabController/PageCarousel.tsx
```diff
@@ -103,6 +101,7 @@ function PageCarousel({...props}) {
       scrollEventThrottle={16}
       contentOffset={initialOffset} // iOS only
       onLayout={scrollToInitial} // Android only
+      onMomentumScrollEnd={() => {}} // workaround for useAnimatedScrollHandler.onMomentumEnd not being called (https://github.com/software-mansion/react-native-reanimated/issues/2735)
```
lidord-wix, line 104: It'll override the `onMomentumScrollEnd` prop that the user might send, no?

lidord-wix (commented):
@M-i-k-e-l 
I wrote small comment

src/components/tabController/PageCarousel.tsx
M-i-k-e-l: This is VERY strange, if the user sends `onMomentumScrollEnd` the workaround stops working (even though this is overriding it).

lidord-wix (approved):

src/components/tabController/PageCarousel.tsx
M-i-k-e-l: Fixed by adding `useCallback`

PR #1823, opened 2022-02-07
Feat/snippets

Inbal-Tish:
## Description
Snippets - adding "snippet" key to the api.json

## Changelog
Snippets - adding "snippet" key to the api.json

### DIFF
--- a/src/components/actionBar/actionBar.api.json
+++ b/src/components/actionBar/actionBar.api.json
@@ -24,5 +24,14 @@
       "default": "true"
     },
     {"name": "style", "type": "ViewStyle", "description": "Component's style"}
+  ],
+  "snippet": [
+    "<ActionBar",
+    "  actions={[",
+    "    {label: $1, onPress: $2},",
+    "    {label: $3, onPress: $4},",
+    "    {label: $5, onPress: $6}",
+    "  ]}",
+    "/>"
   ]
 }
--- a/src/components/actionSheet/actionSheet.api.json
+++ b/src/components/actionSheet/actionSheet.api.json
@@ -71,5 +71,18 @@
       "description": "In iOS, use safe area, in case component attached to the bottom"
     },
     {"name": "testID", "type": "string", "description": "The test id for e2e tests"}
+  ],
+  "snippet": [
+    "<ActionSheet",
+    " title={$1}",
+    " message={$2}",
+    " cancelButtonIndex={$3}",
+    " destructiveButtonIndex={$4}",
+    " options={[",
+    "  {label: '$5', onPress: $6},",
+    "  {label: '$7', onPress: $8},",
+    "  {label: 'Cancel', onPress: $9}",
+    " ]}",
+    "/>"
   ]
 }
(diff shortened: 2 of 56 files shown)


ethanshar (commented):
Looks great! I've been meaning to add something similar (: 

Anyway ,most of the snippets are pretty hollow IMO, I think we should have a team effort to make them a bit more detailed.

Anyway, is there a way to convert the placeholders (e.g `$1`) in the snippets to a meaningful values? 

It'll be great if for example we can have a snippet like this
`<Card backgroundColor={Colors.blue30} />`
instead of 
`<Card backgroundColor={} />`

Inbal-Tish:
> Looks great! I've been meaning to add something similar (:
> 
> Anyway ,most of the snippets are pretty hollow IMO, I think we should have a team effort to make them a bit more detailed.
> 
> Anyway, is there a way to convert the placeholders (e.g `$1`) in the snippets to a meaningful values?
> 
> It'll be great if for example we can have a snippet like this `<Card backgroundColor={Colors.blue30} />` instead of `<Card backgroundColor={} />`

The `$` signed are used in the code snippets extension to mark the position of a cursor and move it along using the tab key. Do we prefer to give that up in favor of a value or present the value in the code snippet extension alongside the cursor?

ethanshar:
> > Looks great! I've been meaning to add something similar (:
> > Anyway ,most of the snippets are pretty hollow IMO, I think we should have a team effort to make them a bit more detailed.
> > Anyway, is there a way to convert the placeholders (e.g `$1`) in the snippets to a meaningful values?
> > It'll be great if for example we can have a snippet like this `<Card backgroundColor={Colors.blue30} />` instead of `<Card backgroundColor={} />`
> 
> The `$` signed are used in the code snippets extension to mark the position of a cursor and move it along using the tab key. Do we prefer to give that up in favor of a value or present the value in the code snippet extension alongside the cursor?

Yes, I think the cursor thing is a nice to have..
But personally as a user, I'd prefer to have a more practical example (both in the docs and snippet extension) than a placeholder that Im not sure what to put there.

Inbal-Tish:
> > > Looks great! I've been meaning to add something similar (:
> > > Anyway ,most of the snippets are pretty hollow IMO, I think we should have a team effort to make them a bit more detailed.
> > > Anyway, is there a way to convert the placeholders (e.g `$1`) in the snippets to a meaningful values?
> > > It'll be great if for example we can have a snippet like this `<Card backgroundColor={Colors.blue30} />` instead of `<Card backgroundColor={} />`
> > 
> > 
> > The `$` signed are used in the code snippets extension to mark the position of a cursor and move it along using the tab key. Do we prefer to give that up in favor of a value or present the value in the code snippet extension alongside the cursor?
> 
> Yes, I think the cursor thing is a nice to have.. But personally as a user, I'd prefer to have a more practical example (both in the docs and snippet extension) than a placeholder that Im not sure what to put there.

Ok. I'll add a value before the $ that's how we'll have both an example and the cursor mark

PR #1833, opened 2022-02-10
TextField: "isFocused" method added

D10S60948:
## Description
Add the ability to check if the `TextField` (`TextInput`) is currently focused.

## Changelog
Add missing `isFocused` method on `Incubator.TextField` ref


ethanshar (commented):
@D10S60948 
I see that the PR build checks fail
Can you please run `npm run prepush` and see what fails?

D10S60948:
I don't think that the build's fail is related to the change I made.
It seems related to a tsc comment on `AvatarHelper`.
Screen Shot 2022-02-10 at 17 53 57

ethanshar:
> I don't think that the build's fail is related to the change I made.
> It seems related to a tsc comment on `AvatarHelper`.
> Screen Shot 2022-02-10 at 17 53 57

 Your PR branch was a little old, I sync it with master. should be ok now

ethanshar (approved):

PR #1839, opened 2022-02-13
Marquee - new component

Inbal-Tish:
## Description
Marquee - new component.

## Changelog
Marquee - new component.


ethanshar:
@Inbal-Tish 
Can we call this component `Marquee` like the one there is in Web. 
https://www.w3schools.in/html-tutorial/marquee-tag/

Inbal-Tish:
> @Inbal-Tish Can we call this component `Marquee` like the one there is in Web. https://www.w3schools.in/html-tutorial/marquee-tag/

Sure.

lidord-wix (commented):
Looks really nice!
Few general comments before I'll get deep into the code: 
- Please add an `api.json` file for the component.
- I think the directions should be flipped - it makes more sense to use `right` when I want the sliding view to slide right, and `left` when I want it to slide left.  
- There's an issue when I press the back button from the example screen and I still see the sliding views on the main screen for a second. it also happens from the opposite side - when getting into the example screen from the main screen, I see the animations before the screen is fully presented.
- Consider adding the following two modes (and you can choose better names for them):
  - `wheel mode` - where the sliding view is presented again before it's completely disappeared (like a vertical wheel).
  - `repeat mode` - where the slide goes right until the right edge and then goes back left until the left edge, again and again (like in the Spotify app)

Inbal-Tish:
> Looks really nice! Few general comments before I'll get deep into the code:
> 
> * Please add an `api.json` file for the component.
> * I think the directions should be flipped - it makes more sense to use `right` when I want the sliding view to slide right, and `left` when I want it to slide left.
> * There's an issue when I press the back button from the example screen and I still see the sliding views on the main screen for a second. it also happens from the opposite side - when getting into the example screen from the main screen, I see the animations before the screen is fully presented.
> * Consider adding the following two modes (and you can choose better names for them):
>   
>   * `wheel mode` - where the sliding view is presented again before it's completely disappeared (like a vertical wheel).
>   * `repeat mode` - where the slide goes right until the right edge and then goes back left until the left edge, again and again (like in the Spotify app)

I fixed the first 3 comments but regarding the modes, I'll wait for the UX requirements before adding any more functionality, etc.

demo/src/screens/componentScreens/MarqueeScreen.tsx
```diff
@@ -0,0 +1,34 @@
...
+        })}
+      </View>
+    );
+  }
+  
+  render() {
+    return (
+      <View flex padding-20 style={{overflow: 'hidden'}}>
+        <Text text50 grey10 marginB-20>Marquee</Text>
+        <Marquee duration={2000}>
```
lidord-wix, line 20: The component should inherit its parent bounds, in this case for example, the component should have `padding-20`

lidord-wix (commented):
Another few comments :)
Please run the `build:dev` script to generate the typings file for `marquee` instead of for `slidingView`

src/components/marquee/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  direction?: Directions;
+  width?: number;
+}
+
+const Marquee = (props: MarqueeProps) => {
+  const {children, duration = 6000, direction = 'right', width = Constants.screenWidth} = props;
+  
+  const isRight = direction === 'right';
+  const from = isRight ? -width : Constants.screenWidth;
+  const to = isRight ? Constants.screenWidth : -width;
```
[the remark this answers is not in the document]
Inbal-Tish: Done

demo/src/screens/componentScreens/MarqueeScreen.tsx
Inbal-Tish: ok

lidord-wix: Now the sliding view is in the right width, but it's still running all over the screen width.
You'll be able to see what I'm talking about if you'll change the screen padding to 120 for example, then the time you wait after it gets out of the screen and until it enters again is too long.

Inbal-Tish: I'm not going to adjust the animation to the container. The user will have to pass the 'width' prop for that

lidord-wix: It doesn't work, try the following example: 
```
        <View bg-red60 width={50}>
          <Marquee width={50}>
            <Text>
              Test
            </Text>
          </Marquee>
        </View>
```

Inbal-Tish: Of course it won't work - if you wrap the component in a container of width 50 it can't move in it, they both have the same size... Give the container width 150 and see the Marqee move in it. The width passed to the Marquee is for calculating the animation (default is screen width)

lidord-wix: I meant that I do see the animation inside the container, but the time I should wait after it gets out of the container and until it enters again is too long because the animation runs all over the screen width (even though we can't see it).
The user can't control that, we have to handle it from the component. 
The animation should be 'flexed' in its container.

Inbal-Tish: When passing the 'width' prop the user controls the animation - the animation will run by the width passed and not the screen width. For example, if a user has a container with 20 paddings he should pass a width of screenWidth - 40.

Inbal-Tish: You know what, I'm closing this PR till I get specific requirements.

lidord-wix: I don't think we have to close it, it is well implemented.
We can ask Diana about this small issue so we can all get on board.

Inbal-Tish: @lidord-wix Since this is a component made without any ux requirements or a real use case I can work with, I'm implementing a basic full-screen-width component. Your comments for extra requirements are missing the point of this small guild-week task and are not in its scope. I do not wish to continue implementing requirements that might not be necessary.

lidord-wix: It is not mentioned in the ticket (and maybe that's the problem) but the first use case is the AudioPlayer component, the title of the audio player should use marquee when its title is too long.

Inbal-Tish: Please attach the guidelines to the ticket

PR #1840, opened 2022-02-13
Feat/dialog improvements

M-i-k-e-l:
## Description
Some improvements to the new Dialog (some from bugs opened and others taken from private components for general use):
Remove extra components from API.
Support `useCustomTheme`.
Add `FadedScrollView` - auto scroll according to content + fader
Support `width` \ `height` props, including percentage (string)

Note: I do not like the implementation for supporting percentage too much, but I did not find a better way to make it work properly.

Consider looking at each commit separately, since at least one is only a refactor.

## Changelog
Incubator.Dialog - support `scrollableProps` (add autoscroll according to content size) + `width` \ `height` props


M-i-k-e-l:
I'll split this up

PR #1842, opened 2022-02-13
Infra/generify faded scroll view

M-i-k-e-l:
## Description
FadedScrollView - make more generic (I will use this in `Incubator.Dialog`, splitting up in to a few PRs).
Look at each commit separately for convenience.

## Changelog
None


src/components/tabController/FadedScrollView.tsx
```diff
@@ -1,47 +1,80 @@
...
+  useGesture?: boolean;
+  children?: React.ReactNode | React.ReactNode[];
+};
 
 type Props = FadedScrollViewProps & ForwardRefInjectedProps;
-
-const FADER_SIZE = 76;
+interface Statics {
+  scrollTo(
+    y?: number | {x?: number | undefined; y?: number | undefined; animated?: boolean | undefined},
```
lidord-wix: Looks like the types are not in the right place here

src/components/tabController/FadedScrollView.tsx
```diff
@@ -1,47 +1,80 @@
...
   } = useScrollReached({
-    horizontal: true,
+    horizontal,
     threshold: 50
   });
 
-  const showStart = scrollEnabled && !isScrollAtStart;
-  const showEnd = scrollEnabled && !isScrollAtEnd;
+  const showStart = (showStartFader && scrollEnabled && !isScrollAtStart) ?? false;
+  const showEnd = (showEndFader && scrollEnabled && !isScrollAtEnd) ?? false;
```
lidord-wix: Why do you need the `?? false` in these two lines?

lidord-wix (commented):
Looks good, I wrote two small comments :)

src/components/tabController/FadedScrollView.tsx
M-i-k-e-l: Not sure what you mean, can you elaborate \ rephrase?

src/components/tabController/FadedScrollView.tsx
M-i-k-e-l: The default for the `Fader` is to be visible, I want it to not be visible unless it's needed.

src/components/tabController/FadedScrollView.tsx
lidord-wix: Can the `y` key get both number and object?

lidord-wix (approved):

PR #1845, opened 2022-02-14
Incubator.Dialog - header refactor

M-i-k-e-l:
## Description
Incubator.Dialog - header refactor

## Changelog
Incubator.Dialog - header refactor


src/incubator/Dialog/DialogHeader.tsx
```diff
@@ -1,30 +1,42 @@
...
 const DialogHeader = (props: DialogHeaderProps = {}) => {
   const {text = {}, renderContent, showKnob = true, showDivider = true, ...others} = props;
 
+  const knob = useMemo(() => {
+    if (showKnob) {
+      return <View style={styles.knob}/>;
+    }
+  }, [showKnob]);
+
   const _renderContent = useMemo(() => {
```
Inbal-Tish, line 19: I would change "_renderContent" to "headerContent" to be extra clear and follow the other component's naming

M-i-k-e-l: Missed that, 10x
Maybe I should rename the prop as well 🤔

M-i-k-e-l: But it's under `headerProps`, WDYT?

Inbal-Tish: I think the prop name is good

Inbal-Tish (approved):

PR #1846, opened 2022-02-14
Feat/incubator dialog support size and custom header

M-i-k-e-l:
## Description
Incubator.Dialog - support `width`, `height` (including percentage) and `customHeader`.
Add `FadedScrollView` - auto scroll according to content + fader.
I've added `height={'40%'}` to the screen, you can change to a number \ to `width` etc.
As we've discussed, I am not very fond of the `useSizeStyle` solution, but did not find a better solution for the percentage part.

## Changelog
Incubator.Dialog - support `width`, `height` (including percentage), `customHeader` and `FadedScrollView` (auto scroll according to content + fader).


src/incubator/Dialog/helpers/useSizeStyle.tsx
```diff
@@ -0,0 +1,65 @@
+import {isUndefined} from 'lodash';
```
ethanshar, line 1: Is this hook meant to convert string percentages to numbers? 
Cause react-native already supports string percentages

src/incubator/Dialog/index.tsx
```diff
@@ -23,14 +39,34 @@ const Dialog = (props: DialogProps) => {
...
+      return customHeader;
+    } else {
+      return <DialogHeader {...headerProps}/>;
+    }
+  }, [headerProps, customHeader]);
+
+  const renderContent = () => {
+    if (enable) {
+      return (
+        <FadedScrollView {...otherScrollableProps} useGesture showEndFader endFaderProps={FADER_PROPS}>
```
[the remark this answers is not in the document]
M-i-k-e-l: I think we had this request at some point, and we have this functionality in private's ActionSheet, I thought it'll be better to implement here for all Dialogs. What's the drawback?

src/incubator/Dialog/helpers/useSizeStyle.tsx
M-i-k-e-l: This is what I've talked with you about; RN does have this functionality, but it does not work because of the different views in the Dialog, I did not find a way to not break other things and have this work 😞

ethanshar: Let's go over it together maybe we can find a solution

PR #1848, opened 2022-02-14
Infra/ Move all colors and DesignTokens to public

lidord-wix:
## Description
Move all colors and DesignTokens to public
Add `supportDarkMode` method to `Colors` class to allow enabling the dark mode tokens from private when the appearance changes 
WOAUILIB-2534

## Changelog
Move all colors and DesignTokens to public

### DIFF
--- a/src/style/colors.ts
+++ b/src/style/colors.ts
@@ -60,8 +60,8 @@ export class Colors {
    * Support listening to Appearance changes
    * and change the design tokens accordingly 
    */
-  supportDarkMode(enable = false) {
-    const designTokensColors = enable && Scheme.getSchemeType() === 'dark' ? designTokensDM : designTokens;
+  enableDarkMode() {
+    const designTokensColors = Scheme.getSchemeType() === 'dark' ? designTokensDM : designTokens;
     Object.assign(this, designTokensColors);
   }
   
--- a/src/style/colors.ts
+++ b/src/style/colors.ts
@@ -60,7 +60,7 @@ export class Colors {
    * Support listening to Appearance changes
    * and change the design tokens accordingly 
    */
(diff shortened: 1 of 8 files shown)


src/style/colorsPalette.ts
```diff
@@ -86,6 +86,8 @@ const colorsPalette = {
   purple70: '#f1daf7',
   purple80: '#faebfd',
   // VIOLETS
+  violet1: '#170E5C',
+  violet5: '#231690',
```
M-i-k-e-l, line 90: Where are these from?

src/style/colors.ts
```diff
@@ -55,6 +56,15 @@ export class Colors {
     Scheme.setScheme(scheme);
   }
 
+  /**
+   * Support listening to Appearance changes
+   * and change the design tokens accordingly 
+   */
+  supportDarkMode(enable = false) {
```
M-i-k-e-l, line 63: This is not clear:
1. Why is there a default for `false`?
2. If you call this (I assume in private), then I am not sure about the name, why is it `support` and not `set`?

src/style/colorsPalette.ts
lidord-wix: Alexey generated them :)

src/style/colors.ts
lidord-wix: I changed it a little bit, no parameter needed, and I renamed it to `enableDarkMode`

src/style/colorsPalette.ts
M-i-k-e-l: I did not see them in the docs 🤷‍♂️

M-i-k-e-l (approved):

PR #1849, opened 2022-02-15
Fix/keyboard tracking view ios

Inbal-Tish:
## Description
KeyboardTrackingView - iOS - fix View's translation when nested in a TabBarController.
Verified on iPhone 12 iOS 14.4 and iPhone 8 iOS 13.7

## Changelog
KeyboardTrackingView - iOS - fix View's translation when nested in a TabBarController.


ethanshar (approved):

ethanshar:
@Inbal-Tish 
Couple of things regarding this fix. 
- It's a fix in our native code, that means we need to advance the the version of `uilib-native` - We have a dedicated package (under `lib/package.json`) because of the engine and the fact they responsible for all native versions. 
- We need of course to update this version in the engine

Inbal-Tish:
> * lib/package.json

Thanks! I added a version bump to this PR. I'll open a PR to the engine and add it to the ticket

PR #1850, opened 2022-02-15
TabController - move to design tokens

M-i-k-e-l:
## Description
TabController - move to design tokens
I did change the color slightly in two places so we won't need to create a new token (this is aligned with private, so it seems ok).

## Changelog
TabController - move to design tokens

### DIFF
--- a/src/components/tabController/TabBar.tsx
+++ b/src/components/tabController/TabBar.tsx
@@ -15,7 +15,7 @@ import {orientations} from '../../commons/Constants';
 import {useDidUpdate} from 'hooks';
 
 const DEFAULT_HEIGHT = 48;
-const DEFAULT_BACKGROUND_COLOR = Colors.white;
+const DEFAULT_BACKGROUND_COLOR = Colors.$backgroundElevated;
 
 const DEFAULT_LABEL_STYLE = {
   ...Typography.text80M,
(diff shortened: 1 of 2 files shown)


src/components/tabController/TabBar.tsx
```diff
@@ -317,19 +317,19 @@ const styles = StyleSheet.create({
...
     width: 70,
     height: 2,
-    backgroundColor: Colors.primary
+    backgroundColor: Colors.$backgroundPrimary
   },
   containerShadow: {
     ...Platform.select({
       ios: {
-        shadowColor: Colors.grey10,
+        shadowColor: Colors.black,
```
lidord-wix, line 325: Why black?

src/components/tabController/TabBar.tsx
```diff
@@ -317,19 +317,19 @@ const styles = StyleSheet.create({
     left: 0,
     width: 70,
     height: 2,
-    backgroundColor: Colors.primary
+    backgroundColor: Colors.$backgroundPrimary
```
lidord-wix, line 320: Do you mean `backgroundPrimaryHeavy`?

lidord-wix (commented):
Looks good, two small comments :) 
And - I know it's a different component, but we should think about how we'll handle the fader color.

src/components/tabController/TabBar.tsx
M-i-k-e-l: Private has this as `black`, and it removed the need for a new token, WDYT?

src/components/tabController/TabBar.tsx
M-i-k-e-l: Indeed

src/components/tabController/TabBar.tsx
lidord-wix: But this way it'll be `black` in all modes.
Maybe `$backgroundNeutralHeavy` will do the work here since it's dark in light mode, and light in dark mode..

M-i-k-e-l: We still don't have a solution for shadows AFAIK, so if we're not handling this in private, why should this be different?

lidord-wix: Ohh so you just align the color to the private one?

M-i-k-e-l: Not really align everything, but this specifically because otherwise it's a new token for this edge case...

lidord-wix: I see, ok..

lidord-wix (approved):

PR #1851, opened 2022-02-15
Remove old docs files

ethanshar:
## Description
Remove old docs files

## Changelog
Remove old docs files


lidord-wix (commented):
Our current private docs are based on these docs, if we'll delete them we'll break the private docs site, no?

ethanshar:
Technically they are based on the npm package `uilib-docs` which still exist 
They won't break, if we'll need to fix something we won't be able.
But the private docs are not maintained anyway... the next step is to move to Docusaurus (:

lidord-wix (approved):

PR #1853, opened 2022-02-16
feat/Sortable grid view

GuyHaviv37:
## Description
Guild week project - following [this JIRA ticket](https://jira.wixpress.com/browse/WOAUILIB-2510).
This component is a drag-and-sort, currently it accepts an array of items - which should have some unique ids and a render function - which will render the items (so items are all props needed to render + id).
The component will calculate a grid based on the complementary props (numOfColumns, itemSpacing, viewWidth).
Each item will be rendered in its assigned grid item (its "square space").
Each grid item is accompanied by a pan gesture handler to enable drag-and-sort.
You can also drag-to-scroll within the ScrollView ! i.e., drag to screen bottom/top to continue the gesture.
Margins (called itemSpacing to align with GridView) are highly recommended to allow non-gesture scrolls.

## Changelog
Component: ```SortableGridView```


src/components/sortableGridView/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <View flex key={`${item.id} - ${index}`}>
+          {renderItem(item)}
+        </View>
+      </SortableGridItemAnimationWrapper>
+    );
+  };
+    
+  return (
+    <GestureHandlerRootView>
+      <Animated.ScrollView
```
ethanshar, line 77: What's the need for ScrollView? (Also why is It Animated.ScrollView?)
Shouldn't the user decide whether to wrap it with ScrollView outside of the component?

src/components/sortableGridView/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+     * Pass the desired grid view width (default is screen wide)
+     */
+    viewWidth?: number;
+}
+
+const SortableGridView: React.FC<SortableGridViewProps> = (props) => {
+  const {items, renderItem, itemSpacing, numOfColumns = DEFAULT_NO_OF_COLUMNS, viewWidth = WINDOW_WIDTH} = props;
+  const scrollViewRef = useAnimatedRef<Animated.ScrollView>();
+  const scrollY = useSharedValue(0);
+  const itemSize = useMemo(() => getItemSize(numOfColumns, viewWidth), [viewWidth, numOfColumns]);
```
ethanshar, line 43: Can't we use internally the GridView component with its `renderCustomItem` prop to render the items here?

src/components/sortableGridView/SortableGridItemAnimationWrapper.tsx
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
```
ethanshar, line 1: I think we can simplify the name and call this component `SortableGridItem`

src/components/sortableGridView/SortableGridItemAnimationWrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+      zIndex,
+      transform: [
+        {translateX: translateX.value},
+        {translateY: translateY.value},
+        {scale}
+      ]
+    };
+  });
+  return (
+    <Animated.View style={[ABSOLUTE_ITEM, animatedStyle]}>
```
ethanshar, line 116: Since we're using now gesture-handler v2, have you considered using its new API 
https://docs.swmansion.com/react-native-gesture-handler/docs/quickstart/quickstart

src/components/sortableGridView/SortableGridItemAnimationWrapper.tsx
GuyHaviv37: np

src/components/sortableGridView/SortableGridItemAnimationWrapper.tsx
GuyHaviv37: I did consider it, but:

- there's not a lot of documentation about it just yet
- I like that in the older API I have context built in, no need for another sharedValue.
- The new API is overpowered to what I managed to do (which is simpler).

Tbh, the new API might be key to making this a truly usable component.
The main issue I think it has is scrolling vs. dragging UX wise.
I would love if this component could have logic such that items are draggable only after a long-press , but I just couldn't make it work - I think the new API can solve it (with some manual gesture handler), but as I said - there's not much documentation or examples so I couldn't make it work during a GW where I learned a lot from scratch.
If this is something that you could follow up on (uilib team) that'd be awesome.

PR #1856, opened 2022-02-16
Infra/ add lodash plugin

lidord-wix:
## Description
Add lodash plugin to reduce bundle size 
Checked on private with version `6.10.0-snapshot.654`

## Changelog
Add lodash plugin to reduce bundle size


M-i-k-e-l (approved):
Wonder if `"node": 6` would have worked 🤔

PR #1857, opened 2022-02-17
Infra/Settings and playground screens migrations

M-i-k-e-l:
## Description
Settings and playground screens migrations

## Changelog
Settings and playground screens migrations


lidord-wix (commented):
Why only those two? 😅
We have many places we use the old TextField...

M-i-k-e-l:
> Why only those two? 😅 We have many places we use the old TextField...

These bothered me in my D2D 🤷‍♂️ 
Feel free to create a task for the rest (or do migrate them).

lidord-wix (approved):

PR #1867, opened 2022-02-24
Fix/ add tintColor to tabbar gradient

lidord-wix:
## Description
Fix the fader color in dark mode TabBar 

## Changelog
Fix the fader color in dark mode TabBar

### DIFF
--- a/src/components/tabController/TabBar.tsx
+++ b/src/components/tabController/TabBar.tsx
@@ -117,7 +117,7 @@ interface Props extends TabControllerBarProps, BaseComponentInjectedProps, Forwa
   children?: ChildProps[] | ChildProps;
 }
 
-const FADER_PROPS = {size: 76};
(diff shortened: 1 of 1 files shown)


src/components/tabController/TabBar.tsx
```diff
@@ -117,7 +117,7 @@ interface Props extends TabControllerBarProps, BaseComponentInjectedProps, Forwa
   children?: ChildProps[] | ChildProps;
 }
 
-const FADER_PROPS = {size: 76};
+const FADER_PROPS = {size: 76, tintColor: Colors.$backgroundElevated};
```
M-i-k-e-l, line 120: I feel like this is not as prominent as the regular scheme, probably because there's less contrast with the text color.
Perhaps this should be `Colors.$backgroundDefault`?

lidord-wix: Agree :)

M-i-k-e-l (approved):

PR #1870, opened 2022-02-27
Migrate Picker implementation to TS and use Function component

ethanshar:
## Description

## Changelog

### DIFF
--- a/src/components/picker/new.tsx
+++ b/src/components/picker/new.tsx
@@ -1,0 +1,330 @@
+// TODO: deprecate all places where we check if _.isPlainObject
+// TODO: deprecate getItemValue prop
+// TODO: deprecate getItemLabel prop
+// TODO: Add initialValue prop
+// TODO: consider deprecating renderCustomModal prop
+// TODO: deprecate onShow cause it's already supported by passing it in pickerModalProps
+import _ from 'lodash';
+import React, {useMemo, useState, useRef, useEffect, PropsWithChildren} from 'react';
+import {LayoutChangeEvent} from 'react-native';
+import {
+  Constants,
+  asBaseComponent,
+  forwardRef,
+  ForwardRefInjectedProps,
+  BaseComponentInjectedProps
+} from '../../commons/new';
+import {LogService} from '../../services';
+// import View from '../../components/view';
+// import Modal from '../modal';
+import ExpandableOverlay, {ExpandableOverlayProps} from '../../incubator/expandableOverlay';
+// import Button from '../../components/button';
+// @ts-expect-error
+import {TextField} from '../inputs';
+import TextFieldMigrator from '../textField/TextFieldMigrator';
+// @ts-expect-error
+import NativePicker from './NativePicker';
+// @ts-expect-error
+import PickerModal from './PickerModal';
+// @ts-expect-error
+import PickerItem from './PickerItem';
+// @ts-expect-error
+import PickerContext from './PickerContext';
+// @ts-expect-error
+import {getItemLabel as getItemLabelPresenter, shouldFilterOut} from './PickerPresenter';
+import {PickerProps, PickerValue, PickerSingleValue, PickerMultiValue, PickerModes, PickerSearchStyle} from './types';
+
+const PICKER_MODES = {
+  SINGLE: 'SINGLE',
+  MULTI: 'MULTI'
+};
+
+const Picker = (props: PropsWithChildren<PickerProps> & ForwardRefInjectedProps & BaseComponentInjectedProps) => {
+  const {
+    mode,
+    selectionLimit,
+    showSearch,
+    searchStyle,
+    searchPlaceholder,
+    renderCustomSearch,
+    useNativePicker,
+    renderPicker,
+    customPickerProps,
+    containerStyle,
+    testID,
+    onChange,
+    onShow,
+    renderCustomModal,
+    forwardedRef,
+    modifiers,
+    enableModalBlur,
+    topBarProps,
+    pickerModalProps,
+    listProps,
+    value,
+    editable,
+    getLabel,
+    getItemLabel,
+    getItemValue,
+    renderItem,
+    children,
+    migrate,
+    migrateTextField
+  } = props;
+
+  const [selectedItemPosition, setSelectedItemPosition] = useState(0);
+  const [items] = useState(Picker.extractPickerItems(props));
+  const [multiDraftValue, setMultiDraftValue] = useState(value as PickerMultiValue);
+  const [multiFinalValue, setMultiFinalValue] = useState(value as PickerMultiValue);
+  const [searchValue, setSearchValue] = useState('');
+  // const [showExpandableModal, setShowExpandableModal] = useState(false);
+
+  const pickerExpandable = useRef<typeof ExpandableOverlay>();
+
+  useEffect(() => {
+    if (mode === Picker.modes.SINGLE && Array.isArray(value)) {
+      LogService.warn('Picker in SINGLE mode cannot accept an array for value');
+    }
+    if (mode === Picker.modes.MULTI && !Array.isArray(value)) {
+      LogService.warn('Picker in MULTI mode must accept an array for value');
+    }
+
+    // TODO: this warning should be replaced by the opposite
+    // we should warn user NOT to pass an object to the value prop
+    // if (props.useNativePicker && _.isPlainObject(props.value)) {
+    //   console.warn('UILib Picker: don\'t use object as value for native picker, use either string or a number');
+    // }
+    if (_.isPlainObject(value)) {
+      LogService.warn('UILib Picker will stop supporting passing object as value in the next major version. Please use either string or a number as value');
+    }
+  }, []);
+
+  const getAccessibilityInfo = () => {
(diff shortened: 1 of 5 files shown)


src/components/picker/PickerItem.tsx
```diff
@@ -30,20 +30,20 @@ const PickerItem = props => {
...
 
   const isItemDisabled = useMemo(() => {
-    return disabled || (!isSelected && context.selectionLimit && context.selectionLimit === selectedCounter);
+    return !!(disabled || (!isSelected && context.selectionLimit && context.selectionLimit === selectedCounter));
   }, [selectedCounter]);
-  
+
   useEffect(() => {
     if (_.isPlainObject(value)) {
       LogService.warn('UILib Picker.Item will stop supporting passing object as value & label (e.g {value, label}) in the next major version. Please pass separate label and value props');
```
Inbal-Tish, line 49: Why leave this log here? You have it in the Picker's hook...

src/components/picker/PickerItem.tsx
```diff
@@ -71,7 +71,10 @@ const PickerItem = props => {
   const _renderItem = () => {
     return (
       <View style={styles.container} flex row spread centerV>
-        <Text numberOfLines={1} style={[styles.labelText, isItemDisabled && styles.labelTextDisabled, labelStyle]}>
+        <Text
+          numberOfLines={1}
+          style={[styles.labelText, isItemDisabled ? styles.labelTextDisabled : undefined, labelStyle]}
```
Inbal-Tish, line 76: style should move to useMemo with dependency on `isItemDisabled` and `labelStyle`

src/components/picker/PickerItemsList.tsx
```diff
@@ -0,0 +1,96 @@
+// TODO: This should be renamed to PickerOverlayContent cause it's
```
Inbal-Tish, line 1: Why is this TODO not included in this refactor? What's holding it back?

src/components/picker/PickerItemsList.tsx
```diff
@@ -0,0 +1,96 @@
...
+// not responsible for rendering the Modal anymore
+import _ from 'lodash';
+// import PropTypes from 'prop-types';
+import React, {useCallback} from 'react';
+import {StyleSheet, FlatList, TextInput} from 'react-native';
+import {Typography, Colors} from '../../style';
+import Assets from '../../assets';
+// import {BaseComponent} from '../../commons';
```
Inbal-Tish, line 9: Why leave the comments here and in line 4?

src/components/picker/helpers/usePickerMigrationWarnings.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const {value, mode} = props;
+  useEffect(() => {
+    if (mode === PickerModes.SINGLE && Array.isArray(value)) {
+      LogService.warn('Picker in SINGLE mode cannot accept an array for value');
+    }
+    if (mode === PickerModes.MULTI && !Array.isArray(value)) {
+      LogService.warn('Picker in MULTI mode must accept an array for value');
+    }
+
+    // TODO: this warning should be replaced by the opposite
```
Inbal-Tish, line 18: This TODO is confusing, please leave only the relevant warning and rewrite the comment if still relevant

src/components/picker/new.tsx
```diff
@@ -0,0 +1,266 @@
+// TODO: deprecate all places where we check if _.isPlainObject
+// TODO: deprecate getItemValue prop
+// TODO: deprecate getItemLabel prop
+// TODO: Add initialValue prop
+// TODO: consider deprecating renderCustomModal prop
+// TODO: deprecate onShow cause it's already supported by passing it in pickerModalProps
```
Inbal-Tish, line 6: 1) Why change `_.isPlainObject` usage?
2) `getItemValue` and `getItemLabel` deprecation started, so we can remove comments.
3) Why not start with the deprecation of `onShow`?

src/components/picker/new.tsx
```diff
@@ -0,0 +1,266 @@
...
+import PickerItemsList from './PickerItemsList';
+import PickerItem from './PickerItem';
+import PickerContext from './PickerContext';
+import usePickerSelection from './helpers/usePickerSelection';
+import usePickerLabel from './helpers/usePickerLabel';
+import usePickerSearch from './helpers/usePickerSearch';
+import usePickerMigrationWarnings from './helpers/usePickerMigrationWarnings';
+import {PickerProps, PickerValue, PickerModes, PickerSearchStyle} from './types';
+
+const PICKER_MODES = {
```
Inbal-Tish, line 32: There is a `PickerModes` enum in types, why hold another one?

src/components/picker/new.tsx
```diff
@@ -0,0 +1,266 @@
...
+          onDone: mode === Picker.modes.MULTI ? () => onDoneSelecting(multiDraftValue) : undefined
+        }}
+        showSearch={showSearch}
+        searchStyle={searchStyle}
+        searchPlaceholder={searchPlaceholder}
+        onSearchChange={_onSearchChange}
+        renderCustomSearch={renderCustomSearch}
+        listProps={listProps}
+        // onShow={onShow}
+        // pickerModalProps={pickerModalProps}
```
Inbal-Tish, line 179: Please remove all unused props

src/components/picker/new.tsx
```diff
@@ -0,0 +1,266 @@
...
+            containerStyle={[paddings, margins, positionStyle, containerStyle]}
+            {...accessibilityInfo}
+            importantForAccessibility={'no-hide-descendants'}
+            value={label}
+            selection={Constants.isAndroid ? {start: 0} : undefined}
+            /* Disable TextField expandable feature */
+            topBarProps={undefined}
+            // expandable={false}
+            // renderExpandable={_.noop}
+            // onToggleExpandableModal={_.noop}
```
Inbal-Tish, line 239: Please remove all unused props

src/components/picker/types.tsx
```diff
@@ -0,0 +1,229 @@
...
+   */
+  label: string;
+  /**
+   * Item's label style
+   */
+  labelStyle?: StyleProp<TextStyle>;
+  /**
+   * Custom function for the item label (e.g (value) => customLabel)
+   */
+  getItemLabel: PickerProps['getItemLabel'];
```
Inbal-Tish, line 179: Is `getItemLabel` prop not deprecated here as well?

src/index.js
```diff
@@ -138,7 +138,7 @@ export default {
     return require('./components/panningViews/panResponderView').default;
   },
   get Picker() {
-    return require('./components/picker').default;
+    return require('./components/picker/new').default;
```
Inbal-Tish, line 141: Where are you routing to the old picker?

Inbal-Tish:
@ethanshar Looks good. I had only small comments. Also, please fix example screens, there are examples with old API that produce warnings.

src/components/picker/PickerItem.tsx
ethanshar: This warning related to the Picker.Item props. 
The hook is only responsible for the Picker props. 
Even though these warnings are related, because people render them separately I need to warn each component separately.

src/components/picker/PickerItem.tsx
ethanshar: Fixed

src/components/picker/PickerItemsList.tsx
ethanshar: Already renamed to PickerItemsList instead, removed the TODO

src/components/picker/new.tsx
ethanshar: 1. For TS, when using comparison like this (`typeof value === 'object'`), Typescript actually interpret the right type of the `value` prop. When using lodash unfortunately it doesn't. 
2. They didn't really started, I just added the @deprecated annotation long time ago. We need to address it accordingly. 
3. That's not the purpose of this PR, notice that most of the TODOs here are old. 

The main purpose of this PR is migrate to TS and function component, which is big as it is, I didn't want to add more stuff.

src/components/picker/new.tsx
ethanshar: Removed

src/components/picker/types.tsx
ethanshar: Not at the moment. 
Anyway, the whole migration of `getItemLabel` is not relevant to this PR

src/index.js
ethanshar: I'm not, it's routing to the new

ethanshar:
> @ethanshar Looks good. I had only small comments. Also, please fix example screens, there are examples with old API that produce warnings.

The example still have old usages so we'll know we didn't break anything while making changes, in the end of the screen there's an example with a migrated usage

src/index.js
Inbal-Tish: So you're keeping the old files as a reference?

Inbal-Tish: Ok. I suggest opening tickets for both completing this migration (and deleting old files) and starting the props migrations needed.

Inbal-Tish (approved):

PR #1872, opened 2022-02-27
Design tokens - add outlineDark

M-i-k-e-l:
## Description
Design tokens - add outlineDark

## Changelog
Design tokens - add outlineDark


lidord-wix (commented):
Which component needs it? 
Since the color of it is grey20 in both modes we can use the system color and not a token (I see that Alexey also wrote it in the tokens sheet)

lidord-wix:
> Which component needs it? Since the color of it is grey20 in both modes we can use the system color and not a token (I see that Alexey also wrote it in the tokens sheet)

Ohh I see, It's for the TabController, I think we can use the system color, wdyt?

M-i-k-e-l:
> Ohh I see, It's for the TabController, I think we can use the system color, wdyt?

But it's not, it has different `styles` for `dark theme` and regular (grey60 - outlineNeutral).

PR #1888, opened 2022-03-11
Infra/picker screen ts

ethanshar:
## Description
Migrate example PickerScreen to TS

## Changelog
Migrate example PickerScreen to TS


Inbal-Tish:
@ethanshar Looks fine. Is there a reason to keep Picker.d.ts in typings/components? Why not use only types.ts in Picker dir?

typings/globalTypes.d.ts
```diff
@@ -1,5 +1,8 @@
-type Constructor<T, TA = any> = new(...args: TA[]) => T;
-type ExtendTypeWith<T extends Constructor<any>, OtherObject extends object> = Constructor<InstanceType<T> & OtherObject, ConstructorParameters<T>>;
+type Constructor<T, TA = any> = new (...args: TA[]) => T;
+type ExtendTypeWith<T extends Constructor<any>, OtherObject extends object> = Constructor<
```
Inbal-Tish, line 2: I don't see why break the line here...

ethanshar: Reverted

Inbal-Tish (approved):

PR #1889, opened 2022-03-13
SkeletonView - adding 'fullWidth' prop

Inbal-Tish:
## Description
SkeletonView - adding 'fullWidth' prop.

## Changelog
SkeletonView - adding 'fullWidth' prop.


M-i-k-e-l (commented):
The code change for this feature seems a little too big for its value.
I also think the same functionality can be achieved by sending `shimmerStyle={{width: '100%'}}` which we can support.
**But more importantly** it seems that both solutions do not actually work, I think that the background is gray and not an actual shimmer effect (you can test this by sending a numeric `width` and enlarging the `height`).

Inbal-Tish:
> The code change for this feature seems a little too big for its value. I also think the same functionality can be achieved by sending `shimmerStyle={{width: '100%'}}` which we can support. **But more importantly** it seems that both solutions do not actually work, I think that the background is gray and not an actual shimmer effect (you can test this by sending a numeric `width` and enlarging the `height`).

I don't understand what's not working...See my video:

https://user-images.githubusercontent.com/33805983/161737058-f5810e49-5ebd-4d49-8328-a90ffb7cd29c.mov

PR #1893, opened 2022-03-14
Infra/ revese color palettes in dark mode

lidord-wix:
## Description
Reverse color palettes in dark mode
WOAUILIB-2585

## Changelog
Reverse color palettes in dark mode

### DIFF
--- a/src/style/colors.ts
+++ b/src/style/colors.ts
@@ -11,6 +11,7 @@ import Scheme, {Schemes, SchemeType} from './scheme';
 
 export class Colors {
   [key: string]: any;
+  shouldSupportDarkMode = false;
 
   constructor() {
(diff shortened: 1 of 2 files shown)


src/style/colors.ts
```diff
@@ -143,6 +147,24 @@ export class Colors {
     return this.getTintedColorForDynamicHex(color, tintKey);
   }
 
+  getInvertedTintKey(tintKey: string | number) {
```
ethanshar, line 150: can I offer a code improvement (: 

You can create an array of our keys (`[1, 5, 10, 20, 30, ..., 80]`) 
Then use `array.indexOf` to detect the given key index, revert the index (as you did before) and return the inverted key..

lidord-wix: done :)

ethanshar (approved):

Issue #1879, opened 2022-03-04
GridView example throws error


MoSattler:
The [GridView example](https://wix.github.io/react-native-ui-lib/docs/components/layoutsAndTemplates/GridView#usage) throw an error: 

> is invalid hex color 

the prop `lastItemOverlayColor` needs to be specified to get it to work.

PR #1900, opened 2022-03-16
Fix issue with GridView lastItemOverlayColor error when undefined

ethanshar:
## Description
Fix issue #1879
- GridView throws an error when passing underfed to lastItemOverlayColor but still passing `lastItemLabel`
- I also removed old code that was copied from private that shouldn't be in the public version 

## Changelog
Fix issue #1879 - Fix error thrown by GridView when passing passing `lastItemLabel` but not `lastItemOverlayColor`

### DIFF
--- a/src/components/gridView/index.tsx
+++ b/src/components/gridView/index.tsx
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import React from 'react';
 import {StyleSheet} from 'react-native';
-import {Colors, Spacings} from 'style';
+import {Spacings} from 'style';
 // TODO: we should use asBaseComponent here instead of using UIComponent directly
 import UIComponent from '../../commons/UIComponent';
 import View from '../view';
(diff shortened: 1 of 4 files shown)


src/components/gridView/index.tsx
```diff
@@ -166,12 +155,7 @@ class GridView extends UIComponent<GridViewProps, GridViewState> {
 
     const imageBorderRadius = _.flow(_.first, item => _.get(item, 'imageProps.borderRadius'))(items);
     return (
-      <View
-        style={[
-          styles.overlayContainer,
-          {backgroundColor: Colors.rgba(overlayColor, 0.6), borderRadius: imageBorderRadius}
-        ]}
-      >
+      <View style={[styles.overlayContainer, {backgroundColor: lastItemOverlayColor, borderRadius: imageBorderRadius}]}>
```
Inbal-Tish, line 158: Here you are changing the ui (it's not breaking change, but a UI breaking change), since the user now needs to pass the color with the alpha, while we did it internally before.

ethanshar: I agree. 
I'm fine with breaking them since it's pretty small, moreover it's a pretty new component and it appear we had issues there anyway. 
We had old code that depended on place color logic, which obviously not relevant to the component code. 
And generally I think it's wrong to decide the alpha for them, therefor they should pass the color as they wish. 

I don't think this small break worth having a migration for it..

Inbal-Tish (approved):

PR #1901, opened 2022-03-16
Revert lib version back to 3.0.4

Inbal-Tish:
## Description
Revert lib version back to 3.0.4

## Changelog
Revert lib version back to 3.0.4


ethanshar:
@Inbal-Tish 
Changing a package version doesn't really revert the changes. 
Anyway, closing this for now. 
As discussed, let's investigate and release a fix when we can

PR #1912, opened 2022-03-20
Fix/ migrate callback refs to object

lidord-wix:
## Description
Migrate callback refs to object, to resolve TS errors

## Changelog
Migrate callback refs to object, to resolve TS errors

### DIFF
--- a/src/components/slider/index.tsx
+++ b/src/components/slider/index.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, {PureComponent, ReactElement, ElementRef} from 'react';
+import React, {PureComponent, ReactElement} from 'react';
 import {
   StyleSheet,
   PanResponder,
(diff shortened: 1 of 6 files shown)


src/components/slider/index.tsx
```diff
@@ -305,12 +305,12 @@ export default class Slider extends PureComponent<SliderProps, State> {
       const nonOverlappingTrackWidth = trackSize.width - this.initialThumbSize.width;
       const _x = Constants.isRTL && disableRTL ? nonOverlappingTrackWidth - x : x; // adjust for RTL
       this._thumbStyles.left = trackSize.width === 0 ? _x : (_x * nonOverlappingTrackWidth) / trackSize.width; // do not render above prefix\suffix icon\text
-      this.thumb.setNativeProps(this._thumbStyles);
+      this.thumb.current?.setNativeProps(this._thumbStyles);
     }
 
     if (this.minTrack) {
```
ethanshar, line 311: You might need to check that `this.minTrack.current` is not undefined and not just `this.minTrack`

src/components/slider/index.tsx
```diff
@@ -327,7 +327,7 @@ export default class Slider extends PureComponent<SliderProps, State> {
 
       const activeOrInactiveStyle = !this.props.disabled ? (start ? activeStyle : style) : {};
       this._thumbStyles.style = _.omit(activeOrInactiveStyle, 'height', 'width');
-      this.thumb.setNativeProps(this._thumbStyles);
+      this.thumb.current?.setNativeProps(this._thumbStyles);
```
ethanshar, line 330: Same here, in the check above you probably need to check `this.thumb.current` 
Technically `this.thumb` is never undefined cause it is `{current: undefined}` when initialized

ethanshar (changes requested):

ethanshar (approved):

PR #1917, opened 2022-03-23
Avatar and Text - design tokens

M-i-k-e-l:
## Description
Avatar - design tokens

WOAUILIB-2596

## Changelog
Avatar - design tokens


src/components/avatar/index.tsx
```diff
@@ -407,7 +407,7 @@ function createStyles(props: AvatarProps) {
       lineHeight: undefined
     },
     ribbon: {
-      backgroundColor: Colors.primary,
+      backgroundColor: Colors.$backgroundPrimaryLight,
```
lidord-wix, line 410: I think it should be `$backgroundPrimaryHeavy`, no?

lidord-wix (commented):
Small comment, also please add `$textDefault` to the screen titles so it'll change in dark mode

src/components/avatar/index.tsx
M-i-k-e-l: Indeed

lidord-wix:
> Small comment, also please add `$textDefault` to the screen titles so it'll change in dark mode

@M-i-k-e-l I think you missed the comment about the screen titles :)

M-i-k-e-l:
> @M-i-k-e-l I think you missed the comment about the screen titles :)

Fixed

lidord-wix (approved):

PR #1935, opened 2022-04-04
TabController - design tokens - fixes

M-i-k-e-l:
## Description
TabController - design tokens - fixes

## Changelog
TabController - design tokens - fixes


src/components/tabController/TabBar.tsx
```diff
@@ -117,7 +117,7 @@ interface Props extends TabControllerBarProps, BaseComponentInjectedProps, Forwa
   children?: ChildProps[] | ChildProps;
 }
 
-const FADER_PROPS = {size: 76, tintColor: Colors.$backgroundDefault};
+const FADER_PROPS = {size: 76, tintColor: Colors.$backgroundElevated};
```
lidord-wix, line 120: Are you sure about it?
I used the `backgroundElevated` originally and you said it looks better with `backgroundDefault`

lidord-wix (commented):
Small comment

src/components/tabController/TabBar.tsx
M-i-k-e-l: It'll be `$backgroundElevated` but will be handled within the `Fader`

lidord-wix: Are you sure that the default fader color should be `$backgroundElevated` and not `$backgroundDefault`? 
maybe it should be `$backgroundDefault` and only here we should override it with `$backgroundElevated`..

M-i-k-e-l: That's what Yulia said in the relevant channel (can't link here).
In any case, if there will be errors we can fix them later.

M-i-k-e-l: Changed

lidord-wix (approved):

PR #1937, opened 2022-04-04
Infra/ migrate chip to designTokens

lidord-wix:
## Description
migrate chip to designTokens

## Changelog
migrate chip to designTokens

### DIFF
--- a/src/components/chip/index.tsx
+++ b/src/components/chip/index.tsx
@@ -145,7 +145,7 @@ const Chip = ({
   borderRadius = BorderRadiuses.br100,
   containerStyle,
   onDismiss,
-  dismissColor,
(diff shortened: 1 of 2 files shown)


src/components/chip/index.tsx
```diff
@@ -202,7 +203,7 @@ const Chip = ({
       >
         <Icon
           source={dismissIcon}
-          tintColor={dismissColor}
+          tintColor={dismissColor || Colors.$iconDefault}
```
Inbal-Tish, line 206: Why not set`Colors.$iconDefault` as the 'dismissColor' prop's default value?

lidord-wix: Fixed :)

Inbal-Tish (approved):

PR #1943, opened 2022-04-05
Infra/ migrate Slider and Stepper to design tokens

lidord-wix:
## Description
Migrate Slider and Stepper to design tokens

## Changelog
Migrate Slider and Stepper to design tokens

### DIFF
--- a/demo/src/screens/componentScreens/SliderScreen.tsx
+++ b/demo/src/screens/componentScreens/SliderScreen.tsx
@@ -71,16 +71,16 @@ export default class SliderScreen extends Component<SliderScreenProps, SliderScr
             minimumValue={-100}
             maximumValue={100}
             value={-30}
-            minimumTrackTintColor={Colors.$backgroundDangerHeavy}
-            thumbTintColor={Colors.$backgroundDangerHeavy}
+            minimumTrackTintColor={Colors.red30}
+            thumbTintColor={Colors.red50}
             containerStyle={styles.slider}
           />
           <Slider
             minimumValue={-300}
             maximumValue={-100}
             value={-130}
-            minimumTrackTintColor={Colors.$backgroundDangerHeavy}
-            thumbTintColor={Colors.$backgroundDangerHeavy}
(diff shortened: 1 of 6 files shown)


demo/src/screens/componentScreens/SliderScreen.tsx
```diff
@@ -59,41 +59,39 @@ export default class SliderScreen extends Component<SliderScreenProps, SliderScr
...
+          <Text $textDefault marginT-30>
+            Negatives
+          </Text>
           <Slider
             minimumValue={-100}
             maximumValue={100}
             value={-30}
-            minimumTrackTintColor={Colors.red30}
-            thumbTintColor={Colors.red50}
+            minimumTrackTintColor={Colors.$backgroundDangerHeavy}
```
Inbal-Tish, line 74: These are just colors for the demo, without any contextual meaning. What's the point of migrating them to design tokes? What I mean is that could be any color, red is just one of them...

Inbal-Tish:
@lidord-wix Hi. All is good, just had a comment/question if we need to migrate demo colors to design token although they do not have a special significance (like, red is not necessarily means error...).

demo/src/screens/componentScreens/SliderScreen.tsx
lidord-wix: Right, it's not a must.
But since we want the demo screens to look as good as possible in dark mode, we change them to design tokens (where possible), just to make a good contrast with the background.

Inbal-Tish: I understand, but I don't see how that makes sense in this context. This is a demo for passing colors to the Slider's track, there is no meaning to the 'danger/warning' tokens here, the red color tints are random. It will be better if you pass 'red60' and 'red80' instead... I don't want us to make this a standard since the screen might look nice but the code doesn't make sense...

lidord-wix: I see.. 
changed it back to reds

Inbal-Tish (approved):

PR #1947, opened 2022-04-06
enh: allow Incubator.WheelPicker to take custom properties for the Fader

Asgaroth:
## Description

`Incubator.WheelPicker` did not allow changing the fader tint. Which makes it hard to style for Dark mode 

Screen Shot 2022-04-04 at 8 11 41 PM

This PR allows users to pass in `tintColor` to the `Incubator.WheelPicker` and forward it to the `<Fader />` component.

## Changelog
- Allows to pass in `tintColor` to the `Incubator.WheelPicker` and forward it to the `<Fader />` component.

### DIFF
--- a/src/incubator/WheelPicker/index.tsx
+++ b/src/incubator/WheelPicker/index.tsx
@@ -6,7 +6,7 @@ import Animated, {useSharedValue, useAnimatedScrollHandler} from 'react-native-r
 import {Colors, Spacings} from 'style';
 import {Constants, asBaseComponent} from '../../commons/new';
 import View from '../../components/view';
-import Fader, {FaderPosition} from '../../components/fader';
+import Fader, {FaderPosition, FaderProps} from '../../components/fader';
 import Item, {ItemProps} from './Item';
 import Text, {TextProps} from '../../components/text';
 import usePresenter from './usePresenter';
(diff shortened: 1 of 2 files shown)


src/incubator/WheelPicker/index.tsx
```diff
@@ -78,6 +78,10 @@ export interface WheelPickerProps {
    */
   separatorsStyle?: ViewStyle;
   testID?: string;
+  /**
+   * Change the default (white) tint color of the fade view.
+   */
+  tintColor?: string;
```
ethanshar, line 84: I'd change the prop name to something more explicit to avoid confusing with other elements. 
Something like `faderTintColor`

Asgaroth: Let me know if this is good enough

src/incubator/WheelPicker/index.tsx
```diff
@@ -78,6 +78,10 @@ export interface WheelPickerProps {
    */
   separatorsStyle?: ViewStyle;
   testID?: string;
+  /**
+   * Change the default (white) tint color of the fade view.
+   */
+  faderTintColor?: string;
```
ethanshar, line 84: Hi @Asgaroth 
I'd suggest instead to add a more general prop like `faderProps` so in case we'll want to support more customizations to the Fader it will be possibly to pass all through a single prop instead of multiple props.

Asgaroth: Thank you, sorry for the back and forth, I should have done this from the beginning.

ethanshar (approved):

PR #1955, opened 2022-04-09
Fix PlatfromColor issue in TabController.TabBarItem

ethanshar:
## Description
There's a bug when using a color that connected to a PlatformColor in useAnimatedStyle hook. 
I opened a issue here that explain it better https://github.com/software-mansion/react-native-reanimated/issues/3164

Anyway, for now, in order to resolve it, I clone the colors in the TabBarItem code. 

## Changelog
Fix usages of PlatformColor in TabController TabBaritem

### DIFF
--- a/src/components/tabController/TabBarItem.tsx
+++ b/src/components/tabController/TabBarItem.tsx
@@ -131,6 +131,8 @@ export default function TabBarItem({
   const sharedLabelStyle = useSharedValue(JSON.parse(JSON.stringify(labelStyle)));
   const sharedSelectedLabelStyle = useSharedValue(JSON.parse(JSON.stringify(selectedLabelStyle)));
 
+  // NOTE: We clone these color values in refs because they might contain a PlatformColor value
+  //       which throws an error (see https://github.com/software-mansion/react-native-reanimated/issues/3164)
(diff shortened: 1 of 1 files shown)


src/components/tabController/TabBarItem.tsx
```diff
@@ -131,6 +131,9 @@ export default function TabBarItem({
   const sharedLabelStyle = useSharedValue(JSON.parse(JSON.stringify(labelStyle)));
   const sharedSelectedLabelStyle = useSharedValue(JSON.parse(JSON.stringify(selectedLabelStyle)));
 
+  const inactiveColor = useRef(_.cloneDeep(labelColor));
+  const activeColor = useRef(_.cloneDeep(!ignore ? selectedLabelColor : inactiveColor.current));
```
Inbal-Tish, line 135: I would give these consts a more specific name like 'animatedLabelColor' and 'animatedSelectedLabelColor' so it will be extra clear what this clone is for.

ethanshar: I think the name is suffice, but I will add a note instead that explains the reason we keep it like this

ethanshar: Added a note

Inbal-Tish (approved):

PR #1956, opened 2022-04-10
Support retrieving color JS value

ethanshar:
## Description
Support retrieving the original string value of color by its key
Good for need to make calculate on the color hex string and want to get it instead of the native platformColor

## Changelog
Added Colors.getColorValue - Support retrieving the original string value of color by its key

### DIFF
--- a/src/style/scheme.ts
+++ b/src/style/scheme.ts
@@ -7,9 +7,10 @@ export type SchemeType = 'default' | 'light' | 'dark';
 export type SchemeChangeListener = (schemeType?: 'light' | 'dark') => void;
 
 class Scheme {
-  currentScheme: SchemeType = 'default';
-  schemes: Schemes = {light: {}, dark: {}};
-  changeListeners: SchemeChangeListener[] = [];
+  private currentScheme: SchemeType = 'default';
+  private schemes: Schemes = {light: {}, dark: {}};
+  private schemesJS: Schemes = {light: {}, dark: {}};
+  private changeListeners: SchemeChangeListener[] = [];
   private usePlatformColors = false;
 
   constructor() {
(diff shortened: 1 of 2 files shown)


src/style/scheme.ts
```diff
@@ -7,9 +7,10 @@ export type SchemeType = 'default' | 'light' | 'dark';
 export type SchemeChangeListener = (schemeType?: 'light' | 'dark') => void;
 
 class Scheme {
-  currentScheme: SchemeType = 'default';
-  schemes: Schemes = {light: {}, dark: {}};
-  changeListeners: SchemeChangeListener[] = [];
+  private currentScheme: SchemeType = 'default';
+  private schemes: Schemes = {light: {}, dark: {}};
+  private schemesJS: Schemes = {light: {}, dark: {}};
```
Inbal-Tish, line 12: I don't understand how `schemesJS` is different from `schemes`... why we need the two?

ethanshar: Did you look how do we store each in `loadScheme` method?

Inbal-Tish: Yes

ethanshar: Basically in `schemes` we keep the native platformColors 
The values that are kept there are not regular strings but a weird object that the native side can handle

That's the reason I also want to keep the plain JS strings in `schemesJS` 
It should serve the `Colors.getColorValue` method when we need to get the original JS string hex value in some cases. 

Why do we need the JS string value? 
One of the cases, is when we call the `Colors.getColorTint` in the private repo. 
With the recent PlatformColor support, this method throws an error because `Colors.getColorTint` expect a string but now gets a weird object. 
So this whole PR meant to let us get the original string value so we can pass to `Colors.getColorTint`

I hope this explains better

Inbal-Tish: Sorry I wasn't clear with my question. I understand the meaning behind the pr and why we want to keep the js schemes. What I was missing is where are you loading these schemesJS with the string values.

ethanshar: Anywhere we call `Colors.loadSchemes` (It calls internally `Scheme.loadSchemes`) and it always accept a maps of strings colors. 
We internally mapped them once to platform colors or strings

Inbal-Tish: Ok

Inbal-Tish (approved):

PR #1958, opened 2022-04-10
Feat/sortable list no scroll

M-i-k-e-l:
## Description
SortableList - a new component (scroll-while-dragging is not supported in this phase).

## Changelog
SortableList - a new component (scroll-while-dragging is not supported in this phase).


src/components/sortableList/SortableListItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+const SortableListItem = (props: Props) => {
+  const {children, index} = props;
+
+  const {itemHeight, itemsOrder, onChange} = useContext(SortableListContext);
+  const {getTranslationByIndexChange, getItemIndexById, getIndexByPosition, getIdByItemIndex} = usePresenter();
+  const translateY = useSharedValue(0);
+  const tempTranslateY = useSharedValue(0);
+  const tempItemsOrder = useSharedValue(itemsOrder.value);
+
+  const onLayout = useCallback((event: LayoutChangeEvent) => {
```
ethanshar, line 39: I noticed that onLayout is being called twice with different heights (minor difference)

src/components/sortableList/SortableListItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+const SortableListItem = (props: Props) => {
+  const {children, index} = props;
+
+  const {itemHeight, itemsOrder, onChange} = useContext(SortableListContext);
+  const {getTranslationByIndexChange, getItemIndexById, getIndexByPosition, getIdByItemIndex} = usePresenter();
+  const translateY = useSharedValue(0);
+  const tempTranslateY = useSharedValue(0);
+  const tempItemsOrder = useSharedValue(itemsOrder.value);
```
ethanshar, line 37: Consider moving these to `useDragAfterLongPressGesture` hook. 
I'd move the logic to the hook, any reason onDragStart, onDragUpdate, onDragEnd shouldn't be part of the hook's code?

src/components/sortableList/SortableListItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const onLayout = useCallback((event: LayoutChangeEvent) => {
+    // Round height for Android
+    const newHeight = Math.round(event.nativeEvent.layout.height);
+    itemHeight.value = newHeight;
+  }, []);
```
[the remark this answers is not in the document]
M-i-k-e-l: TBH this does not bother me, but ok.
What bothered me more is the logic `index === 0 ? onItemLayout : undefined` inside the item, but I was not able to move that to the list, let me know if you have a solution for that.

src/components/sortableList/SortableListItem.tsx
M-i-k-e-l: I did not see that, are you sure you did not have two devices at the time? If you still see this, let's look at it together.

src/components/sortableList/SortableListItem.tsx
ethanshar: Yes, very possible.
But I prefer to do it once the components are stable and we know for sure the logic shared between them is the same.

src/components/sortableList/SortableListItem.tsx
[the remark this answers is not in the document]
ethanshar: In my case, what I did is pass onLayout to all items. 
But save the layout only once and have any callback arriving afterwards do nothing.

src/components/sortableList/SortableListItem.tsx
ethanshar: Yes, you might be right.

demo/src/screens/componentScreens/SortableListScreen.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <Icon source={Assets.icons.demo.chevronRight} tintColor={Colors.$iconDefault}/>
+        </View>
+      </TouchableOpacity>
+    );
+  }, []);
+
+  return (
+    <View flex bg-$backgroundDefault>
+      {renderHeader('Sortable List', {'margin-10': true})}
+      <View flex useSafeArea margin-page>
```
ethanshar, line 49: Setting `margin-page` makes the FlatList's scroll indicator pushed inside. 
Instead it's better to set each Item with padding

src/components/sortableList/SortableListItem.tsx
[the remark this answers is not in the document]
M-i-k-e-l: It is much simpler when using `ScrollView`, here we have a `FlatList`.
If it is possible to do it, then it's either by:
1. Changing the `renderItem` function which will look odd and might affect performance.
2. Do it in the `CellRendererComponent` and that's not what you meant.

src/components/sortableList/SortableListItem.tsx
M-i-k-e-l: So let's keep it as is? Do you want to add a task a TODO or...?

demo/src/screens/componentScreens/SortableListScreen.tsx
M-i-k-e-l: Oops, fixed

src/components/sortableList/SortableListItem.tsx
ethanshar: Why not fix it? Is it a big change? 
If we'll decide not to share code, the code we'll just stuck there..

src/components/sortableList/SortableListItem.tsx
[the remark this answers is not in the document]
ethanshar: Why is it matter if it's FlatList or ScrollView. 
The code I suggested is in the on item layout callback

ethanshar (approved):

PR #1972, opened 2022-04-12
not enable scroll if content is less than 1 pixel than it container

avishain:
## Description
should prevent situations where a friction of pixel enables scroll and hence causes the scrollview-fader appear for no reason

## Example
Screen Shot 2022-04-12 at 15 19 54

### DIFF
--- a/src/commons/withScrollEnabler.tsx
+++ b/src/commons/withScrollEnabler.tsx
@@ -25,7 +25,7 @@ function withScrollEnabler<PROPS, STATICS = {}>(WrappedComponent: React.Componen
     const layoutSize = useRef(0);
 
     const checkScroll = useCallback(() => {
-      const isScrollEnabled = contentSize.current > layoutSize.current;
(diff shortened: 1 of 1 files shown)


M-i-k-e-l:
Hi @avishain,
You could have simply re-opened #1799 

From what we've discussed this is still not reproducible in dev. Does it happen 100% of the time with a release version of your module? If yes, did you try to create a version with this change and it solved your issue?

avishain:
> Hi @avishain, You could have simply re-opened #1799
> 
> If yes, did you try to create a version with this change and it solved your issue?
Not sure how can I create a version with this change. If I'll change it in my node_modules it won't matter as it will be built based on the version in engine. And I can't cook-book with Poco as it's in the open Wix repo + not sure how to create a build-version for it

avishain:
The private version with the fix was QA'd and looks good

M-i-k-e-l (approved):

PR #1977, opened 2022-04-13
Upgrade reanimated and gesture-handler to latest version

ethanshar:
## Description
Upgrade reanimated and gesture-handler to latest versions to make sure there are not issues 
(this is so we can safely upgrade later in the engine)

## Changelog
Upgrade reanimated and gesture-handler to latest versions in our demo app


M-i-k-e-l (commented):
There's an `@ts-expect-error` with an error in `SegmentedControl`
Found one bug: in the new `Dialog`, on Android the Dialog's height is not taking the full height it can (up to maxHeight), not sure why it happens (already clean-installed to the previous version to verify it's a regression, so let me know if you want me to dig into this).

ethanshar:
@M-i-k-e-l 
Can you take a look at the issue with Dialog. 
I want to move to the new versions

M-i-k-e-l (approved):

PR #1991, opened 2022-04-18
Carousel - fix initial page offset (contentOffset)

M-i-k-e-l:
## Description
Carousel - fix initial page offset (contentOffset)
WOAUILIB-2668
Generally I'd like to replace all `calcOffset` usages to the new `getContentOffset` (since `snapToOffsets` seem to have the correct data) but I fear of edge cases and bugs.
TBH, I think we should rewrite this component as a function component.

Testing snippet:
```
<Carousel
  onChangePage={this.onChangePage}
  pageWidth={249 + (Spacings.s5 * 2)}
  itemSpacings={Spacings.s5}
  initialPage={1}
  containerStyle={{height: 160}}
>
  {_.map([...Array(numberOfPagesShown)], (item, index) => (
    <Page style={{backgroundColor: BACKGROUND_COLORS[index]}} key={index}>
      <Text margin-15>CARD {index}</Text>
    </Page>
  ))}
</Carousel>
...
page: {
  flex: 1,
  width: 249,
  borderWidth: 1,
  borderRadius: 8
}
```

## Changelog
Carousel - fix initial page offset (contentOffset)

### DIFF
--- a/src/components/carousel/index.tsx
+++ b/src/components/carousel/index.tsx
@@ -246,6 +246,21 @@ class Carousel extends Component<CarouselProps, CarouselState> {
     }
   };
 
+  getContentOffset = (snapToOffsets: number[] | undefined) => {
+    const {horizontal, initialPage} = this.props;
+    const {initialOffset} = this.state;
+    let contentOffset = initialOffset;
+    if (snapToOffsets && initialPage !== undefined) {
+      const offset = snapToOffsets[initialPage];
+      contentOffset = {
+        x: horizontal ? offset : 0,
+        y: horizontal ? 0 : offset
+      };
+    }
+
+    return contentOffset;
+  }
+
   shouldUsePageWidth() {
     const {loop, pageWidth} = this.props;
     return !loop && pageWidth;
@@ -477,13 +492,13 @@ class Carousel extends Component<CarouselProps, CarouselState> {
 
   renderCarousel() {
     const {containerStyle, animated, horizontal, animatedScrollOffset, ...others} = this.props;
-    const {initialOffset} = this.state;
     const scrollContainerStyle = this.shouldUsePageWidth()
       ? {paddingRight: this.getItemSpacings(this.props)}
       : undefined;
     const snapToOffsets = this.getSnapToOffsets();
     const marginBottom = Math.max(0, this.getContainerPaddingVertical() - 16);
     const ScrollContainer = animatedScrollOffset ? Animated.ScrollView : ScrollView;
+    const contentOffset = this.getContentOffset(snapToOffsets);
     return (
       <View
         animated={animated}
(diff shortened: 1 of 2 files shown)


src/components/carousel/index.tsx
```diff
@@ -246,6 +246,21 @@ class Carousel extends Component<CarouselProps, CarouselState> {
...
+    const {horizontal, initialPage} = this.props;
+    const {initialOffset} = this.state;
+    let contentOffset = initialOffset;
+    if (snapToOffsets && initialPage !== undefined) {
+      const offset = snapToOffsets[initialPage];
+      contentOffset = {
+        x: horizontal ? offset : 0,
+        y: horizontal ? 0 : offset
+      };
+    }
```
ethanshar, line 259: shouldn't this code be part of what we store in the initialOffset? in the `onContainerLayout` method?

src/components/carousel/index.tsx
```diff
@@ -502,8 +517,7 @@ class Carousel extends Component<CarouselProps, CarouselState> {
           horizontal={horizontal}
           pagingEnabled={this.shouldEnablePagination()}
           snapToOffsets={snapToOffsets}
-          contentOffset={initialOffset} // iOS only
-          onContentSizeChange={this.onContentSizeChange}
```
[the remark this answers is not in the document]
M-i-k-e-l: I think that `contentOffset={initialOffset} // iOS only` was actually only working for iOS and it looks like it's working on Android as well now; and I think that `onContentSizeChange` was there to workaround that issue.
Do you know of a case where this is useful?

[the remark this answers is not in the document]
ethanshar: No, but let's keep it commented for now, so if an issue will rise soon it'll be easier to spot it. 
If in the future we'll come across it we'll remove it

src/components/carousel/index.tsx
ethanshar: Not sure it has to go that way. 
We have a method that calcOffset, we use it for setting the initial offset. 
Now you the logic for initial offset changed a little, Can't we create a method for `calcInitialOffset` that returns the relevant offset and set it in the state? 
What are the blockers here?

M-i-k-e-l: The new method (with the new logic) `getContentOffset` depends on `snapToOffsets` which is calculated using `containerWidth` which is obtained in `onContainerLayout`.
Technically from testing now, it seems that only the constructor is occurring before this happens, but it feels unsafe to me, WDYT?

ethanshar: Ok, I see your point. 
So can we rename the new method to `getInitialContentOffset` so it's clear that it's the initial offset

M-i-k-e-l: Sure, but I blame RN:
```
/**
 * Used to manually set the starting scroll offset.
 * The default value is {x: 0, y: 0}
 */
contentOffset?: PointPropType | undefined; // zeros
```

ethanshar (approved):

PR #1995, opened 2022-04-19
Refactor DateTimePicker expandable overlay

ethanshar:
## Description
Refactor DateTimePicker expandable overlay
Remove usage of old TextField expandable prop and replace it with out ExpandableOverlay component. 
This is required so we can eventually migrate to the next TextField implementation

## Changelog
Refactor DateTimePicker expandable overlay

### DIFF
--- a/src/components/dateTimePicker/index.tsx
+++ b/src/components/dateTimePicker/index.tsx
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import moment from 'moment';
 import React, {Component} from 'react';
 import {StyleProp, StyleSheet, ViewStyle} from 'react-native';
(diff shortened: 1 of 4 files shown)


src/components/dateTimePicker/index.tsx
```diff
@@ -41,7 +41,11 @@ export interface DateTimePickerProps {
   /**
    * The onChange callback
    */
-  onChange?: (date: Date) => void;
+  onChange?: (date?: Date) => void;
```
Inbal-Tish, line 44: This is potentially breaking change...

ethanshar: Type change can't be a breaking change, worst thing it can break someone TS check

Inbal-Tish: I know, by 'potentially ' I meant that if you return undefined here instead of a Date it can break whoever using it without checking (basically on Android since on iOS we set a new Date if undefined).

Inbal-Tish (approved):

PR #1997, opened 2022-04-19
Enable useNative when passing activeScale and add its type

ethanshar:
## Description
Expose activeScale prop and enable useNative by default when using it 

WOAUILIB-2221

## Changelog
Expose activeScale prop and enable useNative by default when using it

### DIFF
--- a/src/components/touchableOpacity/index.tsx
+++ b/src/components/touchableOpacity/index.tsx
@@ -31,6 +31,10 @@ export interface TouchableOpacityProps extends Omit<RNTouchableOpacityProps, 'st
      */
     activeBackgroundColor?: string;
     /**
+     * Will apply scale press feedback. This will enforce the useNative prop
+     */
+    activeScale?: number;
+    /**
      * Should use a more native touchable opacity component
      */
     useNative?: boolean;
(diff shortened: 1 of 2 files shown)


src/components/touchableOpacity/index.tsx
```diff
@@ -112,10 +116,10 @@ class TouchableOpacity extends PureComponent<Props, {active: boolean}> {
   }
 
   render() {
-    const {useNative, style, modifiers, forwardedRef, ...others} = this.props;
+    const {useNative, activeScale, style, modifiers, forwardedRef, ...others} = this.props;
     const {borderRadius, paddings, margins, alignments, flexStyle} = modifiers;
 
-    if (useNative) {
+    if (useNative || !!activeScale) {
```
lidord-wix, line 122: It's a rare case but it won't work if the `activeScale` is equal to `0` even though it's a valid value.

lidord-wix (commented):
Small comment :)

src/components/touchableOpacity/index.tsx
ethanshar: Fixed
Though passing `0` to activeScale is really weird (:

lidord-wix (approved):

PR #2003, opened 2022-04-20
Fix Incubator TextField placeholder's disabled color

M-i-k-e-l:
## Description
Fix Incubator TextField placeholder's disabled color (when floating is also disabled).
There are 4 examples below, the 4th has a wrong color for the placeholder, this happens when both `floatingPlaceholder={false}` and  `editable={false}`.

Snippet:
```
<TextField migrate floatingPlaceholder editable placeholder="Placeholder"/>
<TextField migrate floatingPlaceholder editable={false} placeholder="Placeholder"/>
<TextField migrate floatingPlaceholder={false} editable placeholder="Placeholder"/>
<TextField migrate floatingPlaceholder={false} editable={false} placeholder="Placeholder"/>
```

## Changelog
Fix Incubator TextField placeholder's disabled color (when floating is also disabled).

### DIFF
--- a/src/incubator/TextField/index.tsx
+++ b/src/incubator/TextField/index.tsx
@@ -9,7 +9,6 @@ import React, {useMemo} from 'react';
 import {isEmpty, trim, omit} from 'lodash';
 import {asBaseComponent, forwardRef} from '../../commons/new';
 import View from '../../components/view';
-import {Colors} from '../../style';
 import {useMeasure} from '../../hooks';
 import {TextFieldProps, InternalTextFieldProps, ValidationMessagePosition, FieldContextType, TextFieldMethods} from './types';
 import {shouldHidePlaceholder} from './Presenter';
@@ -44,6 +43,7 @@ const TextField = (props: InternalTextFieldProps) => {
     floatingPlaceholderColor,
     floatingPlaceholderStyle,
     floatOnFocus,
+    placeholderColor,
     hint,
     // Label
     label,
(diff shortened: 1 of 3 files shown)


src/incubator/TextField/types.ts
```diff
@@ -215,7 +215,7 @@ export type TextFieldProps = MarginModifiers &
   };
 
 export type InternalTextFieldProps = PropsWithChildren<
-  TextFieldProps & BaseComponentInjectedProps & ForwardRefInjectedProps
+  TextFieldProps & BaseComponentInjectedProps & ForwardRefInjectedProps & {placeholderColor: ColorType}
```
ethanshar, line 218: Shouldn't we add the `placeholderColor` type to `TextFieldProps` with the rest of the props?

M-i-k-e-l: Found the correct prop: `placeholderTextColor`, already there :)

src/incubator/TextField/presets/default.ts
```diff
@@ -7,6 +7,13 @@ const colorByState = {
   disabled: Colors.$textDisabled
 };
 
+const placeHolderTextColorByState = {
```
ethanshar, line 10:
```suggestion
const placeholderTextColorByState = {
```

src/incubator/TextField/index.tsx
```diff
@@ -128,7 +128,7 @@ const TextField = (props: InternalTextFieldProps) => {
             )}
             {children || (
               <Input
-                placeholderTextColor={hidePlaceholder ? 'transparent' : Colors.$textNeutral}
+                placeholderTextColor={hidePlaceholder ? 'transparent' : placeholderTextColor}
```
ethanshar, line 131: technically you're replacing the default value with undefined, no?

M-i-k-e-l: No, I am changing the default for a non-floating placeholder from a `string` (set here to: `Colors.$textNeutral`) to an `object` (set in `default.ts`->`placeholderTextColorByState`); and this way I can set the `disabled` type to look like it's disabled: `Colors.$textDisabled`

ethanshar (approved):

PR #2010, opened 2022-04-25
Picker - multi select - fix behavior when the value is changed by user

### LINKED ISSUE #2005: Picker Value, State not listening.
## Description

I need to clear the Picker value. but state is not listening.

### Related to

- [x] Components
- [ ] Demo
- [ ] Docs
- [ ] Typings

### Code snippet

```
<Picker
    value={recipients}
    onChange={(itemValue) => {setRecipients(itemValue)}}
    mode={Picker.modes.MULTI}
    migrateTextField
    migrate
    showSearch    
  >
  {options.map(item => (
    <Picker.Item key={item.value} value={item.value} label={item.label} />
  ))}
</Picker>

setRecipients([]) - not working.
```

### Environment

- React Native: 0.68.1
- React Native UI Lib: 6.13.3

### Affected platforms

- [x] Android
- [x] iOS
- [ ] Web

M-i-k-e-l:
## Description
Picker - multi select - fix behavior when the value is changed by user
Fixes #2005

## Changelog
Picker - multi select - fix behavior when the value is changed by user

### DIFF
--- a/src/components/picker/helpers/usePickerSelection.tsx
+++ b/src/components/picker/helpers/usePickerSelection.tsx
@@ -1,4 +1,4 @@
-import {RefObject, useCallback, useState} from 'react';
+import {RefObject, useCallback, useState, useEffect} from 'react';
 import _ from 'lodash';
 import {PickerProps, PickerValue, PickerSingleValue, PickerMultiValue} from '../types';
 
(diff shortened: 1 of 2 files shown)


src/components/picker/helpers/usePickerSelection.tsx
```diff
@@ -13,6 +13,13 @@ const usePickerSelection = (props: UsePickerSelectionProps) => {
   const [multiDraftValue, setMultiDraftValue] = useState(value as PickerMultiValue);
   const [multiFinalValue, setMultiFinalValue] = useState(value as PickerMultiValue);
 
+  useEffect(() => {
+    if (multiFinalValue !== value) {
```
ethanshar, line 17: Do we always want to do this or just when the picker is in "multi" mode? 
Could this cause an issue in the "single" mode?

M-i-k-e-l: I think it should not cause an issue, but it's probably more readable and will cause less renders 👍

ethanshar (approved):

PR #2016, opened 2022-04-27
KeyboardAccessoryView - 'iosScrollBehavior' prop deprecation

Inbal-Tish:
## Description
KeyboardAccessoryView - 'iosScrollBehavior' prop deprecation

## Changelog
KeyboardAccessoryView - 'iosScrollBehavior' prop deprecation

### DIFF
--- a/lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
+++ b/lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
@@ -55,7 +55,8 @@ export type KeyboardAccessoryViewProps = kbTrackingViewProps & {
   onKeyboardResigned?: () => void;
   /**
    * @deprecated
-   * Please use 'scrollBehavior' prop instead and take values from 'KeyboardTrackingView.scrollBehaviors' (instead of KeyboardAccessoryView.iosScrollBehaviors)
+   * Please use 'scrollBehavior' prop instead
+   * The scrolling behavior (use KeyboardAccessoryView.iosScrollBehaviors.NONE | SCROLL_TO_BOTTOM_INVERTED_ONLY | FIXED_OFFSET)
    */
   iOSScrollBehavior?: number;
   children?: React.ReactChild;
@@ -67,7 +68,6 @@ export type KeyboardAccessoryViewProps = kbTrackingViewProps & {
  * @gif: https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/KeyboardAccessoryView/KeyboardAccessoryView.gif?raw=true
  */
 class KeyboardAccessoryView extends Component<KeyboardAccessoryViewProps> {
-  // @ts-expect-error
   static iosScrollBehaviors = KeyboardTrackingView.scrollBehaviors;
 
   static defaultProps = {
@@ -181,7 +181,7 @@ class KeyboardAccessoryView extends Component<KeyboardAccessoryViewProps> {
       onItemSelected,
       onRequestShowKeyboard,
       useSafeArea,
-      scrollBehavior = KeyboardAccessoryView.iosScrollBehaviors.FIXED_OFFSET,
+      scrollBehavior = KeyboardTrackingView.scrollBehaviors.FIXED_OFFSET,
       iOSScrollBehavior,
       ...others
     } = this.props;
--- a/lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
+++ b/lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
@@ -16,16 +16,7 @@ import KeyboardUtils from './utils/KeyboardUtils';
 const IsIOS = Platform.OS === 'ios';
 const IsAndroid = Platform.OS === 'android';
 
-const IOS_SCROLL_BEHAVIORS = IsIOS
-  ? {
-    NONE: NativeModules.KeyboardTrackingViewTempManager?.KeyboardTrackingScrollBehaviorNone,
-    SCROLL_TO_BOTTOM_INVERTED_ONLY:
-        NativeModules.KeyboardTrackingViewTempManager?.KeyboardTrackingScrollBehaviorScrollToBottomInvertedOnly,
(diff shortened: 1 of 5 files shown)


lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
```diff
@@ -78,10 +67,10 @@ export type KeyboardAccessoryViewProps = kbTrackingViewProps & {
  * @gif: https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/KeyboardAccessoryView/KeyboardAccessoryView.gif?raw=true
  */
 class KeyboardAccessoryView extends Component<KeyboardAccessoryViewProps> {
-  static iosScrollBehaviors = IOS_SCROLL_BEHAVIORS;
+  // @ts-expect-error
+  static iosScrollBehaviors = KeyboardTrackingView.scrollBehaviors;
 
   static defaultProps = {
-    iOSScrollBehavior: -1,
```
M-i-k-e-l, line 84: Why not change this to `KeyboardAccessoryView.iosScrollBehaviors.FIXED_OFFSET` instead of adding it bellow?

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
```diff
@@ -202,13 +181,15 @@ class KeyboardAccessoryView extends Component<KeyboardAccessoryViewProps> {
...
+      scrollBehavior = KeyboardAccessoryView.iosScrollBehaviors.FIXED_OFFSET,
+      iOSScrollBehavior,
       ...others
     } = this.props;
 
     return (
       <KeyboardTrackingView
         {...others}
-        scrollBehavior={this.getIOSTrackingScrollBehavior()}
+        scrollBehavior={IsIOS ? iOSScrollBehavior || scrollBehavior : undefined}
```
M-i-k-e-l, line 192: Not sure why this was conditioned, we pass other iOS only props to Android, do we really need this?

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
```diff
@@ -63,10 +54,8 @@ export type KeyboardAccessoryViewProps = kbTrackingViewProps & {
    */
   onKeyboardResigned?: () => void;
   /**
-   * iOS only.
-   * The scrolling behavior, use KeyboardAccessoryView.iosScrollBehaviors.X where X is:
-   * NONE, SCROLL_TO_BOTTOM_INVERTED_ONLY or FIXED_OFFSET
-   * default: FIXED_OFFSET
+   * @deprecated
+   * Please use 'scrollBehavior' prop instead and take values from 'KeyboardTrackingView.scrollBehaviors' (instead of KeyboardAccessoryView.iosScrollBehaviors)
```
M-i-k-e-l, line 58: I think it should still be `KeyboardAccessoryView.iosScrollBehaviors`, so the user will not need to import `KeyboardTrackingView` as well.
But probably change this to `scrollBehaviors` as well.

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
```diff
@@ -78,10 +67,10 @@ export type KeyboardAccessoryViewProps = kbTrackingViewProps & {
  * @gif: https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/KeyboardAccessoryView/KeyboardAccessoryView.gif?raw=true
  */
 class KeyboardAccessoryView extends Component<KeyboardAccessoryViewProps> {
-  static iosScrollBehaviors = IOS_SCROLL_BEHAVIORS;
+  // @ts-expect-error
```
M-i-k-e-l, line 70: You can fix this by changing the export in `KeyboardTrackingView` to:
```
export default KeyboardTrackingView as (typeof KeyboardTrackingView & {scrollBehaviors: typeof SCROLL_BEHAVIORS});
```

lib/components/Keyboard/KeyboardInput/keyboardAccessoryView.api.json
```diff
@@ -47,8 +47,9 @@
     {
       "name": "iOSScrollBehavior",
       "type": "number",
-      "description": "The scrolling behavior, use KeyboardAccessoryView.iosScrollBehaviors.X where X is:\nNONE, SCROLL_TO_BOTTOM_INVERTED_ONLY or FIXED_OFFSET\niOS only.",
-      "default": "FIXED_OFFSET"
+      "description": "The scrolling behavior. DEPRECATED: Use 'scrollBehavior' prop instead and take values from 'KeyboardTrackingView.scrollBehaviors' (instead of KeyboardAccessoryView.iosScrollBehaviors)",
```
M-i-k-e-l, line 50: This will need to change as well if we decide to keep the static on `KeyboardAccessoryView`

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
Inbal-Tish: Because I don't want to have a default to 'iOSScrollBehavior' prop so I can use the 'scrollBehavior' prop instead (that holds the default)

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
Inbal-Tish: The comment is wrong, I'll fix this

lib/components/Keyboard/KeyboardInput/KeyboardAccessoryView.tsx
Inbal-Tish: Thanks

lib/components/Keyboard/KeyboardInput/keyboardAccessoryView.api.json
Inbal-Tish: Yes, keeping the static. fixed.

M-i-k-e-l (approved):

PR #2017, opened 2022-04-27
SegmentedControl - move Easing.bezier to the new factory implementation

M-i-k-e-l:
## Description
SegmentedControl - move Easing.bezier to the new factory implementation
WOAUILIB-2521

## Changelog
SegmentedControl - move Easing.bezier to the new factory implementation

### DIFF
--- a/src/components/segmentedControl/index.tsx
+++ b/src/components/segmentedControl/index.tsx
@@ -19,8 +19,7 @@ import {useOrientation} from 'hooks';
 const BORDER_WIDTH = 1;
 const TIMING_CONFIG: WithTimingConfig = {
   duration: 300,
-  // @ts-expect-error TODO: change this to bezierFn or to the new implementation
-  easing: Easing.bezier(0.33, 1, 0.68, 1)
+  easing: Easing.bezier(0.33, 1, 0.68, 1).factory()
 };
 
 export type SegmentedControlItemProps = SegmentProps;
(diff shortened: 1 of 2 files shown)


src/components/segmentedControl/index.tsx
```diff
@@ -19,8 +19,7 @@ import {useOrientation} from 'hooks';
 const BORDER_WIDTH = 1;
 const TIMING_CONFIG: WithTimingConfig = {
   duration: 300,
-  // @ts-expect-error TODO: change this to bezierFn or to the new implementation
-  easing: Easing.bezier(0.33, 1, 0.68, 1)
+  easing: Easing.bezier(0.33, 1, 0.68, 1).factory()
```
ethanshar, line 22: Where this is documented? 
I saw something about moving from Easing to EasingNode. 
What does it fix exactly? the TS error? 

Also, I think the Jira ref in the description is wrong.

M-i-k-e-l: Updated the ticket number.
Where's the docs? Good question... The only thing I found was the below PR and the types for reanimated.

From the description [here](https://github.com/software-mansion/react-native-reanimated/pull/2792):
```
During the performance optimization, I changed the implementation of Easing.bezier function, now return an object instead of a plain function, so I added the old implementation in Easing.bezierFn
```

I tend to think this implementation is more performant, but I'll understand if you'd want to put this on ice and maybe just use `Easing.bezierFn`.

PR #2019, opened 2022-04-27
Modal - support - KeyboardAvoidingView

M-i-k-e-l:
## Description
Modal - support - KeyboardAvoidingView
It seems that there is a limitation for `KeyboardAvoidingView` in a `Modal` see this [SO answer](https://stackoverflow.com/a/65029356/4759619).
WOAUILIB-2719
This will allow using this in the `Incubator.Dialog`, not sure if we want to add support in the current `Dialog` or perhaps tell the user to use the new one.

## Changelog
Modal - support - KeyboardAvoidingView

### DIFF
--- a/src/components/modal/index.tsx
+++ b/src/components/modal/index.tsx
@@ -6,7 +6,9 @@ import {
   Modal as RNModal,
   ModalProps as RNModalProps,
   TouchableWithoutFeedback,
-  GestureResponderEvent
+  GestureResponderEvent,
+  KeyboardAvoidingView,
+  KeyboardAvoidingViewProps
 } from 'react-native';
 import {BlurViewPackage} from '../../optionalDependencies';
 import {Constants, asBaseComponent} from '../../commons/new';
@@ -46,6 +48,10 @@ export interface ModalProps extends RNModalProps {
    * Should add a GestureHandlerRootView (Android only)
    */
   useGestureHandlerRootView?: boolean;
+  /**
+   * Send in order to add a KeyboardAvoidingView (iOS only)
+   */
+  keyboardAvoidingViewProps?: KeyboardAvoidingViewProps;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/components/modal/index.tsx
```diff
@@ -46,6 +48,10 @@ export interface ModalProps extends RNModalProps {
    * Should add a GestureHandlerRootView (Android only)
    */
   useGestureHandlerRootView?: boolean;
+  /**
+   * Send in order to add a KeyboardAvoidingView (iOS only)
+   */
+  keyboardAvoidingViewProps?: KeyboardAvoidingViewProps;
```
ethanshar, line 54: Should we split it into two props? `useKeyboardAvoidingView` and `keyboardAvoidingViewProps`
What if the user doesn't need/want to pass any props to it, but just enable it?

M-i-k-e-l: I was not sure about this choice but technically the user can send an empty object (which will work), WDYT?

M-i-k-e-l:
@ethanshar 
1. Making sure this is seen:
```
This will allow using this in the Incubator.Dialog, not sure if we want to add support in the current Dialog or perhaps tell the user to use the new one.
```
2. Adding the `KeyboardAvoidingViewProps` import breaks the typescript creation, I did not try to update versions yet, any suggestions?

src/components/modal/index.tsx
ethanshar: I'm not a fan, it's not clear enough and it force the user to pass something that isn't trivial
I think in this case I'd go with two props.

ethanshar:
> @ethanshar
> 
> 1. Making sure this is seen:
> 
> ```
> This will allow using this in the Incubator.Dialog, not sure if we want to add support in the current Dialog or perhaps tell the user to use the new one.
> ```
> 
>  2. Adding the `KeyboardAvoidingViewProps` import breaks the typescript creation, I did not try to update versions yet, any suggestions?


1. But we don't use Incubator.Dialog yet. Isn't using it in the old dialog is a matter of passing the prop from Dialog to the Modal? 

2. In what way? Is it something we can `@ts-expect-error` for now?

M-i-k-e-l:
> 1. But we don't use Incubator.Dialog yet. Isn't using it in the old dialog is a matter of passing the prop from Dialog to the Modal?
> 2. In what way? Is it something we can `@ts-expect-error` for now?

1. Forget about it, my mistake.
2. I don't think so, it actually fails the command: `tsc --p tsconfig.build.json` in the build process, not an error in the file.

ethanshar (approved):

PR #2020, opened 2022-04-28
Handle data change and reset sortable grid list order

ethanshar:
## Description

## Changelog

### DIFF
--- a/src/components/sortableGridList/index.tsx
+++ b/src/components/sortableGridList/index.tsx
@@ -3,6 +3,7 @@ import {StyleSheet, FlatListProps, ScrollView, ScrollViewProps, ListRenderItemIn
 import {GestureHandlerRootView} from 'react-native-gesture-handler';
 import {useSharedValue} from 'react-native-reanimated';
 import _ from 'lodash';
+import {useDidUpdate} from 'hooks';
 import {GridListBaseProps} from '../gridList';
 import SortableItem from './SortableItem';
 import usePresenter, {ItemsOrder} from './usePresenter';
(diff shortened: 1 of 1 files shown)


lidord-wix (commented):
Before reviewing the code, I see weird behavior in the example screen, a video attached. 

https://user-images.githubusercontent.com/66782556/166637795-090f7b58-d472-44de-91eb-5601aaf6ac76.mp4

ethanshar:
> Before reviewing the code, I see weird behavior in the example screen, a video attached.

You're right! 
I think that recent update of reanimated version to 2.8.0 makes this bug more frequent. 
Anyway, I also managed to reproduce it on the old reanimated version we were on, so I decided to just fix the issue. 

It's all related to this issue I opened to reanimated library -> https://github.com/software-mansion/react-native-reanimated/issues/3224

Anyway, it should be fixed now in my last commit, I hope I didn't miss anything.

src/components/sortableGridList/SortableItem.tsx
```diff
@@ -40,24 +38,47 @@ function SortableItem(props: PropsWithChildren<SortableItemProps>) {
...
+            translateX.value = withTiming(0, animationConfig);
+            translateY.value = withTiming(0, animationConfig);
+          }
+          dataManuallyChanged.value = false;
+        } else if (newOrder !== prevOrder) {
+          const translation = getTranslationByOrderChange(newOrder, prevOrder);
+          translateX.value = withTiming(translateX.value + translation.x, animationConfig);
+          translateY.value = withTiming(translateY.value + translation.y, animationConfig);
+        }
+      }
```
lidord-wix, line 73: I know we don't use to do it, but I found the conditions here a bit confusing.. consider adding a comment for every condition that explains the cases that are relevant to this condition..

lidord-wix (approved):
Looks really good overall, I approved the PR with a small comment about adding notes in the code. it's up to you if you want to add some or merge it as it is.. :)

src/components/sortableGridList/SortableItem.tsx
ethanshar: Done

PR #2030, opened 2022-05-03
Infra/picker new file

ethanshar:
## Description
Remove old implementation of Picker. 
Fix issues with Picker Dialog when passing useNativePicker - related to issues with extracting dialog props 

## Changelog
Fix error with Picker dialog when passing useNativePicker


src/components/picker/NativePicker.js
```diff
@@ -1,14 +1,12 @@
...
-import {BaseComponent} from '../../commons';
+import React, {Component} from 'react';
 import TextField from '../textField';
 import PickerDialog from './PickerDialog';
 import TouchableOpacity from '../touchableOpacity';
 import {Colors} from '../../style';
 import {WheelPicker} from '../../incubator';
 
-class NativePicker extends BaseComponent {
-  static displayName = 'IGNORE';
```
Inbal-Tish, line 11: Should we add the displayName?

src/components/picker/PickerDialog.js
```diff
@@ -1,17 +1,14 @@
...
 
-import {BaseComponent} from '../../commons';
-import {extractComponentProps} from '../../commons/modifiers';
 import Dialog from '../dialog';
 import View from '../view';
 import Text from '../text';
 import {Colors} from '../../style';
 
-class PickerDialog extends BaseComponent {
-  static displayName = 'IGNORE';
```
Inbal-Tish, line 14: Also here, don't we need the displayName static?

Inbal-Tish:
@ethanshar Looks good, just a litle question regarding the DisplayName static and the picker tests are failing...

src/components/picker/NativePicker.js
ethanshar: Fixed. 
It helps when errors prints callstack

src/components/picker/PickerDialog.js
ethanshar: Fixed

ethanshar:
> @ethanshar Looks good, just a litle question regarding the DisplayName static and the picker tests are failing...

Fix tests, I moved to render tests instead,

src/components/picker/PickerDialog.js
```diff
@@ -1,17 +1,15 @@
...
+class PickerDialog extends Component {
+  static displayName = 'NativePicker';
```
Inbal-Tish, line 12: You have 'NativePicker' already in the NativePicker component. This should be the 'PickerDialog' and the android one should be 'PickerDialog.android', no?

ethanshar: Oops, bad copy-paste. 
And no, both should be `PickerDialog`, they can't co-exist. One is for iOS and the other is for Android

Inbal-Tish (approved):

PR #2033, opened 2022-05-04
Infra/ designTokens disabled Slider

lidord-wix:
## Description
Fix disabled slider border color - design tokens

## Changelog
Fix disabled slider border color - design tokens

### DIFF
--- a/src/components/slider/index.tsx
+++ b/src/components/slider/index.tsx
@@ -515,6 +515,7 @@ export default class Slider extends PureComponent<SliderProps, State> {
         {...this._panResponder.panHandlers}
         style={[
           styles.thumb,
+          disabled && styles.disabledThumb,
           thumbStyle,
           {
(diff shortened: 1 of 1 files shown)


src/components/slider/index.tsx
```diff
@@ -621,6 +622,9 @@ const styles = StyleSheet.create({
     shadowRadius: SHADOW_RADIUS,
     elevation: 2
   },
+  disabledThumb: {
+    borderColor: Colors.$backgroundElevated
```
Inbal-Tish, line 626: Why this token is only for the disabled thumb? The active one is white (as defined in line 618), shouldn't it be tokenized as well?

lidord-wix: For the active slider, the thumb should be white for both dark and light modes. 
For the disabled slider, it should be elevated (white in light mode and grey1 in dark)

Inbal-Tish (approved):

PR #2039, opened 2022-05-10
Use Gesture to handle TabBarItem presses

ethanshar:
## Description
This is a small improvement for performance in TabController component. 
In TabBarItem I replaced the `TouchableOpacity` with gesture-handler `Gesture` component. 
Instead of having a round trip (Native side -> JS side) to invoke the `onPress` callback in order to change the `currIndex` shared value, I now use the Gesture API to do that, this way we avoid the round trip. 

## Changelog
Small performance improvement in TabController transitions

### DIFF
--- a/src/components/tabController/TabBarItem.tsx
+++ b/src/components/tabController/TabBarItem.tsx
@@ -2,13 +2,12 @@
 import React, {useCallback, useContext, useEffect, useRef, useMemo, ReactElement} from 'react';
 import {StyleSheet, TextStyle, LayoutChangeEvent, StyleProp, ViewStyle} from 'react-native';
 import _ from 'lodash';
-import Reanimated, {useAnimatedStyle, useSharedValue} from 'react-native-reanimated';
+import Reanimated, {runOnJS, useAnimatedStyle, useSharedValue} from 'react-native-reanimated';
+import {Gesture, GestureDetector} from 'react-native-gesture-handler';
 import {Colors, Typography, Spacings} from '../../style';
 import Badge, {BadgeProps} from '../badge';
-import _TouchableOpacity from '../touchableOpacity';
+import View from '../view';
 import TabBarContext from './TabBarContext';
-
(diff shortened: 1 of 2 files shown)


src/components/tabController/TabBarItem.tsx
```diff
@@ -187,40 +179,63 @@ export default function TabBarItem({
...
   }, [style]);
 
+  const gesture = Gesture.Tap()
+    .maxDuration(60000)
+    .onEnd(() => {
+      if (!ignore) {
+        setCurrentIndex(index);
+      }
+
+      props.onPress && runOnJS(props.onPress)(index);
```
M-i-k-e-l, line 201: I see this is the same behavior as before, but if a tab is ignored why do we press it?

src/components/tabController/TabBarItem.tsx
```diff
@@ -187,40 +179,63 @@ export default function TabBarItem({
...
+  });
+
   const _style = useMemo(() => {
     const constantWidthStyle = itemWidth.current ? {flex: 0, width: itemWidth.current} : undefined;
-    return [styles.tabItem, style, constantWidthStyle];
+    return [styles.tabItem, style, constantWidthStyle, pressStyle];
   }, [style]);
 
+  const gesture = Gesture.Tap()
+    .maxDuration(60000)
```
M-i-k-e-l, line 195: Is this limit solving a problem?

ethanshar: Yes, it create a proper time between touch down and touch up. 
I didnt find any proper way to do it. 
I need to handle touch down - for the active press background color and opacity
and touch up - is for completing the actual press (and moving the current index)

M-i-k-e-l: Oh, I just tested without it and with `.maxDuration(0)` and now I get it.
I'd expect it'll be possible to have unlimited time...

M-i-k-e-l (approved):

PR #2042, opened 2022-05-11
fix ui flashing when open app from background

s123121:
## Description
In my app with heavy use of react-native-ui-lib, whenever the app change from background to foreground, the ui flash (change from light theme to dark theme or vice versa)

This only happen when use the default scheme. Logging inside the demo app show that the scheme changes from dark to light when the app go to background.

https://github.com/facebook/react-native/issues/28525#issuecomment-897327174

The current work around can not prevent the issue entirely (it keeps happen randomly)

## Changelog
Fix the ui flashing issue with dark mode with default scheme


lidord-wix:
Hello @s123121, 
Any idea how can I reproduce the issue?

s123121:
I attached the video of the issue in my app. Basically, I think if you use a lot of react-native-ui-lib components and all of them listen to colorScheme change you will get this problem. 

In react-native, there is a mechanism (I don't know what it call) that all of the UI change/rerender in background will be delayed and then flushed all at once when the app go back to foreground. You can see that from logging inside the demo app, when you change the app to background, scheme change from dark to light and then change back to dark immediately.

And as I pointed above, this only happens on iOS

https://user-images.githubusercontent.com/3921357/168988227-73c68c9d-1a6c-4e9e-9c9b-43c2d8e5c96e.MP4

s123121:
@lidord-wix I create a test branch to reproduce this very issue here: 
https://github.com/s123121/react-native-ui-lib/tree/test/dark-mode-bug

You can check it out from these steps:
1. go to dark mode scene 
2. go into background
3. go back to the demo app to see the ui flask from light to dark (remember to choose dark mode in your phone system)

In the process of reproduce this issue, I also found out one nasty bug. When go to background, the cpu usage spiked a lot (which is bad for performance and your app will be killed by ios watchdog if the cpu spiked more than 80% for x seconds). Therefore, I also add the safeguard in place for broadcasting scheme change

lidord-wix:
@s123121 
Sorry, but I still can't reproduce it. 
I suspect it might be something with the iOS environment, can you please provide the following info:
```
XCode version: 
iOS device model:  
iOS version: 
```

s123121:
```
XCode version:  13.4.1
iOS device model:  iphone 13
iOS version: 15.5
```

Did you enable dark appearance in Developer menu ?

PR #2046, opened 2022-05-12
Toast - support keyboard awareness (KeyboardAwareInsetsView)

M-i-k-e-l:
## Description
Toast - support keyboard awareness (KeyboardAwareInsetsView)
Known issues: `action` is not supported, sending `pointerEvents: 'box-none'` will cause the following error:
```
UIView base class does not support pointerEvent value: box-none
```
I am unsure if it's our native that needs change in order for this to be fixed.
Changing the text of the `Toast` "on-the-fly" does not work properly.

WOAUILIB-2685

## Changelog
Toast - support keyboard awareness (KeyboardAwareInsetsView)

### DIFF
--- a/src/incubator/toast/index.tsx
+++ b/src/incubator/toast/index.tsx
@@ -13,6 +13,7 @@ import {ToastProps, ToastPresets} from './types';
 import useToastTimer from './helpers/useToastTimer';
 import useToastPresets from './helpers/useToastPresets';
 import useToastAnimation from './helpers/useToastAnimation';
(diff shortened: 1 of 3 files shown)


lidord-wix (commented):
While using the `useKeyboardAwareInsetsView` prop, the Toast is not swipeable anymore, is that intended?

M-i-k-e-l:
> While using the `useKeyboardAwareInsetsView` prop, the Toast is not swipeable anymore, is that intended?

Intended, no. It's the same issue as the `action`. It might be solved in our native but that'll have to be solved in another ticket if it arrises.
Not sure if this is a blocker for UX though...

M-i-k-e-l:
UX decided against this feature, closing.

PR #2050, opened 2022-05-16
Migrate DateTimePicker to a function component

ethanshar:
## Description
In order to get rid of `TextField.extractOwnProps` and in preparation to migrate to the new TextField I migrate this component to be a function component 

## Changelog
Migrate DateTimePicker to a function component


ethanshar:
@Inbal-Tish 
ping

src/components/dateTimePicker/index.tsx
```diff
@@ -108,79 +108,66 @@ export interface DateTimePickerProps {
...
+    locale,
+    is24Hour,
+    minuteInterval,
+    timeZoneOffsetInMinutes,
+    themeVariant,
+    onChange,
+    dialogProps,
+    headerStyle,
+    // @ts-expect-error
+    useCustomTheme,
```
Inbal-Tish, line 133: Why ignore 'useCustomTheme' and not add ThemeComponent to the types? i.e. `(props: DurationPickerProps & ThemeComponen)`

src/components/dateTimePicker/index.tsx
```diff
@@ -108,79 +108,66 @@ export interface DateTimePickerProps {
...
+    testID,
+    ...others
+  } = props;
 
-  constructor(props: DateTimePickerPropsInternal) {
-    super(props);
-    this.chosenDate = props.value;
+  const [value, setValue] = useState(props.value);
```
Inbal-Tish, line 138: I think it will be more readable and less prone to bugs if you extract `value` and rename it `propsValue` instead of using it as `props.value`.

src/components/dateTimePicker/index.tsx
```diff
@@ -108,79 +108,66 @@ export interface DateTimePickerProps {
...
-    this.expandable.current?.toggleExpandable?.();
-  };
-
-  onDonePressed = () => {
-    this.toggleExpandableOverlay();
-    if (Constants.isIOS && !this.chosenDate) {
-      // since handleChange() is not called on iOS when there is no actual change
-      this.chosenDate = new Date();
-    }
+  useEffect(() => {
```
Inbal-Tish, line 148: This useEffect will be called in mount as well, which is unnecessary as you already set it in the useState statement. I suggest using `useDidUpdate` hook instead.

src/components/dateTimePicker/index.tsx
```diff
@@ -199,72 +186,58 @@ class DateTimePicker extends Component<DateTimePickerPropsInternal, DateTimePick
...
+      // since handleChange() is not called on iOS when there is no actual change
+      chosenDate.current = new Date();
+    }
 
-  renderHeader() {
-    // @ts-expect-error
-    const {headerStyle, useCustomTheme} = this.props;
+    onChange?.(chosenDate.current!);
+    setValue(chosenDate.current);
+  }, [toggleExpandableOverlay, onChange]);
```
Inbal-Tish, line 202: Don't you need to add `chosenDate` to the dependency array?

src/components/dateTimePicker/index.tsx
```diff
@@ -275,40 +248,65 @@ class DateTimePicker extends Component<DateTimePickerPropsInternal, DateTimePick
...
+    minimumDate,
+    maximumDate,
+    locale,
+    is24Hour,
+    minuteInterval,
+    timeZoneOffsetInMinutes,
+    themeVariant
+  ]);
 
+  const renderIOSExpandableOverlay = () => {
```
Inbal-Tish, line 264: Why not wrap with useCallback? It is passed as a prop...

src/components/dateTimePicker/index.tsx
ethanshar: It was actually ignored before, I didn't add it
And we don't have `ThemeComponent` type in public (only private). 

We'll eventually get rid of it, if you remember our conversation about deprecating the `useCustomTheme` prop all together.

src/components/dateTimePicker/index.tsx
ethanshar: Right! Fixed!

src/components/dateTimePicker/index.tsx
ethanshar: `chosenDate` is a ref, it has no effect on the callback.

src/components/dateTimePicker/index.tsx
ethanshar: Cause we're not passing the callback but its value

Inbal-Tish: oh I missed the ()...

src/components/dateTimePicker/index.tsx
Inbal-Tish: Actually you have the `useCustomTheme` prop on `asBaseComponent`...

Inbal-Tish (approved):

Issue #1499, opened 2021-08-26
TypeScript Definition Issue - KeyboardAwareScrollView


girishjjain:
### Describe the bug
Module '"react-native-ui-lib"' has no exported member 'KeyboardAwareScrollView'.

### To Reproduce
Add below import to .tsx file:
```
import { Button, KeyboardAwareScrollView, Modal, RadioButton, RadioGroup, TextField, View } from 'react-native-ui-lib';
```

### Expected behavior
TypeScript should not complain about KeyboardAwareScrollView component

### Code snippet
```
import { Button, KeyboardAwareScrollView, Modal, RadioButton, RadioGroup, TextField, View } from 'react-native-ui-lib';
```

### Device (please complete the following information)
 - MacOS 11.5.2

girishjjain:
`Keyboard.KeyboardAwareInsetsView` worked well for me. Thanks for your suggestion @ethanshar.

kkhhaalil:
Hi @ethanshar, I'm getting `Invariant Violation: requireNativeComponent: "KeyboardTrackingViewTemp" was not found in the UIManager.` while using `<Keyboard.KeyboardAwareInsetsView/>`, do you know what might be the reason?

tom555my:
Hi @MichalKrakow, I am also facing this issue.

Inbal-Tish:
@girishjjain Hi. I think it's an export issue. Please follow my PR #2064

M-i-k-e-l:
Hi,

@MichalKrakow, yes, you cannot use any of our **native** components in an expo project.

@kkhhaalil, @tom555my, @jtmozley  are you in an expo environment as well?

PR #2064, opened 2022-05-25
KeyboardAwareScrollView - fix exports

Inbal-Tish:
## Description
KeyboardAwareScrollView - fix exports

## Changelog
KeyboardAwareScrollView - fix exports

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,6 +84,10 @@ export {default as HapticService, HapticType} from './services/HapticService';
 export {default as Hint, HintProps} from './components/hint';
 export {default as Icon, IconProps} from './components/icon';
 export {default as Image, ImageProps} from './components/image';
+// @ts-expect-error
+export {default as KeyboardAwareScrollView} from './components/KeyboardAwareScrollView/KeyboardAwareScrollView';
+// @ts-expect-error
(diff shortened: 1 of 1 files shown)


M-i-k-e-l:
Hi @Inbal-Tish,

If we're trying to solve the OP's issue in #1499, then:
1. They switched to `Keyboard.KeyboardAwareInsetsView` which solved their problem.
2. I think this will have the same effect as a no export, since the user will have no declared props (it will be exported as `any`, as the error on the export suggests). So they will have to add an ignore as well, which is the same situation as it is now.

Inbal-Tish:
> Hi @Inbal-Tish,
> 
> If we're trying to solve the OP's issue in #1499, then:
> 
> 1. They switched to `Keyboard.KeyboardAwareInsetsView` which solved their problem.
> 2. I think this will have the same effect as a no export, since the user will have no declared props (it will be exported as `any`, as the error on the export suggests). So they will have to add an ignore as well, which is the same situation as it is now.

I'm aware of the workaround but if using the KeyboardAware component there is a problem with this export. Try to run our demo screen and you'll see

M-i-k-e-l (approved):

PR #2066, opened 2022-05-25
refector component drivers and add a new component drivers 

ronen0072:
## Description
refector component drivers and add a new component drivers 

## Changelog
refector component drivers and add a new component drivers


src/components/modal/index.tsx
```diff
@@ -90,7 +90,7 @@ class Modal extends Component<ModalProps> {
         <View
           useSafeArea={isScreenReaderEnabled}
           style={!isScreenReaderEnabled && [styles.touchableOverlay, {backgroundColor: overlayBackgroundColor}]}
-          testID={testID}
+          testID={`${testID}.TouchableOverlay`}
```
ethanshar, line 93: This is a breaking change. 
People with tests on the modal might break

ethanshar (changes requested):

ethanshar (approved):

PR #2067, opened 2022-05-26
Fix/dialog new support percentage size

M-i-k-e-l:
## Description
Incubator.Dialog - support percentage size (width \ height)
Also improve some things:
Use `View`'s `useSafeArea`.
Use `View`'s `reanimated`.

## Changelog
Incubator.Dialog - support percentage size (width \ height)


M-i-k-e-l:
Alternative solution (different implementation of the `Incubator.Dialog`) solves this.

PR #2088, opened 2022-06-14
Infra/pan view refactor

M-i-k-e-l:
## Description
Extract `usePanGesture` from `PanView` to allow usage directly (internal, at least for now).

## Changelog
PanView refactor (infra)


src/incubator/panView/usePanGesture.ts
```diff
@@ -0,0 +1,141 @@
...
+  onDismiss?: () => void;
+  /**
+   * Should the direction of dragging be locked once a drag has started.
+   */
+  directionLock?: boolean;
+  /**
+   * Object to adjust the dismiss threshold limits (eg {x, y, velocity}).
+   */
+  threshold?: PanViewDismissThreshold;
+  hiddenLocation: HiddenLocation;
```
ethanshar, line 45: Does it make sense to call it `initialPosition`? 
We refer it as hidden because of the dialog but I think initial is more appropriate and clear.. WDYT?

src/incubator/panView/usePanGesture.ts
```diff
@@ -0,0 +1,141 @@
...
+const SPRING_BACK_ANIMATION_CONFIG = {velocity: 300, damping: 20, stiffness: 300, mass: 0.8};
+
+const usePanGesture = (props: PanGestureProps) => {
+  const {
+    directions = [
+      PanViewDirectionsEnum.UP,
+      PanViewDirectionsEnum.DOWN,
+      PanViewDirectionsEnum.LEFT,
+      PanViewDirectionsEnum.RIGHT
+    ],
```
ethanshar, line 57: Move value to const at the top

ethanshar (approved):
@M-i-k-e-l 
Wrote 2 small comments, anyway approved

src/incubator/panView/usePanGesture.ts
M-i-k-e-l: Actually it's a destination and not initial position; moreover, the logic is relying on it being of the type `HiddenLocation` (having `left`, `right`, `up` and `down`, which will be used to hide the view) so I'm not sure.
Thinking about it, perhaps the name of the component \ hook are wrong.

ethanshar (approved):

PR #2100, opened 2022-06-22
Support useWheelPicker in Picker component

ethanshar:
## Description
Support useWheelPicker in Picker component
This should replace useNativePicker implementation 

## Changelog
Support useWheelPicker prop in Picker component in order to render a wheel picker component

### DIFF
--- a/demo/src/screens/componentScreens/PickerScreen.tsx
+++ b/demo/src/screens/componentScreens/PickerScreen.tsx
@@ -137,7 +137,8 @@ export default class PickerScreen extends Component {
           <Picker
             title="Native Picker"
             placeholder="Pick a Language"
-            useNativePicker
+            // useNativePicker
+            useWheelPicker
             value={this.state.nativePickerValue}
             onChange={nativePickerValue => this.setState({nativePickerValue})}
             rightIconSource={dropdown}
--- a/src/components/picker/index.tsx
+++ b/src/components/picker/index.tsx
@@ -43,6 +43,12 @@ import {
 
 const dropdown = require('./assets/dropdown.png');
 
+const DIALOG_PROPS = {
+  bottom: true,
+  width: '100%',
+  height: 250
+};
+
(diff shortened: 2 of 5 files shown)


demo/src/screens/componentScreens/PickerScreen.tsx
```diff
@@ -137,7 +137,8 @@ export default class PickerScreen extends Component {
           <Picker
             title="Native Picker"
```
Inbal-Tish, line 138: Change the title too. Also, don't you want to keep both examples for native and wheel pickers?

src/components/picker/index.tsx
```diff
@@ -190,6 +197,8 @@ const Picker = (props: PropsWithChildren<PickerProps> & ForwardRefInjectedProps
     return (
       <PickerItemsList
         testID={`${testID}.modal`}
+        useWheelPicker={useWheelPicker}
+        items={items}
```
Inbal-Tish, line 201: Shouldn't you pass 'items' only if 'useWheelPicker' is true?

demo/src/screens/componentScreens/PickerScreen.tsx
ethanshar: Well, I want to deprecate useNativePicker, you mean for the sake of sanity checks - to see nothing was broken with useNativePicker prop?

src/components/picker/index.tsx
ethanshar: I can add a condition, but it doesn't really matter or affect the code.

Inbal-Tish: I think it's better to secure no side-effects

demo/src/screens/componentScreens/PickerScreen.tsx
Inbal-Tish: Sure, if you're deprecating the 'useNativePicker' now there's no point having its example (but the title should be changed to avoid confusion).

Inbal-Tish (approved):

Issue #2048, opened 2022-05-16
StackAggregator change Children


chpppeng:
StackAggregator  is worked well but when I want to change children, It caused  TypeError: undefined is not an object (evaluating 'value.getValue') 

### Related to

- [x] Components
- [x] Demo

### Steps to reproduce

1. Run the sample code ,
2. Click "test“ button to add data to contents ,
3.  TypeError: undefined is not an object (evaluating 'value.getValue') 

#### Actual behavior
When change the children map, it caused exception.

### Code snippet

```
export default class StackAggregatorScreen extends Component {
    state = { contents: [
        'Lorem Ipsum is simply dummy text of the printing and typesetting industry.',
        'Lorem Ipsum is simply dummy text of the printing and typesetting industry.',
        'Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum is simply dummy text of the printing and typesetting industry.',
        'Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum is simply dummy text of the printing and typesetting industry.'
      ] };
  
  onItemPress = (index: number) => {
    console.warn('item pressed: ', index);
  }

  onPress = (index: number) => {
    console.warn('item\'s button pressed: ', index);
  }

  refreshItems=()=>{
      var items = this.state.contents;
        items.push('test')
        this.setState({
            contents: items
        })

  }
  renderItem = (_: string, index: number) => {
    return (
      <View key={index} center padding-12>
        <Button label={`${index}`} marginB-10 size={Button.sizes.small} onPress={() => this.onPress(index)}/>
        <Text>{this.state.contents[index]}</Text>
      </View>
    );
  }

  render() {
    return (
        
      <ScrollView keyboardShouldPersistTaps={'handled'} showsVerticalScrollIndicator={false}>
            <Button label='test' onPress={()=>{this.refreshItems()}} />
        <Text center grey40 text90 marginT-20>Thu, 10 Dec, 11:29</Text>
        <StackAggregator
          containerStyle={{marginTop: 12}}
          onItemPress={this.onItemPress}
        >
          {this.state.contents.map((item, index) => {
            return this.renderItem(item, index);
          })}
        </StackAggregator>
      </ScrollView>
    );
  }
}

```

[...]

PR #2103, opened 2022-06-27
StackAggregator - fix children update

Inbal-Tish:
## Description
StachAggregator - fix children update. 
Moving to a functional component.
fix for issue #2048 

## Changelog
StachAggregator - fix children update.


ethanshar:
@Inbal-Tish 
I noticed that the first time you open the stack it doesn't have an animation. 
It works fine on the master version

Inbal-Tish:
> @Inbal-Tish I noticed that the first time you open the stack it doesn't have an animation. It works fine on the master version

Fixed

src/components/stackAggregator/index.tsx
```diff
@@ -51,275 +52,275 @@ export type StackAggregatorProps = ViewProps & {
...
+    contentContainerStyle,
+    itemBorderRadius,
+    onCollapseWillChange,
+    onCollapseChanged
+  } = props;
+  const itemsCount = React.Children.count(children);
+  const [firstItemHeight, setFirstItemHeight] = useState<number>();
+  const [isCollapsed, setIsCollapsed] = useState(collapsed);
+  
+  LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
```
ethanshar, line 82: You can't trigger a LayoutAnimation like that. It will be called in each render and will also affect everything on the screen, not just the stack aggregator

ethanshar (changes requested):

src/components/stackAggregator/index.tsx
Inbal-Tish: Fixed

ethanshar (approved):

PR #2111, opened 2022-07-06
Allow migration of ChipsInput's TextField

M-i-k-e-l:
## Description
Allow migration of ChipsInput's TextField
WOAUILIB-2818 (SearchInput uses the old ChipsInput which in turn uses the old TextField 🤷‍♂️)

## Changelog
Allow migration of ChipsInput's TextField


Inbal-Tish (approved):

ethanshar:
@M-i-k-e-l 
Notice that there's a new ChipsInput under Incubator. 
I just realized that we never moved to the new ChipsInput in SearchInput..

PR #2112, opened 2022-07-07
Add design tokens in iOS

ethanshar:
## Description
Add design tokens in iOS so we can test platform colors

## Changelog
N/A

### DIFF
--- a/ios/DesignTokens.xcassets/$backgroundDanger.colorset/Contents.json
+++ b/ios/DesignTokens.xcassets/$backgroundDanger.colorset/Contents.json
@@ -1,0 +1,38 @@
+{
+  "colors" : [
+    {
+      "color" : {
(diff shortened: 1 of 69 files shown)


Inbal-Tish:
@ethanshar Do we need to keep the private version of these Design tokens or we can remove the duplications now?

ethanshar:
The design tokens are part of the demo app, they are not included in the public lib package. 
Each demo/app need to add their own native design tokens

Inbal-Tish:
@ethanshar Don't we need to reflect the tokens name change here as well?

ethanshar:
> @ethanshar Don't we need to reflect the tokens name change here as well?

Yes, we do

ethanshar:
I'll create a new PR

PR #2116, opened 2022-07-10
Create a Dialog that uses hooks

M-i-k-e-l:
## Description
Change the `Incubator.Dialog` to use (internal) hooks

## Changelog
Change the `Incubator.Dialog` to use hooks, fixes issues with size given in percentage and not numbers


src/incubator/TransitionView/useAnimatedTransition.ts
```diff
@@ -2,7 +2,7 @@
 import {useEffect, useCallback} from 'react';
 import {runOnJS, useSharedValue, withSpring, withTiming} from 'react-native-reanimated';
 import {HiddenLocation} from '../hooks/useHiddenLocation';
-import {PanningDirections, PanningDirectionsEnum} from '../panView';
+import {PanningDirections, PanningDirectionsEnum, DEFAULT_ANIMATION_CONFIG} from '../panView';
```
ethanshar, line 5: We've might already discuss this...
Don't you think It's a little weird that this hook that sits under TransitionView, takes stuff from another component?
How coupled are these two components? Isn't it better in this case to duplicate the type/configs?

src/incubator/Dialog/DialogHeader.tsx
```diff
@@ -4,11 +4,12 @@ import {StyleSheet} from 'react-native';
 import {asBaseComponent} from '../../commons/new';
 import {Spacings, Colors, BorderRadiuses, Dividers} from 'style';
 import View from '../../components/view';
-import HeaderContent from './HeaderContent';
+import Text from '../../components/text';
 import {DialogHeaderProps} from './types';
 
 const DialogHeader = (props: DialogHeaderProps = {}) => {
   const {text = {}, renderContent, showKnob = true, showDivider = true, ...others} = props;
+  const {title, titleStyle, titleProps, subtitle, subtitleStyle, subtitleProps} = text;
```
ethanshar, line 12: Does dialog header has a `text` prop that contain all of these? 
I think `text` is not a good name here... 

Is it internal or an API we're obligated to?

src/incubator/TransitionView/useAnimatedTransition.ts
```diff
@@ -2,7 +2,7 @@
 import {useEffect, useCallback} from 'react';
```
ethanshar, line 2: Btw, is it possible that we only need the `useAnimatedTransition` hook and don't need the actual TransitionView component anymore?

src/incubator/TransitionView/useAnimatedTransition.ts
M-i-k-e-l: From what I remember we've talked about the props, and that they are exported in another name, so they are the same values but independent if we want to migrate only one of them.
If you mean `DEFAULT_ANIMATION_CONFIG`, then it's not a real dependency at it can be changed (it is also only internal usage so I'm not sure if it's that important).

src/incubator/Dialog/DialogHeader.tsx
M-i-k-e-l: No, this is new, I agree it's not a very good name :/
I wanted to separate some props, do you prefer another name or to put them all under `headerProps?: DialogHeaderProps` (which already exists and has a few props)?

src/incubator/TransitionView/useAnimatedTransition.ts
M-i-k-e-l: It is possible, let me know if you want it removed, I prefer to do it in a different PR.

src/incubator/Dialog/DialogHeader.tsx
```diff
@@ -21,7 +22,24 @@ const DialogHeader = (props: DialogHeaderProps = {}) => {
       return renderContent(props);
     }
 
-    return <HeaderContent text={text}/>;
+    if (title || subtitle) {
```
Effanuel, line 25: Why boolean comparison here is simple, but below it uses `isEmpty`/.

M-i-k-e-l: Good point

src/incubator/Dialog/DialogHeader.tsx
M-i-k-e-l: Ping

M-i-k-e-l: Done

ethanshar (approved):

PR #2117, opened 2022-07-11
fix/grid list list padding

ethanshar:
## Description
Move GridList listPadding logic to useGridLayout hook to share it with SortableGridList

## Changelog
Move GridList listPadding logic to useGridLayout hook to share it with SortableGridList


src/components/gridList/types.ts
```diff
@@ -1,32 +1,35 @@
...
+   */
+  keepItemSize?: boolean;
+  /**
+   * Pass when you want to use a custom container width for calculation
+   */
+  containerWidth?: number;
+  /**
+   * Custom content container style
+   */
+  contentContainerStyle?: ScrollViewProps['contentContainerStyle'];
```
M-i-k-e-l, line 32: Consider adding a `Pick<ScrollViewProps, 'contentContainerStyle'>` at the top instead, in any case this should probably be in the API.

ethanshar: Fixed

M-i-k-e-l (approved):

PR #2134, opened 2022-07-24
Refactor text to optimize performance

ethanshar:
## Description
**Don't Merge**, just review

I did a few things here
- Create new hooks: `useModifiers` and `useThemeProps` - this should help us replace `asBaseComponent`
- Added support in ThemeManager to set a ThemeContext - This should replace the static context type we set on UIComponent
- Refactor Text component to a function component to use the all of the above

Due to this big change, I left with some issues I had to resolve 
- because `Animated.createAnimatedComponent` **can't** accept function components, only class components. I created a new (very) simple HOC called `asClassComponent`. All it does is convert a function component to a class component (we need this for very rare cases, specifically in our WheelPicker.Item component)
- I ignored a TS issue that now appears in FloatingPlaceholder component - I couldn't figure out a solution for this. If you have ideas, do share.. 

## Changelog
Refactor and improve Text component performance

### DIFF
--- a/src/commons/asClassComponent.tsx
+++ b/src/commons/asClassComponent.tsx
@@ -1,0 +1,13 @@
+import React from 'react';
+
+function asClassComponent(WrappedComponent: React.ComponentType<any>): React.ComponentClass<any> {
+  class ClassComponent extends React.Component {
+    render() {
+      return <WrappedComponent {...this.props}/>;
+    }
+  }
+
+  return ClassComponent;
+}
+
+export default asClassComponent;
--- a/src/commons/modifiers.ts
+++ b/src/commons/modifiers.ts
@@ -32,7 +32,7 @@ export interface ExtractedStyle {
   positionStyle?: ReturnType<typeof extractPositionStyle>;
 }
 
-type ModifiersOptions = {
+export type ModifiersOptions = {
   color?: boolean;
   typography?: boolean;
   backgroundColor?: boolean;
--- a/src/commons/modifiers.ts
+++ b/src/commons/modifiers.ts
@@ -32,6 +32,18 @@ export interface ExtractedStyle {
   positionStyle?: ReturnType<typeof extractPositionStyle>;
 }
 
+type ModifiersOptions = {
+  color?: boolean;
+  typography?: boolean;
+  backgroundColor?: boolean;
+  borderRadius?: boolean;
+  paddings?: boolean;
+  margins?: boolean;
+  alignments?: boolean;
+  flex?: boolean;
+  position?: boolean;
+};
+
 const PADDING_VARIATIONS = {
   padding: 'padding',
   paddingL: 'paddingLeft',
@@ -322,9 +334,10 @@ export function extractComponentProps(component: any, props: Dictionary<any>, ig
 }
 
 //@ts-ignore
-export function getThemeProps(props = this.props, context = this.context) {
-  //@ts-ignore
-  const componentName = this.displayName || this.constructor.displayName || this.constructor.name;
+export function getThemeProps(props = this.props, context = this.context, componentDisplayName = '') {
+  const componentName =
+    //@ts-ignore
+    componentDisplayName || this.displayName || this.constructor.displayName || this.constructor.name;
 
   let themeProps;
   if (_.isFunction(ThemeManager.components[componentName])) {
@@ -342,7 +355,7 @@ export function getThemeProps(props = this.props, context = this.context) {
   return {...themeProps, ...props, ...forcedThemeProps};
 }
 
-export function generateModifiersStyle(options = {
+export function generateModifiersStyle(options: ModifiersOptions = {
   color: true,
   typography: true,
   backgroundColor: true,
--- a/src/components/text/index.tsx
+++ b/src/components/text/index.tsx
@@ -78,7 +78,6 @@ const Text = (props: TextProps, ref: any) => {
     color: true,
     backgroundColor: true,
     margins: true,
-    position: true,
     flex: true
   });
 
--- a/src/components/text/index.tsx
+++ b/src/components/text/index.tsx
@@ -1,17 +1,10 @@
-import React, {PureComponent} from 'react';
+import React, {useMemo} from 'react';
 import {Text as RNText, StyleSheet, TextProps as RNTextProps, TextStyle, Animated, StyleProp} from 'react-native';
 import _ from 'lodash';
-import {
-  asBaseComponent,
-  forwardRef,
-  BaseComponentInjectedProps,
-  ForwardRefInjectedProps,
-  MarginModifiers,
-  TypographyModifiers,
-  ColorsModifiers
-} from '../../commons/new';
+import {MarginModifiers, TypographyModifiers, ColorsModifiers} from '../../commons/new';
 import {Colors} from 'style';
 import {TextUtils} from 'utils';
+import {useModifiers, useThemeProps} from 'hooks';
 
 export type TextProps = RNTextProps &
   TypographyModifiers &
@@ -50,7 +43,7 @@ export type TextProps = RNTextProps &
   };
 export type TextPropTypes = TextProps; //TODO: remove after ComponentPropTypes deprecation;
 
-type PropsTypes = BaseComponentInjectedProps & ForwardRefInjectedProps & TextProps;
+const AnimatedText = Animated.createAnimatedComponent(RNText);
 
 /**
  * @description: A wrapper for Text component with extra functionality like modifiers support
@@ -60,23 +53,40 @@ type PropsTypes = BaseComponentInjectedProps & ForwardRefInjectedProps & TextPro
  * @example: https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/TextScreen.tsx
  * @image: https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/Text/Modifiers.png?raw=true, https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/Text/Transformation.png?raw=true, https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/Text/Highlights.png?raw=true
  */
-class Text extends PureComponent<PropsTypes> {
-  static displayName = 'Text';
-  private TextContainer: React.ClassType<any, any, any> = this.props.animated
-    ? Animated.createAnimatedComponent(RNText)
-    : RNText;
+const Text = (props: TextProps, ref: any) => {
+  const themeProps = useThemeProps(props, 'Text');
+  const {
+    animated,
+    children,
+    highlightString,
+    highlightStyle,
+    style,
+    center,
+    uppercase,
+    underline,
+    color: propsColor,
+    ...others
+  } = themeProps;
+  const {
+    margins,
+    typography,
+    backgroundColor,
+    flexStyle,
+    color: modifiersColor
+  } = useModifiers(others, {
+    typography: true,
+    color: true,
+    backgroundColor: true,
+    margins: true,
+    position: true,
+    flex: true
+  });
 
-  // setNativeProps(nativeProps) {
-  //   this._root.setNativeProps(nativeProps); // eslint-disable-line
-  // }
-
-  renderText(children: any): any {
-    const {highlightString, highlightStyle} = this.props;
-
+  const renderText = (children: any): any => {
     if (!_.isEmpty(highlightString)) {
       if (_.isArray(children)) {
-        return _.map(children, child => {
-          return this.renderText(child);
+        return _.map(children, (child: any) => {
+          return renderText(child);
         });
       }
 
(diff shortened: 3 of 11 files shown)


src/incubator/TextField/FloatingPlaceholder.tsx
```diff
@@ -67,6 +67,7 @@ const FloatingPlaceholder = ({
 
   return (
     <View absF style={hidePlaceholder && styles.hidden}>
+      {/* @ts-expect-error TS error is caused due to animatedStyle passed to style prop */}
```
Inbal-Tish, line 70: Not sure it will help but did you try changing the Text's style type to `Animated.AnimatedProps<**StyleProp**<TextStyle>**>**`?

src/incubator/WheelPicker/Item.tsx
```diff
@@ -4,11 +4,11 @@ import Animated, {interpolateColor, useAnimatedStyle} from 'react-native-reanima
 import Text, {TextProps} from '../../components/text';
 import TouchableOpacity from '../../components/touchableOpacity';
 import {Colors, Spacings} from '../../../src/style';
-import {asBaseComponent} from '../../commons/new';
+import {asBaseComponent, asClassComponent} from '../../commons/new';
 import {WheelPickerAlign} from './types';
 
 const AnimatedTouchableOpacity = Animated.createAnimatedComponent(TouchableOpacity);
-const AnimatedText = Animated.createAnimatedComponent(Text);
+const AnimatedText = Animated.createAnimatedComponent(asClassComponent(Text));
```
Inbal-Tish, line 11: Can't we just use our Text component with `animated = true` instead?

src/incubator/TextField/FloatingPlaceholder.tsx
M-i-k-e-l: I think the way to go is (in `Text`):
```
export type TextProps = Animated.AnimatedProps<RNTextProps>...
```
and remove:
```
style?: StyleProp<TextStyle | Animated.AnimatedProps<TextStyle>>;
```
from there

src/hooks/index.ts
```diff
@@ -2,6 +2,8 @@ export {default as useCombinedRefs} from './useCombinedRefs';
 export {default as useToggleValue} from './useToggleValue';
 export {default as useDidUpdate} from './useDidUpdate';
 export {default as useMeasure} from './useMeasure';
+export {default as useModifiers} from './useModifiers';
+export {default as useThemeProps} from './useThemeProps';
```
M-i-k-e-l, line 6: Do we actually want to export them? Is this for private lib usage?

src/style/themeManager.ts
```diff
@@ -1,9 +1,12 @@
 import _ from 'lodash';
+import type {Context} from 'react';
 import Colors from './colors';
 
 
 export class ThemeManager {
```
M-i-k-e-l, line 6: Not sure if you've tested this in private's node_modules, but it might be possible to do:
```
ThemeManager<ContextType = any>
```
(and change the `any`s to `ContextType`)
Which might make usage prettier.

src/style/themeManager.ts
```diff
@@ -1,9 +1,12 @@
...
 
+  private themeContext: Context<any> | undefined;
```
M-i-k-e-l, line 8:
```suggestion
  private themeContext?: Context<ContextType>;
```

src/hooks/useModifiers/index.ts
```diff
@@ -0,0 +1,13 @@
+import {useMemo} from 'react';
+import * as Modifiers from '../../commons/modifiers';
+
+type Options = Parameters<typeof Modifiers.generateModifiersStyle>[0];
```
M-i-k-e-l, line 4: Why not export `ModifiersOptions` from `Modifiers`?

src/incubator/TextField/FloatingPlaceholder.tsx
ethanshar: It didn't work :(

src/incubator/WheelPicker/Item.tsx
ethanshar: Our Text supports `animated` prop to render React Native Animated
But in this case we use Reanimated Text. 
I want to add a `reanimated` prop to Text that will return a Reanimated Text component. 
But i didn't want to do it in this PR. 

in any case, `asClassComponent` is a good HOC, cause it's not the first time I came across this issue
(we had it after migrating AgendaList to a function component)

src/style/themeManager.ts
```diff
@@ -38,6 +41,14 @@ export class ThemeManager {
     return _.get(this.theme, key);
   }
 
+  setThemeContext(context: Context<any>) {
```
lidord-wix, line 44: Where will you set the theme context? in the private config files?

src/commons/modifiers.ts
```diff
@@ -322,9 +334,10 @@ export function extractComponentProps(component: any, props: Dictionary<any>, ig
 }
 
 //@ts-ignore
-export function getThemeProps(props = this.props, context = this.context) {
-  //@ts-ignore
-  const componentName = this.displayName || this.constructor.displayName || this.constructor.name;
+export function getThemeProps(props = this.props, context = this.context, componentDisplayName = '') {
```
lidord-wix, line 337: I know it's not related to this PR, but what was the idea of putting this function here? is it related to the modifiers somehow?

src/incubator/TextField/FloatingPlaceholder.tsx
lidord-wix: try: 
`        style={[styles.placeholder, floatingPlaceholderStyle, animatedStyle] as TextStyle}
`

lidord-wix: try:
`style={[styles.placeholder, floatingPlaceholderStyle, animatedStyle] as TextStyle}
`

M-i-k-e-l: > It didn't work :(

Very weird, it worked for me locally...

src/commons/asClassComponent.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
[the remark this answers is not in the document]
ethanshar: We can do that.. we can then support both reanimated and animated

src/style/colors.ts
```diff
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 //@ts-ignore
 import Color from 'color';
-import {OpaqueColorValue} from 'react-native';
+import type {OpaqueColorValue} from 'react-native';
```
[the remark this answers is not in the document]
ethanshar: Just notice that and added it.. 
technically when u import only a `type` it should remove this import in the transpiled javascript file.. cause it's not really needed

src/hooks/index.ts
ethanshar: Yes

src/style/themeManager.ts
ethanshar: But i'm not instantiating ThemeManager in private, but in public. 
We kinda export it as a singleton instance, so I can't really pass a generics

src/hooks/useModifiers/index.ts
ethanshar: Fixed

src/components/text/index.tsx
```diff
@@ -60,23 +53,40 @@ type PropsTypes = BaseComponentInjectedProps & ForwardRefInjectedProps & TextPro
...
+    typography,
+    backgroundColor,
+    flexStyle,
+    color: modifiersColor
+  } = useModifiers(others, {
+    typography: true,
+    color: true,
+    backgroundColor: true,
+    margins: true,
+    position: true,
```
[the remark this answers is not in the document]
ethanshar: Right, removed

src/components/text/index.tsx
```diff
@@ -60,23 +53,40 @@ type PropsTypes = BaseComponentInjectedProps & ForwardRefInjectedProps & TextPro
...
+    uppercase,
+    underline,
+    color: propsColor,
+    ...others
+  } = themeProps;
+  const {
+    margins,
+    typography,
+    backgroundColor,
+    flexStyle,
```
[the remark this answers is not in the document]
ethanshar: I agree, but that will enforce me changing in all components that use this modifier. 
We can renaming it in a separate PR maybe

src/style/themeManager.ts
ethanshar: Indeed. 
If you don't set, nothing happens.. 
This is basically a good API for us and other consumer to have global theme context

src/incubator/TextField/FloatingPlaceholder.tsx
ethanshar: @M-i-k-e-l Are you sure? Try doing it and run `npm run build:dev` don't depend on the vscode editor

@lidord-wix 
This is basically the same as ignoring, I prefer to add `ts-expect-error` with a commet, maybe the issue will be resolved in the future.

src/style/themeManager.ts
M-i-k-e-l: I did not test it, but I think it might affect the usage of `ThemeManager.getThemeContext()`, it might give us autocomplete with the generics added (used as `ThemeManager<ThemeContextProps>.getThemeContext()`, I think :S ).

ethanshar: Not sure I understand, how would you use that exactly? In private for example?

M-i-k-e-l: I assume that the getter needs to be used in private.
We can postpone this so it's easier to test.

src/incubator/TextField/FloatingPlaceholder.tsx
M-i-k-e-l: Retested now, pretty sure I clean-installed before testing this branch; I hope it's not my env, can you double check please?

src/components/text/index.tsx
```diff
@@ -116,16 +125,25 @@ class Text extends PureComponent<PropsTypes> {
...
+    propsColor,
+    modifiersColor,
+    typography,
+    backgroundColor,
+    flexStyle,
+    margins,
+    center,
+    uppercase,
+    underline,
+    style
```
Effanuel, line 138: In a lot of cases, new `style` reference still gets passed to `<Text />`. There, this `useMemo` will be redundant.

src/hooks/useModifiers/index.ts
```diff
@@ -0,0 +1,11 @@
+import {useMemo} from 'react';
+import * as Modifiers from '../../commons/modifiers';
+
+const useModifiers = (props: any, options: Modifiers.ModifiersOptions) => {
+  const modifiers = useMemo(() => {
+    return Modifiers.generateModifiersStyle(options, props);
+  }, [props]);
```
Effanuel, line 7: Do you really need to depend on all props here?

src/hooks/useThemeProps/index.ts
```diff
@@ -0,0 +1,13 @@
+import {useContext, createContext} from 'react';
+import * as Modifiers from '../../commons/modifiers';
+import {ThemeManager} from 'style';
+
+const EmptyContext = createContext({});
+
+const useThemeProps = (props: any, componentName: string) => {
+  const themeContext = ThemeManager.getThemeContext();
```
Effanuel, line 8: Wouldn't it be better to return empty theme context inside `.getThemeContext` or is that not an option?

src/hooks/useThemeProps/index.ts
```diff
@@ -0,0 +1,13 @@
...
+  const context = useContext(themeContext ?? EmptyContext);
```
Effanuel, line 9: I would expect context to have a method `getThemeProps`, instead of using external function `Modifiers.getThemeProps`

src/style/colors.ts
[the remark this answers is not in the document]
Effanuel: You should consider adding a lint rule, that would force you to have `import type` for types

src/components/text/index.tsx
```diff
@@ -50,7 +43,7 @@ export type TextProps = RNTextProps &
   };
 export type TextPropTypes = TextProps; //TODO: remove after ComponentPropTypes deprecation;
 
-type PropsTypes = BaseComponentInjectedProps & ForwardRefInjectedProps & TextProps;
+const AnimatedText = Animated.createAnimatedComponent(RNText);
```
Effanuel, line 46: Isn't this going to cause worse performance, because now, this will be called every single time, when before, only when `Text` is animated

src/commons/asClassComponent.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+function asClassComponent(WrappedComponent: React.ComponentType<any>): React.ComponentClass<any> {
+  class ClassComponent extends React.Component {
+    render() {
+      return <WrappedComponent {...this.props}/>;
+    }
+  }
+
+  return ClassComponent;
+}
```
Effanuel, line 11: What does this even do?

src/components/text/index.tsx
```diff
@@ -60,23 +53,39 @@ type PropsTypes = BaseComponentInjectedProps & ForwardRefInjectedProps & TextPro
...
+    color: modifiersColor
+  } = useModifiers(others, {
+    typography: true,
+    color: true,
+    backgroundColor: true,
+    margins: true,
+    flex: true
+  });
 
+  const renderText = (children: any): any => {
```
Effanuel, line 84: This looks like a heavy callback. I think you should wrap it with `useMemo`

src/components/text/index.tsx
Effanuel: Are you planning to fix those spots as well? Just an example: https://github.com/wix/react-native-ui-lib/blob/master/src/components/avatar/index.tsx#L269

PR #2136, opened 2022-07-25
Infra/modifiers perf improvement

lidord-wix:
## Description
Update some modifiers extraction methods to improve performance. 
I checked a specific session (on the private demo app, iOS) before and after these changes. the results are in ms: 
```
before: {
  "backgroundColor": 291.95974069833755,
  "color": 1362.02895539999,
  "typography": 878.3454459905624
},
after: {
  "backgroundColor": 54.971199452877045,
  "color": 97.36799198389053,
  "typography": 85.1839787364006
},
```
Also, I added a function to update the modifiers' colors that should be called after loading some colors/scheme.
(and I call it on private in `preset.config.ts`)

## Changelog
Improve modifiers' performance

### DIFF
--- a/src/commons/modifiers.ts
+++ b/src/commons/modifiers.ts
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import {StyleSheet} from 'react-native';
-import {Typography, Colors, Scheme, DesignTokens, BorderRadiuses, Spacings, ThemeManager} from '../style';
+import {Typography, Colors, DesignTokens, BorderRadiuses, Spacings, ThemeManager} from '../style';
 import {BorderRadiusesLiterals} from '../style/borderRadiuses';
 import TypographyPresets from '../style/typographyPresets';
 import {SpacingLiterals} from '../style/spacings';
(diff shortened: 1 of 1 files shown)


src/commons/modifiers.ts
```diff
@@ -100,30 +103,34 @@ export type ContainerModifiers = AlignmentModifiers &
   BorderRadiusModifiers &
   BackgroundColorModifier;
 
+export function updateColorModifiers() {
```
ethanshar, line 106: If you'll think about it, maybe it's better that the Colors class will maintain this logic and provide a cached list of colors keys. 
By moving this logic to the Colors class, you can also update the cached list whenever loadColors is being called - so you don't need to worry about it as a consumer. 
WDYT?

lidord-wix: I wondered about it.. you can see my note in this PR description: https://github.com/wix-private/wix-react-native-ui-lib/pull/2473

ethanshar: Oh ok, I didn't see the private PR yet. 
I think we should call and keep it in the Colors class, what are the cons for doing that?

lidord-wix: The only disadvantage I can think of is that it'll cause an overhead in the `loadColors` method in a case that the user calls it many times in his app..  but I guess it's a rare case and we can ignore it
I'll update it :)

ethanshar (approved):

PR #2143, opened 2022-07-26
Avoid modifiers calculation if there are no boolean truthy props

ethanshar:
## Description
Avoid modifiers calculation if there are no boolean truthy props

## Changelog
Optimize modifiers


lidord-wix:
@ethanshar 
Approved, but the build is red.
Merge once it's passed

Inbal-Tish:
@ethanshar This won't break modifiers support for components that did not yet pass the modifiers options (like Button)?

PR #2145, opened 2022-07-26
Modifiers and ThemeProps hooks

ethanshar:
## Description
Create infra for useModifiers and useThemeProps hooks

## Changelog
Create infra for useModifiers and useThemeProps hooks


ethanshar:
I broke the PR so it won't have dependency to the Text refactor so we can merge this and play with it

src/hooks/useThemeProps/index.ts
```diff
@@ -0,0 +1,13 @@
+import {useContext, createContext} from 'react';
+import * as Modifiers from '../../commons/modifiers';
+import {ThemeManager} from 'style';
+
+const EmptyContext = createContext({});
+
+const useThemeProps = (props: any, componentName: string) => {
+  const themeContext = ThemeManager.getThemeContext();
+  const context = useContext(themeContext ?? EmptyContext);
+  return Modifiers.getThemeProps(props, context, componentName);
```
M-i-k-e-l, line 10: I've just now noticed this is not memoized, it does create a new object and spreads props inside, WDYT?

ethanshar: it's ok it's not memoize because of the spreading.. 
the inner props refs are kept, and that's what important

M-i-k-e-l (approved):

PR #2146, opened 2022-07-26
Infra/refactor view - convert to function component and use modifiers and themeProps hooks

ethanshar:
## Description
Convert View to a function component to reduce render hierarchy 

depends on https://github.com/wix/react-native-ui-lib/pull/2145

## Changelog
Convert View to a function component to reduce render hierarchy


src/components/view/index.tsx
```diff
@@ -46,106 +40,90 @@ export interface ViewProps extends Omit<RNViewProps, 'style'>, ContainerModifier
...
-  alignments: true,
-  flex: true,
-  position: true
-};
+  return (
+    <ViewContainer
+      accessibilityElementsHidden={inaccessible}
+      importantForAccessibility={inaccessible ? 'no-hide-descendants' : undefined}
+      {...others}
+      style={[
```
M-i-k-e-l, line 112: We should probably memoize the `style`

src/components/view/index.tsx
```diff
@@ -46,106 +40,90 @@ export interface ViewProps extends Omit<RNViewProps, 'style'>, ContainerModifier
...
-          alignments,
-          style
-        ]}
-        ref={forwardedRef}
-      >
-        {this.props.children}
-      </Element>
-    );
+    return container;
+  }, [useSafeArea, animated, reanimated]);
```
M-i-k-e-l, line 101: I can see `useSafeArea` changing, but probably not the other two

typings/globalTypes.d.ts
```diff
@@ -14,3 +14,7 @@ declare module 'react-native-measureme';
 interface Extendable {
   [key: string]: any;
 }
+
+interface ThemeComponent {
```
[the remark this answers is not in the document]
ethanshar: Removed

src/components/view/index.tsx
```diff
@@ -46,106 +40,90 @@ export interface ViewProps extends Omit<RNViewProps, 'style'>, ContainerModifier
...
-    } else if (props.animated) {
-      this.Container = Animated.createAnimatedComponent(this.Container);
-    }
-
-    this.state = {
-      ready: !props.renderDelay
-    };
-  }
+function View(props: ViewProps, ref: any) {
+  const themeProps = useThemeProps(props, 'View');
```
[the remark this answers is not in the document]
ethanshar: I think that by default the component name is used as display name

src/components/view/index.tsx
ethanshar: I agree, And if they won't change there's no harm having them here (:

src/components/view/index.tsx
```diff
@@ -46,106 +40,90 @@ export interface ViewProps extends Omit<RNViewProps, 'style'>, ContainerModifier
...
-  }
+  }, []);
 
-  // TODO: do we need this?
-  setNativeProps(nativeProps: any) {
-    //@ts-ignore
-    this._root.setNativeProps(nativeProps); // eslint-disable-line
-  }
+  const ViewContainer = useMemo(() => {
+    const container = useSafeArea && Constants.isIOS ? SafeAreaView : RNView;
```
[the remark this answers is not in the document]
ethanshar: I need the container at the beginning cause createAnimatedComponent wraps it

src/components/view/index.tsx
ethanshar: Done

src/components/view/index.tsx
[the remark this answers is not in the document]
M-i-k-e-l: Oops 🤦‍♂️

src/components/view/index.tsx
M-i-k-e-l: It might have some performance implications, but they are probably minor

M-i-k-e-l (approved):

PR #2150, opened 2022-07-28
Fix/thumbnail resize android

adids1221:
## Description
Fixed image resize mode bug on android  , WOAUILIB-2865

## Changelog
Fixed image resize mode bug on android

### DIFF
--- a/src/components/image/index.tsx
+++ b/src/components/image/index.tsx
@@ -206,7 +206,6 @@ class Image extends PureComponent<Props, State> {
     const shouldFlipRTL = supportRTL && Constants.isRTL;
     const ImageView = this.shouldUseImageBackground() ? ImageBackground : RNImage;
     const {margins} = modifiers;
-    console.log(`useImageInsideContainer: ${useImageInsideContainer}`);
 
     return (
       // @ts-ignore
--- a/src/components/image/index.tsx
+++ b/src/components/image/index.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, {PureComponent} from 'react';
+import React, { PureComponent } from 'react';
 import hoistNonReactStatic from 'hoist-non-react-statics';
 import {
   StyleSheet,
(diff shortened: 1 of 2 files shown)


src/components/image/index.tsx
```diff
@@ -206,7 +206,7 @@ class Image extends PureComponent<Props, State> {
     const shouldFlipRTL = supportRTL && Constants.isRTL;
     const ImageView = this.shouldUseImageBackground() ? ImageBackground : RNImage;
     const {margins} = modifiers;
-    const resizeMode = useImageInsideContainer ? 'contain' : undefined;
+    // const resizeMode = useImageInsideContainer ? 'contain' : undefined;
```
ethanshar, line 209: Don't keep code in a comment (unless there's a good reason)

src/components/image/index.tsx
```diff
@@ -218,10 +218,11 @@ class Image extends PureComponent<Props, State> {
           this.isGif() && styles.gifImage,
           aspectRatio && {aspectRatio},
           !useImageInsideContainer && margins,
+          useImageInsideContainer && {resizeMode: 'contain'},
           style,
           useImageInsideContainer && styles.shrink
         ]}
-        resizeMode={resizeMode}
+        // resizeMode={resizeMode}
```
ethanshar, line 225: Also here, please remove commented code

src/components/image/index.tsx
```diff
@@ -218,10 +218,11 @@ class Image extends PureComponent<Props, State> {
           this.isGif() && styles.gifImage,
           aspectRatio && {aspectRatio},
           !useImageInsideContainer && margins,
+          useImageInsideContainer && {resizeMode: 'contain'},
```
ethanshar, line 221: When setting a static style that doesn't depends on a prop you can move it to the file's StyleSheet

ethanshar (changes requested):

src/components/image/index.tsx
adids1221: `
          useImageInsideContainer && {resizeMode: 'contain'},
          style,
          useImageInsideContainer && styles.shrink
`

In this case I'll add the resizeMode to styles.shrink since they depend on the same boolean right? 

like this:
`
  shrink: {
    flexShrink: 1,
    resizeMode: 'contain'
  }
`

src/components/image/index.tsx
```diff
@@ -206,7 +206,7 @@ class Image extends PureComponent<Props, State> {
     const shouldFlipRTL = supportRTL && Constants.isRTL;
     const ImageView = this.shouldUseImageBackground() ? ImageBackground : RNImage;
     const {margins} = modifiers;
-    // const resizeMode = useImageInsideContainer ? 'contain' : undefined;
+    console.log(`useImageInsideContainer: ${useImageInsideContainer}`);
```
ethanshar, line 209: Don't keep console.log in the code

ethanshar (changes requested):

src/components/image/index.tsx
adids1221: My mistake, fixed

ethanshar (approved):

PR #2152, opened 2022-07-28
Apply ThemeManager ThemeContext in asBaseComponent

ethanshar:
## Description
Apply ThemeManager ThemeContext in asBaseComponent

## Changelog
Apply ThemeManager ThemeContext in asBaseComponent

### DIFF
--- a/src/commons/asBaseComponent.tsx
+++ b/src/commons/asBaseComponent.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import hoistStatics from 'hoist-non-react-statics';
 import * as Modifiers from './modifiers';
-import {Scheme} from '../style';
+import {Scheme, ThemeManager} from '../style';
 import forwardRef from './forwardRef';
 import UIComponent from './UIComponent';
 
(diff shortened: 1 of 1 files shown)


src/commons/asBaseComponent.tsx
```diff
@@ -76,6 +76,10 @@ function asBaseComponent<PROPS, STATICS = {}>(WrappedComponent: React.ComponentT
   BaseComponent.displayName = WrappedComponent.displayName;
   BaseComponent.propTypes = WrappedComponent.propTypes;
   BaseComponent.defaultProps = WrappedComponent.defaultProps;
+  const ThemeContext = ThemeManager.getThemeContext();
+  if (ThemeContext) {
```
M-i-k-e-l, line 80: 1. Do we need the `if`?
2. Do we want to account for `WrappedComponent.contextType`?

ethanshar: It felt safer with if, also typescript didn't like when I just assigned it 

What do you mean be account for? You mean as a breaking change? 
I believe we are the only one using the contextType

M-i-k-e-l: I meant that maybe add something like:
```
else if (WrappedComponent.contextType) {
  BaseComponent.contextType = WrappedComponent.contextType;
```

Approving anyway

M-i-k-e-l (approved):

PR #2153, opened 2022-07-28
Infra/skeleton view

M-i-k-e-l:
## Description
SkeletonView - small changes that should improve performance
From what I've tested it did not seem like the `asBaseComponent` is needed, could you verify I did not miss anything there?
Relates to WOAUILIB-2699

## Changelog
SkeletonView - small changes that should improve performance

### DIFF
--- a/src/components/skeletonView/index.tsx
+++ b/src/components/skeletonView/index.tsx
@@ -4,14 +4,7 @@ import {StyleSheet, Animated, Easing, StyleProp, ViewStyle, AccessibilityProps} 
 import {BorderRadiuses, Colors, Dividers, Spacings} from '../../style';
 import {createShimmerPlaceholder, LinearGradientPackage} from 'optionalDeps';
 import View from '../view';
-import {
-  Constants,
-  asBaseComponent,
-  BaseComponentInjectedProps,
-  AlignmentModifiers,
-  PaddingModifiers,
-  MarginModifiers
-} from '../../commons/new';
+import {Constants, AlignmentModifiers, PaddingModifiers, MarginModifiers} from '../../commons/new';
 import {extractAccessibilityProps} from '../../commons/modifiers';
 
 const LinearGradient = LinearGradientPackage?.default;
@@ -175,9 +168,7 @@ interface SkeletonState {
  * @notes: View requires installing the 'react-native-shimmer-placeholder' and 'react-native-linear-gradient' library
  */
 
-type InternalSkeletonViewProps = SkeletonViewProps & BaseComponentInjectedProps;
-
-class SkeletonView extends Component<InternalSkeletonViewProps, SkeletonState> {
+class SkeletonView extends Component<SkeletonViewProps, SkeletonState> {
   static displayName = 'SkeletonView';
   static defaultProps = {
     size: Size.SMALL,
@@ -191,7 +182,7 @@ class SkeletonView extends Component<InternalSkeletonViewProps, SkeletonState> {
 
   fadeInAnimation?: Animated.CompositeAnimation;
 
-  constructor(props: InternalSkeletonViewProps) {
+  constructor(props: SkeletonViewProps) {
     super(props);
 
     this.state = {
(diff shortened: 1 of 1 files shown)


src/components/skeletonView/index.tsx
```diff
@@ -190,8 +180,18 @@ class SkeletonView extends Component<InternalSkeletonViewProps, SkeletonState> {
   static contentTypes: typeof ContentType = ContentType;
 
   fadeInAnimation?: Animated.CompositeAnimation;
+  contentAccessibilityProps?: any;
+  listItemAccessibilityProps?: any;
 
-  constructor(props: InternalSkeletonViewProps) {
+  getAccessibilityProps(accessibilityLabel: string, accessibilityProps: any) {
```
Effanuel, line 186: In #2154 PR, there is a performance fix by removing accessibility prop when its not needed. Is that not relevant here?

src/components/skeletonView/index.tsx
```diff
@@ -203,9 +203,16 @@ class SkeletonView extends Component<InternalSkeletonViewProps, SkeletonState> {
...
+    } else if (ShimmerPlaceholder === undefined) {
       ShimmerPlaceholder = createShimmerPlaceholder(LinearGradient);
     }
+
+    const accessibilityProps = extractAccessibilityProps(this.props);
+    if (props.template === Template.LIST_ITEM) {
+      this.listItemAccessibilityProps = this.getAccessibilityProps('Loading list item', accessibilityProps);
+    } else {
+      this.contentAccessibilityProps = this.getAccessibilityProps('Loading content', accessibilityProps);
+    }
```
Effanuel, line 215: Why not have it so that `getAccessibilityProps` method is responsible for all things related to accessibility props and move `extractAccessibilityProps` to there?

src/components/skeletonView/index.tsx
```diff
@@ -342,7 +341,7 @@ class SkeletonView extends Component<InternalSkeletonViewProps, SkeletonState> {
     const {style, ...others} = this.props;
 
     return (
-      <View style={[styles.listItem, style]} {...this.getAccessibilityProps('Loading list item')} {...others}>
+      <View style={[styles.listItem, style]} {...this.listItemAccessibilityProps} {...others}>
```
Effanuel, line 344: `useMemo` in `View` for _style will get recalculated every render because of new reference
`style={[styles.listItem, style]}`

src/components/skeletonView/index.tsx
```diff
@@ -190,8 +180,18 @@ class SkeletonView extends Component<InternalSkeletonViewProps, SkeletonState> {
   static contentTypes: typeof ContentType = ContentType;
 
   fadeInAnimation?: Animated.CompositeAnimation;
+  contentAccessibilityProps?: any;
+  listItemAccessibilityProps?: any;
```
Effanuel, line 184: Why are these `any`?

src/components/skeletonView/index.tsx
ethanshar: I agree, we should optimize how and when to add accessibility props. 
We'll do this in a dedicated PR

lidord-wix (approved):

src/components/skeletonView/index.tsx
Inbal-Tish: It's a simple check if `Constants.accessibility.isScreenReaderEnabled`

ethanshar: I know, but it's better to do it in a single place instead of having this check in all of our components. 
We should this once here -> https://github.com/wix/react-native-ui-lib/blob/faead14c8f9bc2c56753ac26af79fee4caf36ad7/src/commons/modifiers.ts#L264

Inbal-Tish: It should be in the `getAccessibilityProps`, not all components use `extractAccessibilityProps` when they pass the props...

M-i-k-e-l: Looking at the code again, it looks like we don't need `extractAccessibilityProps` here as well (`props` are destructured afterwards anyway).
In regards to adding `isScreenReaderEnabled` I seem to remember 'accessibilityLabel' was used in some tests\testing env, not sure it is still true.

ethanshar: You are right, I believe in some cases we use accessibility props in our tests 
In any case, let's handle this separately from this PR. 

Once we'll figure out a solution we'll create a new PR, I want to push forward this PR

PR #2154, opened 2022-07-28
Infra/ colorSwatch perf improvements 

lidord-wix:
## Description
Add `accessibilityInfo` only if the accessibility mode is on
improves performance by ~71%

## Changelog
ColorSwatch - add `accessibilityInfo` only if the accessibility mode is on

### DIFF
--- a/src/components/colorSwatch/index.tsx
+++ b/src/components/colorSwatch/index.tsx
@@ -137,6 +137,7 @@ class ColorSwatch extends PureComponent<Props> {
     const {isSelected} = this.state;
     const Container = onPress ? TouchableOpacity : View;
     const tintColor = this.getTintColor(color);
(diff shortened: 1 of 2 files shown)


src/components/colorSwatch/index.tsx
```diff
@@ -148,7 +149,7 @@ class ColorSwatch extends PureComponent<Props> {
         onPress={this.onPress}
         style={[this.styles.container, style]}
         onLayout={this.onLayout}
-        {...this.getAccessibilityInfo()}
+        {...accessibilityInfo}
```
Effanuel, line 152: Why is the previous approach slow?

lidord-wix: We added a condition to get the accessibility info only when the user is on accessibility mode since the calculation we do for getting the color name is pretty heavy.

Inbal-Tish (approved):

PR #2157, opened 2022-08-01
Infra/animated image improvements

M-i-k-e-l:
## Description
AnimatedImage:
1. Move from a js class to a  tsx function
2. Revert functionality to pre #2079 while keeping the "fix" (using reanimated `entering`).
3. I believe this also fixes a bug in the pre #2079 which did not show the `loader`.
4. This improves load time from what I've tested (relates to `Avatar` - WOAUILIB-2913).

Note: this unfortunately may remove the history from this file, an option is to change to a tsx (while breaking master's `npm run build:dev`) and only then merging this.

## Changelog
AnimatedImage - fix functionality and improve performance


src/components/animatedImage/index.tsx
```diff
@@ -0,0 +1,74 @@
+// TODO: consider unify this component functionality with our Image component
+import React, {useState, useCallback} from 'react';
+import {StyleSheet, StyleProp, ViewStyle, NativeSyntheticEvent, ImageLoadEventData} from 'react-native';
+import Animated, {FadeIn} from 'react-native-reanimated';
+import View from '../../components/view';
+import Image, {ImageProps} from '../../components/image';
+
+const UIAnimatedImage = Animated.createAnimatedComponent<ImageProps>(Image);
```
Effanuel, line 8: Is calling this method expensive?

Effanuel: Anyone importing a type from this file, and not using `import type` will call this method

lidord-wix (approved):
Looks good IMO
I have to say that I don't really feel bad about losing the history here, since it's a small and not that popular component.. but maybe ask Ethan

src/components/animatedImage/index.tsx
M-i-k-e-l: I do not know, but this is a semi-common usage (from what I've seen in another lib this was a HOC).
I think this is only called once for the whole app, am I wrong? So unless you mean that a public user might accidentally do that, it does not matter.

PR #2164, opened 2022-08-03
onPressIn and onPressOut with customValue fix

adids1221:
## Description
TouchableOpacity support passing customValue to onPressIn and onPressOut functions.

## Changelog
TouchableOpacity support passing customValue to onPressIn and onPressOut functions.


src/components/touchableOpacity/index.tsx
```diff
@@ -79,14 +79,22 @@ class TouchableOpacity extends PureComponent<Props, {active: boolean}> {
 
   onPressIn = (...args: any) => {
     this.setState({active: true});
-    //@ts-expect-error
-    this.props.onPressIn?.(...args);
+    if (this.props?.customValue) {
```
ethanshar, line 82: Consider using a ternary for this condition it will make the code a bit short and cleaner (:

src/components/touchableOpacity/index.tsx
```diff
@@ -79,14 +79,22 @@ class TouchableOpacity extends PureComponent<Props, {active: boolean}> {
...
+      //@ts-expect-error
+      this.props.onPressIn?.(...args);
+    }
   };
 
   onPressOut = (...args: any) => {
     this.setState({active: false});
-    //@ts-expect-error
-    this.props.onPressOut?.(...args);
+    if (this.props?.customValue) {
```
ethanshar, line 92: same here (:

adids1221: Last week we agreed to merge without changing.

ethanshar: We still need to pass the customValue inside on an object. You can pass all props, see what we do in onPress

ethanshar (changes requested):

src/components/touchableOpacity/index.tsx
```diff
@@ -41,7 +47,9 @@ export interface TouchableOpacityProps extends Omit<RNTouchableOpacityProps, 'st
    */
   customValue?: any;
   style?: ViewProps['style'];
-  onPress?: (props?: TouchableOpacityProps | any) => void;
+  onPress?: (props?: TouchableOpacityProps) => void;
+  onPressIn?: (props?: TouchableOpacityProps | ((event: GestureResponderEvent) => void)) => void;
+  onPressOut?: (props?: TouchableOpacityProps | ((event: GestureResponderEvent) => void)) => void;
```
ethanshar, line 52: I think the typings here are in correct 
Did it work for you?

ethanshar (commented):
Notice that the PR build fails

ethanshar (approved):

PR #2165, opened 2022-08-04
Fix Stepper theme

adids1221:
## Description
Stepper support for useCustomTheme added.

## Changelog
Stepper support for useCustomTheme.

### DIFF
--- a/src/components/stepper/index.tsx
+++ b/src/components/stepper/index.tsx
@@ -173,11 +173,10 @@ class Stepper extends PureComponent<Props, State> {
   }
 
   renderButton(actionType: ActionType) {
-    const {disabled, small, testID} = this.props;
+    const {disabled, small, testID, useCustomTheme} = this.props;
     const allowStepChange = this.allowStepChange(actionType);
     const minusButton = small ? minusOutlineSmall : minusOutline;
     const plusButton = small ? plusOutlineSmall : plusOutline;
-    const useCustomTheme = this.props?.useCustomTheme;
 
     return (
       <Button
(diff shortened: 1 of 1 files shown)


src/components/stepper/index.tsx
```diff
@@ -173,6 +177,7 @@ class Stepper extends PureComponent<Props, State> {
     const allowStepChange = this.allowStepChange(actionType);
     const minusButton = small ? minusOutlineSmall : minusOutline;
     const plusButton = small ? plusOutlineSmall : plusOutline;
+    const useCustomTheme = !!this.props?.useCustomTheme;
```
ethanshar, line 180: Why do you need to cast this value?

src/components/stepper/index.tsx
```diff
@@ -173,6 +177,7 @@ class Stepper extends PureComponent<Props, State> {
     const allowStepChange = this.allowStepChange(actionType);
     const minusButton = small ? minusOutlineSmall : minusOutline;
     const plusButton = small ? plusOutlineSmall : plusOutline;
+    const useCustomTheme = this.props?.useCustomTheme;
```
ethanshar, line 180: Why do you need keep a const for useCustomTheme?

adids1221: Instead should I pass the the useCsutomValue prop in the Button component?
`
<Button . . . useCustomValue = {this.props?.useCustomTheme} /> 
`
The const is just to have the value from the prop.
Is that a bad practice?

ethanshar (approved):

PR #2173, opened 2022-08-08
Split the renderRibbon to 2 render functions changed the logic

adids1221:
## Description
Fixed customRibbon props render in Avatar component

## Changelog
Fixed customRibbon props render in Avatar component

### DIFF
--- a/src/components/avatar/index.tsx
+++ b/src/components/avatar/index.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, {PropsWithChildren, PureComponent, ReactElement} from 'react';
+import React, {PropsWithChildren, PureComponent} from 'react';
 import {
   StyleSheet,
   ImageSourcePropType,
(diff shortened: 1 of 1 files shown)


src/components/avatar/index.tsx
```diff
@@ -384,7 +389,8 @@ class Avatar extends PureComponent<AvatarProps> {
         </View>
         {this.renderImage()}
         {this.renderBadge()}
-        {this.renderRibbon()}
+        {hasCustomRibbon && this.renderCustomRibbon()}
+        {!hasCustomRibbon && hasRibbonLabel && this.renderRibbon()}
```
ethanshar, line 393: I'd remove here only the condition of `!hasCustomRibbon` 
Basically it means it will move the responsibility to the user. 
If the user pass both ribbonLabel and customRibbon they will have a UI bug and it should be their responsibility to handle it.

ethanshar (commented):
Looks good! 
wrote a small comment

ethanshar (approved):

PR #2178, opened 2022-08-09
[Performance] Defer some calculations only when they are needed for TextField

Effanuel:
## Description
Improving performance of TextField, by optimizing ValidationMessage

## Changelog
* Defer calculation of validation message only when it's needed

### DIFF
--- a/src/incubator/TextField/CharCounter.tsx
+++ b/src/incubator/TextField/CharCounter.tsx
@@ -1,4 +1,4 @@
-import React, {useContext} from 'react';
+import React, {useContext, useMemo} from 'react';
 import {StyleSheet} from 'react-native';
 import _ from 'lodash';
 import Text from '../../components/text';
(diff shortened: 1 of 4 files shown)


src/incubator/TextField/CharCounter.tsx
```diff
@@ -7,12 +7,15 @@ import {CharCounterProps} from './types';
 
 const CharCounter = ({maxLength, charCounterStyle, testID}: CharCounterProps) => {
   const {value} = useContext(FieldContext);
+
+  const textStyle = useMemo(() => [styles.container, charCounterStyle], [charCounterStyle]);
```
ethanshar, line 11: This is a redundant memoization. when value changes, the char counter always changes and re-redner, so u might as well pass a new style. 

No reason to memoize stuff that don't save re-redners. memoize comes in a price of app memory

src/incubator/TextField/index.tsx
```diff
@@ -108,7 +107,7 @@ const TextField = (props: InternalTextFieldProps) => {
             validate={others.validate}
             validationMessage={others.validationMessage}
             validationMessageStyle={validationMessageStyle}
-            retainSpace={retainTopMessageSpace}
```
ethanshar, line 111: what's the purpose of this change?

ethanshar (changes requested):
@Effanuel 
See my comments, also please fix the PR description. 
In the description section add the PR description
In the changelog add the change for the release notes

src/incubator/TextField/index.tsx
Effanuel: To avoid calling `isEmpty` and `trim` on every render  if validation message is at the bottom (which is the default). Do you think this is too small of a change worth having?

PR #2182, opened 2022-08-10
Fix range slider initial position, allow thumb positions to be equal

ajp8164:
The range slider thumbs exhibited an arbitrary gap when they are moved next to each other. This gap should be zero so that the selected range is n..n (e.g. 10..10) when the thumbs are coincident. The arbitrary gap calculation imposed by the prior implementation results in different behavior based on step and range values.

This solution allows the range endpoints to be equal (no gap). The solution also ensures that the thumbs can always be accessed when thumb positions are the same (100% overlap). Fix initial position of minimum range thumb. The prior implementation was using the user defined range value as a "pixel", "x" value which resulted in the min track having the incorrect stop (filled in to the wrong x position).

There could be a new feature added to maintain a "gap" feature. This is out of scope for this PR.

Can ignore the import re-ordering - done by my lint rules.

Changelog
Eliminate the minimum gap between range slider thumbs (minimum gap is now 0).
Fix the initial position of the minimum range thumb.

### DIFF
--- a/src/components/slider/index.tsx
+++ b/src/components/slider/index.tsx
@@ -334,7 +334,7 @@ export default class Slider extends PureComponent<SliderProps, State> {
     if (this.isDefaultThumbActive()) {
       if (this.thumb.current) {
         const {useRange} = this.props;
-        const {trackSize, thumbSize} = this.state;
+        const {trackSize} = this.state;
         const nonOverlappingTrackWidth = trackSize.width - this.initialThumbSize.width;
         const _x = this.shouldForceLTR ? trackSize.width - x : x; // adjust for RTL
         const left = trackSize.width === 0 ? _x : (_x * nonOverlappingTrackWidth) / trackSize.width; // do not render above prefix\suffix icon\text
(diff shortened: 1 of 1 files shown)


src/components/slider/index.tsx
```diff
@@ -161,7 +161,7 @@ export default class Slider extends PureComponent<SliderProps, State> {
   private lastMinValue = this.minInitialValue;
 
   private _thumbStyles: ThumbStyle = {};
-  private _minThumbStyles: ThumbStyle = {left: this.minInitialValue};
+  private _minThumbStyles: ThumbStyle = {left: 0};
```
Inbal-Tish, line 164: Nice. Thanks!

src/components/slider/index.tsx
```diff
@@ -678,7 +678,9 @@ export default class Slider extends PureComponent<SliderProps, State> {
       >
         {this.renderTrack()}
         <View style={styles.touchArea} onTouchEnd={this.handleTrackPress}/>
-        {useRange && this.renderMinThumb()}
+        <View style={[!this.isDefaultThumbActive() ? {zIndex: 1} : {zIndex: 0}, {top: '-50%'}]}>
```
Inbal-Tish, line 681:
```suggestion
        <View style={{zIndex: !this.isDefaultThumbActive() ? 1 : 0, top: '-50%'}}>
```

Inbal-Tish:
@ajp8164 Hi. Please see added features in PR #2248.

PR #2183, opened 2022-08-10
Fixed the ChipsInput field trimmed bug

adids1221:
## Description
ChipsInput field get trimmed when inserting more than 1 line of chips.

## Changelog
ChipsInput field get trimmed when inserting more than 1 line of chips.


adids1221:
Ethan please take a look at this PR, I removed the felxGrow prop from the Input component that fix the bug.
https://github.com/wix/react-native-ui-lib/pull/1734

src/incubator/TextField/Input.tsx
```diff
@@ -45,7 +44,7 @@ const Input = ({
 
 const styles = StyleSheet.create({
   input: {
-    flexGrow: 1,
```
ethanshar, line 48: I checked why we added this flexGrow. 
We did it in this PR
https://github.com/wix/react-native-ui-lib/pull/1734

But unfortunately, I can't reproduce the bug it was meant to fix. 
We can merge the PR and worse case we can revert if the bug appear again. 

Another option is to pass a style that overrides the `flexGrow` from the ChipsInput code.

What do you prefer to do?

src/incubator/ChipsInput/index.tsx
```diff
@@ -142,6 +142,7 @@ const ChipsInput = forwardRef((props: ChipsInputProps, refToForward: React.Ref<a
       ref={fieldRef}
       leadingAccessory={chipList}
       blurOnSubmit={false}
+      style={{flexGrow: undefined}}
```
ethanshar, line 145: Don't pass inline styles
It creates a new object ref every render which triggers subsequent re-renders for inner components that depends on this prop

PR #2191, opened 2022-08-14
Avatar - to function (performance)

M-i-k-e-l:
## Description
Avatar - to function (performance)
I've added a snapshot test for the screen to verify I did not break anything, I figured we can keep it, WDYT?
I saw a significant performance improvement (15-20%) with my tests, please verify you see it too.
WOAUILIB-2913

## Changelog
Avatar - to function (performance)

### DIFF
--- a/src/components/avatar/index.tsx
+++ b/src/components/avatar/index.tsx
@@ -200,8 +200,7 @@ const Avatar = forwardRef<any, AvatarProps>((props: AvatarProps, ref: React.Forw
     if (imageSource) {
       LogService.warn('uilib: imageSource prop is deprecated, use source instead.');
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [imageSource]);
 
   const _source = useMemo(() => {
     return source || imageSource;
--- a/src/components/avatar/index.tsx
+++ b/src/components/avatar/index.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, {PropsWithChildren, useEffect, useMemo} from 'react';
+import React, {PropsWithChildren, useEffect, useMemo, forwardRef} from 'react';
 import {
   StyleSheet,
   ImageSourcePropType,
@@ -13,7 +13,6 @@ import {
 } from 'react-native';
 import {LogService} from '../../services';
 import {Colors, BorderRadiuses} from '../../style';
-import {ForwardRefInjectedProps} from '../../commons/new';
 import {extractAccessibilityProps} from '../../commons/modifiers';
 import Badge, {BadgeProps} from '../badge';
 import View from '../view';
@@ -22,7 +21,7 @@ import Image, {ImageProps} from '../image';
 // @ts-ignore
 import AnimatedImage from '../animatedImage';
 import * as AvatarHelper from '../../helpers/AvatarHelper';
-import {useThemeProps} from 'hooks';
+import {useThemeProps} from '../../hooks';
 
 export enum BadgePosition {
   TOP_RIGHT = 'TOP_RIGHT',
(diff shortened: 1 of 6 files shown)


src/components/avatar/index.tsx
```diff
@@ -155,176 +154,105 @@ export type AvatarProps = Pick<AccessibilityProps, 'accessibilityLabel'> &
...
+    name,
+    backgroundColor,
+    useAutoColors,
+    autoColorsConfig,
+    containerStyle,
+    onPress,
+    children
+  } = themeProps;
+  const {size: _badgeSize, borderWidth = 0} = badgeProps;
+  const badgeSize = _badgeSize || DEFAULT_BADGE_SIZE;
```
lidord-wix, line 197: Consider moving the default value to the destruction (more readable)

src/components/avatar/index.tsx
```diff
@@ -335,92 +263,136 @@ class Avatar extends PureComponent<AvatarProps> {
...
+      return (
+        <Badge testID={`${testID}.onlineBadge`} {...badgeProps} size={badgeSize} containerStyle={_badgePosition}/>
+      );
+    }
+  };
+
+  const ribbon = () => {
+    if (ribbonLabel) {
+      return (
+        <View style={[_ribbonStyle, styles.ribbon, ribbonStyle]}>
```
lidord-wix, line 330: you can memoize the style here as well :)

src/components/avatar/index.tsx
```diff
@@ -335,92 +263,136 @@ class Avatar extends PureComponent<AvatarProps> {
...
+        {!_.isUndefined(text) && (
+          <Text numberOfLines={1} style={textStyle} testID={`${testID}.label`}>
+            {text}
+          </Text>
+        )}
+      </View>
+      {image()}
+      {badge()}
+      {_customRibbon()}
+      {ribbon()}
```
lidord-wix, line 368: IMO, `renderImage`, `renderBadge`, `renderCustomRibbon`, and `renderRibbon` are better names when we talk about callable functions and not just memoized values.. WDYT?

lidord-wix (commented):
Looks good, 
I left few small comments, mostly about naming :)

src/components/avatar/index.tsx
M-i-k-e-l: It was there, this is actually a (not so readable :/) fix for a test, if the input is `0` the result should be `DEFAULT_BADGE_SIZE`, let me know if you have a better suggestion.

src/components/avatar/index.tsx
```diff
@@ -155,176 +154,105 @@ export type AvatarProps = Pick<AccessibilityProps, 'accessibilityLabel'> &
...
+  const {size: _badgeSize, borderWidth = 0} = badgeProps;
+  const badgeSize = _badgeSize || DEFAULT_BADGE_SIZE;
+
+  useEffect(() => {
+    if (imageSource) {
       LogService.warn('uilib: imageSource prop is deprecated, use source instead.');
     }
-  }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
M-i-k-e-l: I don't expect the `imageSource` to change, this is the equivalent for the usage in the `constructor`.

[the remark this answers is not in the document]
lidord-wix: But the user can pass `imageSource={undefined}` until he loads the data, and then pass `imageSource={someSource}` after the first mount.. we want to catch that and I don't think it's so rare

src/components/avatar/index.tsx
lidord-wix: I see, so let's keep it that way

lidord-wix (approved):

PR #2194, opened 2022-08-14
Fix range slider initial values

ajp8164:
## Description
Fixes range slider to allow initial min and max thumb positions. The prior implementation always sets the initial thumb positions to the track min and max values which is typically not the same as the values selected by the user. This change allows the range slider to be reconstructed from values set in a previous session.

## Changelog
Fix range slider to allow initial min and max thumb positions to be set on component mount.


ajp8164:
Bad set of commits

PR #2195, opened 2022-08-14
Fix range slider initial values

ajp8164:
## Description
Fixes range slider to allow initial min and max thumb positions. The prior implementation always sets the initial thumb positions to the track min and max values which is typically not the same as the values selected by the user. This change allows the range slider to be reconstructed from values set in a previous session.

## Changelog
Fix range slider to allow initial min and max thumb positions to be set on component mount.

### DIFF
--- a/src/components/slider/index.tsx
+++ b/src/components/slider/index.tsx
@@ -1,23 +1,24 @@
-import _ from 'lodash';
-import React, {PureComponent, ReactElement} from 'react';
 import {
-  StyleSheet,
-  PanResponder,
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@ajp8164 Thanks for the PR. I like the idea but seems not to work well, see my screenshot of a range slider when setting min and max initial values to 30 and 80 respectively: 
Screen Shot 2022-09-08 at 11 40 35

Inbal-Tish:
@ajp8164 Hi. Please see added features in PR #2248.

PR #2200, opened 2022-08-18
Optimize use of asBaseComponent in Picker and TextField

ethanshar:
## Description
Optimize use of asBaseComponent in Picker and TextField
Picker technically gets modifiers supports from TextField, so there's no need to use asBaseComponent there. 
I replaced it with `useThemeProps` hook to only add theme support

In TextField, I optimized the usage of asBaseComponent to get only relevant modifiers 
## Changelog
Optimize use of asBaseComponent in Picker and TextField


src/components/picker/index.tsx
```diff
@@ -50,7 +49,9 @@ const DIALOG_PROPS = {
   height: 250
 };
 
-const Picker = (props: PropsWithChildren<PickerProps> & ForwardRefInjectedProps & BaseComponentInjectedProps) => {
+const Picker = (props: PropsWithChildren<PickerProps> & ForwardRefInjectedProps /*  & BaseComponentInjectedProps */
+) => {
```
Inbal-Tish, line 53: No need to break line here... Why leave the commented BaseComponentInjectedProps? If we don't use it, let's delete it

Inbal-Tish (approved):

Inbal-Tish:
@ethanshar Approved. Please see a small comment and check the failing build

src/components/picker/index.tsx
ethanshar: Fixed

Inbal-Tish:
@ethanshar I enabled auto-merge, but worth taking a look there...

PR #2207, opened 2022-08-21
New floating button behavior

adids1221:
## Description
New behavior for the FloatingButton

## Changelog
New behavior for the FloatingButton


ethanshar (commented):
@adids1221 
I think there was a little mistake. 
You replaced the current implementation (top+bottom buttons) with the new one. 
The idea (I think) was to support the new layout (left+right)

src/components/floatingButton/index.tsx
```diff
@@ -123,7 +142,24 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
+      return (
+        <Button
+          outline
+          flex
+          size={Button.sizes.large}
+          testID={`${testID}.secondaryButton`}
+          {...secondaryButton}
+          style={[
+            styles.shadow,
+            {marginTop: Spacings.s4, marginBottom: Spacings.s7, marginLeft: 20, backgroundColor: bgColor}
```
ethanshar, line 157: move static styling to StyleSheet and apply only dynamic style as inline style

src/components/floatingButton/index.tsx
```diff
@@ -86,21 +96,32 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
-  }
+  };
 
   renderButton() {
-    const {bottomMargin, button, secondaryButton, testID} = this.props;
-    const bottom = secondaryButton ? Spacings.s4 : bottomMargin || Spacings.s8;
+    const {bottomMargin, button, secondaryButton, testID, buttonLayout} = this.props;
+
+    const secondaryHorizontal = secondaryButton && buttonLayout === FloatingButtonLayouts.HORIZONTAL;
+    const secondaryVertical = secondaryButton && buttonLayout === FloatingButtonLayouts.VERTICAL;
```
ethanshar, line 114: This is much better and way more readable. 
I would improve a little more. 
Since you use the same logic in multiple places (renderButton and renderSecondary), I think it'll be better to centralize the logic to a method. 

What we usually do is create a getter method that can be used throughout the component's code. 
It looks like this
```
get isSecondaryVertical() {
    return ...
  }
```

This way, in the  future, if the logic changes, you change it in one place.

PR #2209, opened 2022-08-21
Fixed demo app dark mode, changed to design tokens

adids1221:
## Description
Fixed demo app dark mode

## Changelog
Fixed demo app dark mode

### DIFF
--- a/demo/src/screens/MainScreen.js
+++ b/demo/src/screens/MainScreen.js
@@ -316,7 +316,7 @@ class MainScreen extends Component {
       return <SectionItem item={item} onPress={this.openScreen} onLongPress={this.setDefaultScreen}/>;
     } else {
       return (
-        <View paddingH-s5 marginV-s1 height={20} bg-grey80>
+        <View paddingH-s5 marginV-s1 height={20} bg-$backgroundNeutralLight>
           <Text text80M>{item.title}</Text>
         </View>
       );
@@ -387,7 +387,7 @@ class MainScreen extends Component {
 
         {showNoResults && (
           <View padding-20>
-            <Text grey40 text50>
+            <Text $textNeutralLight text50>
               Sorry, nothing was found. Try Button or something..
             </Text>
           </View>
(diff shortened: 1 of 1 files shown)


demo/src/screens/MainScreen.js
```diff
@@ -413,19 +413,19 @@ const SectionItem = React.memo(props => {
       activeOpacity={1}
       style={Dividers.d10}
     >
-      <Text style={[item.deprecate && styles.entryTextDeprecated]} grey10 text80>
+      <Text style={[item.deprecate && styles.entryTextDeprecated]} text80>
         {item.title}
       </Text>
-      <Icon source={chevronIcon} style={{tintColor: Colors.grey10}} supportRTL/>
+      <Icon source={chevronIcon} style={{tintColor: Colors.$iconDefault}} supportRTL/>
```
ethanshar, line 419: No need to set the default token, it's already set be default (:

demo/src/screens/MainScreen.js
```diff
@@ -413,19 +413,19 @@ const SectionItem = React.memo(props => {
...
     </TouchableOpacity>
   );
 });
 
 const SectionHeader = React.memo(props => {
   const {section} = props;
   return (
-    <View backgroundColor={'white'}>
-      <Text back marginV-20 marginH-20 text60M>
+    <View bg-$backgroundDefault>
```
ethanshar, line 427: Consider removing the background and keeping it transparent so it'll take the background from the view behind it.

adids1221: The reason I didn't removed the background is because if the background will be transparent you will see the Item beneath the header, the header will override the item title and you will see 2 text one override the other one.
It should stay with background color.

demo/src/screens/MainScreen.js
adids1221: Without the tintColor prop set to iconDefault token in dark mode I don't see the icon.

ethanshar: Ok, I understand why. 
It's because we set the default color in private and not public.

ethanshar (approved):

PR #2214, opened 2022-08-23
Fetch stars count from repo

ethanshar:
## Description
We set our stars count in our docs manually (through the docs config)
I changed it to fetch stars count from repo info

## Changelog
Docs - get stars count from github repo info

### DIFF
--- a/docuilib/src/components/MainSection.tsx
+++ b/docuilib/src/components/MainSection.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
 import Link from '@docusaurus/Link';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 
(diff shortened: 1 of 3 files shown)


docuilib/src/components/MainSection.tsx
```diff
@@ -1,14 +1,31 @@
...
+  }, []);
+
+  const fetchStarsCount = async () => {
+    const response = await fetch('https://api.github.com/repos/wix/react-native-ui-lib');
+    const data = await response.json();
+    const _starsCount = (data.stargazers_count / 1000).toFixed(1);
+
+
+    localStorage.setItem(STARS_COUNT_KEY, _starsCount.toString());
+    setStarsCount(_starsCount);
```
M-i-k-e-l, line 27: I'm assuming it's not critical for a browser, but should we wrap with try / catch?

M-i-k-e-l (approved):

docuilib/src/components/MainSection.tsx
ethanshar: Don't think it's necessary

PR #2215, opened 2022-08-23
Added listKey prop to flat list of wheel picker

Effanuel:
## Description
* Added listKey prop to WheelPicker flatlist, so that it can be rendered inside parent flatlist

## Changelog
* Added listKey prop to FlatList of `WheelPicker` component


ethanshar (approved):
I changed the key format a little, anyway approved

PR #2220, opened 2022-08-27
Make moment an optional dependency

frw:
## Description
This PR makes the moment package an optional dependency so as to decrease bundle size for those that do not use moment to format date/time.

Potentially breaking change: Projects that do not have moment as a direct dependency will find when upgrading that the supplied `dateFormat`/`timeFormat` would no longer have any effect on the displayed date/time, although there will be an error message logged to console to indicate that this is happening.

Tested by running the example app and ensured the dates/times are still rendered properly
Screenshot_20220827-171143_R N U I L I B

## Changelog
Make moment an optional dependency


src/components/dateTimePicker/index.tsx
```diff
@@ -173,17 +172,29 @@ function DateTimePicker(props: DateTimePickerPropsInternal) {
...
-              ? moment(value).format(dateFormat)
-              : value.toLocaleDateString();
+          if (dateFormatter) {
+            return dateFormatter(value);
+          }
+          if (!dateFormat) {
+            return value.toLocaleDateString();
+          }
+          if (!moment) {
+            console.error(`RNUILib DateTimePicker component with dateFormat requires installing "moment" dependency`);
```
M-i-k-e-l, line 182: I think adding this and the other `error` to a `useEffect` (either the existing one or a new one with `[dateFormat, timeFormat]` as a dependency array) will simplify the code and make it more readable.

M-i-k-e-l (commented):
Hi @frw,
Thanks a lot for the contribution!
I've added a small comment (please let me know if you do not have the capacity to fix that).

In any case we cannot merge this into master IMO since it's a breaking change, but we should definitely merge it for the next major release.

src/components/dateTimePicker/index.tsx
frw: Hi @M-i-k-e-l,

Thanks for reviewing. Definitely makes sense to have it in the next major release.
I've made the changes to the code as you've suggested. Let me know if there's anything else you'd like me to change.

M-i-k-e-l: Looks great!

M-i-k-e-l:
Look at unifying with #1805

dougg0k:
Why not replace with a better option like https://day.js.org/ or even https://date-fns.org/ if dayjs are no longer enough.

M-i-k-e-l:
@frw 
Thanks a lot for your contribution; apologies, but we will be merging this as part of our v7 ([link](https://github.com/wix/react-native-ui-lib/pull/2485)), so unfortunately you'll not get the due credit, hope it's ok.

@dougg0k, I think this component will be replaced so I am not sure we want to invest in this further.

PR #2229, opened 2022-08-31
Fix/ demo app colors

lidord-wix:
## Description
Fix demo app colors for dark mode

## Changelog
Fix demo app colors for dark mode

### DIFF
--- a/demo/src/demoApp.js
+++ b/demo/src/demoApp.js
@@ -42,39 +42,40 @@ function getDefaultNavigationStyle() {
     statusBar: {
       visible: true,
       style: 'light',
-      backgroundColor: Colors.primary // for Android
+      backgroundColor: Colors.$backgroundPrimaryHeavy // for Android
     },
     layout: {
-      backgroundColor: Colors.white,
-      orientation: ['portrait', 'landscape']
+      backgroundColor: Colors.$backgroundDefault,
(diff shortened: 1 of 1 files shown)


demo/src/demoApp.js
```diff
@@ -42,39 +42,40 @@ function getDefaultNavigationStyle() {
...
       style: 'light',
-      backgroundColor: Colors.primary // for Android
+      backgroundColor: Colors.$backgroundPrimaryHeavy // for Android
     },
     layout: {
-      backgroundColor: Colors.white,
-      orientation: ['portrait', 'landscape']
+      backgroundColor: Colors.$backgroundDefault,
+      orientation: ['portrait', 'landscape'],
+      componentBackgroundColor: Colors.$backgroundDefault
```
M-i-k-e-l, line 50: What's this for?

lidord-wix: run the demo app and toggle to dark mode (without changing the config file)

M-i-k-e-l: So you prevent the color change?

M-i-k-e-l (approved):

demo/src/demoApp.js
lidord-wix: I set the right colors instead the defaults we have today

PR #2231, opened 2022-09-01
Feat/marquee new

adids1221:
## Description
New Marquee component

## Changelog
New Marquee component

### DIFF
--- a/src/components/marquee/index.tsx
+++ b/src/components/marquee/index.tsx
@@ -1,0 +1,112 @@
+import React, {useCallback, useEffect, useState} from 'react';
+import {LayoutChangeEvent, LayoutRectangle, StyleSheet} from 'react-native';
+import {useSharedValue, useAnimatedStyle, withTiming, withRepeat, Easing} from 'react-native-reanimated';
+import View from '../view';
+import Text from '../text';
+import {MarqueeDirections, MarqueeProps} from './types';
+
+const DEFAULT_DURATION = 3000;
+const DEFAULT_DURATION_PER_WORD = 250;
+
+function Marquee(props: MarqueeProps) {
+  const {label, labelStyle, direction = MarqueeDirections.LEFT, duration, numberOfReps = -1, containerStyle} = props;
+
+  const calcDuration = () => {
+    const numOfWords = label.split(' ').length;
+    return DEFAULT_DURATION + DEFAULT_DURATION_PER_WORD * numOfWords;
+  };
+
+  const isHorizontal = direction === MarqueeDirections.LEFT || direction === MarqueeDirections.RIGHT;
+  const fixedDuration = duration || (isHorizontal ? calcDuration() : DEFAULT_DURATION);
+
+  const [viewLayout, setViewLayout] = useState<LayoutRectangle | undefined>(undefined);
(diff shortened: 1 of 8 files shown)


src/components/marquee/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      }
+    }
+  }, [viewLayout, textLayout]);
+
+  return (
+    <View style={[styles.container, containerStyle]} onLayout={onLayoutView}>
+      <View
+        reanimated
+        style={[
+          {position: 'absolute', width: !isHorizontal || textLayout?.width ? textLayout?.width : '400%'},
```
ethanshar, line 96: consider moving this part to useAnimatedStyle. 
It might not be animated, but it will unified the style into a single object.

src/components/marquee/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+          translateStyle
+        ]}
+      >
+        <Text style={[styles.text, labelStyle]} onLayout={onLayoutText}>
+          {label}
+        </Text>
+      </View>
+      <Text style={[styles.text, labelStyle, {color: 'transparent'}]} numberOfLines={1}>
```
ethanshar, line 104: move `color: transparent` style to stylesheet with a proper name

src/components/marquee/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+        {label}
+      </Text>
+    </View>
+  );
+}
+
+export {MarqueeProps, MarqueeDirections};
+
+export default asBaseComponent<MarqueeProps>(forwardRef(Marquee));
```
ethanshar, line 113: asBaseComponent add two important features: modifiers, theme props. 
As I see here you are not using modifier (yet) and not sure yet if we need to support theme props for such component. Better start without `asBaseComponent`, we can always add it after if we need to

ethanshar, line 113: No reason to wrap here with our forwardRef HOC, you don't use this as well in the component so it's not necessary. 
If you are not clear on how forwardRef works and when to use it, let's go over it

ethanshar (commented):
@adids1221 
The code looks great, very clean and simple. Anyway, I wrote a few comments, please go over them.

Few couple of things
- You need to add an api.json file to the component so it appears in our docs
- I noticed some behavior in the component. The last iteration, when the text goes back to its natural state, the animation is way too fast.

src/components/marquee/index.tsx
adids1221: do temporaryText is a good name for this style?

src/components/marquee/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [viewLayout, setViewLayout] = useState<LayoutRectangle | undefined>(undefined);
+  const [textLayout, setTextLayout] = useState<LayoutRectangle | undefined>(undefined);
+
+  const offset = useSharedValue<number | undefined>(undefined);
+
+  let initialOffset = 0;
+  let axisX = false;
+  let axisY = false;
+
+  if (direction === MarqueeDirections.RIGHT || direction === MarqueeDirections.LEFT) {
```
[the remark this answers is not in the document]
ethanshar: You are checking this twice. 
You can use the `isHorizontal` const

ethanshar (approved):

PR #2233, opened 2022-09-01
WIP - web demo app

OrenZak:
## Description
Create basic web support, and a demo app to showcase the support of current state.

## Changelog
- Added new package -> `webDemo`
- using a snapshot version of this branch, cuz i've changed few files to be able to render on web.


ethanshar (commented):
@OrenZak 
Can you explain a little, why do we need all of the files under `src/blur`? It appears you created a web alternative for the blur-view library, is that right? 

Another question,  what's the purpose of all the components under src/webComs?

OrenZak:
> @OrenZak Can you explain a little, why do we need all of the files under `src/blur`? It appears you created a web alternative for the blur-view library, is that right?
> 
> Another question, what's the purpose of all the components under src/webComs?

1. we need that because we they are missing web support, and its a dependency you have.  so i created this folder as a polyfill to the blur lib.

2. webComps, are just demo components i used, so they are using the components with some data

OrenZak:
> > > @OrenZak Can you explain a little, why do we need all of the files under `src/blur`? It appears you created a web alternative for the blur-view library, is that right?
> > > Another question, what's the purpose of all the components under src/webComs?
> > 
> > 
> > 
> > 1. we need that because they are missing web support, and its a dependency you have.  so i created this folder as a polyfill to the blur lib.
> > 2. webComps, are just demo components i used, so they are using the components with some data
> 
> As for 1, we use the "blur" package optionally. Only in specific cases, is there a real reason to polyfill it, can't we ignore the relevant feature while on Web? For example [here](https://github.com/wix/react-native-ui-lib/blob/a5be6ab775d3ee66d21a3491e3b33cbc56286979/src/components/card/index.tsx#L124), you can see we only use the blur package only when the enableBlur is passed. IMO, we can just ignore such props on Web, WDYT?

i think that we will need to support blur for web.
we can ignore for now i guess.

ethanshar:
> > > > @OrenZak Can you explain a little, why do we need all of the files under `src/blur`? It appears you created a web alternative for the blur-view library, is that right?
> > > > Another question, what's the purpose of all the components under src/webComs?
> > > 
> > > 
> > > 
> > > 1. we need that because they are missing web support, and its a dependency you have.  so i created this folder as a polyfill to the blur lib.
> > > 2. webComps, are just demo components i used, so they are using the components with some data
> > 
> > 
> > As for 1, we use the "blur" package optionally. Only in specific cases, is there a real reason to polyfill it, can't we ignore the relevant feature while on Web? For example [here](https://github.com/wix/react-native-ui-lib/blob/a5be6ab775d3ee66d21a3491e3b33cbc56286979/src/components/card/index.tsx#L124), you can see we only use the blur package only when the enableBlur is passed. IMO, we can just ignore such props on Web, WDYT?
> 
> i think that we will need to support blur for web. we can ignore for now i guess.

Let's ignore it. I don't want to maintain a web implementation for another library

ethanshar:
> > i think that we will need to support blur for web. we can ignore for now i guess.
> 
> Let's ignore it and remove its code. I don't want to maintain a web implementation for another library

@OrenZak Are you planning to handle this one?

OrenZak:
> > > i think that we will need to support blur for web. we can ignore for now i guess.
> > 
> > 
> > Let's ignore it and remove its code. I don't want to maintain a web implementation for another library
> 
> @OrenZak Are you planning to handle this one?

yes, will do.

ethanshar (approved):

PR #2236, opened 2022-09-04
Fix selected card iconColor

lidord-wix:
## Description
Fix selected card iconColor (set tintColor to ignore the default coloring of the private Icon config)
to test it you should run private on the `darkMode` branch
WOAUILIB-3064

## Changelog
Fix selected card iconColor

### DIFF
--- a/src/components/card/index.tsx
+++ b/src/components/card/index.tsx
@@ -231,8 +231,8 @@ class Card extends PureComponent<PropTypes, State> {
         {!selectionOptions.hideIndicator && (
           <View style={[this.styles.selectedIndicator, {backgroundColor: selectionColor}]}>
             <Icon
-              style={this.styles.selectedIcon}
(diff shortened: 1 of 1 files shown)


src/components/card/index.tsx
```diff
@@ -231,8 +231,8 @@ class Card extends PureComponent<PropTypes, State> {
         {!selectionOptions.hideIndicator && (
           <View style={[this.styles.selectedIndicator, {backgroundColor: selectionColor}]}>
             <Icon
-              style={this.styles.selectedIcon}
               source={_.get(selectionOptions, 'icon', DEFAULT_SELECTION_PROPS.icon)}
+              tintColor={_.get(selectionOptions, 'iconColor', DEFAULT_SELECTION_PROPS.iconColor)}
```
ethanshar, line 235: While we're here, maybe worth removing `_.get` usage and move to optional chaining

lidord-wix: Done :)

ethanshar (approved):

PR #2244, opened 2022-09-07
docsite - add "note" entry

Inbal-Tish:
## Description
docsite - add "note" to components and props api.

## Changelog
docsite - add "note" to components and props api.


Inbal-Tish:
@ethanshar The site is not responsive when adding the <h4> (marked with ####) to the script.

ethanshar:
@Inbal-Tish 
The error is coming from FloatingButton doc file. 
You can see the error when you run `npm start` in docuilib folder. 
Screen Shot 2022-09-12 at 13 47 46

The reason it started now is because you added support to show the `note`. 
The error is because of the format of how we documented the `<testID>` prop. We need to avoid `<` and `>` chars. 

See how I fixed it in Incubator.Dialog api.json file

ethanshar (approved):

ethanshar:
@Inbal-Tish Approved, looks good 👌 
We'll need to update the private script as well

PR #2245, opened 2022-09-07
Add build:local script

M-i-k-e-l:
## Description
Add a script to build changes locally.
Usage:
1. `npm run build:local`
2. Extract `react-native-ui-lib-5.0.0.tgz` (preferably in another location).
3. Replace `react-native-ui-lib` in your `node_modules` (probably backup or delete your current `node_modules/react-native-ui-lib`).
4. Delete `react-native-ui-lib-5.0.0.tgz`.
5. `package.json` will change, take extra care if you have changes there.

Note: I've added a warning if the user has untracked files (which will be deleted without this).

CC @Inbal-Tish @lidord-wix @adids1221 

## Changelog
None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 
 # Do not track local package
 react-native-ui-lib-5.0.0.tgz
+react-native-ui-lib.tgz
 
 # Xcode
 #
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 #
 .DS_Store
 
+# Do not track local package
+react-native-ui-lib-5.0.0.tgz
+
 # Xcode
 #
 build/
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,6 @@
 #
 .DS_Store
 
-# Do not track local package
-react-native-ui-lib-5.0.0.tgz
-react-native-ui-lib.tgz
-
 # Xcode
 #
 build/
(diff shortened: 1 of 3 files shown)


ethanshar:
@M-i-k-e-l 
Reminder: check for changes in package.json. And revert the change we do as part of the build

M-i-k-e-l:
> check for changes in package.json. And revert the change we do as part of the build

Done.
Sadly it is impossible to change the folder name (`package`) without expanding it, so I chose not to do it.

.gitignore
```diff
@@ -2,6 +2,10 @@
 #
 .DS_Store
 
+# Do not track local package
+react-native-ui-lib-5.0.0.tgz
```
ethanshar, line 6: Will it always be 5.0.0?

scripts/createLocalPackage.sh
```diff
@@ -0,0 +1,39 @@
...
+rm -f react-native-ui-lib.tgz
+# Build javascript files
+npm run build
+# Package into a tar
+npm pack
+# Clean ALL untracked files
+git clean -f
+git checkout package.json
+# Rename tar
+mv react-native-ui-lib-5.0.0.tgz react-native-ui-lib.tgz
```
ethanshar, line 39: Maybe mv to the user Downloads folder?

ethanshar (approved):

ethanshar:
Wrote some comment, anyway approved. 
nice work (:

.gitignore
M-i-k-e-l: I believe it comes from our `package.json`'s `version`, so it should not change since we do not really update it.

scripts/createLocalPackage.sh
M-i-k-e-l: Good idea, this should probably remove the need for the ignore, and solve the above comment.

PR #2260, opened 2022-09-20
Upgrate RN types and fix issues

ethanshar:
## Description
Update @types/react-native to 0.68.5 
Also set @types/react to v17 to avoid "JSX Children" errors coming from v18
Fix relevant issues

## Changelog
Update @types/react-native to 0.68.5 in demo environment


src/components/wizard/types.tsx
```diff
@@ -46,7 +46,7 @@ export interface WizardStepProps {
   /**
    * Icon to replace the (default) index
    */
-  icon?: ImageProps;
+  icon?: ImageProps['source'];
```
Inbal-Tish, line 49: Isn't this a breaking change? Can it be that someone passed other ImageProps that won't apply now? I don't see where we use this prop in the Wizard component so I can't really tell...

ethanshar: It's actually a fix as far as I see it. The weird thing is that it didn't threw an error before. 
We use it in WizardStep, [here](https://github.com/wix/react-native-ui-lib/blob/cd69d7cf5ab6dced4f92370ee3ce21680f98255c/src/components/wizard/WizardStep.tsx#L85)

We pass the icon to the image source prop so its type has to be of `ImageProps['source'];`

Inbal-Tish (approved):

PR #2265, opened 2022-09-28
[Guild Week] Added generator for basic design tokens

Effanuel:
## Description
* My current guild week task is to [create a util for generating design tokens](https://jira.wixpress.com/browse/WOAUILIB-2550). Added a util method to `Colors` class, since that method depends mostly on other `Colors` class method.

## Changelog
* Added `generateTokens` method for generating design tokens based on chosen color.

### DIFF
--- a/src/style/__tests__/colors.spec.js
+++ b/src/style/__tests__/colors.spec.js
@@ -124,11 +124,11 @@ describe('style/Colors', () => {
     });
   });
 
-  describe('generateTokens(...)', () => {
+  describe('generateDesignTokens(...)', () => {
     it('should generate design tokens from dark color for light theme', () => {
       const chosenColor = '#860D86';
       expect(uut.isDark(chosenColor)).toEqual(true);
-      expect(uut.generateTokens(chosenColor, 'light')).toEqual({
+      expect(uut.generateDesignTokens(chosenColor, 'light')).toEqual({
         $backgroundPrimaryHeavy: chosenColor,
         $backgroundPrimaryLight: '#FFFAFF',
         $backgroundPrimaryMedium: '#FACCFA',
@@ -142,7 +142,7 @@ describe('style/Colors', () => {
     it('should generate design tokens from light color for light theme', () => {
       const chosenColor = '#E9BEE7';
       expect(uut.isDark(chosenColor)).toEqual(false);
-      expect(uut.generateTokens(chosenColor, 'light')).toEqual({
+      expect(uut.generateDesignTokens(chosenColor, 'light')).toEqual({
         $backgroundPrimaryHeavy: '#A4379F',
         $backgroundPrimaryLight: '#F6E4F5',
         $backgroundPrimaryMedium: '#E9BEE7',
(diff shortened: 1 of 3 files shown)


src/style/__tests__/colors.spec.js
```diff
@@ -123,4 +123,62 @@ describe('style/Colors', () => {
...
+      expect(uut.isDark(chosenColor)).toEqual(true);
+      expect(uut.generateTokens(chosenColor, 'light')).toEqual({
+        $backgroundPrimaryHeavy: chosenColor,
+        $backgroundPrimaryLight: '#FFFAFF',
+        $backgroundPrimaryMedium: '#FACCFA',
+        $iconPrimary: chosenColor,
+        $iconPrimaryLight: '#F16FF1',
+        $outlinePrimary: chosenColor,
+        $textPrimary: chosenColor
+      });
```
Effanuel, line 139: These returned colors don't really match the ones in [figma](https://www.figma.com/file/Krv1pLl7kq2L52vMRwd498/Guidelines?node-id=11032%3A571101). I assume `generatePalette()` would need fixing here.

src/style/colors.ts
```diff
@@ -199,10 +199,58 @@ export class Colors {
...
+  generateColorPalette = _.memoize((color: string): string[] => {
+    const palette = this.generatePalette(color);
     return this.shouldSupportDarkMode && Scheme.getSchemeType() === 'dark' ? _.reverse(palette) : palette;
   });
 
+  generateTokens(color: string, mode: 'light' | 'dark') {
+    return mode === 'light' ? this.generateLightModeTokens(color) : this.generateDarkModeTokens(color);
+  }
+
+  private generateLightModeTokens(color: string) {
```
Effanuel, line 214: `generateLightModeTokens` and `generateDarkModeTokens` methods are very similar, but I would rather have them separate, then to couple them

src/style/__tests__/colors.spec.js
lidord-wix: Actually I think the fix should be in the figma board, I'll take it with the relevant UX.

lidord-wix (commented):
Looks very good! thank you for taking this task :) 

Consider adding another method to apply the new design tokens automatically (we can assume that most of the users that will use the `generateTokens()` method will load them right after using the `loadSchemes()` method, so we can offer a method that gets color and load the relevant light and dark design tokens).

lidord-wix:
> Looks very good! thank you for taking this task :)
> 
> Consider adding another method to apply the new design tokens automatically (we can assume that most of the users that will use the `generateTokens()` method will load them right after using the `loadSchemes()` method, so we can offer a method that gets color and load the relevant light and dark design tokens).

@Effanuel I think you missed this comment :)

src/style/colors.ts
```diff
@@ -43,6 +43,18 @@ export class Colors {
...
+   * Load light and dark schemes based on generated design tokens
+   * @param color - palette color
+   */
+  loadDesignTokens(color: string) {
+    Scheme.loadSchemes({
+      light: this.generateLightModeTokens(color),
+      dark: this.generateDarkModeTokens(color)
+    });
+    Object.assign(this, Scheme.getScheme());
+  }
```
lidord-wix, line 56:
```suggestion
    this.loadSchemes({
      light: this.generateLightModeTokens(color),
      dark: this.generateDarkModeTokens(color)
    });
  }
```

lidord-wix (approved):
Looks great!
see my last comment :)

PR #2275, opened 2022-10-03
Fixed textField placeholder color, support in design token

adids1221:
## Description
TextField presenter now supports design tokens.

## Changelog
TextField presenter now supports design tokens.


lidord-wix (approved):

ethanshar:
@adids1221 @lidord-wix 
I came across this PR and I suspect it might have an issue...
There's a conversion of color to a string, but it's inside an if statement that verifies the color is an object. so it's must be an object. why do we convert it to a string.. returning an object can lead to a bug

PR #2279, opened 2022-10-06
ExpandableSection support using separator when expanded

adids1221:
## Description
ExpandableSection support using a separator when expanded, feature for List.CollapsableItem

## Changelog
ExpandableSection support using a separator when expanded, feature for List.CollapsableItem

### DIFF
--- a/src/components/expandableSection/index.tsx
+++ b/src/components/expandableSection/index.tsx
@@ -63,7 +63,7 @@ function ExpandableSection(props: ExpandableSectionProps) {
   }, [expanded]);
 
   const renderSeparator = () => {
-    return <View style={styles.separator}/>;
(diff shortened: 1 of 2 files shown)


src/components/expandableSection/index.tsx
```diff
@@ -73,5 +83,9 @@ export default ExpandableSection;
 const styles = StyleSheet.create({
   container: {
     overflow: 'hidden'
+  },
+  separator: {
+    ...Dividers.d10,
+    marginLeft: Spacings.s10
```
lidord-wix, line 89: You can move the margin to a modifier and pass the dividers object without spreading, I think it's better :) 
```suggestion
  separator: Dividers.d10
```

lidord-wix (approved):
Approved, I just wrote a small comment :)

PR #2280, opened 2022-10-06
New getSystemColorByHex function in colors

adids1221:
## Description
New getSystemColorByHex function in colors

## Changelog
New getSystemColorByHex function in colors


lidord-wix (commented):
Looks Great! 
Two things:
1. Consider using lodash `_.findKey()` instead of looping all the colors for each search (I guess it's more optimized, but I never checked that).
2. Please add some tests of this new function for some trivial and edge cases.

src/style/colors.ts
```diff
@@ -131,6 +131,15 @@ export class Colors {
     return ColorName.name(color)[1];
   }
 
+  getSystemColorByHex(colorsOptions: string[], colorValue: string) {
+    const color = colorStringValue(colorValue);
+    const isColorFound = (c: string) => c.toString() === color;
+    const system_color = _.findKey(this, isColorFound);
```
lidord-wix, line 137: Now when you moved all the logic to the `return`, you can add the callback inside the `findKey` itself: 
```
const system_color = _.findKey(this, c => c.toString() === color);
```

src/style/colors.ts
```diff
@@ -131,6 +131,15 @@ export class Colors {
     return ColorName.name(color)[1];
   }
 
+  getSystemColorByHex(colorsOptions: string[], colorValue: string) {
```
lidord-wix, line 134: I think that the first argument of this function should be the `colorValue`, and the second argument can be optional (if the user doesn't pass the white list, all the colors are valid)
And also consider changing the `colorsOptions` name to something more meaningful, `validColors` for example.
WDYT?

lidord-wix (commented):
I added two comment, also I think you missed the comment about the tests :)

lidord-wix (commented):
Looks good! but I think you missed my comment about the tests once again :)

src/style/__tests__/colors.spec.js
```diff
@@ -123,4 +124,39 @@ describe('style/Colors', () => {
       expect(palette).toEqual(['#661A66', '#8F248F', '#B82EB7', '#D148D1', '#DB71DB', '#E699E6', '#F0C2F0', '#FFE5FF']);
     });
   });
+
+  describe('getSystemColorByHex', () => {
+    it('should return system color from valid colors in the array', () => {
```
lidord-wix, line 129:
```suggestion
    it('should return the system color, if the color is included in the validColors array', () => {
```

src/style/__tests__/colors.spec.js
```diff
@@ -123,4 +124,39 @@ describe('style/Colors', () => {
...
+      expect(uut.getSystemColorByHex('#FFFFFF', SYSTEM_COLORS)).toEqual('white');
+      expect(uut.getSystemColorByHex('#A6ACB1', SYSTEM_COLORS)).toEqual('grey40');
+      expect(uut.getSystemColorByHex('#20303C', SYSTEM_COLORS)).toEqual('grey10');
+      expect(uut.getSystemColorByHex('#000000', SYSTEM_COLORS)).toEqual('black');
+    });
+
+    it('should return undefined if color not include in validColors', () => {
```
lidord-wix, line 136:
```suggestion
    it('should return undefined if the color is not included in validColors', () => {
```

src/style/__tests__/colors.spec.js
```diff
@@ -123,4 +124,39 @@ describe('style/Colors', () => {
...
+      expect(uut.getSystemColorByHex('#00A87E', SYSTEM_COLORS)).toEqual(undefined);
+      expect(uut.getSystemColorByHex('#FFC50D', SYSTEM_COLORS)).toEqual(undefined);
+      expect(uut.getSystemColorByHex('#FB6413', SYSTEM_COLORS)).toEqual(undefined);
+    });
+
+    it('should return premium, fit, dark colors when validColors include premium, fit, dark', () => {
+      expect(uut.getSystemColorByHex('#116DFF', [...SYSTEM_COLORS, 'blue'])).toEqual('blue30');
+      expect(uut.getSystemColorByHex('#00A87E', [...SYSTEM_COLORS, 'green'])).toEqual('green30');
+      expect(uut.getSystemColorByHex('#FFC50D', [...SYSTEM_COLORS, 'yellow'])).toEqual('yellow30');
+      expect(uut.getSystemColorByHex('#FB6413', [...SYSTEM_COLORS, 'orange'])).toEqual('orange30');
```
lidord-wix, line 147: You can add these examples to the first `it` (it tests the same case)
also, you don't need all 4 tests if it's not testing something different :)

src/style/__tests__/colors.spec.js
```diff
@@ -123,4 +124,39 @@ describe('style/Colors', () => {
...
+    });
+
+    it('should return system color without validColors array', () => {
```
lidord-wix, line 150: I'd change the text a bit, you can mention here that without the array it searches the color in all the colors

src/style/__tests__/colors.spec.js
```diff
@@ -123,4 +124,39 @@ describe('style/Colors', () => {
...
+      expect(uut.getSystemColorByHex('#116DFF')).toEqual('blue30');
+      expect(uut.getSystemColorByHex('#00A87E')).toEqual('green30');
+      expect(uut.getSystemColorByHex('#FFC50D')).toEqual('yellow30');
+      expect(uut.getSystemColorByHex('#FB6413')).toEqual('orange30');
```
lidord-wix, line 154: Also here, there's no need in 4 tests if they're all the same :)

src/style/__tests__/colors.spec.js
```diff
@@ -123,4 +124,39 @@ describe('style/Colors', () => {
...
+    });
+
+    it('should return undefined for color that does not exist', () => {
+      expect(uut.getSystemColorByHex('#ff99cxac')).toEqual(undefined);
+      expect(uut.getSystemColorByHex('#zce333e1')).toEqual(undefined);
```
lidord-wix, line 159: Use a valid hex here (but something that we don't have in out palette),

lidord-wix (commented):
some last comments :)

lidord-wix (commented):
I actually wanted you to add a test case that checks that one hex brings one color with one validColors array and a different color with a different array, for example, I would expect those tests to pass:
```
      expect(uut.getSystemColorByHex('#5A48F5')).toEqual('violet30');
      expect(uut.getSystemColorByHex('#5A48F5', SYSTEM_COLORS)).toEqual(undefined);
      expect(uut.getSystemColorByHex('#5A48F5', ['primary'])).toEqual('primary');
```
but the last one doesn't pass, so I'm afraid that something in this function is still missing..
let me know if it's not clear :)

ethanshar:
@adids1221 @lidord-wix 
Can you explain the purpose of this new API? Where it's being used? I think the API is a little confusing

adids1221:
> @adids1221 @lidord-wix Can you explain the purpose of this new API? Where it's being used? I think the API is a little confusing

In the new Colors screen when the user press on any design token toast will pop with: the design token name, hex color, and system color name.
This API let us retrieve the system color based on the hex value, it looks for the key (the system color) based on the value as the Colors object holds the colors (name: hex).

ethanshar:
> This API let us retrieve the system color based on the hex value, it looks for the key (the system color) based on the value as the Colors object holds the colors (name: hex).

@adids1221 
Let's go over this f2f, I think the API can be a little confusing and once we export something people can use it - not sure if we want that

PR #2294, opened 2022-10-20
Set default value to Toast zIndex prop

adids1221:
## Description
Set default value to Toast zIndex prop.
This is the solution to the floating button that opens the Toast message, the Toast gets rendered beneath the button in Android. 
Ticket: 3100.

## Changelog
Set default value to Toast zIndex prop.

### DIFF
--- a/src/incubator/toast/index.tsx
+++ b/src/incubator/toast/index.tsx
@@ -1,5 +1,12 @@
 import React, {PropsWithChildren, useCallback, useEffect, useMemo, useRef, useState} from 'react';
-import {ActivityIndicator, StyleSheet, findNodeHandle, AccessibilityInfo, ViewStyle, LayoutChangeEvent} from 'react-native';
+import {
+  ActivityIndicator,
+  StyleSheet,
+  findNodeHandle,
+  AccessibilityInfo,
+  ViewStyle,
+  LayoutChangeEvent
+} from 'react-native';
 import _ from 'lodash';
 import {Constants, asBaseComponent} from '../../commons/new';
 import {useDidUpdate} from '../../hooks';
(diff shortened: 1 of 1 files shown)


src/incubator/toast/index.tsx
```diff
@@ -23,7 +30,7 @@ const Toast = (props: PropsWithChildren<ToastProps>) => {
     icon,
     iconColor,
     preset,
-    zIndex,
+    zIndex = Constants.isAndroid ? 100 : undefined,
```
Inbal-Tish, line 33: This change will be applied to all the Toast instances, but the issue is not relevant for all cases so why not address this more specifically or instruct the user to pass it in his usecase?

adids1221: I and Miki discussed about that, and the conclusion was that the toast behavior should be on Top of the elements (we don't have any case we want the toast beneath an element), so it is relevant in all of the cases.
Do you agree with that?

Inbal-Tish: The Toast component is currently displayed on top of other components, the issue you have relating to a Toast on top of a FloatingButton. And the issue is that the FloatingButton has a zIndex of 99 so you need the Toast to be set to 100 (try a smaller number and see that it doesn't work). So I don't see the need to set this value as a default to the Toast component but to instruct the user to pass zIndex = 100 in this use case (or treat it privately in our "toggleToast" method).

adids1221: After consulting with Ethan, I'll make PR in the private repo, and set the default zIndex to 100 in the config file.
We will set the default zIndex to 100 in the public repo in the next major version.

PR #2298, opened 2022-10-23
fix Gemfile syntax error and creating a Gemfile.lock

Inbal-Tish:
## Description
fix Gemfile syntax error and creating a Gemfile.lock

## Changelog
fix Gemfile syntax error and creating a Gemfile.lock


ethanshar:
Should we push the Gemfile.lock to git?

Inbal-Tish:
> Should we push the Gemfile.lock to git?

No, we shouldn't, but now I wonder why it's not in the .gitignore then...

ethanshar (approved):

PR #2302, opened 2022-10-25
Fix/ loadDesignTokens API

lidord-wix:
## Description
Fix Colors.loadDesignTokens API
@Effanuel FYI - we decided to change the API a bit to support adding features in the future easily  

## Changelog
Fix Colors.loadDesignTokens API

### DIFF
--- a/src/style/colors.ts
+++ b/src/style/colors.ts
@@ -49,10 +49,10 @@ export class Colors {
    * Load light and dark schemes based on generated design tokens
    * @param color - palette color
    */
-  loadDesignTokens(color: string) {
+  loadDesignTokens({primaryColor}: {primaryColor: string}) {
     this.loadSchemes({
-      light: this.generateLightModeTokens(color),
-      dark: this.generateDarkModeTokens(color)
+      light: this.generateDesignTokens({primaryColor}),
+      dark: this.generateDesignTokens({primaryColor, dark: true})
     });
   }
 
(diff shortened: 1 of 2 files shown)


src/style/colors.ts
```diff
@@ -49,10 +49,10 @@ export class Colors {
    * Load light and dark schemes based on generated design tokens
    * @param color - palette color
    */
-  loadDesignTokens(color: string) {
+  loadDesignTokens({primaryColor}: {primaryColor: string}) {
```
Inbal-Tish, line 52: Why not make the type a defined type, like `export type TokensOptions: {primaryColor: string}`, and have a parameter like `loadDesignTokens(options: TokensOptions)`? Then you can use `options?.primaryColor` etc.

src/style/colors.ts
```diff
@@ -220,38 +220,19 @@ export class Colors {
...
-      $iconPrimary: mainColor,
-      $iconPrimaryLight: color50,
-      $textPrimary: mainColor,
-      $outlinePrimary: mainColor
-    };
-  }
-
-  private generateDarkModeTokens(color: string) {
-    const colorPalette: string[] = _.reverse(this.generatePalette(color));
+  private generateDesignTokens({primaryColor, dark}: {primaryColor: string; dark?: boolean}) {
```
Inbal-Tish, line 223: Why the parameter here is an object? This method is private so no worries about changing its signature in the future. The parameters could be simply `primaryColor: string and dark: boolean`.

src/style/colors.ts
```diff
@@ -220,38 +220,19 @@ export class Colors {
...
+    const colorPalette: string[] = dark
+      ? _.reverse(this.generatePalette(primaryColor))
```
Inbal-Tish, line 225: You can generate the palette outside of the condition and reverse the array if it is for a dark palette, this way you save the code duplication

src/style/colors.ts
```diff
@@ -220,38 +220,19 @@ export class Colors {
...
+      : this.generatePalette(primaryColor);
     const color30 = colorPalette[2];
     const color50 = colorPalette[4];
     const color70 = colorPalette[6];
     const color80 = colorPalette[7];
 
-    const mainColor = this.isDark(color) ? color30 : color;
+    let mainColor = this.isDark(primaryColor) ? primaryColor : color30;
```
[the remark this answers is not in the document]
lidord-wix: I rewrote this condition and ended up with this one:
```
const mainColor = this.isDark(primaryColor) === !!dark ? color30 : primaryColor
```
because I want to use color30 only if the color is light in light mode or dark in dark mode, but I'm not sure if it makes the code more readable than that:

```
let mainColor = isPrimaryColorDark ? primaryColor : color30;
if (dark) {
  mainColor = isPrimaryColorDark ? color30 : primaryColor;
}
```
WDYT?

src/style/colors.ts
lidord-wix: Done

src/style/colors.ts
lidord-wix: Done

src/style/colors.ts
lidord-wix: I actually tried it and found it less readable that `generateDesignTokens()` will need to get the `primaryColor`, the `palette`, and the `dark` boolean as parameters. it gets all the logic out of the method. 
And I think it's not that bad to call it twice since it happens only once when the app starts.

Inbal-Tish: I agree that the palette should be generated in the method, but you can still do it once and only reverse the array if dark = true. For example:
`let colorPalette = this.generatePalette(primaryColor);
  if (dark) {
    colorPalette = _.reverse(colorPalette);
  }
`

src/style/colors.ts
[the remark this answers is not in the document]
Inbal-Tish: if `cont isPrimaryColorDark = this.isDark(primaryColor)` is the line before then it readable to do:
`let mainColor = isPrimaryColorDark ? primaryColor : color30;
if (dark) {
  mainColor = isPrimaryColorDark ? color30 : primaryColor;
}`

src/style/colors.ts
lidord-wix: Ohh right!
done

ethanshar:
@lidord-wix 
I think we should release this as a hotfix so people won't get a chance to use the previous API

lidord-wix:
> @lidord-wix I think we should release this as a hotfix so people won't get a chance to use the previous API

Done :) 
6.22.3

PR #2321, opened 2022-11-07
getTintColor support options, should reverse on dark prop

adids1221:
## Description
getTintColor support options object with should reverse on dark prop.

## Changelog
getTintColor support options object with should reverse on dark prop for the new Loader logic - Theme preset.

### DIFF
--- a/src/style/colors.ts
+++ b/src/style/colors.ts
@@ -12,7 +12,7 @@ import Scheme, {Schemes, SchemeType} from './scheme';
 
 export type DesignToken = {semantic?: [string]; resource_paths?: [string]; toString: Function};
 export type TokensOptions = {primaryColor: string};
-export type GetColorTintOptions = {shouldReverseOnDark?: boolean};
+export type GetColorTintOptions = {avoidReverseOnDark?: boolean};
 
 export class Colors {
   [key: string]: any;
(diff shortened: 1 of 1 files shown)


src/style/colors.ts
```diff
@@ -12,6 +12,7 @@ import Scheme, {Schemes, SchemeType} from './scheme';
 
 export type DesignToken = {semantic?: [string]; resource_paths?: [string]; toString: Function};
 export type TokensOptions = {primaryColor: string};
+export type ColorTintOptions = {shouldReverseOnDark: boolean};
```
ethanshar, line 15: I'd rename it to `GetColorTintOptions` to make it more accurate

ethanshar: Also, `shouldReverseOnDark` should probably be optional

src/style/colors.ts
```diff
@@ -162,7 +163,9 @@ export class Colors {
     return validColors ? undefined : results[0];
   }
 
-  getColorTint(colorValue: string | OpaqueColorValue, tintKey: string | number) {
+  getColorTint(colorValue: string | OpaqueColorValue,
+    tintKey: string | number,
+    options: ColorTintOptions = {shouldReverseOnDark: true}) {
```
ethanshar, line 168: Try to declare options in a way that works with your default logic
for example: instead of calling the option `shouldReverseOnDark` and set its default to true, maybe have an option called `avoidReverseOnDark`, this way its falsy default value will work with your logic. 

Why is it important? cause if you will add more options in the future, let's call it `anotherOption` and the user will only pass that option like this `getColorTint('#efefef', {anotherOption: true})` then your `shouldReverseOnDark` option will become undefined/false and the user won't understand why the default behavior has changed.  

sorry for the long explanation, let me know if it's not clear...

src/style/colors.ts
```diff
@@ -162,7 +163,8 @@ export class Colors {
     return validColors ? undefined : results[0];
   }
 
-  getColorTint(colorValue: string | OpaqueColorValue, tintKey: string | number) {
+  getColorTint(colorValue: string | OpaqueColorValue, tintKey: string | number, options?: GetColorTintOptions) {
+    const avoidReverseOnDark = !!options?.shouldReverseOnDark;
```
ethanshar, line 167: I think you misunderstood me here.. 
What I meant is that the option name will be avoidReverseOnDark, this way you don't need to do any special manipulation, and if the user passes nothing it will behave as we meant to

PR #2328, opened 2022-11-14
ColorPicker now support backgroundColor prop

adids1221:
## Description
ColorPicker now supports backgroundColor prop.

## Changelog
ColorPicker now supports backgroundColor prop.

### DIFF
--- a/src/components/colorPalette/index.tsx
+++ b/src/components/colorPalette/index.tsx
@@ -381,12 +381,7 @@ class ColorPalette extends PureComponent<Props, State> {
   }
 
   render() {
-    const {backgroundColor} = this.props;
-    return (
-      <View flex style={{backgroundColor}}>
-        {this.usePagination ? this.renderPaginationContent() : this.renderScrollableContent()}
-      </View>
-    );
+    return this.usePagination ? this.renderPaginationContent() : this.renderScrollableContent();
   }
 }
 
--- a/src/components/colorPalette/index.tsx
+++ b/src/components/colorPalette/index.tsx
@@ -53,6 +53,10 @@ interface Props {
   onValueChange?: (value: string, options: object) => void;
   style?: StyleProp<ViewStyle>;
   testID?: string;
+  /**
+   * Give the ColorPalette a background color
+   */
+  backgroundColor?: string;
 }
 export type ColorPaletteProps = Props;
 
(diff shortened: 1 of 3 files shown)


src/components/colorPicker/index.tsx
```diff
@@ -84,12 +87,12 @@ class ColorPicker extends PureComponent<Props> {
...
     this.setState({show: false});
-  }
+  };
 
   render() {
-    const {initialColor, colors, value, testID, accessibilityLabels} = this.props;
+    const {initialColor, colors, value, testID, accessibilityLabels, backgroundColor} = this.props;
     const {show} = this.state;
-
+    const bgColor = backgroundColor;
```
Inbal-Tish, line 95: Why do you need this extra const?

src/components/colorPicker/index.tsx
```diff
@@ -36,13 +35,17 @@ interface Props extends ColorPickerDialogProps {
...
+    dismissButton?: string;
+    doneButton?: string;
+    input?: string;
   };
   style?: StyleProp<ViewStyle>;
   testID?: string;
+  /**
+   * Give the ColorPicker a background color
+   */
+  backgroundColor?: string;
```
Inbal-Tish, line 48: We have a 'style' prop here that we're not passing anywhere, why do we need 'backgroundColor' prop and not use the style prop for that?

src/components/colorPalette/index.tsx
```diff
@@ -331,19 +336,19 @@ class ColorPalette extends PureComponent<Props, State> {
...
   renderScrollableContent() {
-    const {containerStyle, ...others} = this.props;
+    const {containerStyle, backgroundColor, ...others} = this.props;
     const {scrollable, contentWidth} = this.state;
 
     return (
       <ScrollBar
         ref={this.scrollBar}
-        style={[containerStyle, styles.scrollContainer]}
+        style={[containerStyle, styles.scrollContainer, {backgroundColor}]}
```
Inbal-Tish, line 345: Why not pass the backgroundColor in the containerStyle and just not override it? Both 'style' and 'gradientColor' can get the bg color from the same source `const bg = style?.backgroundColor || Colors.$backgroundDefault;`

src/components/colorPicker/index.tsx
adids1221: At first, I tried to use the style prop to pass the backgroundColor but I have ts error when trying to use the background prop.
```
const bg = style?.backgroundColor || Colors.$backgroundDefault;

    
Property 'backgroundColor' does not exist on type 'false | ViewStyle | RegisteredStyle<ViewStyle> | RecursiveArray<Falsy | ViewStyle | RegisteredStyle<ViewStyle>>'.
Property 'backgroundColor' does not exist on type 'false'.
```
I need to use this color for the button and pass it to the ColorPalette via style, in ColorPalette I have the same error since the containerStyle prop is the same type as the style prop in the colorPicker.

This is why I created the background prop.

src/components/colorPalette/index.tsx
```diff
@@ -375,7 +381,12 @@ class ColorPalette extends PureComponent<Props, State> {
   }
 
   render() {
-    return this.usePagination ? this.renderPaginationContent() : this.renderScrollableContent();
+    const {backgroundColor} = this.props;
+    return (
+      <View flex style={{backgroundColor}}>
```
Inbal-Tish, line 386: Why do we need this View wrapper?

src/components/colorPicker/index.tsx
```diff
@@ -100,8 +103,9 @@ class ColorPicker extends PureComponent<Props> {
           animatedIndex={this.animatedIndex}
           onValueChange={this.onValueChange}
           testID={`${testID}-palette`}
+          backgroundColor={backgroundColor}
         />
-        <View style={styles.buttonContainer}>
+        <View style={[styles.buttonContainer, {backgroundColor}]}>
```
Inbal-Tish, line 108: BackgroundColor should not be passed here but to the component View wrapper (line 97)

Inbal-Tish (approved):

PR #2329, opened 2022-11-14
SortableList SortableGridList - add initialData

M-i-k-e-l:
## Description
SortableList SortableGridList - add initialData
* `initialData` is taken over `data`
* Unfortunately I had to make `data` optional (not in the API), otherwise the typescript would be too complex.

WOAUILIB-3120

## Changelog
SortableList SortableGridList - add initialData

### DIFF
--- a/demo/src/screens/componentScreens/SortableGridListScreen.tsx
+++ b/demo/src/screens/componentScreens/SortableGridListScreen.tsx
@@ -100,7 +100,7 @@ class SortableGridListScreen extends Component {
         </View>
         <View flex>
           <SortableGridList
-            data={items}
+            initialData={items}
             renderItem={this.renderItem}
             // numColumns={2}
             maxItemWidth={140}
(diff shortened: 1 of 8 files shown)


ethanshar:
@M-i-k-e-l 
I'm actually having second thoughts on this change. 
Unlike props like `initialValue`, isn't `initialData` prop is weird as a main prop for the component, especially when it's based on FlatList `data` prop.

M-i-k-e-l:
> @M-i-k-e-l I'm actually having second thoughts on this change. Unlike props like `initialValue`, isn't `initialData` prop is weird as a main prop for the component, especially when it's based on FlatList `data` prop.

We can change the name to `initialValue` if you like it more (or try to think of something else), but I think that preventing bugs is more important here, and since it's based on `FlatList` (even more so in `SortableList`) we should make sure the user is aware it's not the `data` they are used to.

ethanshar:
I just wonder if it's confusing to the user. 
We want `initial...` prop so they'll know to not pass it on each change
On the other hand, if they do change the data by adding/removing items, they do need to pass it..

M-i-k-e-l:
As discussed, we're going to look at the docs and make sure they are clear

PR #2340, opened 2022-11-21
Fix marquee screen now have scroll view

adids1221:
## Description
Marquee screen now has a scroll view.

## Changelog
Marquee screen now has a scroll view.
The screen does not fit the private demo app.

### DIFF
--- a/demo/src/screens/componentScreens/MarqueeScreen.tsx
+++ b/demo/src/screens/componentScreens/MarqueeScreen.tsx
@@ -60,7 +60,7 @@ export default class MarqueeScreen extends Component<{}> {
 
   render() {
     return (
-      <ScrollView style={{flex: 1, padding: 20}}>
+      <ScrollView style={styles.container}>
         <Text h1 center margin-20 $textDefault>
           Marquee
         </Text>
@@ -89,6 +89,10 @@ export default class MarqueeScreen extends Component<{}> {
 }
 
 const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    padding: 20
+  },
   containerHorizontal: {
     borderWidth: 1,
     borderColor: 'black',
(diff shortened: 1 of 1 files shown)


demo/src/screens/componentScreens/MarqueeScreen.tsx
```diff
@@ -62,28 +61,32 @@ export default class MarqueeScreen extends Component<{}> {
...
-          {renderMultipleSegmentOptions.call(this, 'Number Of Reps', 'numOfReps', [
-            {label: 'Infinite', value: -1},
-            {label: '1', value: 1},
-            {label: '3', value: 3},
-            {label: '5', value: 5}
-          ])}
-          <View marginV-s2>
-            {renderBooleanOption.call(this, 'Direction Horizontal: Left To Right/Right To Left', 'directionHorizontal')}
-            {renderBooleanOption.call(this, 'Direction Vertical:  Bottom To Up/Up To Bottom', 'directionVertical')}
+        <ScrollView>
```
lidord-wix, line 64: Why not replace the wrapping View with a ScrollView?

adids1221: You're right I'll just pass the style to the ScrollView

demo/src/screens/componentScreens/MarqueeScreen.tsx
```diff
@@ -61,7 +60,7 @@ export default class MarqueeScreen extends Component<{}> {
 
   render() {
     return (
-      <View flex padding-page>
+      <ScrollView style={{flex: 1, padding: 20}}>
```
lidord-wix, line 63: It is more performant to use react-native's StyleSheet instead of inline style.

lidord-wix (approved):
Approved.
I left a small comment, it's not so important because it's in the screen, but many users copy our examples in their code so it's better to keep the best practices :)

PR #2344, opened 2022-11-24
Incubator.TextField - should float with defaultValue

M-i-k-e-l:
## Description
Incubator.TextField - FloatingPlaceholder should float with defaultValue
Adding `shouldFloat` and fixing its logic

WOAUILIB-3123

Handling edge cases, using the below `PlaygroundScreen`:
1. Switching between `TextField`s should:
    a. float\un-float if both are empty
    b. float\un-float on an empty one and remain un-floated on one with a `value` \ `defaultValue`
2. Changing `defaultValue` should affect the floating status
3. Deleting the `defaultValue` (to an empty `value`) should un-float

```
import {View, Text, Card, Incubator, Button} from 'react-native-ui-lib';

  state = {
    defaultValue: 'initial value'
  };
  render() {
    return (
      <View bg-grey80 flex padding-20>
        <View marginT-20>
          <Incubator.TextField
            migrate
            marginH-s5
            placeholder={'placeholder'}
            defaultValue={this.state.defaultValue}
            floatingPlaceholder
            floatOnFocus
          />
          <Incubator.TextField migrate marginH-s5 placeholder={'placeholder'} floatingPlaceholder floatOnFocus/>
        </View>
        <Card height={100} center padding-20>
        </Card>
        <View flex center>
          <Button
            marginV-20
            label="Button"
            onPress={() => {
              this.setState({defaultValue: undefined});
            }}
          />
        </View>
      </View>
    );
  }
}
```

## Changelog
Incubator.TextField - FloatingPlaceholder should float with defaultValue

### DIFF
--- a/src/incubator/TextField/FloatingPlaceholder.tsx
+++ b/src/incubator/TextField/FloatingPlaceholder.tsx
@@ -28,10 +28,8 @@ const FloatingPlaceholder = ({
     left: 0
   });
 
-  const shouldFloat = useMemo(() => {
-    return (floatOnFocus && context.isFocused) || context.hasValue || !isEmpty(defaultValue);
-  }, [floatOnFocus, context.isFocused, context.hasValue, defaultValue]);
-
+  const useDefaultValue = !isEmpty(defaultValue) && context.value === undefined; // To consider a user that has deleted the defaultValue (and then the placeholder should un-float)
+  const shouldFloat = (floatOnFocus && context.isFocused) || context.hasValue || useDefaultValue;
   const animation = useRef(new Animated.Value(Number(shouldFloat))).current;
   const hidePlaceholder = !context.isValid && validationMessagePosition === ValidationMessagePosition.TOP;
 
--- a/src/incubator/TextField/FloatingPlaceholder.tsx
+++ b/src/incubator/TextField/FloatingPlaceholder.tsx
@@ -1,9 +1,8 @@
-import {isEmpty} from 'lodash';
 import React, {useContext, useRef, useCallback, useState, useMemo} from 'react';
 import {Animated, LayoutChangeEvent, StyleSheet, Platform} from 'react-native';
 import {useDidUpdate} from 'hooks';
 import {FloatingPlaceholderProps, ValidationMessagePosition} from './types';
-import {getColorByState} from './Presenter';
+import {getColorByState, shouldPlaceholderFloat} from './Presenter';
 import {Colors} from '../../style';
 import {Constants} from '../../commons/new';
 import View from '../../components/view';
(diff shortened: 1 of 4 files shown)


src/incubator/TextField/FloatingPlaceholder.tsx
```diff
@@ -18,24 +19,29 @@ const FloatingPlaceholder = ({
...
   testID
 }: FloatingPlaceholderProps) => {
   const context = useContext(FieldContext);
   const [placeholderOffset, setPlaceholderOffset] = useState({
     top: 0,
     left: 0
   });
-  const animation = useRef(new Animated.Value(Number((floatOnFocus && context.isFocused) || context.hasValue))).current;
+
+  const shouldFloat = useMemo(() => {
```
ethanshar, line 31: There's no need to memoize primitive values (booleans, strings, etc...) 😬

M-i-k-e-l: Wouldn't that break the `useDidUpdate` dependencies? Unless I put the dependencies from here, which is a little weird.

M-i-k-e-l: I think you're right but I found a bug :/

src/incubator/TextField/types.ts
```diff
@@ -86,6 +86,7 @@ export interface FloatingPlaceholderProps {
   floatOnFocus?: boolean;
   validationMessagePosition?: ValidationMessagePosition;
   extraOffset?: number;
+  defaultValue?: TextInputProps['defaultValue'];
```
Inbal-Tish, line 89: 'TextInputProps' are the OLD TextField's prop... why use them in the incubator?

src/incubator/TextField/index.tsx
```diff
@@ -125,6 +125,7 @@ const TextField = (props: InternalTextFieldProps) => {
           <View flexG /* flex row */>
             {floatingPlaceholder && (
               <FloatingPlaceholder
+                defaultValue={others.defaultValue}
```
Inbal-Tish, line 128: Why not pass the input's 'value'? Why do we need a separate prop?

src/incubator/TextField/FloatingPlaceholder.tsx
```diff
@@ -18,24 +19,27 @@ const FloatingPlaceholder = ({
   floatOnFocus,
   validationMessagePosition,
   extraOffset = 0,
+  defaultValue,
```
Inbal-Tish, line 22: The FlotingPlaceholder only needs to know only if to float or not, why doesn't the TextField component controls it? Like passing in 'shouldFloat' prop as boolean that will be updated by the state of the input's value?

src/incubator/TextField/types.ts
M-i-k-e-l: No, it's RN's props

src/incubator/TextField/index.tsx
M-i-k-e-l: This is a prop from RN which we inherit and the user that opened the bug is using

src/incubator/TextField/FloatingPlaceholder.tsx
M-i-k-e-l: You're saying that `shouldFloat` should be coming from somewhere else? Maybe the `Presenter` like in `ValidationMessage`?

Inbal-Tish: Yes. I think the floating placeholder should only get a boolean and the logic will be in the component that uses it

src/incubator/TextField/types.ts
Inbal-Tish: Ok...

Inbal-Tish (approved):

PR #2353, opened 2022-11-29
Upgrade reanimated from 2.8 to 2.13

ethanshar:
## Description
Upgrade reanimated from 2.8 to 2.13

## Changelog
Upgrade reanimated from 2.8 to 2.13 in our demo app


Inbal-Tish:
@ethanshar Did we upgrade it in the engine already?

Inbal-Tish (approved):

ethanshar:
@Inbal-Tish 
No, I want to upgrade uilib first, see everything is ok and only then upgrade the engine.

PR #2370, opened 2022-12-13
NumberInput - support theme and generify useThemeProps

M-i-k-e-l:
## Description
NumberInput - support theme and generify useThemeProps
When adding `useThemeProps` to `NumberInput` I've noticed that the props are not being used (I had a ` // @ts-expect-error` which was supposedly fixed when it should not have been), so I added generics to `useThemeProps`; I've fixed some errors and ignored one in `asBaseComponent`.

## Changelog
⚠️ NumberInput - support theme and generify useThemeProps

### DIFF
--- a/src/components/numberInput/index.tsx
+++ b/src/components/numberInput/index.tsx
@@ -1,7 +1,7 @@
 import {isEmpty} from 'lodash';
 import React, {useMemo, useCallback, useState, useEffect} from 'react';
 import {StyleSheet, StyleProp, ViewStyle} from 'react-native';
-import {useDidUpdate} from 'hooks';
+import {useDidUpdate, useThemeProps} from 'hooks';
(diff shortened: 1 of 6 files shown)


src/components/numberInput/index.tsx
```diff
@@ -47,6 +47,7 @@ export type NumberInputProps = React.PropsWithRef<
 };
 
 function NumberInput(props: NumberInputProps, ref: any) {
+  const themeProps = useThemeProps(props, 'NumberInput');
```
ethanshar, line 50: Didn't you add generics to useThemeProps so you can send it here?

M-i-k-e-l: I did, but it works without it 🪄 
Pretty sure that it gets the type from the input `props` and that's why I got the errors in `Picker` and `Avatar`

ethanshar (approved):

PR #2374, opened 2022-12-14
Fix/ TextField for react-native-web

lidord-wix:
## Description
Fix Incubator.TextField for react-native-web - make it pressable + remove extra outline
WOAUILIB-3243

## Changelog
Fix Incubator.TextField for react-native-web

### DIFF
--- a/src/incubator/TextField/Input.tsx
+++ b/src/incubator/TextField/Input.tsx
@@ -58,8 +58,7 @@ const styles = StyleSheet.create({
   input: {
     flexGrow: 1,
     textAlign: Constants.isRTL ? 'right' : 'left',
-    // The next line is required to remove the default outline of react-native-web
-    outlineWidth: 0,
+    outlineWidth: Constants.isWeb ? 0 : undefined,
     // Setting paddingTop/Bottom separately fix height issues on iOS with multiline
     paddingTop: 0,
     paddingBottom: 0,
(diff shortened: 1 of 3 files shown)


src/incubator/TextField/Input.tsx
```diff
@@ -58,6 +58,8 @@ const styles = StyleSheet.create({
   input: {
     flexGrow: 1,
     textAlign: Constants.isRTL ? 'right' : 'left',
+    // The next line is required to remove the default outline of react-native-web
+    outlineWidth: 0,
```
Inbal-Tish, line 62: Did you check if you can do this: `outlineWidth: Platform.OS === 'web' ? 0 : undefined,`?

adids1221 (approved):

src/incubator/TextField/Input.tsx
lidord-wix: I guess I can, but I preferred not to add this check many times in the code.
(this is the first PR and maybe we'll have many more in the near future)
And since it doesn't affect the native I thought that's ok..
WDYT?

Inbal-Tish: In this particular case, this code is not affecting the native but could be a case where it will. Also, I think we need to mark web-related code somehow and this is a good one (obviously we need to add this check to our constants and just check `isWeb`). We can take it with the rest of the devs but I think we need to decide about it as this is the first in many adjustments for web support.

lidord-wix: Agree.
I added Constants.isWeb

PR #2379, opened 2022-12-15
Fix/incubator dialog fader visibility and more

M-i-k-e-l:
## Description
You can look at each commit separately
Fixes for the fade view (most important is moving the `withTiming` to the `useAnimatedStyle`), the others are making the code a little better.
Do not set pointerEvents by default (allow via props) - this fixes a bug in Android where clicking on the header closes the dialog (it goes through to the fade view)

WOAUILIB-3268

## Changelog
Fix Incubator.Dialog's fader view not being shown sometimes


src/incubator/Dialog/index.tsx
```diff
@@ -163,7 +164,7 @@ const Dialog = (props: DialogProps, ref: ForwardedRef<DialogImperativeMethods>)
   const renderDialog = () => {
     return (
       <PanGestureHandler onGestureEvent={isEmpty(directions) ? undefined : panGestureEvent}>
-        <View pointerEvents={'box-none'} reanimated style={style} onLayout={onLayout} ref={setRef} testID={testID}>
+        <View {...containerProps} reanimated style={style} onLayout={onLayout} ref={setRef} testID={testID}>
```
Inbal-Tish, line 167: Isn't removing `pointerEvents={'box-none'}` the fix? So not sure why the 'containerProps' are needed... Assuming that's for passing 'pointerEvents', in what scenario would the user need to pass `pointerEvents={'box-none'}` in the container props?

M-i-k-e-l: In private we have a case were we render our own `header` which only part of it has `pointerEvents: 'box-none'` - that's the use case.

From the description:
> Do not set pointerEvents by default (allow via props) - this fixes a bug in Android where clicking on the header closes the dialog (it goes through to the fade view)

You can look at master in private on `ActionSheet` (Android only) - click on the header where there's no text

Inbal-Tish (approved):

PR #2381, opened 2022-12-18
setLocalScheme

lidord-wix:
## Description
Add `setLocalScheme` method to our Config class to support using the scheme from the local storage.

## Changelog
Add `setLocalScheme` method to our Config class to support using the scheme from the local storage.

### DIFF
--- a/src/commons/Config.ts
+++ b/src/commons/Config.ts
@@ -1,4 +1,5 @@
 import {SchemeType} from '../style';
+import {AsyncStoragePackage} from 'optionalDeps';
 interface ConfigOptions {
   /**
    * Should use platform colors for design tokens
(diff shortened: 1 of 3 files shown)


src/commons/Config.ts
```diff
@@ -18,10 +19,18 @@ class Config {
     this.setConfig({});
   }
 
-  public setConfig(options: ConfigOptions) {
+  public async setConfig(options: ConfigOptions) {
     const {usePlatformColors = false, appScheme = 'light'} = options;
     this.usePlatformColors = usePlatformColors;
-    this.appScheme = appScheme;
+    this.appScheme = await AsyncStoragePackage?.getItem?.('appScheme') || appScheme;
```
ethanshar, line 25: You don't check here if AsyncStoragePackage (you check only in setLocalScheme) so technically this would throw error for all users who don't have this package. 
I would create a small function here for validating the package so you won't have to do it many times

ethanshar, line 25: I would also add an option called "useLocalScheme", so not all users have to await the result of async storage if they don't care for it.

ethanshar (commented):
Wrote couple of comments. 
In general, think on how to add it this "local storage" feature as an opt-in so users won't get it by default - since this might affect the behavior of how we set the scheme (see my comments)

src/commons/Config.ts
ethanshar, line 25: consider btw using a more unique id for the storage item, something like `rnuilib.appScheme`. 
So it won't conflict by any chance with other items people might set

ethanshar (approved):

PR #2384, opened 2022-12-19
Image - pass width and height to image style

Inbal-Tish:
## Description
When passing 'width' and 'height' the image doesn't render (only when they are passed in style).
Image - pass width and height to image style.

## Changelog
⚠️ Image - pass 'width' and 'height' to image style to fix layout props not applied to Image without 'style'.


M-i-k-e-l:
The example below seems to be working fine on master
```
<Image
  source={{uri: 'https://static.wixstatic.com/media/d911269bdf7972c9a59ba30440cb3789.jpg_128'}}
  width={50}
  height={50}
/>
```

Can you please provide an example that fails?

Inbal-Tish:
> The example below seems to be working fine on master
> 
> ```
> <Image
>   source={{uri: 'https://static.wixstatic.com/media/d911269bdf7972c9a59ba30440cb3789.jpg_128'}}
>   width={50}
>   height={50}
> />
> ```
> 
> Can you please provide an example that fails?

This example doesn't work on master...

src/components/image/index.tsx
```diff
@@ -233,6 +236,7 @@ class Image extends PureComponent<Props, State> {
           aspectRatio && {aspectRatio},
           !useImageInsideContainer && margins,
           useImageInsideContainer && styles.containImage,
+          !cover && sizeProps && {width, height},
```
M-i-k-e-l, line 239: I don't think the `cover` is needed here, if the user makes a mistake they can fix it (this will be bugged without an overlay as well)

M-i-k-e-l: Also, personally I prefer to remove the extra `const sizeProps`.
Maybe just `{width, height}` or `(width || height) && {width, height}`?

M-i-k-e-l (commented):
Please add the relevant example and a ⚠️

src/components/image/index.tsx
Inbal-Tish: 1) If we know passing width and height not working on 'cover' prop, why not condition it? Why make the user figure it out (or not) by themseves? 
2) What do you mean by 'this will be bugged without an overlay as well'? This solution works whether you pass overlay or not...
3) Done

demo/src/screens/componentScreens/ImageScreen.tsx
```diff
@@ -121,7 +121,9 @@ class ImageScreen extends Component<{}, State> {
         cover={cover}
         overlayType={overlayType !== 'none' ? overlayType : undefined}
         overlayIntensity={overlayIntensity}
-        style={!cover && {width: DEFAULT_SIZE, height: DEFAULT_SIZE}}
+        // style={!cover && {width: DEFAULT_SIZE, height: DEFAULT_SIZE}}
+        width={!cover ? DEFAULT_SIZE : undefined}
+        height={!cover ? DEFAULT_SIZE : undefined}
```
M-i-k-e-l, line 126: Was this on purpose?

src/components/image/index.tsx
M-i-k-e-l: I think I meant that if the user adds width and forgets height (or vice versa, or adds a width with cover etc) they will have bugs that they will need to fix.

M-i-k-e-l (approved):
Added a small comment, approved

demo/src/screens/componentScreens/ImageScreen.tsx
Inbal-Tish: Yes, I remove the 'style' prop

src/components/image/index.tsx
Inbal-Tish: Yes, that's for sure

PR #2387, opened 2022-12-20
Fix/ add webStyle to TextField

lidord-wix:
## Description
TextField - fix web style 

## Changelog
TextField - fix web style

### DIFF
--- a/src/incubator/TextField/Input.tsx
+++ b/src/incubator/TextField/Input.tsx
@@ -41,7 +41,7 @@ const Input = ({
 
   return (
     <TextInput
-      style={[styles.input, !!inputColor && {color: inputColor}, style]}
+      style={[styles.input, !!inputColor && {color: inputColor}, style, Constants.isWeb && styles.webStyle]}
       {...props}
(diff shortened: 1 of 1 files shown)


src/incubator/TextField/Input.tsx
```diff
@@ -69,6 +68,10 @@ const styles = StyleSheet.create({
         textAlignVertical: 'center'
       }
     })
+  },
+  webStyle: {
```
Inbal-Tish, line 72: Does this solve your issue? I believe we meant to get the 'webStyle' to a const outside of the stylesheet, as the stylesheet is typed. You can move it outside the component or create a memorized const as this is not a dynamic style

lidord-wix: Yes it solves the issue (with the ts-expect-error) and keeps the style in the stylesheet so I think it is more readable and clean.

Inbal-Tish (approved):

Inbal-Tish:
@lidord-wix Approved. Please fix build

PR #2388, opened 2022-12-22
fix: TextField defaultValue change

AmitShwarts:
## Description
Currently when trying to change default value then it is not rendered.
This PR will fix this issue.

## Changelog
Fix defaultValue not changing


lidord-wix (commented):
Can you share an example? 
When I render a simple example of `TextField` and update the `defaultValue` prop it does the update as I expected..

AmitShwarts:
> Can you share an example? When I render a simple example of `TextField` and update the `defaultValue` prop it does the update as I expected..

Sorry, forgot to mention that it happens on web.
i updated the webDemo to show this use case.

src/incubator/TextField/__tests__/index.spec.js
```diff
@@ -83,4 +84,33 @@ describe('TextField', () => {
...
+      renderTree.getByDisplayValue('someDefaultValue');
+    });
+
+    it('value should equal value on first render when given', () => {
+      const renderTree = render(<TestCase {...props} defaultValue={undefined}/>);
+
+      renderTree.getByDisplayValue('someValue');
+    });
+
+    it('on web should reset defaultValue when prop changed after first render', () => {
```
lidord-wix, line 107: I suggest adding the same test without setting `Constants.isWeb` to `true` since the same behavior should also apply on mobile

lidord-wix (commented):
I updated the changes from this PR in the node modules and I can't see any change when pressing the 'update default value' button.. 
Did you check it?

AmitShwarts:
> I updated the changes from this PR in the node modules and I can't see any change when pressing the 'update default value' button.. Did you check it?

yes it worked well for me but now it's not.
i think it's because a change in master that was merged in the mean time.
i will fix it and update you.

lidord-wix (approved):

PR #2390, opened 2022-12-26
Fix dummyPlaceholder style

M-i-k-e-l:
## Description
After looking at `MainInput` size issue we've discussed, I think the dummy placeholder should have the `inputStyle` for the typography.
Technically it can be `[typographyStyle, others.style, styles.dummyPlaceholder]` but I think that will be less clear.

## Changelog
Fix dummyPlaceholder style


ethanshar (approved):

ethanshar:
@M-i-k-e-l Approved. 
Need to fix conflicts

PR #2396, opened 2022-12-29
Fix/ Avatar renderRibbon

lidord-wix:
## Description
Avatar - don't render ribbon when there's a custom ribbon
(the bug is visible on the private's avatar screen, on master, and on production) 

## Changelog
Avatar - don't render ribbon when there's a custom ribbon


M-i-k-e-l (approved):

M-i-k-e-l:
Private snapshots should will probably need updating

PR #2410, opened 2023-01-08
fix: web load svg tag data

OrenZak:
## Description
Fix loading svg data tag on web

## Changelog
* refactor svgImage/index.web.ts to allow render a svg tag data
*  added demo button with svg data tag as icon source


src/components/button/ButtonConstants.ts
```diff
@@ -22,3 +24,8 @@ export const MIN_WIDTH = {
 };
 
 export const DEFAULT_SIZE = ButtonSize.large;
+
+export const ICON_SOURCE_MARGIN = {
```
adids1221, line 28: This `ICON_SOURCE_MARGIN` constant is for SVG image usage, instead of having this constant in the `ButtonConstants` file can we add this functionality to the `asCss` function in the `svgImage` index file?

In the `asCss` we will change the margin from `8` to `8px` instead of dealing with it inside the `ButtonConstants` file.

webDemo/package.json
```diff
@@ -28,7 +28,7 @@
     "react-native-shimmer-placeholder": "^2.0.8",
     "react-native-svg": "^12.1.0",
     "react-native-svg-transformer": "^0.14.3",
-    "react-native-ui-lib": "snapshot",
+    "react-native-ui-lib": "6.27.3-snapshot.2347",
```
adids1221, line 31: The example does work.
We need to change the version to `master` or `ga`.

src/optionalDependencies/JsCssPackage.web.ts
```diff
@@ -0,0 +1,13 @@
+let JsCssPackage: {postcss: any, cssjs: any} | undefined;
```
lidord-wix, line 1: I don't think you need two files for this package.. we can have only the:
`src/optionalDependencies/JsCssPackage.ts`
in the mobile we'll not use it anyway since the usage of this package is only from the web implementation of the SVG

src/optionalDependencies/index.web.ts
```diff
@@ -0,0 +1,9 @@
+export {default as NetInfoPackage} from './NetInfoPackage';
+export {default as HapticFeedbackPackage} from './HapticFeedbackPackage';
+export {PickerPackage, CommunityPickerPackage} from './PickerPackage';
+export {default as SvgPackage} from './SvgPackage';
+export {createShimmerPlaceholder} from './ShimmerPackage';
+export {default as LinearGradientPackage} from './LinearGradientPackage';
+export {default as AsyncStoragePackage} from './AsyncStoragePackage';
+export {default as JsCssPackage} from './JsCssPackage';
```
lidord-wix, line 8: I'd call it `PostcssPackage` just to keep it aligned with the package name

webDemo/package.json
```diff
@@ -28,9 +28,11 @@
     "react-native-shimmer-placeholder": "^2.0.8",
     "react-native-svg": "^12.1.0",
     "react-native-svg-transformer": "^0.14.3",
-    "react-native-ui-lib": "snapshot",
+    "react-native-ui-lib": "6.28.2-snapshot.2397",
```
lidord-wix, line 31: Are you sure you want to push it? I think you should keep using `snapshot` here

src/optionalDependencies/JsCssPackage.web.ts
OrenZak: i know but i want to imply that by code.

webDemo/package.json
OrenZak: nope, it just was so you can test locally :)

adids1221 (approved):

PR #2411, opened 2023-01-09
fix: TextField validateOnChange with defaultValue executed on start

AmitShwarts:
## Description
Fix bug: validatingOnChange executed on start when providing defaultValue

## Changelog
Fix: validatingOnChange executed on start when providing defaultValue

### DIFF
--- a/src/incubator/TextField/useFieldState.ts
+++ b/src/incubator/TextField/useFieldState.ts
@@ -23,14 +23,14 @@ export default function useFieldState({
     if (Constants.isWeb && !props.value && props.defaultValue && props.defaultValue !== value) {
       setValue(props.defaultValue);
 
-      if (validateOnChange) {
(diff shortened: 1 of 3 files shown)


src/incubator/TextField/useFieldState.ts
```diff
@@ -42,7 +42,7 @@ export default function useFieldState({
     if (props.value !== value) {
       setValue(props.value);
 
-      if (validateOnChange) {
+      if (validateOnChange && (!Constants.isWeb || value !== props.defaultValue)) {
```
lidord-wix, line 45: From what I've seen, it also reproduces on mobile
Please verify it and if it's true, add this condition constantly :)

lidord-wix (commented):
I added 2 small comments :)

PR #2412, opened 2023-01-09
Slider to reanimated

Inbal-Tish:
## Description
Slider to reanimated

## Changelog
Slider to reanimated


src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,575 @@
...
+    } else {
+      return {
+        width: activeTrackWidth.value
+      };
+    }
+  });
+
+  /** renders */
+
+  const renderRangeThumb = () => {
```
ethanshar, line 445: The default thumb and range thumb are pretty similar and can be extracted to a separate sub component

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,575 @@
+import isNumber from 'lodash/isNumber';
```
ethanshar, line 1: Run document formatting on the file, it is not formatted

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,575 @@
...
+    activeTrackWidth.value = x - defaultThumbStart.value.x;
+    
+    onChange({min: defaultThumbStart.value.x, max: x});
+  };
+
+  const defaultThumbGesture = Gesture.Pan()
+    .onBegin(() => {
+      if (disabled) {
+        return;
+      }
```
ethanshar, line 308: Consider instead calling the `enabled` method on the gesture like this 
```
defaultThumbGesture.enabled(!disabled);
```

ethanshar (commented):
Overall the code looks good and the Slider works great
In any case, couple of things
- The code looks good but it's pretty long and difficult to review. We should break it a little, separate to components and move some logic out to presenter/hooks. For example the Thumbs code definitely can moved to a separate sub component
- I noticed a weird bug. After moving a thumb the value numbers are still running for a second.. have you noticed it?

src/incubator/Slider/index.tsx
Inbal-Tish: Done

src/incubator/Slider/index.tsx
Inbal-Tish: Cool. Thanks

src/incubator/Slider/index.tsx
Inbal-Tish: I know. The animated styles are not updated properly when I pass them as a parameter...

Inbal-Tish:
> Overall the code looks good and the Slider works great In any case, couple of things
> 
> * The code looks good but it's pretty long and difficult to review. We should break it a little, separate to components and move some logic out to presenter/hooks. For example the Thumbs code definitely can moved to a separate sub component
> * I noticed a weird bug. After moving a thumb the value numbers are still running for a second.. have you noticed it?

If we want to break the component we should consider using context, it will take about a day to do that...
About the bug, the slider position update is faster than the value's, what you see is probably a delay in the js updates. If I stop the value updates the value would be off

ethanshar:
> > Overall the code looks good and the Slider works great In any case, couple of things
> > 
> > * The code looks good but it's pretty long and difficult to review. We should break it a little, separate to components and move some logic out to presenter/hooks. For example the Thumbs code definitely can moved to a separate sub component
> > * I noticed a weird bug. After moving a thumb the value numbers are still running for a second.. have you noticed it?
> 
> If we want to break the component we should consider using context, it will take about a day to do that... About the bug, the slider position update is faster than the value's, what you see is probably a delay in the js updates. If I stop the value updates the value would be off

Ok, so let's break it. I don't want to merge it like this. 

And about the bug, what do you suggest then? Not sure it's ok at the moment, it looks buggy like this

Inbal-Tish:
@ethanshar I moved some updates to worklets, can you please pull and recheck the bug. I think it helped a bit with the delay

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,390 @@
...
+    })
+    .onUpdate(e => {
+      onSeekStart?.();
+      let newX = defaultThumbStart.value.x + e.translationX * (shouldDisableRTL ? 1 : rtlFix);
+      if (newX < 0) { // bottom edge
+        newX = 0;
+      } else if (!useRange && newX > trackSize.value.width) { // top edge
+        newX = trackSize.value.width;
+      }
+      if (newX <= rangeThumbStart.value.x - rangeGap && newX >= 0) { // range
```
ethanshar, line 265: I found this part a little difficult to understand
Why are you comparing against the rangeSlider value without first checking if useRange is even enabled?
Also, not sure I understand how activeTrackWidth works.. can you elaborate on this part

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,390 @@
...
+        defaultThumbOffset.value = {x: newX, y: 0};
+        activeTrackWidth.value = useRange ? rangeThumbStart.value.x - newX : newX;
+        runOnJS(onChange)(useRange ? {min: newX, max: rangeThumbStart.value.x} : newX);
```
ethanshar, line 268: Consider having a dedicate `useAnimatedReaction` that handles the data and the onChange callback
There you can normalize the (decimal) values so the onChange won't trigger that often. It might even fix the weird behavior you have now 

```
  useAnimatedReaction(() => {
    return Math.round(defaultThumbOffset.value);
  },
  (selected, previous) => {
     // Report onChange only when the value changes 
  },
```

src/incubator/Slider/Thumb.tsx
```diff
@@ -0,0 +1,51 @@
...
+const THUMB_SIZE = 24;
+const THUMB_BORDER_WIDTH = 6;
+const SHADOW_RADIUS = 4;
+
+export interface Props extends ViewProps {
+  animatedStyle?: any;
+  gesture?: any;
+}
+
+const Thumb = (props: Props) => {
```
ethanshar, line 16: You can extract much more logic and code into this component. This is not enough. 
There are still lots of similar duplicate code for default and range thumbs in the main file. 

Try to isolate the Thumb component and think of it as a standalone. 

From what I see this component should accept
- value
- available range to move in (For a simple Slider we should pass the whole slider range. For a range slider, this value should always change)

Once you move the relevant data to this component you'll be able to move most of the code from the main file. Style, gesture, etc..

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,390 @@
...
+      {...accessibilityProps}
+    >
+      {_renderTrack()}
+      {renderDefaultThumb()}
+      {useRange && renderRangeThumb()}
+    </View>
+  );
+};
+
+// @ts-expect-error
```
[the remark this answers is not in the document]
Inbal-Tish: Only you

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,390 @@
...
+          updateRangeThumb(locationX);
+        }
+      }
+    } else {
+      updateDefaultThumb(locationX);
+    }
+  }, []);
+
+  /** gestures and animations */
+
```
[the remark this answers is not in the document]
Inbal-Tish: Ok

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,390 @@
...
+  const activeThumb = useSharedValue(ThumbType.DEFAULT);
+  const isPressedDefault = useSharedValue(false);
+  const defaultThumbOffset = useSharedValue({x: 0, y: 0});
```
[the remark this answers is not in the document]
Inbal-Tish: I was thinking of a vertical slider as well, as a new feature. I wouldn't want to change it again

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,390 @@
...
+  const defaultThumbStart = useSharedValue({x: 0, y: 0});
+  const isPressedRange = useSharedValue(false);
+  const rangeThumbOffset = useSharedValue({x: 0, y: 0});
+  const rangeThumbStart = useSharedValue({x: 0, y: 0});
+
+  const defaultThumbStyle: StyleProp<ViewStyle> = [
+    styles.thumb,
+    {backgroundColor: disabled ? Colors.$backgroundDisabled : thumbTintColor || Colors.$backgroundPrimaryHeavy}
+  ];
+  const _thumbStyle = useSharedValue(JSON.parse(JSON.stringify(StyleSheet.flatten(thumbStyle || defaultThumbStyle))));
```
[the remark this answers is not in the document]
Inbal-Tish: Cool

src/incubator/Slider/index.tsx
Inbal-Tish: 'range' here in the comment refers to a range of the x value, not range Thumb i.e. when the x is in the range of 0 and trackWidth.
activeTrackWidth is the width of the "colored" track, it goes from 0 to x on a default slider and between the default thumb position and range thumb position in the range slider. When the thumbs are moving I update it's width to flow the thumbs

src/incubator/Slider/index.tsx
```diff
@@ -0,0 +1,390 @@
...
+      activeThumb.value = ThumbType.DEFAULT;
+    })
+    .onUpdate(e => {
+      onSeekStart?.();
+      let newX = defaultThumbStart.value.x + e.translationX * (shouldDisableRTL ? 1 : rtlFix);
+      if (newX < 0) { // bottom edge
+        newX = 0;
+      } else if (!useRange && newX > trackSize.value.width) { // top edge
+        newX = trackSize.value.width;
+      }
```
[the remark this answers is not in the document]
Inbal-Tish: I don't see how I can capture the complexity in this simple tool, nor how I can reuse it, so I'm not sure how the code will be more readable using it...

src/incubator/Slider/index.tsx
Inbal-Tish: Why normalize the decimal values? The 'step' prop is the one to use if the user wishes to round the values... I've considered throttling the 'onChange' but that will make the slider inaccurate.

src/incubator/Slider/Thumb.tsx
Inbal-Tish: If I move the gesture logic to the thumb then I will have to have callbacks on every phase so the main Slider can update the track and other dependent logic, and will have to keep updating the Thumb's value. This will create code duplications other than reducing the length of the code...

src/incubator/Slider/index.tsx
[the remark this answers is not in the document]
Inbal-Tish: Done

src/incubator/Slider/index.tsx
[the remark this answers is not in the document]
ethanshar: It is still after useMemo and useImperativeHandle, move it right after props extraction like we do in all of our components

src/incubator/Slider/index.tsx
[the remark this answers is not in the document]
ethanshar: Ok, now I see the error when I remove it. 
Very weird, we do this in many other components..

PR #2423, opened 2023-01-15
fix: touchable opacity ignores event on press

AmitShwarts:
## Description
When pressing on TouchableOpacity it ignores the event and pass only the props.

## Changelog
Fix: Don't ignore the event passed to TouchableOpacity onPress


lidord-wix (commented):
Looks harmless.. but can I ask why you need the event itself? 
Do you have an example?

AmitShwarts:
> Looks harmless.. but can I ask why you need the event itself? Do you have an example?

Yes, the Viewer expects to get the nativeEvent and then connect some logic from Velo handlers.
without the nativeEvent the press event can't processed as expected.

lidord-wix (approved):

PR #2427, opened 2023-01-19
fix value override

lidord-wix:
## Description
TextField - fix value override https://github.com/wix/react-native-ui-lib/pull/2388
WOAUILIB-3377

## Changelog
TextField - fix value override


AmitShwarts (commented):
@lidord-wix 
looks fine to me, i checked it still works as expected

PR #2434, opened 2023-01-23
Add support to inject BI logger to our LogService

ethanshar:
## Description
I refactored our LogService to be a Class 
And also added support to inject a generic BI logger. This is order to support BI events in our components 

## Changelog
Support injecting a BI logger in our LogService

### DIFF
--- a/src/services/LogService.ts
+++ b/src/services/LogService.ts
@@ -1,13 +1,44 @@
-function warn(message: string) {
-  if (__DEV__) {
-    console.warn(message);
-  }
+interface BILogger {
+  log: (event: any) => void;
 }
+class LogService {
+  biLogger: BILogger | undefined;
 
-function error(message: string) {
-  if (__DEV__) {
-    console.error(message);
+  injectBILogger = (biLogger: BILogger) => {
+    this.biLogger = biLogger;
+  };
+
+  biLog = (event: any) => {
+    this.biLogger?.log(event);
   }
+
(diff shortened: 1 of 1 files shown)


ethanshar:
@M-i-k-e-l Can you please review this one, it's pretty small. I'd like to push this forward.. 🙏

src/services/LogService.ts
```diff
@@ -1,13 +1,44 @@
-function warn(message: string) {
-  if (__DEV__) {
-    console.warn(message);
-  }
+interface BILogger {
+  log: (event: any) => void;
```
M-i-k-e-l, line 2: Do we need this to be any?
We can make it generic with the following:
```
interface BILogger<T> {
  log: (event: T) => void;
}
class LogService<T = any> {
  biLogger: BILogger<T> | undefined;

  injectBILogger = (biLogger: BILogger<T>) => {
    this.biLogger = biLogger;
  };

  biLog = (event: T) => {
    this.biLogger?.log(event);
  }
```

M-i-k-e-l (approved):
Approved with a  saggestion

src/services/LogService.ts
ethanshar: How would that work? Since I export this class as a singleton there's no where to set the LogService generic type. No?

M-i-k-e-l: Very true, did not find a different way to do it...

PR #2441, opened 2023-02-01
Tests/incubator dialog support headless tests v2

M-i-k-e-l:
## Description
All changed in this PR are required for this to work except this one:
```
if (!modalVisibility) {
  return null;
}
```

## Changelog
Incubator.Dialog - support headless tests

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -17,9 +17,30 @@ jest.mock('react-native-reanimated', () => {
 });
 global.__reanimatedWorkletInit = jest.fn();
 jest.mock('react-native-gesture-handler',
-  () => ({
-    FlatList: require('react-native').FlatList
-  }),
+  () => {
+    jest.requireActual('react-native-gesture-handler/jestSetup');
+    const GestureHandler = jest.requireActual('react-native-gesture-handler');
+    GestureHandler.Gesture.Pan = () => {
+      const PanMock = {
+        _handlers: {}
+      };
+
+      const getDefaultMockedHandler = handlerName => handler => {
+        if (typeof handler === 'function') {
+          PanMock._handlers[handlerName] = handler;
+        }
+        return PanMock;
+      };
+
+      PanMock.onStart = getDefaultMockedHandler('onStart');
+      PanMock.onUpdate = getDefaultMockedHandler('onUpdate');
+      PanMock.onEnd = getDefaultMockedHandler('onEnd');
+      PanMock.prepare = jest.fn();
+      return PanMock;
+    };
+
+    return GestureHandler;
+  },
   {virtual: true});
 jest.mock('@react-native-picker/picker', () => ({Picker: {Item: {}}}));
 jest.mock('react-native', () => {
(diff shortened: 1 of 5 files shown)


jest-setup.js
```diff
@@ -17,16 +17,42 @@ jest.mock('react-native-reanimated', () => {
...
   {virtual: true});
 jest.mock('@react-native-picker/picker', () => ({Picker: {Item: {}}}));
 jest.mock('react-native', () => {
   const reactNative = jest.requireActual('react-native');
   reactNative.NativeModules.KeyboardTrackingViewTempManager = {};
   return reactNative;
 });
+jest.mock('./src/incubator/Dialog', () => {
+  const Dialog = jest.requireActual('./src/incubator/Dialog');
+  Dialog.default._inTest = true;
```
ethanshar, line 53: Is it me or you don't use this after all?

src/incubator/Dialog/index.tsx
```diff
@@ -216,6 +217,10 @@ const Dialog = (props: DialogProps, ref: ForwardedRef<DialogImperativeMethods>)
     <View testID={`${testID}.overlayFadingBackground`} absF reanimated style={overlayStyle} pointerEvents="none"/>
   );
 
+  if (!modalVisibility) {
```
ethanshar, line 220: This solves the tests issue? I don't understand? 

Will this means the layout calculation will happen every time the visibility changes?

jest-setup.js
M-i-k-e-l: Used [here](https://github.com/wix/react-native-ui-lib/pull/2441/files#diff-2692883c2610fa29ab975894c726a1d04f3c41c3f64a2d0c7ded257267600efbR19)

src/incubator/hooks/useHiddenLocation.ts
```diff
@@ -10,13 +10,26 @@ export interface HiddenLocation extends HiddenLocationRecord {
   wasMeasured: boolean;
 }
 
+// Adding this for headless tests that are not triggering onLayout
+// Unfortunately Dialog._inTest only exist after Dialog was created,
+// so this cannot be done statically
+function getWasMeasuredDefaultValue() {
+  let wasMeasuredDefaultValue = false;
+  if (__DEV__) {
+    wasMeasuredDefaultValue = require('../Dialog').default._inTest ?? false;
```
ethanshar, line 19: You are requiring a component that use this hook (doesn't it cause cyclic require issue?)
And you also reference Dialog inside a general hook, since useHiddenLocation is not a dedicated hook of Dialog. It's a bit weird don't you think?

jest-setup.js
ethanshar: Right missed it.

src/incubator/hooks/useHiddenLocation.ts
M-i-k-e-l: Technically it is a cycle, yes, and I thought it's not optimal as well.
Excuses: this is the solution we've talked about and this is only in dev.

Another option I thought about is sending the defaultValue via a prop with a default value, but I'm pretty sure this means the boolean value cannot be set on Dialog itself.

PR #2443, opened 2023-02-01
Incubator.Dialog - allow headless testing mocking

M-i-k-e-l:
## Description
All changed in this PR are required for this to work except this one:
```
if (!modalVisibility) {
  return null;
}
```

Added `typings/global.d.ts` to solve a TS error (it is not exported); we can add an ignore if you prefer.

## Changelog
Incubator.Dialog - support headless tests

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -1,4 +1,5 @@
 import {NativeModules, AccessibilityInfo} from 'react-native';
+global.__UI_LIB_TESTING__ = true;
 
 NativeModules.StatusBarManager = {getHeight: jest.fn()};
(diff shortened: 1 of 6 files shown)


jest-setup.js
```diff
@@ -1,4 +1,5 @@
 import {NativeModules, AccessibilityInfo} from 'react-native';
+global.__UI_LIB_TESTING__ = true;
```
ethanshar, line 2: Can we have less underscores 😬 

I think `global.UILIB_TESTING` or `global._UILIB_TESTING` is enough

ethanshar (approved):

PR #2446, opened 2023-02-02
Checkbox + RadioButton - fix long labels are stretching outside

Inbal-Tish:
## Description
Checkbox + RadioButton - fix long labels are stretching outside their container

## Changelog
Checkbox + RadioButton - fix long labels are stretching outside their container

### DIFF
--- a/src/components/checkbox/index.tsx
+++ b/src/components/checkbox/index.tsx
@@ -242,7 +242,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     return label ? (
       <View row centerV style={containerStyle}>
         {this.renderCheckbox()}
-        <Text style={[this.styles.checkboxLabel, labelStyle]} {...labelProps} onPress={this.onPress}>
+        <Text flexS style={[this.styles.checkboxLabel, labelStyle]} {...labelProps} onPress={this.onPress}>
           {label}
         </Text>
       </View>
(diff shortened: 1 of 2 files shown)


lidord-wix (commented):
Seems like it didn't solve the problem, see the attached screenshots.
Also on the RadioButton, the spacing between the label and the icon is now bigger (`Text on left `example, with the `iconOnRight` prop)
Screen Shot 2023-02-05 at 15 47 34
Screen Shot 2023-02-05 at 15 46 10

Inbal-Tish:
@lidord-wix This wasn't the fix... The issue was that the text is going outside of the Checkbox parent container, not its own container as this is a line break issue and you cannot control it 100% as the text is dynamic. What the fix did is made the text break before it gets to the parent edge. See screenshots.
Without the fix
With the fix

Inbal-Tish:
Regarding the RadioButton and the spacing between the label and the icon when passing `contentOnLeft` - this will also be dynamic and depends on the text and line break:

Long word that goes to the second line: 

Simulator Screen Shot - iPhone 13 - 2023-02-06 at 09 23 35

Short word that stays at the first line:

Simulator Screen Shot - iPhone 13 - 2023-02-06 at 09 23 42

Inbal-Tish:
Anyways I'll use flex shrink so it will look better...

PR #2450, opened 2023-02-07
Dialog and Modal fix for web

adids1221:
## Description
The picker component supports the web, fixed issues: useDialog mode, Multi-select mode, and TopBar icons fixed.

## Changelog
The picker component supports the web, fixed issues: useDialog mode, Multi-select mode, and TopBar icons fixed.


src/components/icon/index.tsx
```diff
@@ -50,16 +55,36 @@ const Icon = (props: Props) => {
...
+          margins,
+          iconSize,
+          shouldFlipRTL && styles.rtlFlipped,
+          !!tintColor && {
+            tintColor
+          }
+        ]}
+      />
+    );
+  }
```
ethanshar, line 75: It looks like this render is exactly the same as the default render at the bottom..

src/components/icon/index.tsx
```diff
@@ -37,10 +37,15 @@ export type IconProps = ImageProps &
...
+const defaultWebIconSize = 16;
+
+const Icon = forwardRef((props: Props, ref: any) => {
   const {size, tintColor, style, supportRTL, source, assetGroup, assetName, modifiers, ...others} = props;
   const {margins} = modifiers;
-  const iconSize = size ? {width: size, height: size} : undefined;
+  const webIconSize = Constants.isWeb
+    ? {width: size || defaultWebIconSize, height: size || defaultWebIconSize}
+    : undefined;
+  const iconSize = size ? {width: size, height: size} : webIconSize;
```
ethanshar, line 48: I suggest encapsulating this logic under a hook for useIconSize. It will handle all the logic and return the final size. 
Right now it's getting a little unreadable

src/components/modal/index.tsx
```diff
@@ -80,7 +80,7 @@ class Modal extends Component<ModalProps> {
   renderTouchableOverlay() {
     const {testID, overlayBackgroundColor, onBackgroundPress, accessibilityLabel = 'Dismiss'} = this.props;
     if (_.isFunction(onBackgroundPress) || !!overlayBackgroundColor) {
-      const isScreenReaderEnabled = Constants.accessibility.isScreenReaderEnabled;
+      const isScreenReaderEnabled = !Constants.isWeb ? Constants.accessibility.isScreenReaderEnabled : false;
```
ethanshar, line 83: Consider changing this expression to 
```suggestion
      const isScreenReaderEnabled = !Constants.isWeb && Constants.accessibility.isScreenReaderEnabled;
```

src/components/picker/helpers/usePickerLabel.tsx
```diff
@@ -15,10 +16,13 @@ const usePickerLabel = (props: UsePickerLabelProps) => {
...
     const itemsByValue = _.keyBy(items, 'value');
-
-    return _.flow(arr =>
+    const selectedItems = _.flow(arr =>
       _.map(arr, item => (_.isPlainObject(item) ? getItemLabel?.(item) || item?.label : itemsByValue[item]?.label)),
     arr => _.join(arr, ', '))(value);
+    if (Constants.isWeb) {
+      return selectedItems ? selectedItems : undefined;
+    }
+    return selectedItems;
```
ethanshar, line 25: Why do you need this?

src/components/picker/helpers/usePickerLabel.tsx
```diff
@@ -38,7 +42,9 @@ const usePickerLabel = (props: UsePickerLabelProps) => {
 
     // otherwise, extract from picker items
     const selectedItem = _.find(items, {value});
-
+    if (Constants.isWeb) {
+      return _.get(selectedItem?.value || selectedItem, 'label');
+    }
```
ethanshar, line 47: Is this needed? 
We had a migration in v7 regarding the picker values and label. 
This is pure logic, I don't see a reason for web to work different from mobile in this case

src/components/checkbox/index.tsx
```diff
@@ -84,6 +85,8 @@ export interface CheckboxProps extends TouchableOpacityProps {
   containerStyle?: StyleProp<ViewStyle>;
 }
 
+export type Props = CheckboxProps;
```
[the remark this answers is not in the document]
adids1221: No, changed back to `CheckboxProps`

src/components/icon/index.tsx
adids1221: I'll move it into `renderImage` function.
Should I move the Svg into `renderSvg` function also?

src/components/button/index.tsx
```diff
@@ -290,6 +291,11 @@ class Button extends PureComponent<Props, ButtonState> {
       if (typeof iconSource === 'function') {
         return iconSource(iconStyle);
       } else {
+        if (Constants.isWeb) {
```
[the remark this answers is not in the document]
adids1221: Yes, if we will user Icon and the user don't pass any `tintColor` the Icon will get the default color.
So if we change it to Icon it will break some users.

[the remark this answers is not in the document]
ethanshar: Ok, we need to migrate to Icon IMO, we can do it v8. 
Can you add it to the v8 spreadsheet

src/components/icon/index.tsx
ethanshar: Sure, it can make a little more organized 👌

src/components/icon/index.tsx
ethanshar: See if you can do something like that when you destruct the size prop from props with a default prop
```
const {
  size = Constants.isWeb ? defaultWebIconSize : undefined,
  tintColor,
  style,
  .....
}

```

ethanshar (approved):
I replied to some of the comments, small stuff. 
Anyway, approved.

src/components/button/index.tsx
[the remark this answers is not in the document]
adids1221: Done added to the spreadsheet.

PR #2452, opened 2023-02-08
Colors - getTintColor - get key from a single digit (ex. green1, green5)

Inbal-Tish:
## Description
Colors - fix getTintColor - get key from a single digit (ex. green1, green5)

## Changelog
Colors - fix getTintColor - get key from a single digit (ex. green1, green5)


Inbal-Tish:
@lidord-wix Second option was to rename colors like 'green1' 'green01'. What do you think?

lidord-wix:
> @lidord-wix Second option was to rename colors like 'green1' 'green01'. What do you think?

I prefer the current solution since we already use '<color>1' in many places

lidord-wix (approved):

PR #2454, opened 2023-02-09
Add FlashList to optional dependencies

ethanshar:
## Description
Add FlashList to optional dependencies
Use FlashListPackage in Incubator.Calendar component

## Changelog
Add FlashList to optional dependencies

### DIFF
--- a/src/incubator/Calendar/Agenda.tsx
+++ b/src/incubator/Calendar/Agenda.tsx
@@ -1,6 +1,7 @@
 import React, {useContext, useCallback, useRef} from 'react';
 import {runOnJS, useAnimatedReaction, useSharedValue} from 'react-native-reanimated';
-import {FlashList, ViewToken} from '@shopify/flash-list';
+import {FlashListPackage} from 'optionalDeps';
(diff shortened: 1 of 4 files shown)


src/incubator/Calendar/Agenda.tsx
```diff
@@ -1,17 +1,20 @@
...
 import {InternalEvent, Event, DateSectionHeader, UpdateSource} from './types';
 import CalendarContext from './CalendarContext';
 
+const {FlashList} = FlashListPackage;
+
 // TODO: Fix initial scrolling
 function Agenda() {
   const {data, selectedDate, setDate, updateSource} = useContext(CalendarContext);
-  const flashList = useRef<FlashList<InternalEvent>>(null);
+  const flashList = useRef<FlashListType<InternalEvent>>(null);
```
M-i-k-e-l, line 17: Do we prefer this or maybe export from the `optionalDeps` as `FlatList` so we'll have most types

ethanshar: As discussed, we can't export typings with a require.

Inbal-Tish (approved):

PR #2462, opened 2023-02-12
refactor TouchableOpacity forwarding event on onPress and onLongPress

AmitShwarts:
## Description
change the way we forward the event to onPress and onLongPress handler to send event along with props, instead spreading the event together with props

## Changelog
refactor TouchableOpacity forwarding event on onPress and onLongPress

### DIFF
--- a/src/components/touchableOpacity/index.tsx
+++ b/src/components/touchableOpacity/index.tsx
@@ -47,10 +47,16 @@ export interface TouchableOpacityProps
    */
   customValue?: any;
   style?: ViewProps['style'];
-  onPress?: (props?: TouchableOpacityProps & {event: GestureResponderEvent} | any) => void;
-  onPressIn?: (props?: TouchableOpacityProps | GestureResponderEvent | any) => void | RNTouchableOpacityProps['onPressIn'];
-  onPressOut?: (props?: TouchableOpacityProps | GestureResponderEvent | any) => void | RNTouchableOpacityProps['onPressOut'];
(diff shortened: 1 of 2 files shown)


src/components/touchableOpacity/index.tsx
```diff
@@ -155,12 +159,12 @@ class TouchableOpacity extends PureComponent<Props, {active: boolean}> {
     );
   }
 
-  onPress(event: BaseSyntheticEvent) {
-    this.props.onPress?.({...this.props, ...event});
+  onPress(event: GestureResponderEvent) {
```
lidord-wix, line 162: Why you changed it from `BaseSyntheticEvent` to `GestureResponderEvent`?

AmitShwarts: this is the correct typing in React Native:
```
    onLongPress?: ((event: GestureResponderEvent) => void) | undefined;

    /**
     * Called when the touch is released,
     * but not if cancelled (e.g. by a scroll that steals the responder lock).
     */
    onPress?: ((event: GestureResponderEvent) => void) | undefined;

    onPressIn?: ((event: GestureResponderEvent) => void) | undefined;

    onPressOut?: ((event: GestureResponderEvent) => void) | undefined;
```

lidord-wix: OK, please just verify it's the same type for web events (you can run the web demo with this change). 
I thought you picked the `BaseSyntheticEvent` last time for that reason

lidord-wix (approved):

PR #2464, opened 2023-02-13
Incubator.TextField - readonly - allow editable to be undefined

M-i-k-e-l:
## Description
Incubator.TextField - readonly - allow editable to be undefined

## Changelog
None


src/incubator/TextField/Input.tsx
```diff
@@ -30,6 +30,7 @@ const Input = ({
   const placeholderTextColor = getColorByState(props.placeholderTextColor, context);
   const value = formatter && !context.isFocused ? formatter(props.value) : props.value;
   const disabled = props.editable === false || readonly;
+  const editable = (props.editable !== undefined || readonly) ? !disabled : undefined;
```
ethanshar, line 33: It's turning to be a little un-readable
can we do something like this?

```
const editable = props.editable === undefined || props.editable || !readonly;
const disabled = !editable;
```

PR #2465, opened 2023-02-13
feat: extended styling options for segments

ylinssen:
## Description

- Added option to pass _styling_ directly to the _Segment_ component.
- Added setting for adjusting _Segment_'s height to parent's.

For providing users more control over _SegmentedControl's_ (_Segment_) styling.

Simulator Screen Shot – iPhone 14 Pro

## Changelog
- Added option to provide _styling_ directly to the Segment component via SegmentedControlItemProps.
- Added setting for adjusting to parent's height (e.g., segment's height to 100%) via SegmentedControlItemProps.

### DIFF
--- a/src/components/segmentedControl/index.tsx
+++ b/src/components/segmentedControl/index.tsx
@@ -68,6 +68,10 @@ export type SegmentedControlProps = {
    */
   iconOnRight?: boolean;
   /**
+   * Should segments fill parent's height
+   */
+  fillParentY?: boolean;
+  /**
    * Trailing throttle time of changing index in ms.
    */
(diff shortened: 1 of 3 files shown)


ylinssen:
@Inbal-Tish @M-i-k-e-l

M-i-k-e-l:
Hi @ylinssen,
I think we have a very similar example in our example screen to the image you've given,  could you please specify what you are trying to achieve and cannot.
Maybe an image of the design and the result you are getting so we can compare the two.

🤔 I see there are conflicts, maybe your feature was already added?

PR #2467, opened 2023-02-15
v7_TabBar - remove component

Inbal-Tish:
## Description
TabBar - remove component

## Changelog
TabBar - remove component (use TabController instead)


lidord-wix (commented):
I think that in such cases we should keep the old code but remove the component export.
@ethanshar, right?

ethanshar:
> I think that in such cases we should keep the old code but remove the component export.
> @ethanshar, right?

I think this one is safe to remove. 
It's old and undocumented for a while now

lidord-wix (approved):

PR #2468, opened 2023-02-15
v7_TextField

ethanshar:
## Description
Set migrate prop to true by default in TextFieldMigrator and set the new typings. 
Also export TextFieldMethods as type

## Changelog
Enable TextField migration by default


M-i-k-e-l (approved):
How do we feel about moving the `IncubatorTextFieldScreen` from the incubator?
If that's a yes then I suggest using `TextField` and not `Incubator.TextField` and using `TextFieldMethods` in the refs (maybe we should do that either way).

ethanshar:
> How do we feel about moving the `IncubatorTextFieldScreen` from the incubator?
> If that's a yes then I suggest using `TextField` and not `Incubator.TextField` and using `TextFieldMethods` in the refs (maybe we should do that either way).

Done

ethanshar:
> I just noticed that the `TextFieldMigrator` warns about using `centered` - I think it can be removed.
> BTW, I already approved the PR.

Cool, fixed. Also removed transformer from there, it is also supported with formatter prop

PR #2484, opened 2023-02-15
V7/ ThemeManager

lidord-wix:
## Description
ThemeManager deprecations

## Changelog
ThemeManager deprecations


src/style/themeManager.ts
```diff
@@ -3,32 +3,16 @@ import type {Context} from 'react';
...
-    CTA: {
-      textColor: string,
-      disabledColor: string,
-      backgroundColor: string
-    },
-    titleColor: string,
-    subtitleColor: string,
-    dividerColor: string,
-    components: Extendable
+  primaryColor: string;
```
ethanshar, line 6: any reason to keep `primaryColor`?

lidord-wix: I think we use it in private

lidord-wix: Yes, we do :)

ethanshar (approved):

PR #2490, opened 2023-02-16
V7 typescript

ethanshar:
## Description
Upgrade typescript to version V7 4.9.5 and fix errors

## Changelog
Upgrade typescript to version V7 4.9.5 and fix errors


demo/src/screens/MenuStructure.js
```diff
@@ -175,10 +172,9 @@ export const navigationData = {
     title: 'Incubator (Experimental)',
     screens: [
       {title: 'Calendar', tags: 'calendar', screen: 'unicorn.components.IncubatorCalendarScreen'},
-      {title: 'ChipsInput (New)', tags: 'chips input', screen: 'unicorn.components.IncubatorChipsInputScreen'},
+      {title: 'ChipsInput', tags: 'chips input', screen: 'unicorn.components.ChipsInputScreen'},
```
Inbal-Tish, line 175: Should move under 'form'

src/components/button/Button.driver.ts
```diff
@@ -17,7 +17,7 @@ export class ButtonDriver extends ComponentDriver {
   }
 
   isPressable = async () => {
-    if (this.exists()) {
+    if (await this.exists()) {
```
Inbal-Tish, line 20: I think Miki just change the onPress to be sync as it actually is sync. Don't we need to remove the async instead?

Inbal-Tish: https://github.com/wix/react-native-ui-lib/pull/2459/files
It was merged after the Adi cut the version (I wasn't aware that we're cutting a version yesterday) so maybe we should include it in v7

src/components/text/Text.driver.ts
```diff
@@ -11,7 +11,7 @@ export class TextDriver extends ComponentDriver {
   }
 
   isPressable = async () => {
-    if (this.exists()) {
+    if (await this.exists()) {
```
Inbal-Tish, line 14: Same here

Inbal-Tish (approved):

Inbal-Tish:
@ethanshar Wow, these are the only ts errors we had! Just one comment about onPress sync instead of async, but approved anyway

demo/src/screens/MenuStructure.js
ethanshar: Why is related to this PR? I didn't change there anything

ethanshar: hmm I think it was because the PR was targeting master and not v7

src/components/button/Button.driver.ts
ethanshar: @Inbal-Tish This fix is relevant. The isExists is async and we address it a sync function..

M-i-k-e-l: As Inbal stated it looks like it was merged to master after we started working on v7.

Inbal-Tish: Correct

ethanshar: I don't see any delta. 
Which change are you referring to exactly? why did we merged changes while working on v7? 
anyway, do we need to cherry pick it to v7?

Inbal-Tish: I merged the PR since no one told me that we were cutting a version. The change is in the PR I added here

PR #2504, opened 2023-02-27
RN71 upgrade

Inbal-Tish:
## Description
RN71 upgrade

## Changelog
RN71 upgrade


Inbal-Tish:
@ethanshar I disabled the Incubator.Dialog test for now. I'll keep looking for a solution for it but I don't think it should block this upgrade

src/commons/Constants.ts
```diff
@@ -131,15 +131,14 @@ const constants = {
       (screenHeight >= 812 || screenWidth >= 812);
   },
   /* Orientation */
+  dimensionsEventListener: undefined,
```
ethanshar, line 134: Is `dimensionsEventListener` being used anywhere?

src/components/view/index.tsx
```diff
@@ -142,6 +142,7 @@ function View(props: ViewProps, ref: any) {
 
   return (
     <ViewContainer
+      //@ts-ignore
```
ethanshar, line 145: Is it something that might be fixed by react-native types?
Maybe use `@ts-expect-error` instead?

ethanshar (commented):
Looks good. 
Wrote some comments. 

I do have an issue on IOS (Android is fine) in the main screen when scrolling down the list of screens it throws an error, something related to the sections list. 
Do you get it?

ethanshar:
@Inbal-Tish One more thing. 
It's worth someone with the new Mac to checkout this PR and see the new version works fine. either @M-i-k-e-l  or @adimorwix  can check it, they have the new mac

Inbal-Tish:
> Looks good. Wrote some comments.
> 
> I do have an issue on IOS (Android is fine) in the main screen when scrolling down the list of screens it throws an error, something related to the sections list. Do you get it?

Fixed!

src/incubator/Dialog/__tests__/index.spec.tsx
```diff
@@ -49,13 +49,13 @@ describe('Incubator.Dialog', () => {
     const component = _TestCase({onDismiss});
     const dialogDriver = new ComponentDriver({component, testID: 'dialog'});
     expect(await dialogDriver.exists()).toBeFalsy();
```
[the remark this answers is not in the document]
Inbal-Tish: Ticket #3447 created

src/components/view/index.tsx
Inbal-Tish: Changed. Don't really know where this error comes from, might be a RN bug...

src/commons/Constants.ts
Inbal-Tish: Yes, It is used for fixing the layout when orientation change in Carousel, ColorPalette, GridView, Slider, Wizard, and useOrientation hook.

demo/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
Inbal-Tish: Removed. I wasn't sure about this one and forgot to recheck after I finished with fixing the builds... Thanks!

.eslintrc.js
```diff
@@ -43,6 +43,7 @@ module.exports = {
     '@typescript-eslint/no-empty-function': 0,
     '@typescript-eslint/camelcase': 0,
     '@typescript-eslint/indent': 0,
-    '@typescript-eslint/explicit-module-boundary-types': 0
+    '@typescript-eslint/explicit-module-boundary-types': 0,
+    semi: [0, 'never']
```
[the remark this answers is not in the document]
Inbal-Tish: It was hiding 27 errors so I though we were not enforcing it and now it's started erroring when I updated the library. I can fix these errors and remove this

Inbal-Tish:
> @Inbal-Tish One more thing. It's worth someone with the new Mac to checkout this PR and see the new version works fine. either @M-i-k-e-l or @adimorwix can check it, they have the new mac

Ok. I'll ask them to check it.

M-i-k-e-l:
Both platforms installed and small sanity was performed

ethanshar (approved):

Issue #2494, opened 2023-02-17
Could not find a declaration file for module 'react-native-ui-lib/keyboard'


whalemare:
## Description

Incorrect typings in docs or library

### Related to

- [ ] Components
- [ ] Demo
- [ ] Docs
- [x] Typings

### Steps to reproduce

1. Open documentation link: https://wix.github.io/react-native-ui-lib/docs/getting-started/setup#how-does-it-work
2. See usage example
```
...
import {KeyboardTrackingView, KeyboardAwareInsetsView, KeyboardRegistry, KeyboardAccessoryView, KeyboardUtils} from 'react-native-ui-lib/keyboard';
...
```
3. Try to import like docs says in some TypeScript file

#### Expected behavior

Imported without TypeScript errors

#### Actual behavior

TypeScript file has error

```
Could not find a declaration file for module 'react-native-ui-lib/keyboard'
```

## More Info

Next import is resolves without errors, but its just prop types:

```
import { KeyboardTrackingViewProps } from 'react-native-ui-lib/lib/components/Keyboard'
```

### Code snippet

```
No applicable
```

### Environment

- React Native version: 0.71
- React Native UI Lib version: 7.0.0

### Affected platforms

- [] Android
- [] iOS
- [] Web

gentlee:
This bug is still valid, can't just export KeyboardTrackingView from /keyboard as in getting started (facepalm).

PR #2506, opened 2023-02-27
Export the Keyboard's component as the documentation describes

reinaldonetof:
## Description

Working with this library, I noticed that the components from the keyboard aren't imported as the documentation https://wix.github.io/react-native-ui-lib/docs/getting-started/setup#how-does-it-work

To resolve this issue I created a file keyboard.ts and exported each component as const. Now we can import as described
`import { KeyboardAccessoryView } from 'react-native-ui-lib/keyboard';`

Closes https://github.com/wix/react-native-ui-lib/issues/2494

## Changelog
*Add a quick message for our users about this change (include Component name, relevant props and general purpose of the PR)*

### DIFF
--- a/keyboard.d.ts
+++ b/keyboard.d.ts
@@ -1,0 +1,2 @@
+import Keyboard from './lib/components/Keyboard';
+export = Keyboard;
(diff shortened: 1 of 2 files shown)


gentlee:
This bug is still valid, can't just export KeyboardTrackingView from /keyboard as in getting started (facepalm).

PR #2510, opened 2023-03-01
New colors screen

adids1221:
## Description
New colors screen.

## Changelog
New colors screen.


demo/src/screens/foundationScreens/ColorsScreen.js
```diff
@@ -1,115 +1,308 @@
...
+  TextField,
+  Incubator
+} from 'react-native-ui-lib';
 
+const {Toast} = Incubator;
 
-const {width} = Dimensions.get('window');
-const baseColors = ['grey', 'blue', 'cyan', 'green', 'yellow', 'orange', 'red', 'purple', 'violet'];
+const SYSTEM_COLORS = ['grey', 'violet', 'blue', 'green', 'red', 'yellow', 'orange'];
+const PRIVATE_COLORS = ['dine', 'fit', 'premium'];
```
Inbal-Tish, line 20: Let's move this to a custom prop and use 'passProps' to pass it from outside

demo/src/screens/foundationScreens/ColorsScreen.js
```diff
@@ -1,115 +1,313 @@
...
+  } else if (key.startsWith('$outline')) {
+    TOKENS_ARRAYS.Outline.push(key);
+  }
+}
 
+class ColorsScreen extends Component {
   constructor(props) {
     super(props);
 
+    this.extraColors = this.props.extraColors;
```
Inbal-Tish, line 42: if you're in the c'tor you don't need the 'this' before props, they are passed as a parameter. By the way, you don't need a c'tor at all as you can get the props using 'this' (as you did here).

demo/src/screens/foundationScreens/ColorsScreen.js
```diff
@@ -1,115 +1,313 @@
...
+  };
+
+  dismissToast = () => {
+    this.setState({showToast: false});
+  };
+
+  onTokenPress = value => {
+    const systemColorName = Colors.getSystemColorByHex(Colors[value].toString(), [
+      ...SYSTEM_COLORS,
+      ...(this.extraColors || []),
```
Inbal-Tish, line 71: Did you check if it crashes when you spread undefined? If it is, you can set the empty array as the const default to avoid this condition... it will look and read better

adids1221: Yes, I checked for crashes, this is why I used the condition inside the spread here.
Do you mean something like this ?
`const {extraColors=[]} = this.props`

Inbal-Tish:
@adids1221 Looks good. 2 little comments

demo/src/screens/foundationScreens/ColorsScreen.js
Inbal-Tish: Yes

Inbal-Tish (approved):

PR #2521, opened 2023-03-09
initialNumToRender prop added to WheelPicker FlatList props

adids1221:
## Description
SectionWheelPicker passes `flatListProps` and `style` to solve the item not rendering issue in wheelPicker.

## Changelog
SectionWheelPicker new prop `androidBigDataSection`, pass this prop when using Android to solve wheelPicker not render all the section items after selecting a value.


adids1221:
> Hi @adids1221,
> 
> I don't really like the Android only prop (even if the bug occurs only on Android). Further more, I think the solution should be to allow passing the required props to solve the issue and then let the user know what they should pass (this is the same solution as the other ticket you've mentioned, see #2461).

We need the Android only prop but in the Duration Picker.
`DurationPicker` will have Android only prop as a boolean, and the logic must be handled in this case in the `sectionWheelPicker`.
If this prop passed from the `DurationPicker` then the `sectionWheelPicker` will go over each section, extract the section's length, and set it to the `flatList` props we want to pass to the `WheelPicker` in this case.

TLDR:
We must have this prop in the DurationPicker and the logic, in this case, must be handled by us in the `sectionWheelPicker`.

M-i-k-e-l:
As discussed, this can be handled in the `DurationPicker` (only)

PR #2542, opened 2023-03-27
Replacing 'react-native-text-size' with wix fork

Inbal-Tish:
## Description
Replacing 'react-native-text-size' with wix fork

DO NOT MERGE YET!!!

## Changelog
⚠️ Replacing 'react-native-text-size' with wix fork


ethanshar:
Just making sure. 
Once this is merged.. if the engine still have the old dependency, what will happen? It's a native dep.. no?

Inbal-Tish:
> Just making sure. Once this is merged.. if the engine still have the old dependency, what will happen? It's a native dep.. no?

I marked it as "pending" as I need the engine to include the new native dep. Once the engine will add the new it will have both repos and we could marge this PR to use the new one.

ethanshar (approved):

PR #2547, opened 2023-03-30
Fix/grid related fixes

M-i-k-e-l:
## Description
useGridLayout - allow configuration (related to WOAUILIB-3451)
GridView - fix item size calculation (related to WOAUILIB-3451, can be seen with private changes, in the private repo in the dynamic width section on iPad Pro 9.7-inch)

## Changelog
GridView - fix item size calculation

### DIFF
--- a/src/components/gridList/useGridLayout.ts
+++ b/src/components/gridList/useGridLayout.ts
@@ -1,6 +1,6 @@
 import {useMemo} from 'react';
 import {Spacings} from 'style';
-import {useOrientation} from 'hooks';
+import {useOrientation, useThemeProps} from 'hooks';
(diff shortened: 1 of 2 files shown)


src/components/gridList/useGridLayout.ts
```diff
@@ -16,7 +16,7 @@ const useGridLayout = (props: GridListBaseProps) => {
     keepItemSize,
     containerWidth,
     contentContainerStyle
-  } = props;
+  } = useThemeProps(props, 'useGridLayout');
```
ethanshar, line 19: I commented about it in the private PR. If we don't have any alternatives let's go for it..

ethanshar (approved):

PR #2548, opened 2023-04-02
Fix/ avatar label sizes

lidord-wix:
## Description
Fix Avatar label sizes
WOAUILIB-3468

## Changelog
Fix Avatar label sizes

### DIFF
--- a/demo/src/screens/__tests__/__snapshots__/AvatarScreen.spec.js.snap
+++ b/demo/src/screens/__tests__/__snapshots__/AvatarScreen.spec.js.snap
@@ -456,7 +456,10 @@ exports[`AvatarScreen renders screen 1`] = `
                 undefined,
                 [
                   {
-                    "fontSize": 16,
+                    "fontFamily": "System",
+                    "fontSize": 14,
+                    "fontWeight": "700",
+                    "lineHeight": 20,
                   },
                   {
                     "backgroundColor": "transparent",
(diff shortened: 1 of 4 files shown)


src/helpers/AvatarHelper.ts
```diff
@@ -1,6 +1,33 @@
...
+      break;
+    case size < 100:
+      typography = Typography.text50H;
+      break;
+    default:
+    case size >= 100:
+      typography = Typography.text40H;
+      break;
+  }
+  return typography;
```
M-i-k-e-l, line 29: I like this feature! Please add it to the dev sync.
I assume you've talked about the UX on this change, right?

demo/src/screens/__tests__/__snapshots__/AvatarScreen.spec.js.snap
M-i-k-e-l, line 1: You'll probably need to fix private as well

src/helpers/AvatarHelper.ts
lidord-wix: Yes with Diana :)

demo/src/screens/__tests__/__snapshots__/AvatarScreen.spec.js.snap
lidord-wix: Yup

M-i-k-e-l (approved):

PR #2551, opened 2023-04-04
Incubator.Dialog - fix test (add mocks)

M-i-k-e-l:
## Description
There is some awkward code in this PR, but after talking with Niryo, this is what we're doing in the private mocks.

## Changelog
None

### DIFF
--- a/GestureDetectorMock.tsx
+++ b/GestureDetectorMock.tsx
@@ -1,0 +1,67 @@
+import React from 'react';
+import {TouchableOpacity} from 'react-native';
+
+type Props = {
+  gesture: any;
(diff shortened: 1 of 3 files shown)


Inbal-Tish (approved):

Inbal-Tish:
@M-i-k-e-l This is crazy! where did you find this solution??

M-i-k-e-l:
> @M-i-k-e-l This is crazy! where did you find this solution??

Stole some from NirYo (private mocks) and added onto it.
We got some feedback in private, so we might need to adjust it here as well, not sure yet.

PR #2558, opened 2023-04-16
fix wheelpicker in android

adids1221:
## Description
WheelPicker missing data in Android bug fix, passing `maxToRenderPerBatch` to the `FlatList` when using Android.

## Changelog
WheelPicker missing data in Android bug fix, passing `maxToRenderPerBatch` to the `FlatList` when using Android.
Solve [RN FlastList issue](https://github.com/facebook/react-native/issues/15990).
`WOAUILIB-3440`

### DIFF
--- a/src/components/WheelPicker/wheelPicker.api.json
+++ b/src/components/WheelPicker/wheelPicker.api.json
@@ -2,7 +2,7 @@
   "name": "WheelPicker",
   "category": "form",
   "description": "A customizable WheelPicker component",
-  "note": "When using Android by default the FlatList will have 'maxToRenderPerBatch' prop set to items.<br/>length to solve FlatList bug on Android, you can override it by passing your own 'flatListProps' with 'maxToRenderPerBatch' prop.<br/>See the RN FlatList issue for more info: https://github.com/facebook/react-native/issues/15990",
+  "note": "When using Android by default the FlatList will have <code>maxToRenderPerBatch</code> prop set to <code>items.length</code> to solve FlatList bug on Android, you can override it by passing your own 'flatListProps' with 'maxToRenderPerBatch' prop.<br/>See the RN FlatList issue for more info: https://github.com/facebook/react-native/issues/15990",
   "example": "https://github.com/wix/react-native-ui-lib/blob/master/demo/src/screens/componentScreens/WheelPickerScreen.tsx",
   "images": [],
   "props": [
--- a/src/components/WheelPicker/wheelPicker.api.json
+++ b/src/components/WheelPicker/wheelPicker.api.json
@@ -33,7 +33,12 @@
       "default": "center"
     },
     {"name": "separatorsStyle", "type": "ViewStyle", "description": "Extra style for the separators"},
-    {"name": "testID", "type": "string", "description": "test identifier"}
+    {"name": "testID", "type": "string", "description": "test identifier"},
+    {
+      "name": "flatListProps",
+      "type": "FlatListProps",
+      "description": "Props to be sent to the FlatList. When using Android by default the FlatList will have 'maxToRenderPerBatch' prop set to items.length to solve FlatList bug on Android, you can override it by passing your own 'flatListProps' with 'maxToRenderPerBatch' prop"
+    }
   ],
   "snippet": [
     "<WheelPicker",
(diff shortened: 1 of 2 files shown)


adids1221:
@M-i-k-e-l If you want to test the solution you can use version `2.76143.0`.
I'm not sure about the warning sign in the changelog, tell me what you think.

M-i-k-e-l (commented):
The solution looks good generally and it is working, however I would not want it forced on all of our users.
Can you think of another solution?

M-i-k-e-l:
> @M-i-k-e-l If you want to test the solution you can use version `2.76143.0`. I'm not sure about the warning sign in the changelog, tell me what you think.

Generally I'd avoid two lines in the changelog since I don't think our script handles that well.

adids1221:
> The solution looks good generally and it is working, however I would not want it forced on all of our users. Can you think of another solution?

There are no performance issues when the list is small (or less than 20 items).
In that case, forcing it on all the users will fix the bug and the performance will impact only "big data" lists.

I don't think we need it as a prop inside WheelPicker since it will force us to go over many components which use the WheelPicker.

M-i-k-e-l:
> There are no performance issues when the list is small (or less than 20 items). In that case, forcing it on all the users will fix the bug and the performance will impact only "big data" lists.
> 
> I don't think we need it as a prop inside WheelPicker since it will force us to go over many components which use the WheelPicker.

1. We don't need a new prop.
2. What if someone has a long list and only uses iOS? Why should they suffer performance degradation?

M-i-k-e-l:
> > > There are no performance issues when the list is small (or less than 20 items). In that case, forcing it on all the users will fix the bug and the performance will impact only "big data" lists.
> > > I don't think we need it as a prop inside WheelPicker since it will force us to go over many components which use the WheelPicker.
> > 
> > 
> > 
> > 1. We don't need a new prop.
> > 2. What if someone has a long list and only uses iOS? Why should they suffer performance degradation?
> 
> the `androidFlatListProps` object will return the `maxRenderPerBatch` prop only if the user has an Android device. I'm using this line `if (Constants.isAndroid)` inside the `useMemo`. So if the user uses an IOS device it will return undefined.

True 🤦 🙃 But I still think this should be in private and not affect all the public users

adids1221:
@M-i-k-e-l As we spoke the user can pass `flatListProps={{maxToRenderPerBatch: undefined}}` to the `WheelPicker` and override the prop we pass to solve this issue.

M-i-k-e-l:
> @M-i-k-e-l As we spoke the user can pass `flatListProps={{maxToRenderPerBatch: undefined}}` to the `WheelPicker` and override the prop we pass to solve this issue.

Lets document this properly

adids1221:
@M-i-k-e-l After checking with Ethan, I add the RN issue link to the component note.

src/components/WheelPicker/wheelPicker.api.json
```diff
@@ -2,6 +2,7 @@
   "name": "WheelPicker",
   "category": "form",
   "description": "A customizable WheelPicker component",
+  "note": "When using Android by default the FlatList will have 'maxToRenderPerBatch' prop set to items.<br/>length to solve FlatList bug on Android, you can override it by passing your own 'flatListProps' with 'maxToRenderPerBatch' prop.<br/>See the RN FlatList issue for more info: https://github.com/facebook/react-native/issues/15990",
```
M-i-k-e-l, line 5: I did not test it but I assume `items.<br/>length` will look strange, it should probably be in the same line and preferably with code look (if possible in the `note`): `items.length`

M-i-k-e-l (approved):

src/components/WheelPicker/wheelPicker.api.json
adids1221: Fixed.
Screen Shot 2023-05-01 at 16 29 13

adids1221: Fixed, checked.
Wrapped with `<code>Text...</code>`

M-i-k-e-l:
> > @M-i-k-e-l If you want to test the solution you can use version `2.76143.0`. I'm not sure about the warning sign in the changelog, tell me what you think.
> 
> Generally I'd avoid two lines in the changelog since I don't think our script handles that well.

@adids1221 
Please avoid multiline in the changelog in the future

PR #2560, opened 2023-04-16
export FlashList with default value

mendyEdri:
## Description
Export `FlashList` optional dependency from `index.web` as well, plus set FlashList return value with a default object to fix deconstruct javascript error.  

## Changelog
Fix `FlashList`'s javascript error when rendering web components


mendyEdri:
@adids1221 I'm not sure why I get this build error 🙁

PR #2561, opened 2023-04-17
Feat/ SortableList - add ItemMargin prop

lidord-wix:
## Description
SortableList - add ItemMargin prop
https://github.com/wix/react-native-ui-lib/discussions/2538

## Changelog
SortableList - add ItemMargin prop


YuyouWu:
Hi @lidord-wix. I checked out `7.3.0-snapshot.2948` and tried it out. I'm getting some weird behavior when reordering items. Are you able to reproduce this? 
https://user-images.githubusercontent.com/16201149/232669274-edc86e92-12f7-41a0-b342-d8c80f0e92fc.MP4

lidord-wix:
> Hi @lidord-wix. I checked out `7.3.0-snapshot.2948` and tried it out. I'm getting some weird behavior when reordering items. Are you able to reproduce this? https://user-images.githubusercontent.com/16201149/232669274-edc86e92-12f7-41a0-b342-d8c80f0e92fc.MP4

@YuyouWu Actually I can't reproduce it
this is how it works in my case: 
https://user-images.githubusercontent.com/66782556/232724744-ecf026d6-60ac-4985-a317-69a4cf3971f7.mp4
can you please share a snippet code of your case?

YuyouWu:
This is my `renderItem`:
```  const renderItem = useCallback(
    ({ item, index: _index }: { item: Folder; index: number }) => {
      return (
        <View
          style={{
            backgroundColor: "black",
          }}
        >
          <Drawer
            rightItems={[
              {
                text: "Delete",
                background: "red",
                //TODO: Add animation for deleting folder
                onPress: () => {
                  deleteFolder(item.id).then(() => {
                    getFolderData().then((res) => {
                      setFolders(res);
                    });
                  });
                },
              },
            ]}
            style={{
              borderRadius: 5,
            }}
          >
            <TouchableHighlight
              onPress={() => {
                //Set currentFolderId to item.id
                setCurrentFolderId(item.id);
                navigation.navigate({
                  name: "TodosContainer",
                  params: {
                    folder: item,
                  },
                });
              }}
              style={{
                height: 62,
                backgroundColor: "#111111",
                borderRadius: 5,
              }}
            >
              <Text
                marginH-10
                color="white"
                style={{ marginTop: "auto", marginBottom: "auto" }}
              >
                {item.name}
              </Text>
            </TouchableHighlight>
          </Drawer>
        </View>
      );
    },
    []
  );
```
and this is my `SortableList` :
```
  return (
    <View useSafeArea>
      <SortableList
        data={folders}
        renderItem={renderItem}
        keyExtractor={keyExtractor}
        onOrderChange={onOrderChange}
        scale={1.02}
        itemMargins={{marginBottom: 10, marginLeft: 10, marginRight: 10}}
        style={{
          height: "100%",
          marginTop: 5,
        }}
      />
    </View>
  );
```

Weird thing is if I go into `node_modules\react-native-ui-lib\src\components\sortableList\SortableListItem.js` and save the file without changes, the list will behave as expected like you've shown. I'm not sure if this is my code or env issue. @lidord-wix The PR look good to me though. Feel free to merge if not reproducible in other devices.

YuyouWu:
@lidord-wix That worked! Thank you for the fix.

YuyouWu (approved):

lidord-wix:
> #2538

The elevation on Android works in my implementation.. and it doesn't work when changing the background to transparent.
Also, IMO it is more correct logically since the item is indeed smaller (and not at the same height with transparent margins) 
WDYT?

M-i-k-e-l:
@lidord-wix 
I did not find a better way so we'll probably go with this one.
But I'm not a fan of the prop name, maybe we should make it more general so we can add more things later if we want to?
I'm thinking about `itemProps`, we'll be able to add `height` to it as well.
WDYT?

lidord-wix:
> @lidord-wix I did not find a better way so we'll probably go with this one. But I'm not a fan of the prop name, maybe we should make it more general so we can add more things later if we want to? I'm thinking about `itemProps`, we'll be able to add `height` to it as well. WDYT?

Done :)

M-i-k-e-l (approved):

PR #2563, opened 2023-04-17
Add segmentStyle prop to SegmentedControl

lidord-wix:
## Description
SegmentedControl - add segmentStyle prop (and update example)
https://github.com/wix/react-native-ui-lib/discussions/2495

## Changelog
SegmentedControl - add segmentStyle prop

### DIFF
--- a/src/components/segmentedControl/index.tsx
+++ b/src/components/segmentedControl/index.tsx
@@ -169,7 +169,7 @@ const SegmentedControl = (props: SegmentedControlProps) => {
           selectedIndex={animatedSelectedIndex}
           activeColor={activeColor}
           inactiveColor={inactiveColor}
-          segmentStyle={segmentStyle}
+          style={segmentStyle}
           {...segments?.[index]}
           testID={testID}
         />
--- a/src/components/segmentedControl/index.tsx
+++ b/src/components/segmentedControl/index.tsx
@@ -13,7 +13,7 @@ import Reanimated, {
 import {Colors, BorderRadiuses, Spacings} from '../../style';
 import {Constants, asBaseComponent} from '../../commons/new';
 import View from '../view';
-import Segment, {SegmentedControlItemProps as SegmentProps} from './segment';
+import Segment, {SegmentedControlItemProps} from './segment';
 
 const BORDER_WIDTH = 1;
 const TIMING_CONFIG: WithTimingConfig = {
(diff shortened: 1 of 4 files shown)


src/components/segmentedControl/segment.tsx
```diff
@@ -49,6 +49,10 @@ export type SegmentProps = SegmentedControlItemProps & {
    * onLayout function.
    */
   onLayout?: (index: number, event: LayoutChangeEvent) => void;
+  /**
+   * Additional style for the segment.
+   */
+  segmentStyle?: StyleProp<ViewStyle>;
```
Inbal-Tish, line 55: Why it is not simply called 'style' as this is the Segment component?

Inbal-Tish, line 55: Why this is not called simply "style" as it is the Segment component? 
I see in the API file that there is a "style" prop but I don't see that it is used anywhere...

src/components/segmentedControl/index.tsx
```diff
@@ -164,6 +169,7 @@ const SegmentedControl = (props: SegmentedControlProps) => {
           selectedIndex={animatedSelectedIndex}
           activeColor={activeColor}
           inactiveColor={inactiveColor}
+          segmentStyle={segmentStyle}
```
Inbal-Tish, line 172: If it's an individual style for each segment, why it's not passed in the segment, meaning part of the `SegmentedControlItemProps` and not the `SegmentedControlProps`? It's confusing, or maybe its just the naming...

src/components/segmentedControl/segment.tsx
lidord-wix: Renamed to `style` inside `Segment`.
The `style` prop in the API file is for the whole component.

src/components/segmentedControl/index.tsx
lidord-wix: Maybe it is confusing because I actually wanted one style prop for all of them. 😅 
Maybe `segmentsStyle` in plural?

Inbal-Tish: Yes, I guess `segmentsStyle` is better...

lidord-wix: Done :)

src/components/segmentedControl/index.tsx
```diff
@@ -98,6 +102,7 @@ const SegmentedControl = (props: SegmentedControlProps) => {
     outlineColor = activeColor,
     outlineWidth = BORDER_WIDTH,
     throttleTime = 0,
+    segmentsStyle: segmentsStyleProp,
```
Inbal-Tish, line 105: Why do you need the rename here?

lidord-wix: There's another `segmentsStyle` in this file

Inbal-Tish: Ok...

Inbal-Tish (approved):

PR #2568, opened 2023-04-18
SortableList - add driver and test

M-i-k-e-l:
## Description
SortableList - add driver and test
WOAUILIB-3444

## Changelog
SortableList - add driver and test

### DIFF
--- a/src/components/sortableList/SortableListItem.driver.ts
+++ b/src/components/sortableList/SortableListItem.driver.ts
@@ -3,7 +3,7 @@ import {ComponentDriver} from '../../testkit/Component.driver';
 
 /**
  * Please run clear after each test
- * */
+ */
 export class SortableListItemDriver extends ComponentDriver {
   dragUp = async (indices: number) => {
     this.validateIndices(indices);
--- a/src/components/sortableList/SortableListItem.driver.ts
+++ b/src/components/sortableList/SortableListItem.driver.ts
@@ -1,0 +1,35 @@
+import _ from 'lodash';
+import {ComponentDriver} from '../../testkit/Component.driver';
+
+/**
+ * Please run clear after each test
+ * */
+export class SortableListItemDriver extends ComponentDriver {
+  dragUp = async (indices: number) => {
+    this.validateIndices(indices);
+    const data = _.times(indices, index => {
+      return {
+        translationY: -52 * (index + 1)
+      };
+    });
+
+    await this.uniDriver.selectorByTestId(this.testID).then(driver => driver.drag(data));
+  };
+
+  dragDown = async (indices: number) => {
+    this.validateIndices(indices);
+    const data = _.times(indices, index => {
+      return {
+        translationY: 52 * (index + 1)
+      };
+    });
+
+    await this.uniDriver.selectorByTestId(this.testID).then(driver => driver.drag(data));
+  };
+
+  private validateIndices = (indices: number) => {
+    if (indices <= 0 || !Number.isInteger(indices)) {
+      throw Error('indices must be a positive integer');
+    }
+  };
+}
--- a/src/components/sortableList/__tests__/index.spec.tsx
+++ b/src/components/sortableList/__tests__/index.spec.tsx
@@ -1,0 +1,85 @@
+import _ from 'lodash';
+import React, {useCallback} from 'react';
+import Text from '../../text';
+import View from '../../view';
+import SortableList from '../index';
+import {ComponentDriver, SortableListItemDriver} from '../../../testkit';
+
+const defaultProps = {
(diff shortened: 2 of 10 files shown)


src/components/sortableList/SortableListItem.driver.ts
```diff
@@ -0,0 +1,35 @@
+import _ from 'lodash';
+import {ComponentDriver} from '../../testkit/Component.driver';
+
+/**
+ * Please run clear after each test
+ * */
```
lidord-wix, line 6:
```suggestion
 */
```

src/components/sortableList/SortableListItem.driver.ts
```diff
@@ -0,0 +1,35 @@
...
+export class SortableListItemDriver extends ComponentDriver {
+  dragUp = async (indices: number) => {
+    this.validateIndices(indices);
+    const data = _.times(indices, index => {
+      return {
+        translationY: -52 * (index + 1)
```
lidord-wix, line 12: Will it always be 52? 
The user can pass a different item height, no?

src/components/sortableList/SortableListItem.driver.ts
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    await this.uniDriver.selectorByTestId(this.testID).then(driver => driver.drag(data));
+  };
+
+  dragDown = async (indices: number) => {
+    this.validateIndices(indices);
+    const data = _.times(indices, index => {
+      return {
+        translationY: 52 * (index + 1)
```
lidord-wix, line 23: Same here

src/components/sortableList/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <View flex>
+      <View flex useSafeArea>
+        <SortableList data={ITEMS} renderItem={renderItem} {...defaultProps} {...props}/>
+      </View>
+    </View>
+  );
+};
+
+const _TestCase = props => <TestCase {...props}/>;
```
lidord-wix, line 37: What do you need this const for? 
you can use the TestCase itself..

lidord-wix (commented):
Looks good overall, I wrote some small comments, and:
1. Why does the changelog "None"? 
2. Do we need to add some mocks in private or mobile-mocks? (will it potentially break something?) 
3. Consider creating a private version for the module to check that it answers their needs :)

src/components/sortableList/SortableListItem.driver.ts
M-i-k-e-l: It starts with 52 and then there's `onLayout` on the first item.
This does not happen in tests (at least for now).
It could change if we change the component (I rather have the tests fail than export a const from the code just for the tests).

src/components/sortableList/SortableListItem.driver.ts
M-i-k-e-l: Same

src/components/sortableList/__tests__/index.spec.tsx
M-i-k-e-l: I copied from Dialog's test, where it caused an error (I think something about hooks location); just tested and it looks like it's passing, so I removed from both.

M-i-k-e-l:
> Looks good overall, I wrote some small comments, and:
> 
> 1. Why does the changelog "None"?
> 2. Do we need to add some mocks in private or mobile-mocks? (will it potentially break something?)
> 3. Consider creating a private version for the module to check that it answers their needs :)

Almost forgot :)
1. I don't think this is something we should publicise ATM, our drivers are still in "incubator" state AFAIK.
2. It should not break anything, I am still working on private stuff but it should not block this IMO (this is not used AFAIK).
3. Yes, hopefully I'll get to that soon, but this is a step in the right direction at least.

lidord-wix:
> > Looks good overall, I wrote some small comments, and:
> > 
> > 1. Why does the changelog "None"?
> > 2. Do we need to add some mocks in private or mobile-mocks? (will it potentially break something?)
> > 3. Consider creating a private version for the module to check that it answers their needs :)
> 
> Almost forgot :)
> 
> 1. I don't think this is something we should publicise ATM, our drivers are still in "incubator" state AFAIK.
> 2. It should not break anything, I am still working on private stuff but it should not block this IMO (this is not used AFAIK).
> 3. Yes, hopefully I'll get to that soon, but this is a step in the right direction at least.

1. The changelog in the release notes really helps us to locate bugs during the release cycle, I'd add it even if it's not relevant to most of our users...
2. 👍🏽 
3. 👍🏽

lidord-wix (approved):

PR #2570, opened 2023-04-19
CharCounter now working with emojis (16-bit code)

adids1221:
## Description
CharCounter fix to count emojis, counting 16-bit code symbols.

## Changelog
CharCounter fix to count emojis, counting 16-bit code symbols.

### DIFF
--- a/src/incubator/TextField/CharCounter.tsx
+++ b/src/incubator/TextField/CharCounter.tsx
@@ -1,4 +1,4 @@
-import React, {useContext} from 'react';
+import React, {useContext, useMemo} from 'react';
 import {StyleSheet} from 'react-native';
 import _ from 'lodash';
 import Text from '../../components/text';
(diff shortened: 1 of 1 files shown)


src/incubator/TextField/CharCounter.tsx
```diff
@@ -7,13 +7,14 @@ import {CharCounterProps} from './types';
 
 const CharCounter = ({maxLength, charCounterStyle, testID}: CharCounterProps) => {
   const {value} = useContext(FieldContext);
+  const length = useMemo(() => (_.isUndefined(value) ? 0 : value.length), [value]);
```
ethanshar, line 10: Memoizing the length is a little overkill, also it's primitive value (number) so there no real reason to memoize it

ethanshar: Also consider simplifying the code by doing something like this `value?.length ?? 0`

ethanshar (changes requested):

ethanshar (approved):

PR #2583, opened 2023-05-08
Width and height props added to Icon, passed only in case of svg icon

adids1221:
## Description
Icon, new width, and height props will be used only if the Icon is an SVG image. 

## Changelog
Icon, new width, and height props will be used only if the Icon is an SVG image.


src/components/icon/index.tsx
M-i-k-e-l, line 1: Can you please link (in the PR description) the reason for this PR?

src/components/icon/index.tsx
```diff
@@ -27,6 +27,16 @@ export type IconProps = ImageProps &
      * whether the icon should flip horizontally on RTL
      */
     supportRTL?: boolean;
+    /**
+     * Width and height of the icon will be used only in case that the icon is svg image.
```
M-i-k-e-l, line 31: This will only be shown in the width docs, maybe write something specific for each one?

src/components/icon/index.tsx
M-i-k-e-l: Do we want this in the API file (I am not sure)?

src/components/icon/index.tsx
```diff
@@ -77,6 +87,8 @@ const Icon = forwardRef((props: Props, ref: any) => {
     return renderImage();
   }
 
+  console.log(`isSvg(source)L`, isSvg(source));
```
M-i-k-e-l, line 90: I assume this was added by mistake

src/components/icon/index.tsx
```diff
@@ -27,6 +27,15 @@ export type IconProps = Omit<ImageProps, 'source'> &
...
     supportRTL?: boolean;
+    /**
+     * 
+     * The icon width, width will be used only in case that the icon is svg image.
+     */
+    width?: string | number;
+    /**
+     * The icon height, height will be used only in case that the icon is svg image.
+     */
+    height?: string | number;
```
M-i-k-e-l, line 38: Why did you decide to add the types (which will only work for SVG) and not set width\height from size to the SVG image?

M-i-k-e-l (commented):
Sorry for taking this long...

src/components/icon/index.tsx
adids1221: Since RN deprecated the `width` and `height` props from `ImageProps` I thought we should add them to the component so the user could still pass them.
On the other hand, you are right we can take the size and pass it as width and height.
Should we go on the size solution ?

M-i-k-e-l: I think that if we can achieve the same result with less props it's better, especially if they are only applicable in specific situations.
It'll probably require a little more code, but I doubt it's a lot more code.

M-i-k-e-l (approved):
The code looks good, you can merge if you've tested it (on both platforms)

adids1221:
> The code looks good, you can merge if you've tested it (on both platforms)

Checked on both platforms 👍

PR #2598, opened 2023-05-17
Incubator.Slider - fix onValueChange throttle

Inbal-Tish:
## Description
Incubator.Slider - fix onValueChange throttle
ColorPicker - add 'migrate' prop to use new SliderImplementation

## Changelog
Incubator.Slider - fix onValueChange throttle
ColorPicker - add 'migrate' prop to use new SliderImplementation

### DIFF
--- a/src/components/colorPicker/ColorPickerDialog.tsx
+++ b/src/components/colorPicker/ColorPickerDialog.tsx
@@ -51,6 +51,10 @@ interface Props extends DialogProps {
     doneButton?: string,
     input?: string
   };
+  /** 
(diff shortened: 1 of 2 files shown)


M-i-k-e-l (approved):

M-i-k-e-l:
@Inbal-Tish 
Please avoid multiline in the changelog in the future since our release script does not handle it well

Inbal-Tish:
> @Inbal-Tish Please avoid multiline in the changelog in the future since our release script does not handle it well

Let's fix the script then

PR #2606, opened 2023-05-28
Fixed component typings

The-White-Fang:
- exported types from globalTypes.d.ts
- created a new file module.d.ts to declare modules like '.png'
- updated components/code to use exported types instead of gloabal types

## Description
Fixes the issue where components wouldn't annotate their props. 

## Changelog
*Add a quick message for our users about this change (include Component name, relevant props and general purpose of the PR)*
- added export in `globalTypes.d.ts` to export types from the file (according to [this issue](https://github.com/Microsoft/TypeScript/issues/5112) the contents of the `d.ts` files are never copied to build and nor are referenced)
- moved module declaration for `.png` files to `modules.d.ts`
- updated code to import the changed types` scope


ethanshar:
Fixed by another PR
Thank you for contributing

PR #2608, opened 2023-05-29
TextField and Switch/ forward id prop

AmitShwarts:
## Description
we need to be able to pass `id` prop to the TextField container.
currently `id` is typed but passed to the `Input` component, so we declared new prop `containerId` in order to not break anyone.

## Changelog
TextField - add `containerProps` prop
Switch - add `id` prop

### DIFF
--- a/src/components/switch/index.tsx
+++ b/src/components/switch/index.tsx
@@ -56,6 +56,7 @@ export type SwitchProps = {
   thumbStyle?: StyleProp<ViewStyle>;
   style?: StyleProp<ViewStyle>;
   testID?: string;
+  id?: string;
 }
 
 /**
--- a/src/incubator/TextField/index.tsx
+++ b/src/incubator/TextField/index.tsx
@@ -46,7 +46,7 @@ const TextField = (props: InternalTextFieldProps) => {
   const {
     modifiers,
     // General
-    containerId,
+    containerProps = {},
     fieldStyle: fieldStyleProp,
     dynamicFieldStyle,
     containerStyle,
@@ -124,7 +124,7 @@ const TextField = (props: InternalTextFieldProps) => {
 
   return (
     <FieldContext.Provider value={context}>
-      <View id={containerId} style={[margins, positionStyle, containerStyle, centeredContainerStyle]}>
+      <View {...containerProps} style={[margins, positionStyle, containerStyle, centeredContainerStyle]}>
         <Label
           label={label}
           labelColor={labelColor}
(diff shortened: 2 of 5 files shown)


src/incubator/TextField/index.tsx
```diff
@@ -123,7 +124,7 @@ const TextField = (props: InternalTextFieldProps) => {
 
   return (
     <FieldContext.Provider value={context}>
-      <View style={[margins, positionStyle, containerStyle, centeredContainerStyle]}>
+      <View id={containerId} style={[margins, positionStyle, containerStyle, centeredContainerStyle]}>
```
lidord-wix, line 127: I suggest changing it to `containerProps` to allow passing any prop the user wants to the container (of type of `ViewProps` without the style since we have the `containerStyle` prop)

src/incubator/TextField/types.ts
```diff
@@ -164,6 +164,10 @@ export type TextFieldProps = MarginModifiers &
   // FieldStateProps &
   ValidationMessageProps &
   Omit<CharCounterProps, 'maxLength' | 'testID'> & {
+    /**
+     * Pass id to the container
+     */
+    containerId?: string;
```
lidord-wix, line 170: See my other comment about the name :) 
Also, please put it after the `containerStyle` prop and also add it to the `textField.api.json` file (this file is for the docs site)

src/components/switch/index.tsx
```diff
@@ -56,6 +56,7 @@ export type SwitchProps = {
   thumbStyle?: StyleProp<ViewStyle>;
   style?: StyleProp<ViewStyle>;
   testID?: string;
+  id?: string;
```
lidord-wix, line 59: I guess it's a leftover, right?

lidord-wix: You can leave it here, please add it to the Switch's API file and I'll update the PR's changelog and name :)

lidord-wix (commented):
Last small comments :)

src/incubator/TextField/index.tsx
```diff
@@ -46,6 +46,7 @@ const TextField = (props: InternalTextFieldProps) => {
   const {
     modifiers,
     // General
+    containerProps = {},
```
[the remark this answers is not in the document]
AmitShwarts: because i'm spreading it and IMO this is better:
```
<View {...containerProps} ...>
```
than:
```
<View {...(containerProps ?? {})} ...>
```

WDYT?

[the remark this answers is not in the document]
lidord-wix: We can just keep it empty
```
const {
  ...
  contatinerProps,
  ...
  }  = ...

<View {...containerProps} ...>
```

lidord-wix (approved):

PR #2612, opened 2023-06-07
Fix how we handle style in our Svg image component for web

ethanshar:
## Description
Fix how we handle style in our Svg image component for web
There was a potential issue with how we merge the incoming style prop that can be either an object or an array 

## Changelog
SvgImage (web) - fix issue with style being passed as object


ethanshar:
@OrenZak 
Why were you cloning the style object that was passed to SvgImage component?

OrenZak:
> @OrenZak Why were you cloning the style object that was passed to SvgImage component?

not sure actually, but if you have a test component for that, it will be awsome.

OrenZak (approved):
🥇

PR #2613, opened 2023-06-07
Include TouchableOpacityProps in SwitchProps

ethanshar:
## Description
Switch props passed on to TouchableOpacity component but typing wise didn't support TouchableOpacity props

## Changelog
Switch - fix props types to include TouchableOpacity props


lidord-wix (approved):

lidord-wix:
@ethanshar merged. 
Why do we need to HF it? it's a 5-year-old code, no?

ethanshar:
> @ethanshar merged.
> Why do we need to HF it? it's a 5-year-old code, no?

Yes, but it was raised now because we started using this type in private.. 
But I guess we can wait with to next version anyway..

PR #2619, opened 2023-06-14
Infra/test kit add props

M-i-k-e-l:
## Description
Add props to the testkit - this improves the quality of the tests (already found some issues).
I'm not sure about the `TextField`'s `isPressable` being removed but we can add it later (it's wrong anyway since `onPress` does not exist).
Relates to WOAUILIB-2351

## Changelog
None

## Additional info
None

### DIFF
--- a/src/components/slider/index.tsx
+++ b/src/components/slider/index.tsx
@@ -259,7 +259,7 @@ export default class Slider extends PureComponent<SliderProps, State> {
       accessibilityLabel: 'Slider',
       accessible: true,
       accessibilityRole: 'adjustable' as AccessibilityRole,
-      accessibilityStates: disabled ? ['disabled'] : [],
+      accessibilityState: disabled ? {disabled} : undefined,
       accessibilityActions: [
         {name: 'increment', label: 'increment'},
         {name: 'decrement', label: 'decrement'}
--- a/src/components/slider/slider.driver.ts
+++ b/src/components/slider/slider.driver.ts
@@ -2,5 +2,5 @@ import {SliderProps} from './index';
 import {ComponentDriver} from '../../testkit';
 
 export class SliderDriver extends ComponentDriver<SliderProps> {
-  isDisabled = async () => (await this.getElementProps()).accessibilityStates[0] === 'disabled';
+  isDisabled = async () => (await this.getElementProps()).accessibilityState?.disabled === true;
 }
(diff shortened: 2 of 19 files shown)


src/testkit/index.ts
```diff
@@ -11,3 +11,4 @@ export {HintDriver} from '../components/hint/Hint.driver';
 export {RadioButtonDriver} from '../components/radioButton/RadioButton.driver';
 export {RadioGroupDriver} from '../components/radioGroup/RadioGroup.driver';
 export {SortableListItemDriver} from '../components/sortableList/SortableListItem.driver';
+export {SliderDriver} from '../components/slider/slider.driver';
```
Inbal-Tish, line 14: better use the incubator slider driver. The old Slider will be replaced on the next major version

src/components/slider/slider.driver.ts
```diff
@@ -1,5 +1,6 @@
+import {SliderProps} from './index';
 import {ComponentDriver} from '../../testkit';
 
-export class SliderDriver extends ComponentDriver {
-  isDisabled = async () => (await this.getElementProps()).accessibilityStates[0] === 'disabled';
+export class SliderDriver extends ComponentDriver<SliderProps> {
+  isDisabled = async () => (await this.getElementProps()).accessibilityState?.disabled === true;
```
Inbal-Tish, line 5: Did you check if it works? I remember the array did work...

src/components/slider/index.tsx
```diff
@@ -259,7 +259,7 @@ export default class Slider extends PureComponent<SliderProps, State> {
       accessibilityLabel: 'Slider',
       accessible: true,
       accessibilityRole: 'adjustable' as AccessibilityRole,
-      accessibilityStates: disabled ? ['disabled'] : [],
+      accessibilityState: disabled ? {disabled} : undefined,
```
Inbal-Tish, line 262: Did you check if it works? I remember the array did work...

src/testkit/index.ts
M-i-k-e-l: There is no driver for the incubator slider, or did I miss it?
I might write it in the next phase and replace this

src/components/slider/index.tsx
M-i-k-e-l: Yes

src/testkit/index.ts
Inbal-Tish: Right, my bad

Inbal-Tish (approved):

PR #2630, opened 2023-06-21
Fixed TabBar scroll when there is selected index

adids1221:
## Description
TabBar selectedIndex changed to the currentPage instead of initialIndex.
When the items sent to the TabController change the selectedIndex will stay on the current page.

## Changelog
TabBar stays on the selected index when items change.


M-i-k-e-l (approved):
Approved, please add the ticket number (not link) to the `Additional info` section

PR #2638, opened 2023-06-27
ColorPicker - add resultInHexString

M-i-k-e-l:
## Description
Add an option to return the full hex string:
```
Should the result be returned only with 6 characters (#FFFFFF and never #FFF)
```
Can't say I like the name too much, I would love a suggestion :)

## Changelog
ColorPalette - add resultInHexString
ColorPicker - add resultInHexString

## Additional info
WOAUILIB-3628


Inbal-Tish:
@M-i-k-e-l I'm not sure I understand the need to return a 6-digit hex... If the user wants the return value in 6-digit-hex can't he just do `Colors.getHexString(value)` on the returned value? Why does it have to be in the component? What am I missing?

Inbal-Tish:
@M-i-k-e-l I would consider passing the 6-digit hex in the onValueChange options object instead

PR #2641, opened 2023-06-27
Feature: allow setting the placeholder scale

TNAJanssen:
## Description
I want to override the scale of the floating placeholder

## Changelog
Add floatingPlaceholderScale to TextField

## Additional info
I have not tested it yet.

### DIFF
--- a/src/components/textField/index.tsx
+++ b/src/components/textField/index.tsx
@@ -444,7 +444,7 @@ export default class TextField extends BaseInput {
                   {
                     scale: floatingPlaceholderState.interpolate({
                       inputRange: [0, 1],
-                      outputRange: [1, FLOATING_PLACEHOLDER_SCALE]
+                      outputRange: [1, floatingPlaceholderScale]
                     })
                   },
                   {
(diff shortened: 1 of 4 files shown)


src/components/textField/index.tsx
```diff
@@ -44,7 +44,6 @@ const CHAR_COUNTER_COLOR_BY_STATE = {
 const LABEL_TYPOGRAPHY = Typography.text80;
 const ICON_SIZE = 24;
 const ICON_LEFT_PADDING = 6;
-const FLOATING_PLACEHOLDER_SCALE = 0.875;
```
Inbal-Tish, line 47: You can leave this const and use it in the defaultProps

src/incubator/TextField/FloatingPlaceholder.tsx
```diff
@@ -9,15 +9,14 @@ import View from '../../components/view';
 import Text from '../../components/text';
 import FieldContext from './FieldContext';
 
-const FLOATING_PLACEHOLDER_SCALE = 0.875;
```
Inbal-Tish, line 12: You can leave this const and use it in the defaultProps

Inbal-Tish:
@TNAJanssen Hi. When changing the scale it not only changes the font size, but also the location of the floatingPlaceholder. Please try this change and add a video here.

Issue #2428, opened 2023-01-19
Adding Picker.Items after initial render fails to retrieve labels for selected items


siteassist-benjamin:
### Related to

Unfortunately there is no mention of this online, I tried very hard to find it.

### Steps to reproduce

```
const [templates, setTemplates] = useState([])

API.getTemplates().then(res => setTemplates(res))

return (
      <Picker
        value={value}
        getLabel={(item) => {
          // console.log("templates", templates);
          const name = templates.find((template) => template.id === item)?.name;
          return name;
        }}
        templates={templates}
      >
        {templates?.map(({ id, name }) => {
          return <PickerItem key={id} value={id} label={name} />;
        })}
      </Picker>
```

#### Expected behavior

Expected behavior would be for the input storing the currently selected value to show the selected label

#### Actual behavior

When an item is selected from the modal picker list - which has all the correct labels, the input field showing the selected item after the modal is closed is blank

### Screenshots/Video

![Screenshot 2023-01-19 at 19 48 54](https://user-images.githubusercontent.com/102516120/213533406-3c187c29-978e-4356-a3bf-92ea42bfbb78.png)

### Environment

- React Native version: 0.69.7
- React Native UI Lib version: 6.21.0 

### Affected platforms

- [x ] Android
- [x ] iOS
- [ n/a] Web

oottoohh:
same to me, any solution for this ?

illank86:
it works on iphone simulator? but on android device, its not. any solution for this?

brandon-hass:
It looks like Picker items get set as state only on initial creation: [src/components/picker/index.tsx](https://github.com/wix/react-native-ui-lib/blob/master/src/components/picker/index.tsx#L98). In our app, we've been able to work around this by updating the key prop on the Picker when the items change to force a state reset.

jiteshgolecha:
Any updates?

jiteshgolecha:
I have found a workaround, add renderPicker to show the label.

`<Picker 
                    topBarProps={{ useSafeArea: true }}
                    showSearch={true} 
                    value={selectedProduct}
                    placeholder="Products" 
                    onSearchChange={query => fetchProducts(query)} 
                    floatingPlaceholder
                    fieldStyle={styles.withUnderline2} 
                    renderPicker={selectedItem => { 
                        console.log("selectedItem", selectedItem); 
                        var itm = _.find(productList, {id: selectedItem})
                        console.log("itm", itm); 
                        return (
[...]

PR #2646, opened 2023-06-28
update picker items on children change

trofima:
## Description
fix for https://github.com/wix/react-native-ui-lib/issues/2428

### DIFF
--- a/src/components/picker/index.tsx
+++ b/src/components/picker/index.tsx
@@ -96,13 +96,13 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
   const {preset} = others;
 
   const [selectedItemPosition, setSelectedItemPosition] = useState(0);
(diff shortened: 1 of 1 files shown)


src/components/picker/index.tsx
```diff
@@ -96,13 +96,13 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
+  const [items, setItems] = useState(pickerItems);
 
   useEffect(() => {
-    if (propItems) {
-      setItems(propItems);
-    }
-  }, [propItems]);
+    setItems(pickerItems);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [propItems, children]);
```
ethanshar, line 105: I think there's a possibility that `children` always change and therefor will always trigger setItems

ethanshar: @adimorwix 
Did you get a chance to test it?

Issue #2557, opened 2023-04-15
No cancel button on Picker component


castusoft:
Picker's `useWheelPicker` modal is missing cancelLabel button even if I set it. Plus no matter how I apply styles, it is not getting it.

```
<Picker
            label="Trip type"
            useWheelPicker
            value={tripType}
            onChange={nativePickerValue => setTripType(nativePickerValue)}
            trailingAccessory={<ReloadSvg width={24} height={24} fill='black'/>}
            containerStyle={{borderWidth: 1, borderColor: Colors.grey50, padding: 3}}
            topBarProps={{
              doneLabel: 'Ok',
              cancelLabel: 'NO',
              title: 'Trip type',
              useSafeArea: true,
              includeStatusBar: true,
              titleStyle: {color: 'red'},
              containerStyle: {backgroundColor: 'red'}
          }}
          >
            {_.map(options, option => (
              <Picker.Item key={option.value} value={option.value} label={option.label} disabled={option.disabled}/>
            ))}
          </Picker>
```

### Related to

- [ x ] Components
- [ ] Demo
- [ ] Docs
- [ ] Typings

### Screenshots/Video
![IMG_5B7E5B224ACF-1](https://user-images.githubusercontent.com/60087368/232196199-cab3d465-7496-41b1-bfb3-742481d4da15.jpeg)

### Environment

- React Native version: 0.71.6
- React Native UI Lib version: ^7.2.4

### Affected platforms

- [ ] Android
- [ x ] iOS
- [ ] Web

kilimchoi:
is there a way to style the OK button? it's shifted to the right too much.

PR #2648, opened 2023-06-30
Wheel picker custom cancel button and title and container custom styling

trofima:
## Description
Wheel picker custom cancel button and title and container custom styling
Fixes the https://github.com/wix/react-native-ui-lib/issues/2557

Still there is no support of custom cancel/done buttons but it was not requested in the issue anyway


lidord-wix (commented):
Looks good! please see my small comment :)

PR #2651, opened 2023-07-02
Feat/color picker add full color to result

M-i-k-e-l:
## Description
Add hexString to results from ColorSwatch, ColorPalette and ColorPicker
Please let me know if you have a better name than ColorExtraData

## Changelog
ColorSwatch - Add hexString to onPress result
ColorPalette - Add hexString to onPress result
ColorPicker - Add hexString to onValueChange result

## Additional info
WOAUILIB-3628

### DIFF
--- a/src/components/colorPalette/ColorPalette.api.json
+++ b/src/components/colorPalette/ColorPalette.api.json
@@ -31,7 +31,7 @@
     },
     {
       "name": "onValueChange",
-      "type": "(value: string, options: object) => void",
+      "type": "(color: string, extraData: ColorExtraData) => void",
       "description": "Invoked once when value changes by selecting one of the swatches in the palette"
     },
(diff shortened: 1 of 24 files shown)


src/components/colorPalette/ColorPalette.api.json
```diff
@@ -31,7 +31,7 @@
     },
     {
       "name": "onValueChange",
-      "type": "(value: string, options: object) => void",
+      "type": "(color: string, extraData: ColorExtraData) => void",
```
Inbal-Tish, line 34: Since the props is named 'onValueChange' I'd leave it as 'value'. Also since the user probably already extracting 'value' here

Inbal-Tish (approved):

Inbal-Tish:
@M-i-k-e-l Approved. Looks great! See my comment regarding naming

PR #2655, opened 2023-07-04
testkit - new iteration

M-i-k-e-l:
## Description
Some improvements  from previous implementation:
1. Much simpler.
2. Using `renderTree` - allows us to use `rerender`; should make it easier to integrate with internal libs.
3. Better encapsulation while keeping usability.
      a. Only real base methods should be in the base driver (`Component.driver`) and others should be separated in their own hooks (`usePressableDriver` & `useDraggableDriver`). 
      b. Multiple wrapping:

```
usePressableDriver<ComponentProps, DraggableDriverResult<ComponentProps>>(useDraggableDriver<ComponentProps>(useComponentDriver(props)));
```

Align `GestureDetectorMock` to the internal implementation (for future integration).
Adding a couple of new drivers based on the new infra: `Text.driver.new.ts` & `SortableListItem.driver.new.ts`

I did remove (did not import from the old version) a lot of error handling, not sure if it'll have impact but I rather add them only if we see it's needed.

## Changelog
None

## Additional info
WOAUILIB-2351


src/components/sortableList/SortableListItem.driver.new.ts
```diff
@@ -0,0 +1,38 @@
...
+import {useDraggableDriver} from '../../testkit/new/useDraggable.driver';
+
+export const SortableListItemDriver = (props: ComponentProps) => {
+  const driver = useDraggableDriver<SortableListItemProps>(useComponentDriver(props));
+
+  const dragUp = async (indices: number) => {
+    validateIndices(indices);
+    const data = _.times(indices, index => {
+      return {
+        translationY: -52 * (index + 1)
```
Inbal-Tish, line 13: Isn't the 52 should be a variable defined in the SortableListItem component props?

Inbal-Tish:
@M-i-k-e-l I love it! So readable and clean and easy to use. Great work!

src/components/sortableList/SortableListItem.driver.new.ts
M-i-k-e-l: Fixed

Inbal-Tish (approved):

adids1221 (approved):
Great Work!

M-i-k-e-l:
> Looks really good! one question - when creating a component driver, is it possible to use two types of drivers? for example `usePressableDriver` and `useDraggableDriver`?

Yes, it's in the description (I did not test on a real component but it should work)

Issue #2457, opened 2023-02-10
[Text, IOS]String with newline breaks the alignment


mk0116:
## Description

[Text]String with newline breaks the alignment.

When I add new line character('\n') in the text and nest a Text, it breaks the alignment.
The first line should be aligned center but left.
This only happens on iOS.
It works fine on Android.

### Related to

- [v] Components
- [ ] Demo
- [ ] Docs
- [ ] Typings

### Code snippet

```
<Text
    textColor
    center>
    <Text bold>test</Text>
    {'\nrest string'}
</Text>
```

### Screenshots/Video

<img width="255" alt="스크린샷 2023-02-10 오전 9 58 44" src="https://user-images.githubusercontent.com/11452913/217973363-7c138b8c-7b56-4ecf-91c8-eac051762389.png">

### Environment

- React Native version: 0.70.3
- React Native UI Lib version: 6.29.0

### Affected platforms

- [ ] Android
- [v] iOS
- [ ] Web

mk0116:
This has not been fixed yet.

Inbal-Tish:
@mk0116 Hi. What result do you expect to get?

mk0116:
Hello,

I expect all elements within the parent Text component to be aligned
centrally.

However, it is evident that the first line is aligning to the left instead
of the center.

Interestingly, when I use the standard Text component, the alignment is
correctly set to the center.

Additionally, I have been unable to figure out a method to align the nested
text components with react-native-ui-lib

Thank you.




2023년 7월 10일 (월) 오후 3:59, Inbal Tish ***@***.***>님이 작성:

mk0116:
Thank you so much!
Because of that I stopped my migration. That will help me a lot.


2023년 7월 10일 (월) 오후 9:34, Inbal Tish ***@***.***>님이 작성:

Inbal-Tish:
@mk0116 Hi. Issue solved on version 7.6.0. Thank you for reporting!

mk0116:
Thank you!!!

2023년 7월 24일 (월) 오후 5:02, Inbal Tish ***@***.***>님이 작성:

PR #2665, opened 2023-07-10
Text - remove default text alignment

Inbal-Tish:
## Description
⚠️ Text - remove default text alignment
I've checked with 'animated' prop and don't see any effect as the RN default is 'auto'

## Changelog
⚠️ Text - remove default text alignment

## Additional info
solves #2457

### DIFF
--- a/src/components/text/index.tsx
+++ b/src/components/text/index.tsx
@@ -157,7 +157,6 @@ class Text extends PureComponent<PropsTypes> {
 const styles = StyleSheet.create({
   container: {
     backgroundColor: 'transparent',
-    textAlign: 'left',
     color: Colors.$textDefault
   },
   centered: {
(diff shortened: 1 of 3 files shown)


src/components/text/index.tsx
```diff
@@ -157,7 +157,6 @@ class Text extends PureComponent<PropsTypes> {
 const styles = StyleSheet.create({
   container: {
     backgroundColor: 'transparent',
-    textAlign: 'left',
```
ethanshar, line 160: Did you check RTL? 
Also, did you check behavior on Android, it tends to be different sometimes..

ethanshar (approved):
Approved. 
But see my comment, also notice it will change private uilib snapshot, so be ready to fix these after it's merged

src/components/text/index.tsx
Inbal-Tish: Actually Android is acting fine with and without this line. Yep, problem in RTL - good catch!

Inbal-Tish:
This solution breaks text alignment in RTL (only works on Android if the string language is actually on RTL language).

PR #2667, opened 2023-07-11
feat: [MAW-257] svgImage support tintColor on web

OrenZak:
## Description

Add web support for `tintColor` on svg tag data

## Changelog

Add web support for `tintColor` on svg tag data

### DIFF
--- a/src/components/svgImage/index.web.tsx
+++ b/src/components/svgImage/index.web.tsx
@@ -1,4 +1,5 @@
 import React, {useState} from 'react';
+import {StyleSheet} from 'react-native';
 import Image from '../image';
 import {isSvg, isSvgUri, isBase64ImageContent} from '../../utils/imageUtils';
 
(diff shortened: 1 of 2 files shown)


src/components/svgImage/index.web.tsx
```diff
@@ -31,7 +30,7 @@ function SvgImage(props: SvgImageProps) {
   };
 
   if (isSvgUri(data)) {
-    return <img {...others} src={data.uri} style={styleObj}/>;
+    return <img {...others} src={data.uri} style={StyleSheet.flatten(style)}/>;
```
ethanshar, line 33: I think that the flatter is not necessary here.. only for the postCSS thing

src/components/svgImage/index.web.tsx
```diff
@@ -44,13 +43,13 @@ function SvgImage(props: SvgImageProps) {
         />
       );
     }
-    return <img {...others} src={data} style={styleObj}/>;
+    return <img {...others} src={data} style={StyleSheet.flatten(style)}/>;
```
ethanshar, line 46: Also here, not sure Flatten is necessary here...

ethanshar (approved):

src/components/svgImage/index.web.tsx
OrenZak: we need that, its not supported on React

tested in: 
https://playcode.io/react
```
import React from 'react';

export function App(props) {
  return (
    <div className='App'>
      <img src="https://reactjs.org/logo-og.png" alt="react logo" style={[{ width: '400px'}]}/>
    </div>
  );
}
```

PR #2668, opened 2023-07-11
Demo - adding RTL screen

Inbal-Tish:
## Description
Demo - adding RTL screen

## Changelog
Demo - adding RTL screen - Hebrew language

### DIFF
--- a/demo/src/index.js
+++ b/demo/src/index.js
@@ -210,6 +210,9 @@ module.exports = {
       get SpacingsScreen() {
         return require('./screens/foundationScreens/SpacingsScreen').default;
       },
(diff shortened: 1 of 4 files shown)


adids1221:
We have a similar screen in the private repo, I think we should use this screen or move the private screen to the public repo instead of having two RTL screens.

Also, I found this issue if the simulator language is not RTL language and you toggle force RTL in the settings screen the Slider is flaky and not working as it should be.
Simulator Screen Shot - iPhone 14 - 2023-07-11 at 16 10 41

adids1221 (approved):
Approved

PR #2670, opened 2023-07-12
Text - default alignment

Inbal-Tish:
## Description
Text - remove default textAlign style to fix iOS child Text not inheriting from its parent.
⚠️ Note: on Android, the Text will not align to the right on RTL if the actual language is not RTL language (back to its default behavior in RN)

## Changelog
Text - remove default textAlign style to fix iOS child Text not inheriting from its parent.

## Additional info
Solves #2457


ethanshar (approved):

ethanshar:
@Inbal-Tish 
Approved. 
You have conflicts. 

Notice that once this is merged, it will break private snapshot tests, so I suggest to prepare a PR that fix them

PR #2671, opened 2023-07-16
NumberInput - support focus

M-i-k-e-l:
## Description
NumberInput - support focus

## Changelog
None

## Additional info
WOAUILIB-3406

### DIFF
--- a/demo/src/screens/componentScreens/NumberInputScreen.tsx
+++ b/demo/src/screens/componentScreens/NumberInputScreen.tsx
@@ -161,8 +161,8 @@ const NumberInputScreen = () => {
   }, [label, textStyle, validate, validationMessage]);
 
   return (
-    <TouchableWithoutFeedback onPress={RNKeyboard.dismiss}>
-      <View centerH>
+    <TouchableWithoutFeedback style={{flex: 1}} onPress={RNKeyboard.dismiss}>
+      <View centerH flex>
         <Text text40 margin-s10>
           Number Input
         </Text>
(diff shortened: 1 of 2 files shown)


lidord-wix (commented):
Looks good. 
I'd add the underline color and the autoFocus here by default since it's part of the component design and not a "private" UX.

M-i-k-e-l:
> Looks good. I'd add the underline color and the autoFocus here by default since it's part of the component design and not a "private" UX.

Actually the focused color is part of private (you can test the public vs private TextField - they are not the same).
Regarding the `autoFocus` - the design does not require it by default; maybe it should, but since our `MainInput` does not do it I assume it's the same behavior.

lidord-wix:
> > Looks good. I'd add the underline color and the autoFocus here by default since it's part of the component design and not a "private" UX.
> 
> Actually the focused color is part of private (you can test the public vs private TextField - they are not the same). Regarding the `autoFocus` - the design does not require it by default; maybe it should, but since our `MainInput` does not do it I assume it's the same behavior.

About the underline - in the textField you have another indication that you're focused (the caret)
Agree about the autoFocus

M-i-k-e-l:
> > > Looks good. I'd add the underline color and the autoFocus here by default since it's part of the component design and not a "private" UX.
> > 
> > 
> > Actually the focused color is part of private (you can test the public vs private TextField - they are not the same). Regarding the `autoFocus` - the design does not require it by default; maybe it should, but since our `MainInput` does not do it I assume it's the same behavior.
> 
> About the underline - in the textField you have another indication that you're focused (the caret) Agree about the autoFocus

Good point, fixed

lidord-wix (approved):

PR #2672, opened 2023-07-16
Checkbox - add validation state invoked by validate()

Inbal-Tish:
## Description
Checkbox - add validation state invoked by validate() - see example in screen

## Changelog
Checkbox - add validation state invoked by validate()

## Additional info
#3662


demo/src/screens/componentScreens/CheckboxScreen.tsx
```diff
@@ -67,6 +70,32 @@ export default class CheckboxScreen extends Component {
...
+            <Checkbox
+              disabled
+              ref={this.checkbox1}
+              value={this.state.value7}
+              onValueChange={value7 => this.setState({value7})}
+              iconColor={Colors.green10}
+              label={'This is disabled checkbox'}
+            />
+            <Button disabled size={'small'} label={'Validate'} onPress={() => this.checkbox1.current?.validate()}/>
+          </View>
```
M-i-k-e-l, line 97: What is this (second) example testing?

src/components/checkbox/index.tsx
```diff
@@ -197,13 +209,21 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     }
   };
 
+  getErrorColor = () => !this.state.isValid && Colors.$textDanger;
```
M-i-k-e-l, line 212: There are two colors for this state: `$outlineDanger` and `$textDangerLight`.

src/components/checkbox/index.tsx
```diff
@@ -139,6 +147,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
       ]
     };
   }
+  validationState = false;
```
M-i-k-e-l, line 150: Not sure why `validationState` is needed, I think this check can cause issues with `Form` (we might need to add an `isValid` function for it, which might get buggy from this).

src/components/checkbox/index.tsx
```diff
@@ -122,7 +129,8 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     super(props);
 
     this.state = {
-      isChecked: new Animated.Value(this.props.value ? 1 : 0)
+      isChecked: new Animated.Value(this.props.value ? 1 : 0),
+      isValid: true
```
M-i-k-e-l, line 133: If we want to support things like `validateOnStart` as we do elsewhere, this should be conditional on initial state.

src/components/checkbox/index.tsx
```diff
@@ -197,13 +209,21 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     }
   };
 
+  getErrorColor = () => !this.state.isValid && Colors.$textDanger;
+
   getBorderStyle() {
-    const borderColor = {borderColor: this.getColor()};
+    const borderColor = {borderColor: this.getErrorColor() || this.getColor()};
```
M-i-k-e-l, line 215: I find this less readable than:
```
this.state.isValid ? this.getColor() : Colors.$outlineDanger
```

src/components/checkbox/index.tsx
```diff
@@ -197,13 +209,21 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
     const borderStyle = [this.styles.container, {borderWidth: DEFAULT_BORDER_WIDTH}, borderColor];
 
     return borderStyle;
   }
 
+  getLabelStyle = () => {
+    return {
+      color: this.props.disabled ? Colors.$textDisabled : this.getErrorColor() || Colors.$textDefault
```
M-i-k-e-l, line 223: Same as above, I don't like the `||` solution, it is less readable

src/components/checkbox/index.tsx
```diff
@@ -241,14 +261,21 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     return label ? (
       <View row centerV style={containerStyle}>
         {this.renderCheckbox()}
-        <Text flexS style={[this.styles.checkboxLabel, labelStyle]} recorderTag={'unmask'} {...labelProps} onPress={this.onPress}>
+        <Text flexS style={[this.styles.checkboxLabel, this.getLabelStyle(), labelStyle]} recorderTag={'unmask'} {...labelProps} onPress={this.onPress}>
```
M-i-k-e-l, line 264: I'd move the `this.styles.checkboxLabel` (with the array) to the function and memoize it.

demo/src/screens/componentScreens/CheckboxScreen.tsx
Inbal-Tish: That there is no validation on a disabled checkbox

src/components/checkbox/index.tsx
Inbal-Tish: fixed

src/components/checkbox/index.tsx
Inbal-Tish: `validationState` is a private flag that makes sure the validation check will only be performed after the user invoked the validate() method. The Form should only perform the validate().

src/components/checkbox/index.tsx
Inbal-Tish: `validateOnStart` was not a requirement

src/components/checkbox/index.tsx
Inbal-Tish: Non of the styles on this component are memorized. We should take it to a different task

src/components/checkbox/index.tsx
```diff
@@ -272,10 +272,15 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
     this.validationState = true;
+    // Validation: value must be true (checked)
     if (!this.props.disabled && !this.props.value) {
-      this.setState({isValid: false});
+      this.setState({valid: false});
     }
   }
+
+  isValid() {
+    return this.state.valid;
```
M-i-k-e-l, line 282: Can't you keep the state variable as `isValid` (it is prettier), or do you not want to because you find it confusing?

src/components/checkbox/index.tsx
M-i-k-e-l: I don't see the value that it adds, is there an example it is useful in\for?

src/components/checkbox/index.tsx
Inbal-Tish: I thought it was confusing...

M-i-k-e-l:
If I add the following code and press it before doing anything I'll receive a `valid` state which is incorrect.
```
<Button
  size={'small'}
  label={`Is valid: ${this.state.isValid}`}
  onPress={() => this.setState({isValid: this.checkbox.current?.isValid()})}
/>
```

src/components/checkbox/index.tsx
M-i-k-e-l: It looks to me like it is serving no purpose, and if this is the case we should remove it.

Inbal-Tish:
> If I add the following code and press it before doing anything I'll receive a `valid` state which is incorrect.
> 
> ```
> <Button
>   size={'small'}
>   label={`Is valid: ${this.state.isValid}`}
>   onPress={() => this.setState({isValid: this.checkbox.current?.isValid()})}
> />
> ```

Why? The component is valid by default until it moves to the validation state, which means the user invoked the validate() method and it passed/failed the validation check. If the user doesn't invoke the validate() I will not show error colors if the component is unchecked and the component will keep acting as valid (as it always did)

src/components/checkbox/index.tsx
```diff
@@ -122,7 +140,10 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     super(props);
 
     this.state = {
-      isChecked: new Animated.Value(this.props.value ? 1 : 0)
+      isChecked: new Animated.Value(this.props.value ? 1 : 0),
+      showError: false,
+      isValid: !(props.required && !props.value),
+      validationState: false
```
M-i-k-e-l, line 146: Look at moving this out of state (as before)

src/components/checkbox/__tests__/index.spec.tsx
```diff
@@ -24,57 +27,253 @@ describe('Checkbox renderer test', () => {
...
+        isValid = checkboxRef.current?.isValid();
+
+        expect(isValid).toBe(false);
+      });
+
+      it('should be valid after validate when initial value is true', async () => {
+        const props = {value: true, required: true};
+        const component = <TestCase {...props}/>;
+        //eslint-disable-next-line
+        const driver = new CheckboxDriver({component, testID});
```
M-i-k-e-l, line 217: Can we removing these ignores?

src/components/checkbox/__tests__/index.spec.tsx
```diff
@@ -24,57 +27,253 @@ describe('Checkbox renderer test', () => {
...
+      const component = <TestCase {...props}/>;
+      const driver = new CheckboxDriver({component, testID});
+  
+      await driver.press();
+  
+      expect(onValueChange).toHaveBeenCalledTimes(1);
+      expect(onValueChange).toHaveBeenCalledWith(true);
+    });
+  
+    it.each`
```
M-i-k-e-l, line 42: Please prettify the file

src/components/checkbox/__tests__/index.spec.tsx
```diff
@@ -24,57 +27,253 @@ describe('Checkbox renderer test', () => {
...
+      it('should not been called', async () => {
+        const props = {onChangeValidity};
+        const component = <TestCase {...props}/>;
+        const driver = new CheckboxDriver({component, testID});
+        await driver.press();
 
-    await driver.press();
+        expect(onChangeValidity).not.toHaveBeenCalled();
+      });
+      it('should not been called with required prop', async () => {
```
M-i-k-e-l, line 98: This does not make sense IMO, maybe it comes from here:
```
    if (this.state.validationState && prevState.isValid !== this.state.isValid) {

```

src/components/checkbox/index.tsx
```diff
@@ -140,11 +161,15 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
 
-  componentDidUpdate(prevProps: CheckboxProps) {
-    const {value} = this.props;
+  componentDidUpdate(prevProps: CheckboxProps, prevState: CheckboxState) {
+    const {value, onChangeValidity} = this.props;
     if (prevProps.value !== value) {
       this.animateCheckbox(value);
     }
+    
+    if (this.state.validationState && prevState.isValid !== this.state.isValid) {
```
M-i-k-e-l, line 170: Why do we not notify the user before `validate` was pressed?
Does it affect the `Form`?

src/components/checkbox/index.tsx
M-i-k-e-l: I saw you reverted it, did it fail at something?

src/components/checkbox/__tests__/index.spec.tsx
Inbal-Tish: It doesn't let me remove the const -> Error: Do not use 'new' for side effects.eslint[no-new](https://eslint.org/docs/rules/no-new)

M-i-k-e-l: Maybe ignore that one, it is more readable IMO

src/components/checkbox/__tests__/index.spec.tsx
Inbal-Tish: As long as we don't invoke `checkboxRef.current?.validate()`, even if 'required' is passed, 'onChangeValidity' should not be invoked

src/components/checkbox/index.tsx
```diff
@@ -122,7 +140,10 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     super(props);
 
     this.state = {
-      isChecked: new Animated.Value(this.props.value ? 1 : 0)
+      isChecked: new Animated.Value(this.props.value ? 1 : 0),
+      showError: false,
+      isValid: !(props.required && !props.value),
```
Inbal-Tish, line 145: Changing to `isValid: !(props.required && !props.value)` fails the tests

Inbal-Tish: I think I know why. If it is required and the value is false it isValid should be true since the user didn't call 'validate' yet and till it does, isValid is true

src/components/checkbox/index.tsx
M-i-k-e-l: I think this should be treated as a different mechanism, the user can `validate` or they can track the validity state (like we do in the `FormScreen`)

src/components/checkbox/index.tsx
M-i-k-e-l: It should be `!props.required || props.value` and not `!props.required || !props.value`

Inbal-Tish: Yes, that's what I did... sorry. Falling the tests.

src/components/checkbox/index.tsx
Inbal-Tish: So you're saying once the user passed 'required' the onValidityChange' should be called, no matter if the user invoked 'validate' or not. Right?

Inbal-Tish: Changed to call if required

src/components/checkbox/index.tsx
Inbal-Tish: Test failed since this condition can returns value and value could be undefined. Fixed it by setting default value to false

src/components/checkbox/__tests__/index.spec.tsx
```diff
@@ -24,57 +28,263 @@ describe('Checkbox renderer test', () => {
...
+        expect(onChangeValidity).toHaveBeenCalledWith(true);
+      });
+
+      it('should not been called after invoking validate()', async () => {
+        const props = {required: true, onChangeValidity};
+        const component = <TestCase {...props}/>;
+        //eslint-disable-next-line
+        new CheckboxDriver({component, testID});
+
+        checkboxRef.current?.validate();
```
M-i-k-e-l, line 124: If you add a `?` you don't know that `validate` was actually called (what if `current` is `undefined`?

src/components/checkbox/index.tsx
```diff
@@ -86,8 +94,17 @@ export interface CheckboxProps extends TouchableOpacityProps {
   indeterminate?: boolean;
 }
 
+interface CheckboxMethods {
+  validate: () => void;
```
M-i-k-e-l, line 98: I see in `TextField` there's a `onValidationFailed` method that **is** called when the validation fails, WDYT about adding it? Unlike `onChangeValidity` it is called if `validate` fails.
It can be another PR...

src/components/checkbox/index.tsx
```diff
@@ -172,10 +196,19 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
   onPress = () => {
-    const {disabled} = this.props;
-
+    const {disabled, value, required, onValueChange, onChangeValidity} = this.props;
+    
     if (!disabled) {
-      this.props.onValueChange?.(!this.props.value);
+      const newValue = !value;
+
+      onValueChange?.(newValue);
```
M-i-k-e-l, line 204: I feel like this should be called together with `onChangeValidity` after the state was changed, I realize this would make the code a little more awkward since you'll have to add an `else` with another `onValueChange`, so I am conflicted, WDYT?

M-i-k-e-l (approved):

src/components/checkbox/index.tsx
```diff
@@ -195,20 +198,23 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     }).start();
   }
 
+  setValidation(newValue: boolean) {
```
M-i-k-e-l, line 201: I'd call it `testValidation`

M-i-k-e-l (approved):

src/components/checkbox/index.tsx
Inbal-Tish: I'll keep it simple until we'll get such a request

src/components/checkbox/index.tsx
Inbal-Tish: onValueChange will be called either way, the delay doesn't seem much of an issue

src/components/checkbox/index.tsx
Inbal-Tish: I think 'set' better describes what the func does

ethanshar:
@Inbal-Tish 
This PR breaks tests in private. 
Can you please take a look. 
You will need to update snapshot version in private using `yarn up react-native-ui-lib@snapshot`

PR #2677, opened 2023-07-23
fixed the ts error on ExpandableOverlay dialogProps in Picker.

adids1221:
## Description
Picker - ExpandableOverlay had ts error over the dialogProps.
We had the same error inside ExpandableOverlay `renderDialog`, it's coming from the `null` type for `height` prop in the old `Dialog` props.

## Changelog
Picker ts error fix.


M-i-k-e-l:
Lets fix this as we've discussed

M-i-k-e-l (approved):

Issue #2230, opened 2022-08-31
VSCode not suggesting the component's props


thanhloc-17:
This problem occur when I update new version or RNUILib, then I type some prop and VSCode not suggest any props
<img width="568" alt="image" src="https://user-images.githubusercontent.com/20339875/187686212-6067f88a-f3f4-492f-ae6f-2c0be1f8b2eb.png">

With the others package, the IntelliSense still working well.

Thanks

ramssutharr:
having same issue on react native v0.70 .

qkudev:
Me too!

thanhloc-17:
up

tarikpnr:
any news on this? @ethanshar

lakardion:
Same on my side. I get key and ref as the only two available props
It is happening on ALL imported components from the library:
![image](https://user-images.githubusercontent.com/52294448/197536560-c06982da-8178-44dd-b4b4-80d60f509740.png)

The-White-Fang:
any updates or hints on what might be wrong?

The-White-Fang:
possible duplicate of https://github.com/wix/react-native-ui-lib/issues/2202 ?

KishoreBarik:
same issue here

thanhloc-17:
up

thanhloc-17:
up

chrismerino:
Issue still persists on RN 0.70.5 and react-native-ui-lib: 7.1.0

ethanshar:
Hi, is this issue still relevant? 
I don't experience it

jkaveri:
I still experience it on `7.3.5`

jkaveri:
<img width="1205" alt="Screenshot 2023-05-14 at 11 45 32" src="https://github.com/wix/react-native-ui-lib/assets/884709/25a28694-bdc8-4e18-be17-087d9ced6a88">
It looks like the components is losing props type.

The-White-Fang:
> Hi, is this issue still relevant? I don't experience it

Could you share your `tsconfig.json` please? We may have to add something to it to make it work. 
Starting a new project from scratch and installing the lib doesn't work for us. 

A workaround I found was to remove the global type `ThemeComponent` and place it in an `export` statement. But that seems like something that will never be in the official repo.
I am still looking into it and will let you know if I can find something else to fix it.

jkaveri:
my tsconfig below:

```
{
  "extends": "expo/tsconfig.base",
[...]

PR #2678, opened 2023-07-24
declare other global typings

ethanshar:

### DIFF
--- a/typings/globalTypes.d.ts
+++ b/typings/globalTypes.d.ts
@@ -6,10 +6,10 @@ type Dictionary<TYPE> = {[key: string]: TYPE};
 declare module '*.png';
 declare module 'react-native-measureme';
 
-interface Extendable {
+declare interface Extendable {
   [key: string]: any;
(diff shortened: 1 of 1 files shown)


src/typings/module.d.ts
```diff
@@ -8,3 +8,7 @@ declare namespace JSX {
     fsTagName?: string;
   }
 }
+
+// This support importing png files, typing wise
+declare module '*.png';
+declare module 'react-native-measureme';
```
M-i-k-e-l, line 14: I think we can remove this

src/typings/module.d.ts
```diff
@@ -8,3 +8,7 @@ declare namespace JSX {
     fsTagName?: string;
```
M-i-k-e-l, line 8: This should probably be in common

M-i-k-e-l (approved):
A couple of small notes

M-i-k-e-l:
Fixes #2230

PR #2681, opened 2023-07-24
Picker in wheel picker mode supports top bar props

adids1221:
## Description
Picker in WheelPicker mode supports `topBarProps`.

## Changelog
Picker in WheelPicker mode supports `topBarProps`.

## Additional info
https://github.com/wix/react-native-ui-lib/issues/2557

### DIFF
--- a/src/components/picker/PickerItemsList.tsx
+++ b/src/components/picker/PickerItemsList.tsx
@@ -7,6 +7,7 @@ import Modal from '../modal';
 import View from '../view';
 import Text from '../text';
 import Icon from '../icon';
+import Button from '../button';
 import WheelPicker from '../WheelPicker';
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@adids1221 I'm not sure I understand. We're supporting passing topBarProps but we're not rendering TopBar? Why?

adids1221:
> @adids1221 I'm not sure I understand. We're supporting passing topBarProps but we're not rendering TopBar? Why? Do we want to support all of these props or just cancel button?

We are supporting TopBar props when the user isn't using WheelPicker.
When using WheelPicker we are giving the option to render `done` button but not a `cancel` button.
I'll consult the UX but for now, let's support the `cancel` button.

Inbal-Tish (approved):

PR #2683, opened 2023-07-27
Chip - change margin to padding

lidord-wix:
## Description
Using a margin cause a cut in the chip in some cases. moving to padding keeps the same UI without this bug.

## Changelog
Fix cut chips 

## Additional info
WOAUILIB-3665

### DIFF
--- a/src/components/chip/index.tsx
+++ b/src/components/chip/index.tsx
@@ -278,7 +278,7 @@ const Chip = ({
               marginRight: Spacings.s2
             };
           } else {
-            return {marginHorizontal: Spacings.s3};
+            return {paddingHorizontal: Spacings.s3};
           }
         case 'avatar':
           return {


src/components/chip/index.tsx
```diff
@@ -278,7 +278,7 @@ const Chip = ({
               marginRight: Spacings.s2
             };
           } else {
-            return {marginHorizontal: Spacings.s3};
+            return {paddingHorizontal: Spacings.s3};
```
Inbal-Tish, line 281: What about the other cases? Does it work fine there?

Inbal-Tish:
@lidord-wix How did you manage to reproduce it here? Do you have a code snippet or in the demo screen?

src/components/chip/index.tsx
lidord-wix: I couldn't see any issue with the other cases.

lidord-wix:
> @lidord-wix How did you manage to reproduce it here? Do you have a code snippet or in the demo screen?

I was able to reproduce it only in the private

src/components/chip/index.tsx
Inbal-Tish: Ok. Let's do that for now

Inbal-Tish (approved):

PR #2684, opened 2023-07-27
No static locale infra

M-i-k-e-l:
## Description
We should avoid `defaultProps` that are localized and are directly assigned.

## Changelog
Deprecate `ColorPicker.accessibilityLabels` in favor of `ColorPicker.getAccessibilityLabels`
Deprecate `ColorSliderGroup.labels` in favor of `ColorSliderGroup.getLabels`

## Additional info
WOAUILIB-3659

### DIFF
--- a/src/components/colorPicker/colorPicker.api.json
+++ b/src/components/colorPicker/colorPicker.api.json
@@ -26,10 +26,10 @@
       "description": "Callback for the picker's color palette change"
     },
     {
-      "name": "accessibilityLabels",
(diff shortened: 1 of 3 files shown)


Inbal-Tish:
> We should avoid `defaultProps` that are localized and are directly assigned.

Why?

Inbal-Tish (approved):

M-i-k-e-l:
> > We should avoid `defaultProps` that are localized and are directly assigned.
> 
> Why?

The `defaultProps` are called staticky which triggers (if localization s used) the start of our `i18n` file which should not be initialized this early (because of the new translation migration)

PR #2699, opened 2023-08-07
Picker support useDialog

adids1221:
## Description
Picker support `useDialog` prop to render picker in Dialog instead of Modal.

## Changelog
Picker support `useDialog` prop to render picker in Dialog instead of Modal.

### DIFF
--- a/src/components/picker/index.tsx
+++ b/src/components/picker/index.tsx
@@ -64,6 +64,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
     renderCustomSearch,
     // useNativePicker,
     useWheelPicker,
+    useDialog,
     renderPicker,
     customPickerProps,
     containerStyle,
@@ -287,7 +288,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
     <PickerContext.Provider value={contextValue}>
       <ExpandableOverlay
         ref={pickerExpandable}
-        useDialog={useWheelPicker}
+        useDialog={useDialog || useWheelPicker}
         modalProps={modalProps}
         dialogProps={DIALOG_PROPS}
         expandableContent={expandableModalContent}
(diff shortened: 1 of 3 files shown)


ethanshar (commented):
The code looks good. 
But the UI seem a little off, can you take a look? 
When picker opens as dialog, the header (where the X is) is a little high, where the height is coming from?

src/components/picker/index.tsx
```diff
@@ -284,46 +287,50 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
-            value={label}
-            selection={Constants.isAndroid ? {start: 0} : undefined}
-            /* Note: Disable TextField expandable feature */
-            // topBarProps={undefined}
-          >
-            {renderPickerInnerInput()}
-          </TextFieldMigrator>
-        )}
-      </ExpandableOverlay>
+      {
```
ethanshar, line 292: please remove the curly brackets
If you need this for the ts-expect-error you can add it like this 
```
{/* @ts-expect-error */}
```

src/components/picker/index.tsx
```diff
@@ -221,11 +222,13 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
         testID={`${testID}.modal`}
         useWheelPicker={useWheelPicker}
+        useDialog={useDialog}
         items={useItems ? items : undefined}
         topBarProps={{
           ...topBarProps,
           onCancel: cancelSelect,
-          onDone: mode === PickerModes.MULTI ? () => onDoneSelecting(multiDraftValue) : undefined
+          onDone: mode === PickerModes.MULTI ? () => onDoneSelecting(multiDraftValue) : undefined,
+          includeStatusBar: useDialog ? false : undefined
```
ethanshar, line 231: any reason why not 
```
includeStatusBar: !useDialog
```

PR #2705, opened 2023-08-09
wheel picker onlayout fix

adids1221:
## Description
WheelPicker - `onLayout` is called more than once, replacing the `onLayout` function with `useEffect`.

## Changelog
WheelPicker with a long list of data jumps back to `initalValue` fix.

## Additional info

### DIFF
--- a/src/components/WheelPicker/index.tsx
+++ b/src/components/WheelPicker/index.tsx
@@ -205,10 +205,6 @@ const WheelPicker = ({
     setTimeout(() => scrollToOffset(index, animated), 100);
   };
 
-  // const scrollToPassedIndex = useCallback(() => {
-  //   scrollToIndex(currentIndex, false);
-  // }, []);
-
   const selectItem = useCallback((index: number) => {
     scrollToIndex(index, true);
   },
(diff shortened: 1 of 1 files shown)


src/components/WheelPicker/index.tsx
```diff
@@ -200,9 +205,9 @@ const WheelPicker = ({
     setTimeout(() => scrollToOffset(index, animated), 100);
   };
 
-  const scrollToPassedIndex = useCallback(() => {
-    scrollToIndex(currentIndex, false);
-  }, []);
+  // const scrollToPassedIndex = useCallback(() => {
```
Inbal-Tish, line 208: Why to keep it commented?

Inbal-Tish (approved):

Inbal-Tish:
@adids1221 Approved with a small comment. You can merge at will

src/components/WheelPicker/index.tsx
adids1221: I thought it will be good to keep it comment (as documentation for us) and the `onLayout` prop in the `FlatList` in case of any issue with it.
Should I delete it?

Inbal-Tish: I don't see a need to keep it. If the fix will cause issues we can revert it.

Issue #2706, opened 2023-08-10
"Slider value is not in range" triggering falsely when useRange is true


sebastianjung:
## Description
console spits out a warning of "Slider value is not in range" in cases when useRange is used and the initialMinimumValue and initialMaximumValue is definitely in the range.

checkProps function checks for props.value which is not used when useRange is true. This is probably the reason for the bug.

```js
checkProps(props) {
    if (props.minimumValue >= props.maximumValue) {
      console.warn('Slider minimumValue must be lower than maximumValue');
    }
    if (props.value < props.minimumValue || props.value > props.maximumValue) {
      console.warn('Slider value is not in range');
    }
  }
```

### Steps to reproduce

using the component like this should trigger the bug:
```jsx
<Slider
                minimumValue={1}
                initialMinimumValue={3}
                initialMaximumValue={6}
                maximumValue={8}
                step={1}
                thumbTintColor={Colors.primary}
                maximumTrackTintColor={Colors.gray5}
                minimumTrackTintColor={Colors.primary}
                useRange
                onRangeChange={({ min, max }) => {
                  console.log('range changed')
                }}
              />
```

Inbal-Tish:
@sebastianjung Hi. The 'value' and 'initialValue' props default is 0. When you set only 'minValue' to 1, 'value' is not in the range.

sebastianjung:
@Inbal-Tish Thank you for your comment.

When using 'useRange' prop the value and initialValue Prop should not be used, right? Only for sliders with one Thumb, right?

I'm probably just confused here.

Inbal-Tish:
Sure. But you are passing initial value... I'll add the check but I suggest not passing initialvalue without value

Inbal-Tish:
@sebastianjung 'nimimumValue' prop should be considered in both cases, but I see your point regarding the 'value' prop. I'll add the check. Thanks

PR #2709, opened 2023-08-11
Slider - fix value out of range warning in range slider

Inbal-Tish:
## Description
Slider - fix value out of range warning in range slider

## Changelog
Slider - fix value out of range warning in range slider

## Additional info
solves #2706 2706


adids1221 (approved):
Approved

PR #2712, opened 2023-08-14
limit picker number of lines

liatnetach:
## Description
Picker - fix ellipsis on long text in `filterType='filter'`, add `numberOfLines` to the text.

## Changelog
Picker - fix ellipsis on long text in filterType set to filter.

## Additional info


liatnetach:
> @liatnetach the solution isn't working for me. Does it reproduce and work in your env?

in addition to maxWidth from our end, yes 🙏🏻

adids1221 (commented):
Approved

adids1221 (approved):

PR #2725, opened 2023-08-24
on filtered children change cb

adids1221:
## Description
Picker `onSearchChange` callback now supports `filteredChildren` prop to get the filtered list after a search.

## Changelog
Picker `onSearchChange` callback now supports `filteredChildren` prop.


lidord-wix (approved):
Approved, see the conflict
BTW, is there a ticket for that?

adids1221:
> Approved, see the conflict BTW, is there a ticket for that?

Yes, the request came from the support channel, I've posted the PR there and I'll update the user ticket.

PR #2727, opened 2023-08-28
Fix image svg

mendyEdri:
## Description
Until now, all images with url used `SvgImage` component, because source was a url and `isSvg` returned `true`.

## Changelog
Added support for `Image` web


package.json
OrenZak, line 1: I think its good to add a demo item to the webDemo app

src/components/image/index.tsx
```diff
@@ -226,23 +226,31 @@ class Image extends PureComponent<Props, State> {
...
+      width && {width},
+      height && {height},
+      cover && styles.coverImage,
+      this.isGif() && styles.gifImage,
+      aspectRatio && {aspectRatio},
+      !useImageInsideContainer && margins,
+      useImageInsideContainer && styles.containImage,
+      style,
+      useImageInsideContainer && styles.shrink
+    ];
```
OrenZak, line 241: this wasn't change at all right ?

src/components/image/index.tsx
```diff
@@ -226,23 +226,31 @@ class Image extends PureComponent<Props, State> {
...
+
+    if (!this.shouldUseImageBackground() && Constants.isWeb) {
```
OrenZak, line 243: `Constants.isWeb` this will be always false, as you have `index.web.tsx`

src/components/image/index.web.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const finalSource = source.uri ?? source;
+
+    return (
+      // @ts-expect-error
+      <View style={imageViewStyle}>
+        {/* @ts-expect-error */}
+        <img 
+          {...others} 
+          src={finalSource} 
+          style={StyleSheet.flatten([imageViewStyle, styles.containImage]) as CSSProperties}
```
OrenZak, line 176: why do we need to always flat this ?

package.json
OrenZak: also can you create a version, so we can test it before the merge ? in our Editor
@AmitShwarts, can you give it a try once, @mendyEdri  shares a version ?

src/components/image/index.tsx
mendyEdri: Nope

src/components/image/index.web.tsx
mendyEdri: Same as you guys did, thought it's for flattening any sort of imageProps passed etc

package.json
mendyEdri: 7.8.0-snapshot.3463

src/components/image/index.web.tsx
lidord-wix, line 1: Please format the file :)

src/components/image/index.web.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  isGif() {
+    if (Constants.isAndroid) {
+      const {source} = this.props;
+      const url = _.get(source, 'uri');
+      const isGif = /(http(s?):)([/|.|\w|\s|-])*\.gif/.test(url ?? '');
+      return isGif;
+    }
+  }
```
lidord-wix, line 81: I believe we can remove it here

src/components/image/index.web.tsx
lidord-wix, line 1: Something looks different in the webDemo project in the `Animated Image` example, when I run it with the changes from this PR.

src/components/image/index.web.tsx
```diff
@@ -0,0 +1,236 @@
...
+      useImageInsideContainer && styles.containImage,
+      style,
+      useImageInsideContainer && styles.shrink
+    ];
+
+    const finalSource = source.uri ?? source;
+
+    return (
+      // @ts-expect-error
+      <View style={imageViewStyle}>
```
lidord-wix, line 171: Why do you add this extra `View` and not pass the style to the `img` component?

lidord-wix (commented):
I added some comments, but... I have to say that I don't like this solution, it'll require Uilib to maintain two image files, and I'm confident that it will be forgotten. 
react-native-web supports RN's Image, so we should re-check what doesn't work and close the gaps. 
Until now we added the .web files only when something isn't supported on the web and I think we should keep doing it instead of duplicating to code.

src/components/image/index.web.tsx
OrenZak: i'm not sure we need it here, on the Icon comp it was because the button passes some default styles as array. 
not sure its the case here.

package.json
AmitShwarts: @mendyEdri as we just talked, we're getting lot's of issues of width, height and resizeMode.
all of the page layout breaks for some reason because of this.. not sure why

src/components/image/index.web.tsx
mendyEdri: @OrenZak is right

mendyEdri:
> I added some comments, but... I have to say that I don't like this solution, it'll require Uilib to maintain two image files, and I'm confident that it will be forgotten. react-native-web supports RN's Image, so we should re-check what doesn't work and close the gaps. Until now we added the .web files only when something isn't supported on the web and I think we should keep doing it instead of duplicating to code.

I don't mind. atm in production sources with type of string (either source={url} or images actual data) are rendering the SVG component which causes issues and left us with production bugs.

Also, the current mobile image implementation cannot render ImageBackground properly in web.
moreover, using the mobile image, rendered styling weird, and wasn't affecting by image-ratio prop.

I suggest to assign someone from the team to address it :)
thanks!

OrenZak:
> > I added some comments, but... I have to say that I don't like this solution, it'll require Uilib to maintain two image files, and I'm confident that it will be forgotten. react-native-web supports RN's Image, so we should re-check what doesn't work and close the gaps. Until now we added the .web files only when something isn't supported on the web and I think we should keep doing it instead of duplicating to code.
> 
> I don't mind. atm in production sources with type of string (either source={url} or images actual data) are rendering the SVG component which causes issues and left us with production bugs.
> 
> Also, the current mobile image implementation cannot render ImageBackground properly in web. moreover, using the mobile image, rendered styling weird, and wasn't affecting by image-ratio prop.
> 
> I suggest to assign someone from the team to address it :) thanks!

@ethanshar can you help ?

ethanshar (commented):
I don't think we want to go with a duplicate file for the Image component just for web. 
These type of solutions won't scale, we can't maintain two identical files for two platforms.. 

Either we come up with a better solution, or please open a detailed ticket and we'll tailor this properly.

ethanshar:
Is this PR still relevant? can we close it?

mendyEdri:
> Is this PR still relevant? can we close it? 

It is. 
Removing it, still breaking images in web-editor

ethanshar:
But we don't use this fix, it's not merged. 
Is there still an issue with SVG?

mendyEdri:
> But we don't use this fix, it's not merged. 
> 
> Is there still an issue with SVG? 

We do, we use this version in web-editor.
And yes, there is still an issue

ethanshar:
@mendyEdri 
Is this PR still relevant?

mendyEdri:
> @mendyEdri 
> 
> Is this PR still relevant? 
> 
> 

Not anymore 💪

PR #2728, opened 2023-08-28
fix: [WEB] SvgImage update styles

AmitShwarts:
## Description
This PR is fixing 3 bugs in web svg component:
1. when rendering more than 1 svg, the style of the last one will override all of them
2. the style is initialized once, and doesn't been updated
3. the height of the svg wrapper bigger than the svg's height - caused an offest inside the button. the svg took too much space
image

**changes:**
1. add component to the web demo with `IconButton` that changes size
2. add unique id to each svg component
3. apply style also on the div wrapper
4. update the css when the style changes
## Changelog

fix: [WEB] Svg image unique style tag and update styles


adids1221 (commented):
Left one comment about the useEffect

adids1221 (approved):
Approved

PR #2734, opened 2023-09-06
Colors - generate palette - pass options object

Inbal-Tish:
## Description
Colors - generate palette - pass options object

## Changelog
Colors - generate palette - pass options object

### DIFF
--- a/src/style/colors.ts
+++ b/src/style/colors.ts
@@ -15,9 +15,12 @@ export type DesignToken = {semantic?: [string]; resource_paths?: [string]; toStr
 export type TokensOptions = {primaryColor: string};
 export type GetColorTintOptions = {avoidReverseOnDark?: boolean};
 export type GetColorByHexOptions = {validColors?: string[]};
-export type PaletteOptions = {
-  lightnessFix?: boolean;
-  saturationFix?: boolean;
+export type GeneratePaletteOptions = {
+  /** Whether to adjust the lightness of very light colors (generating darker palette) */
+  adjustLightness?: boolean;
+  /** Whether to adjust the saturation of colors with high lightness and saturation (unifying saturation level throughout palette) */
+  adjustSaturation?: boolean;
+  /** Whether to add two extra darker colors usually used for dark mode (generating a palette of 10 instead of 8 colors) */
   darkModeColors?: boolean;
 }
 export class Colors {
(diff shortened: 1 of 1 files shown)


src/style/colors.ts
```diff
@@ -15,7 +15,11 @@ export type DesignToken = {semantic?: [string]; resource_paths?: [string]; toStr
 export type TokensOptions = {primaryColor: string};
 export type GetColorTintOptions = {avoidReverseOnDark?: boolean};
 export type GetColorByHexOptions = {validColors?: string[]};
-
+export type PaletteOptions = {
```
ethanshar, line 18: Rename to GeneratePaletteOptions so it's more associated with the method

src/style/colors.ts
```diff
@@ -15,7 +15,11 @@ export type DesignToken = {semantic?: [string]; resource_paths?: [string]; toStr
...
+  lightnessFix?: boolean;
```
ethanshar, line 19: Maybe make this option more generic and instead of passing `lightnessFix` as boolean you can have a  `lightnessThreshold` option where you can pass the actual  threshold value you want. 
This way you easily control it from the outside without changing the source code each time

src/style/colors.ts
```diff
@@ -15,7 +15,11 @@ export type DesignToken = {semantic?: [string]; resource_paths?: [string]; toStr
...
+  saturationFix?: boolean;
+  darkModeColors?: boolean;
```
ethanshar, line 21: The options is not very clear and doesn't sound like a boolean option. 
Maybe rename to something like `forDarkModeColors` 
I would also consider adding a short comment description about each option, so it will appear in the typings info

src/style/colors.ts
```diff
@@ -15,7 +15,11 @@ export type DesignToken = {semantic?: [string]; resource_paths?: [string]; toStr
 export type TokensOptions = {primaryColor: string};
 export type GetColorTintOptions = {avoidReverseOnDark?: boolean};
 export type GetColorByHexOptions = {validColors?: string[]};
-
+export type PaletteOptions = {
+  lightnessFix?: boolean;
+  saturationFix?: boolean;
```
ethanshar, line 20: consider renaming to a more boolean name like `adjustSaturation`, `shouldAdjustSaturation`
Or also here, think how the user can pass a numeric value instead of a boolean where they can control the adjustment. If the user pass `undefined` you can treat it as false of course.

ethanshar (commented):
Wrote mostly naming comments and ideas on how to improve the API and make it more generic

src/style/colors.ts
Inbal-Tish: Done

src/style/colors.ts
Inbal-Tish: It's not going to be as easy as passing a lightnessThreshold since the fix only applies if it meets a set range. The user will have to adjust the range and pass the value to adjust to. The boolean is about whethere I apply this range check or not. Maybe rename it adjustLightness?

src/style/colors.ts
Inbal-Tish: Done

src/style/colors.ts
Inbal-Tish: Added comments. How about renaming to 'addDarkestColors'?

src/style/colors.ts
ethanshar: ok, let's go with `adjustLightness`

src/style/colors.ts
ethanshar: Much better, maybe `addDarkestTints`

ethanshar (approved):

PR #2741, opened 2023-09-11
ColorPickerDialog - migrate Dialog to use new implementation

Inbal-Tish:
## Description
ColorPickerDialog - migrate Dialog to use the new incubator implementation

## Changelog
ColorPickerDialog - migrate Dialog to use the new incubator implementation


src/components/colorPicker/ColorPickerDialog.tsx
```diff
@@ -338,9 +338,9 @@ class ColorPickerDialog extends PureComponent<Props, State> {
         centerH
         onDismiss={this.onDismiss}
         containerStyle={styles.dialog}
-        panDirection={PanningProvider.Directions.DOWN}
+        direction={PanningProvider.Directions.DOWN}
```
M-i-k-e-l, line 341: This is not needed anymore

src/components/colorPicker/ColorPickerDialog.tsx
```diff
@@ -338,9 +338,9 @@ class ColorPickerDialog extends PureComponent<Props, State> {
...
         testID={`${testID}.dialog`}
-        supportedOrientations={['portrait', 'landscape', 'landscape-left', 'landscape-right']} // iOS only
+        modalProps={{supportedOrientations: ['portrait', 'landscape', 'landscape-left', 'landscape-right']}} // iOS only
```
M-i-k-e-l, line 343: Can we make this a `const` please?

M-i-k-e-l (commented):
When pressing done (the checkmark) the dialog is closed immediately without animation

src/components/colorPicker/ColorPickerDialog.tsx
Inbal-Tish: Not really, there's a type issue here since this is not a string array per se

M-i-k-e-l: I think it's best to add it as a const and cast it:
```
import {ModalProps} from '../../components/modal';
...
const MODAL_PROPS = {supportedOrientations: ['portrait', 'landscape', 'landscape-left', 'landscape-right']} as ModalProps;
```

M-i-k-e-l (approved):
Approved but I think we should make it a const (see the comment)

M-i-k-e-l (approved):

PR #2742, opened 2023-09-11
chore: infer SVG size from style

AmitShwarts:
## Description

allow SVG to infer size from the style object

## Changelog

Now SVG can infer size from style

### DIFF
--- a/src/components/svgImage/index.tsx
+++ b/src/components/svgImage/index.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import {ImageStyle, StyleProp, StyleSheet} from 'react-native';
 import {isSvg, isSvgUri} from '../../utils/imageUtils';
 import {SvgPackage} from '../../optionalDependencies';
 
@@ -12,6 +13,7 @@ export interface SvgImageProps {
    */
   tintColor?: string | null;
   data: any; // TODO: I thought this should be string | React.ReactNode but it doesn't work properly
+  style?: StyleProp<ImageStyle>;
 }
 
 function SvgImage(props: SvgImageProps) {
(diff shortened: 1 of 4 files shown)


src/components/svgImage/index.tsx
```diff
@@ -28,7 +30,12 @@ function SvgImage(props: SvgImageProps) {
   if (isSvgUri(data)) {
     return <SvgCssUri {...others} uri={data.uri}/>;
   } else if (typeof data === 'string') {
-    return <SvgXml xml={data} {...others}/>;
+    const flattenStyle = StyleSheet.flatten(props.style) as Record<string, any>;
```
M-i-k-e-l, line 33: We use `flatten` very rarely, why is this required here?

AmitShwarts: to extract the `width` and `height` from the provided styles.
do you have suggestion how to do it differently?

OrenZak: are we sure its coming with more than one style object ?
if not, and we always have just single style object (and not as array).?
we can remove that.

AmitShwarts: yes i'm sure. we can't remove it

M-i-k-e-l: Can you please provide the scenario this is used in?

AmitShwarts: Yes, for example when using Button with SVG Icon, we're passing `iconStyle` with `width` & `height`.
in this case it's been passed to the SVG component, and without this PR the height and width is not respected.

AmitShwarts:
```
    <Button
      id={id}
      testID={TestIds.buttonElement}
      style={style}
      color={color}
      iconSource={iconSource}
      iconStyle={{width: Spacings.s6, height: Spacings.s6}}
      round
      size={Button.sizes.large}
    />
```

M-i-k-e-l: As discussed, lets allow passing `iconProps` to the `Button` instead, which will allow sending `width` and `height`

M-i-k-e-l (approved):

PR #2748, opened 2023-09-18
fix/svg image respect width height props on web

AmitShwarts:
## Description

on web, SvgImage didn't respect `width` & `height` props

## Changelog

SvgImage now respects `width` & `height` props on web

## Additional info

in order to see the changes in the web demo you need to apply the changes manually in node_modules of the web-demo


src/components/svgImage/index.web.tsx
```diff
@@ -25,13 +27,13 @@ function SvgImage(props: SvgImageProps) {
       const {postcss, cssjs} = PostCssPackage;
       const styleObj: Record<string, any> = StyleSheet.flatten(style);
       postcss()
-        .process(styleObj, {parser: cssjs})
+        .process({width, height, ...styleObj}, {parser: cssjs})
```
M-i-k-e-l, line 30: Maybe add someone that looked at web recently for this part

src/components/icon/index.tsx
```diff
@@ -53,10 +61,12 @@ const Icon = forwardRef((props: Props, ref: any) => {
     assetName,
     modifiers,
     recorderTag,
+    width,
+    height,
     ...others
   } = props;
   const {margins} = modifiers;
-  const iconSize = size ? {width: size, height: size} : undefined;
+  const iconSize = size ? {width: width ?? size, height: height ?? size} : undefined;
```
M-i-k-e-l, line 69: The props might be missing but, but doesn't they work without this change? The props are spread in both cases, so if you do not send `size` it should work.
* I did not test it.

AmitShwarts: it doesn't work without it because it was spreading iconSize after spreading the props

src/components/svgImage/index.web.tsx
AmitShwarts: @adids1221

src/components/icon/index.tsx
AmitShwarts: on web size is always defined, that's why we need to give width & height priority.
if size isn't defined (mobile and no one passed size) then width and height will be passed when spreading the props

M-i-k-e-l: If this is only for the SVG part, I suggest changing this line:
```
const renderSvg = () => <SvgImage fsTagName={recorderTag} data={source} {...props} {...iconSize}/>;
```
to
```
const renderSvg = () => <SvgImage fsTagName={recorderTag} data={source} {...iconSize} {...props}/>;
```

M-i-k-e-l: This will remove the need for all the other changes

webDemo/src/App.tsx
```diff
@@ -55,10 +55,14 @@ const itemsToRender: ItemToRender[] = [
           size={Button.sizes.large}
           iconSource={svgData}
           iconStyle={{
-            width: size,
-            height: size,
+            // width: size,
+            // height: size,
```
adids1221, line 59: If the `iconStyle` object isn't relevant anymore let's remove it.
Delete the commented lines.

src/components/icon/index.tsx
```diff
@@ -25,6 +25,14 @@ export type IconProps = Omit<ImageProps, 'source'> &
      * the icon size
      */
     size?: number;
+    /**
+     * the icon height
+     */
+    height?: number | string;
```
adids1221, line 31: About the `string` type, when I pass the string I get this error:
```
JSON value of type NSString cannot be converted to a YGValue.
```

To get over this error the user should pass `value%` as the `width` or `height` value and this is not the result we want.

Let's make it only a `number` type.

src/components/icon/icon.api.json
```diff
@@ -11,6 +11,8 @@
     {"name": "assetGroup", "type": "string", "description": "the asset group, default is icons"},
     {"name": "tintColor", "type": "string", "description": "The icon tint"},
     {"name": "size", "type": "number", "description": "The icon size"},
+    {"name": "height", "type": "number | string", "description": "The icon height"},
```
adids1221, line 14: also change the type here to `number`.

adids1221 (commented):
I wrote a few small comments

adids1221:
Also, snapshot version to check in the webDemo: `7.8.0-snapshot.3520`

src/components/icon/index.tsx
M-i-k-e-l: @adids1221 what string did you use?

PR #2749, opened 2023-09-19
chore: [WEB] SvgImage respect width & height props

AmitShwarts:
## Description

on web, SvgImage didn't respect `width` & `height` props.
currently we need to provide the `width` & `height` as props and also in style, so it would be respected both in web and mobile.

## Changelog

SvgImage now respects `width` & `height` props on web

## Additional info

in order to see the changes in the web demo you need to apply the changes manually in node_modules of the web-demo
instead of #2748


adids1221 (approved):
Approved

PR #2752, opened 2023-09-21
Add a way to delete fonts from disk

M-i-k-e-l:
## Description
`DynamicFonts` - add a way to delete fonts from disk
Move\add some debug logs
Rename `buildFontName` into `buildFontInput`

## Changelog
None

## Additional info
None

### DIFF
--- a/lib/components/DynamicFonts/FontDownloader.ts
+++ b/lib/components/DynamicFonts/FontDownloader.ts
@@ -35,6 +35,13 @@ export default class FontDownloader {
     this.fs = RNFS;
   }
 
+  private log(message?: any, ...optionalParams: any[]) {
+    const {debug} = this.props;
+    if (debug) {
+      console.log(message, optionalParams);
+    }
+  }
+
   private getPrivateFolder() {
     const {dynamicFontsFolder} = this.props;
     return (
@@ -97,35 +104,27 @@ export default class FontDownloader {
     fontName: string,
     fontExtension: FontExtension,
     timeout: number): Promise<string> {
-    const {downloadErrorMessage, debug} = this.props;
+    const {downloadErrorMessage} = this.props;
     await this.createPrivateFolderIfNeeded();
     const fileName = this.getFileName(fontName, fontExtension);
     const downloadLocation = this.getPrivateFilePath(fileName);
 
     try {
-      if (debug) {
-        console.log(fontName, 'Starting to download');
-      }
+      this.log(fontName, 'Starting to download');
       const result = await this.fs.downloadFile(this.getDownloadFontOptions(fontUri, downloadLocation, timeout))
         .promise;
       if (result.statusCode === 200) {
-        if (debug) {
-          console.log(fontName, 'Finished downloading');
-        }
+        this.log(fontName, 'Finished downloading');
         return downloadLocation;
       } else {
-        if (debug) {
-          console.log(fontName, 'Error downloading statusCode:', result.statusCode);
-        }
+        this.log(fontName, 'Error downloading statusCode:', result.statusCode);
         return Promise.reject({
           source: 'uilib:FontDownloader:downloadFont',
           message: `${downloadErrorMessage} fontName: ${fontName} statusCode: ${result.statusCode}`
         });
       }
     } catch (error) {
-      if (debug) {
-        console.log(fontName, 'Error downloading error:', error);
-      }
+      this.log(fontName, 'Error downloading error:', error);
       return Promise.reject({
         source: 'uilib:FontDownloader:downloadFont',
         message: `${downloadErrorMessage} fontName: ${fontName} error: ${JSON.stringify(error)}`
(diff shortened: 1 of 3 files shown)


lib/components/DynamicFonts/FontDownloader.ts
```diff
@@ -97,26 +97,38 @@ export default class FontDownloader {
...
     const downloadLocation = this.getPrivateFilePath(fileName);
 
     try {
+      if (debug) {
+        console.log(fontName, 'Starting to download');
+      }
       const result = await this.fs.downloadFile(this.getDownloadFontOptions(fontUri, downloadLocation, timeout))
         .promise;
       if (result.statusCode === 200) {
+        if (debug) {
```
Inbal-Tish, line 112: We can't use logService here, to log only on DEV? I would add a simple console.log there to use here

lib/components/DynamicFonts/index.ts
```diff
@@ -118,13 +112,13 @@ export default class DynamicFonts {
     }
   }
 
-  private buildFontName(rootUri: string,
+  private buildFontInput(rootUri: string,
```
Inbal-Tish, line 115: I don't understand this name, shouldn't it be more like 'getFontData' or something?

lib/components/DynamicFonts/index.ts
```diff
@@ -145,11 +139,31 @@ export default class DynamicFonts {
         fontsLoaded = await this.getFonts(fonts);
       } catch (error) {
         if (debug) {
-          console.log(`getFontFamily failed (try #${tryCounter}) error: ${error}`);
+          console.log(`getFontFamily failed (try #${tryCounter}) error:`, error);
```
Inbal-Tish, line 142:
```suggestion
          console.log(`getFontFamily failed (try #${tryCounter}) error: `, error);
```

lib/components/DynamicFonts/index.ts
```diff
@@ -145,11 +139,31 @@ export default class DynamicFonts {
...
+    if (await this.fontDownloader.isFontDownloaded(fontName, fontExtension)) {
+      await this.deleteFontFromDisk(fontName, fontExtension);
+    }
+  }
+
+  public async deleteFontFamily(fontNames: string[],
+    fontExtension: FontExtension,
+    fontNamePrefix?: string): Promise<void> {
+    fontNames.forEach(async fontName => {
+      await this.deleteFontFromDisk(fontName, fontExtension, fontNamePrefix);
```
Inbal-Tish, line 166: Here you're not checking if 'isFontDownloaded' before deleting it?

lib/components/DynamicFonts/FontDownloader.ts
M-i-k-e-l: It might be possible but I don't like it, it will open a precedent of using `src` in the `lib` which can cause cycles.
Adding all logs is a little spammy IMO.

lib/components/DynamicFonts/index.ts
M-i-k-e-l: I wanted to reflect the type `GetFontInput`.
I think I'll change `GetFontInput` -> `GetFontOptions` and `buildFontInput` -> `buildFontOptions` (or `getFontOptions` if you prefer).
WDYT?

lib/components/DynamicFonts/index.ts
M-i-k-e-l: It adds a space in the `log`

lib/components/DynamicFonts/index.ts
Inbal-Tish: Are they options or required?

lib/components/DynamicFonts/index.ts
Inbal-Tish: Usually not

lib/components/DynamicFonts/index.ts
M-i-k-e-l: I'm thinking of removing this test altogether since a non-existing is checked for later anyway, WDYT?

lib/components/DynamicFonts/index.ts
M-i-k-e-l: Add `console.log('bla', 'bli');` and you'll get:
```
 LOG  bla bli
```

lib/components/DynamicFonts/index.ts
M-i-k-e-l: Good point, mostly required

lib/components/DynamicFonts/FontDownloader.ts
```diff
@@ -97,26 +97,38 @@ export default class FontDownloader {
...
+          console.log(fontName, 'Finished downloading');
+        }
         return downloadLocation;
       } else {
+        if (debug) {
```
Inbal-Tish, line 117: It doesn't warn you that the else and if should be one the same line?

lib/components/DynamicFonts/index.ts
Inbal-Tish: Music to my ears!

lib/components/DynamicFonts/index.ts
Inbal-Tish: If it's checked later, yes remove it

lib/components/DynamicFonts/FontDownloader.ts
Inbal-Tish, line 112: I would still move it to another function that checks if debug is true and logs fontName, massage and other data

lib/components/DynamicFonts/index.ts
M-i-k-e-l: Removed, but added a permission check

lib/components/DynamicFonts/FontDownloader.ts
M-i-k-e-l: Created a `log` function in each file, not ideal but I don't want to create a util for this at this time

Inbal-Tish (approved):

PR #2757, opened 2023-09-28
Colors - generateColorPalette - avoid reversing on dark mode

Inbal-Tish:
## Description
Colors - generateColorPalette - add the option to avoid reversing on dark mode

## Changelog
Colors - generateColorPalette - add the option to avoid reversing on dark mode


ethanshar (approved):
Approved 
We will have to add unit tests for this logic

Inbal-Tish:
> Approved We will have to add unit tests for this logic

Yes, you're right. I added tests for light mode to make sure the logic didn't change

PR #2760, opened 2023-10-03
DynamicFonts - add isFontFamilyDownloaded

M-i-k-e-l:
## Description
DynamicFonts - add isFontFamilyDownloaded

## Changelog
None

## Additional info
None

### DIFF
--- a/lib/components/DynamicFonts/index.ts
+++ b/lib/components/DynamicFonts/index.ts
@@ -173,4 +173,23 @@ export default class DynamicFonts {
   public async isFontDownloaded(fontName: string, fontExtension: FontExtension): Promise<boolean> {
     return await this.fontDownloader.isFontDownloaded(fontName, fontExtension);
   }
+
+  public async isFontFamilyDownloaded(rootUri: string,
+    fontNames: string[],
+    fontExtension: FontExtension,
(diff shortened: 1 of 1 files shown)


lib/components/DynamicFonts/index.ts
```diff
@@ -173,4 +173,23 @@ export default class DynamicFonts {
   public async isFontDownloaded(fontName: string, fontExtension: FontExtension): Promise<boolean> {
     return await this.fontDownloader.isFontDownloaded(fontName, fontExtension);
   }
+
+  public async isFontFamilyDownloaded(rootUri: string,
```
ethanshar, line 177: It sounds like you named the method with the purpose you need it for while technically this method checks if multiple fonts are downloaded, right?

ethanshar (approved):
Approved. 
But if I understand the code correctly, I'd rename this method, or maybe unify it with the method above

lib/components/DynamicFonts/index.ts
M-i-k-e-l: It checks if the whole font family is downloaded, yes; all the methods in this API are this way.
We can try to think of a different API if you like, but I'm merging for now.

PR #2766, opened 2023-10-16
Fixed chips scroll on press

NitzanWix:
## Description
Fixed bug where clicking on the chips wouldn't scroll to the appropriate section.

## Changelog
Tiny change in scrollToSection function


ethanshar (approved):
Cool. 
It worked before, so I can only assume it's an issue with react-native after we upgraded to rn71
Anyway, looks good 👌

PR #2772, opened 2023-10-17
Fix/chip right icon margin

NitzanWix:
## Description
Added horizontal margins to the right icon in the chip component. The icon was really close to the label when rendering the icon without setting margins manually.

## Changelog
Chip - Added default horizontal margins to right icon.

## Additional info
WOAUILIB-3772

### DIFF
--- a/expoDemo/package.json
+++ b/expoDemo/package.json
@@ -7,36 +7,35 @@
     "eject": "expo eject"
   },
   "dependencies": {
-    "@expo/metro-config": "~0.10.0",
+    "@expo/metro-config": "^0.7.1",
     "@react-native/normalize-color": "^2.1.0",
     "@react-navigation/native": "^6.1.5",
     "@react-navigation/stack": "^6.3.15",
-    "expo": "^49.0.13",
-    "expo-status-bar": "~1.6.0",
+    "expo": "^48.0.4",
+    "expo-status-bar": "~1.4.4",
     "fuzzysearch": "^1.0.3",
     "lodash": "^4.17.21",
     "metro-config": "^0.75.1",
-    "moment": "^2.29.4",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "react-native": "0.72.5",
-    "react-native-gesture-handler": "~2.12.0",
+    "react-native": "0.71.3",
+    "react-native-gesture-handler": "~2.9.0",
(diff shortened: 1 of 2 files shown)


src/components/chip/index.tsx
```diff
@@ -292,6 +292,13 @@ const Chip = ({
           return {
             marginRight: Spacings.s2
           };
+        case 'iconSource':
+          if (label || leftElement || rightElement) {
+            return {
+              marginLeft: Spacings.s1,
+              marginRight: Spacings.s2
+            };
```
ethanshar, line 300: overall It looks much better. 
But I think it mostly solve the right icon
notice that the use can either pass `rightIconSource` or `iconSource` (rendered on the left)
I think those margins works good for right icons and maybe should be opposite for left icon

src/components/chip/index.tsx
```diff
@@ -255,22 +255,25 @@ const Chip = ({
...
+                marginRight: Spacings.s3
+              };
+            }
+            if (rightIconSource || rightElement) {
+              return {
+                marginLeft: Spacings.s3,
+                marginRight: marginFromElement
+              };
+            }
           }
```
ethanshar, line 277: A small suggestion for making this a little more readable
you can define two consts
```
const addLeftMargin = !!(iconSource || leftElement)
const addRightMargin = !!(rightIconSource || rightElement)
```
Then the conditions will much more readable. Also if there will new elements in either left or right we can simply add them to the const definition and that's it

Last, consider using ternary to make things a little shorter, for example 
```
const marginStyle = {
  marginHorizontal: addLeftMargin && addLeftMargin ? Spacings.1 : 0,
 marginLeft: ...
 marginRight: ... 
}
``

NitzanWix: Did you mean
`marginHorizontal: addLeftMargin && addRightMargin ? Spacings.s1 : 0`?
About the last part. I don't really understand what you mean, If I use marginHorizontal and margin left or right in the same object marginHorizontal will always be overridden by the left and right margins. So I don't really know how to use them together in the same object.

expoDemo/package.json
ethanshar, line 1: Can you revert this file changes, it's not relevant to the PR

NitzanWix: Weird. Idk where it came from. Done

ethanshar (approved):

Issue #2770, opened 2023-10-17
Indication for Mandatory TextField


SupriyaPKalghatgi:
## Description

Currently I can find `validate` and its related prop to know if the textfield has unexpected input / input format.
But can there be an indication as to which all fields are mandatory in a given input form?
Something like `*` with the input field, shared in screenshot below

### Related to

- Components

### Screenshots/Video

<img width="335" alt="image" src="https://github.com/wix/react-native-ui-lib/assets/95462949/c91bba09-dfc4-46a0-ba81-e3a4fd988f1d">

### Environment

- Expo version: ^49.0.7,
- React Native version: 0.72.3
- React Native UI Lib version: ^7.8.0

### Affected platforms

- Android
- iOS

SupriyaPKalghatgi:
@ethanshar Can you please check this

ethanshar:
@NitzanWix 
Let's consider adding this as a feature to the TextField component

NitzanWix:
Please refer to the last commit. You can now add a prop "showMandatoryIndication" to the TextField to show a star if the TextField is required (i.e. has a 'required' validator).

SupriyaPKalghatgi:
@NitzanWix The default color of `*` should be `red` if `showMandatoryIndication` is true
Instead I found it to be same color as that of TextInput Label.
Anything that I am missing?

NitzanWix:
Due to our UI guidelines and some layout problems we cannot explicitly expose props for that use case.
**HOWEVER**, a workaround to achieve the behavior you want could be using the highlightString and highlightStyle of the Text component (via the labelProps).
The following TextField will get you what you want.
```
<TextField
  placeholder="Enter School name"
  label="School name"
  validate={'required'}
  labelProps={{highlightString: '*', highlightStyle: {color: 'red'}}}
  showMandatoryIndication
/>
```
**Please mind that if your label contains an '*' it would also be colored.**
@SupriyaPKalghatgi

PR #2775, opened 2023-10-22
Feat/indication for mandatory field

NitzanWix:
## Description
Added mandatory indication to the text field. The star (*) symbol is rendered conditionally if the textfield is required, i.e. has a required validator and showMandatoryIndication prop is present.

## Changelog
TextField - Supports a mandatory star indication. A red star indication will show if the the TextField is required and showMandatoryIndication props is passed.

## Additional info
Issue:
https://github.com/wix/react-native-ui-lib/issues/2770

### DIFF
--- a/src/incubator/TextField/Label.tsx
+++ b/src/incubator/TextField/Label.tsx
@@ -28,7 +28,7 @@ const Label = ({
   const style = useMemo(() => {
     return [styles.label, labelStyle, floatingPlaceholder && styles.dummyPlaceholder];
   }, [labelStyle, floatingPlaceholder]);
-  const shouldRenderIdication = context.isMandatory && showMandatoryIndication;
+  const shouldRenderIndication = context.isMandatory && showMandatoryIndication;
 
   if ((label || floatingPlaceholder) && !forceHidingLabel) {
     return (
@@ -39,7 +39,7 @@ const Label = ({
         recorderTag={'unmask'}
         {...labelProps}
       >
-        {shouldRenderIdication ? label?.concat('*') : label}
+        {shouldRenderIndication ? label?.concat('*') : label}
       </Text>
     );
   }
--- a/src/incubator/TextField/Label.tsx
+++ b/src/incubator/TextField/Label.tsx
@@ -28,6 +28,7 @@ const Label = ({
   const style = useMemo(() => {
     return [styles.label, labelStyle, floatingPlaceholder && styles.dummyPlaceholder];
   }, [labelStyle, floatingPlaceholder]);
+  const shouldRenderIdication = context.isMandatory && showMandatoryIndication;
 
   if ((label || floatingPlaceholder) && !forceHidingLabel) {
     return (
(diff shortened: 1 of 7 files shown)


src/incubator/TextField/types.ts
```diff
@@ -162,13 +162,15 @@ export interface InputProps
    */
   readonly?: boolean;
 }
+type MandatoryField = {showMandatoryIndication?: boolean}
```
ethanshar, line 165: since this type contain a single property you might as well add it to the main TextFieldProp list directly

src/incubator/TextField/index.tsx
```diff
@@ -124,18 +126,42 @@ const TextField = (props: InternalTextFieldProps) => {
...
+      labelStyle={_labelStyle}
+      labelProps={labelProps}
+      floatingPlaceholder={floatingPlaceholder}
+      validationMessagePosition={validationMessagePosition}
+      testID={`${props.testID}.label`}
+    />
+  );
+  const textFieldIsRequired =
+    (typeof others.validate === 'string' && others.validate === 'required') ||
+    (Array.isArray(others.validate) && others.validate.includes('required'));
```
ethanshar, line 142: Let's move this part of logic to the hook that responsible for the field state (useFieldState)
You can add isMandatory to the fieldState object

src/incubator/TextField/index.tsx
```diff
@@ -124,18 +126,42 @@ const TextField = (props: InternalTextFieldProps) => {
     return [inputStyle, styles.dummyPlaceholder];
   }, [inputStyle]);
 
-  return (
-    <FieldContext.Provider value={context}>
-      <View {...containerProps} style={[margins, positionStyle, containerStyle, centeredContainerStyle]}>
+  let textFieldLabel = (
+    <Label
```
ethanshar, line 130: Instead of handling this logic in the main index, move it inside the Label component, so it's more encapsulated and less messy for the index file. 

Moreover, consider not rendering two labels, but concatenate the * with the original label
I would not color the * with red, today we already have a solution for this in private (in TextField.config) where we don't color the * so better keep it aligned with our UI guidelines. 

If you still want to support coloring the asterisk for the community (was it even requested by the user?) there are a bit simpler ways doing that..

ethanshar (changes requested):

src/incubator/TextField/index.tsx
NitzanWix: Done. I had to pass the showMandatoryIndication to the label though.
On regard to the red asterisk there was an image attached to the issue where the asterisk was red so I thought this is what we wanted. I removed the coloring for now.

NitzanWix:
I moved the isMandatory to the useForm hook and the adding of the asterisk and the adding of the asterisk in the label component like we talked.
I added some tests as well - More for practice but they passed.

src/incubator/TextField/Label.tsx
```diff
@@ -28,6 +28,7 @@ const Label = ({
   const style = useMemo(() => {
     return [styles.label, labelStyle, floatingPlaceholder && styles.dummyPlaceholder];
   }, [labelStyle, floatingPlaceholder]);
+  const shouldRenderIdication = context.isMandatory && showMandatoryIndication;
```
ethanshar, line 31: Notice you have a typo here (shouldRenderIdication -> shouldRenderIndication)
Best way to spot typos is to install VSCode extension `Code Spell Checker`

ethanshar (approved):
Wrote small comment on a typo
Anyway approved. Merge after fixing the type

PR #2777, opened 2023-10-23
Fix/colorpicker debounce

NitzanWix:
## Description
Added migrate prop to the ColorPickerDialog. The prop wasn't passed so the new slider was not used.

## Changelog
ColorPickerDialog now uses a new slider for better performance.

## Additional info
WOAUILIB-3631


src/components/slider/index.tsx
ethanshar, line 1: Are the changes in this file are only formatting?

ethanshar (approved):

src/components/slider/index.tsx
NitzanWix: Yes. Should I just revert them?

ethanshar: you can keep it

PR #2784, opened 2023-10-29
Expandable testids

NitzanWix:
## Description
Added testid prop to the expandable section component. The testid is assigned to the touchable opacity component.

## Changelog
ExpandableSection - Added testing identifier prop, testID.

## Additional info
WOAUILIB-3816

### DIFF
--- a/src/components/touchableOpacity/index.tsx
+++ b/src/components/touchableOpacity/index.tsx
@@ -48,6 +48,10 @@ export interface TouchableOpacityProps
    * Custom value of any type to pass on to TouchableOpacity and receive back in onPress callback
    */
   customValue?: any;
+  /**
+   * Testing identifier
+   */
+  testID?: string;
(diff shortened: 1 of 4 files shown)


src/components/touchableOpacity/index.tsx
```diff
@@ -48,6 +48,10 @@ export interface TouchableOpacityProps
    * Custom value of any type to pass on to TouchableOpacity and receive back in onPress callback
    */
   customValue?: any;
+  /**
+   * Testing identifier
+   */
+  testID?: string;
```
ethanshar, line 54: I don't think you don't need to define a testID prop on TouchableOpacity, it already inherit it from React Native's TouchableOpacity

NitzanWix: You're right. Fixed.

ethanshar (approved):

PR #2786, opened 2023-10-31
Picker - add onItemPress

M-i-k-e-l:
## Description
Picker - add onItemPress
Relates to the FontPicker

## Changelog
Picker - add onItemPress

## Additional info
None

### DIFF
--- a/src/components/picker/helpers/usePickerSelection.tsx
+++ b/src/components/picker/helpers/usePickerSelection.tsx
@@ -3,13 +3,14 @@ import _ from 'lodash';
 import {PickerProps, PickerValue, PickerSingleValue, PickerMultiValue, PickerModes} from '../types';
 
 interface UsePickerSelectionProps
-  extends Pick<PickerProps, 'migrate' | 'value' | 'onChange' | 'getItemValue' | 'topBarProps' | 'mode'> {
+  extends Pick<PickerProps, 'migrate' | 'value' | 'onChange' | 'getItemValue' | 'topBarProps' | 'mode' | 'onItemPress'> {
   pickerExpandableRef: RefObject<any>;
   setSearchValue: (searchValue: string) => void;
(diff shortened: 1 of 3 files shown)


src/components/picker/types.tsx
```diff
@@ -95,6 +95,10 @@ export type PickerBaseProps = Omit<NewTextFieldProps, 'value' | 'onChange'> & {
     itemProps: PickerItemProps & {isSelected: boolean; isItemDisabled: boolean},
     label?: string
   ) => React.ReactElement;
+  /**
+   * A callback to be used when an item has been pressed, will stop selection if false is returned
+   */
+  onItemPress?: (item: PickerValue) => Promise<boolean>;
```
ethanshar, line 101: assuming people dont need onItemPress to be asynchronous, typing wise they will still need to return a promise and that can be annoying, no?
if my logic is synchronous, TS won't complain?

src/components/picker/helpers/usePickerSelection.tsx
```diff
@@ -20,7 +21,11 @@ const usePickerSelection = (props: UsePickerSelectionProps) => {
     }
   }, [value]);
 
-  const onDoneSelecting = useCallback((item: PickerValue) => {
+  const onDoneSelecting = useCallback(async (item: PickerValue) => {
+    // Using !(await onItemPress?.(item)) does not work properly when onItemPress is not sent
+    if (onItemPress && !(await onItemPress(item))) {
```
ethanshar, line 26: We already have an onPress callback in PickerItemProps, do you think we should use that?

M-i-k-e-l: The other way I found to do this is shown in [another PR](https://github.com/wix/react-native-ui-lib/pull/2787), you choose

PR #2793, opened 2023-11-06
Add TS support for shorthand gap attribute

GuyHaviv37:
## Description
Add shorthand prop support for [gap property](https://reactnative.dev/docs/flexbox#row-gap-column-gap-and-gap)

## Changelog
View - TS support for shorthand `gap` style.

### DIFF
--- a/src/components/view/index.tsx
+++ b/src/components/view/index.tsx
@@ -36,6 +36,10 @@ export interface ViewProps extends Omit<RNViewProps, 'style'>, ReanimatedProps, 
    */
   height?: string | number;
   /**
+   * Set flexbox gap
+   */
+  gap?: number;
+  /**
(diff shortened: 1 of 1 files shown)


src/components/view/index.tsx
```diff
@@ -35,6 +35,10 @@ export interface ViewProps extends Omit<RNViewProps, 'style'>, ReanimatedProps,
    * TODO: probably isn't needed
    */
   height?: string | number;
+  /**
+   * Set flexbox gap
+   */
+  gap?: number;
```
ethanshar, line 41: Adding the type doesn't necessarily means that it'll work.. and if it works the type should be part of RN's View types 
Passing `gap` works for you? (both on iOS and Android)

PR #2808, opened 2023-11-15
Dialog - BG and overlay should not be static

M-i-k-e-l:
## Description
Dialog - BG and overlay should not be static

## Changelog
Dialog - BG and overlay should not be static

## Additional info
None


Inbal-Tish:
@M-i-k-e-l This fix is the right one, but does fall on the statics issue we have with the DK colors. Though I don't think we can find another solution, we do need to talk about it and decide this is our approach (which means colors could NOT be defined in global scope, styleSheet, or style files)

Inbal-Tish (approved):

Issue #2807, opened 2023-11-14
Textfield built in validators should allow empty strings


brianangulo:
## Description

When using the built in validators for the TextField component such as `validate={['email']}`, `validate={['number']}` it throws as invalid when the fields are cleared to empty string. This happens despite the fields being not required

### Related to

- [X] Components
- [ ] Demo
- [ ] Docs
- [ ] Typings

### Steps to reproduce

Use built in validators on a TextField component

#### Expected behavior

I would expect for validation to be positive if the field is not labeled with the validation 'required' and is cleared/deleted

#### Actual behavior

For fields not marked with 'required' validators but with other built in ones it still shows an error if the field is cleared

### Code snippet

```ts
<TextField
            placeholder="email@email.com..."
            label="Email"
            onChangeText={(val: string) => handleFieldChange('email', val)}
            value={employeeData.email || ''}
            labelStyle={{
              fontSize,
              lineHeight: fontSize,
              color: theme.colors.text,
              marginBottom: '2%',
            }}
            validate={['email']}
            validationMessage={['Must be a valid email']}
            validationMessageStyle={{ color: theme.colors.error }}
            style={{ fontSize, lineHeight: fontSize, color: theme.colors.text }}
            placeholderTextColor={theme.colors.info}
            validateOnChange
            enableErrors
            onChangeValidity={(isValid) =>
              handleOnValidationChange(isValid, 'email')
            }
          />

```

### Screenshots/Video

![Screenshot 2023-11-14 at 1 12 05 PM](https://github.com/wix/react-native-ui-lib/assets/59832188/57589a1c-83f6-49e3-a460-f011bf553113)

### Environment

  expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 13.6
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 18.18.0 - ~/.nvm/versions/node/v18.18.0/bin/node
      Yarn: 1.22.19 - ~/.nvm/versions/node/v18.18.0/bin/yarn
      npm: 9.8.1 - ~/.nvm/versions/node/v18.18.0/bin/npm
      Watchman: 2023.09.25.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.13.0 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 23.0, iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0
    IDEs:
[...]

PR #2811, opened 2023-11-17
fix: Allow empty strings for email and number validation in validators.ts

abdulbasithqb:
## Description

Updated the `email` and `number` validators in `validator.ts` to allow empty strings. Now, empty email and number strings are considered valid.

## Motivation

issue #2807

## Changelog

- Modified the `email` validator logic to allow empty strings explicitly.
- Modified the `number` validator logic to consider empty strings as valid.

## Additional info

This change ensures a more permissive validation for both email and number fields. If a field is not required, no error message will be displayed when users clear their input.
**If you encounter any issues or have further suggestions, please let me know.**

### Results

<table border="0">
 <tr>
    <td><b style="font-size:30px">1. Invalid Email </b></td>
    <td><b style="font-size:30px">2. Invalid Number</b></td>
 </tr>
 <tr>
    <td></td>
    <td></td>
 </tr>
</table>

<table border="0">
 <tr>
    <td><b style="font-size:30px">1. Email  Cleared </b></td>
    <td><b style="font-size:30px">2. Number Cleared</b></td>
 </tr>
 <tr>
    <td></td>
     <td></td>
 </tr>
</table>

### DIFF
--- a/src/components/textField/validators.ts
+++ b/src/components/textField/validators.ts
@@ -8,7 +8,8 @@ const priceRegEx = /^[0-9]{1,9}([.][0-9]{1,2})?$/;
 
 const validators = {
   required: (value = '') => !_.isEmpty(value),
-  email: (value = '') => {
+  email: (value = '') => new EmailValidator().isValid(value),
+  optionalEmail : (value = '') => {
     // Allow empty strings
     if (_.isEmpty(value)) {
       return true;
--- a/src/components/textField/validators.ts
+++ b/src/components/textField/validators.ts
@@ -16,7 +16,7 @@ const validators = {
     return new EmailValidator().isValid(value);
   },
   url: (value = '') => urlRegEx.test(value),
-  number: (value = '') =>  integerRegEx.test(value) || decimalNumberRegEx.test(value),
+  number: (value = '') => integerRegEx.test(value) || decimalNumberRegEx.test(value),
   price: (value = '') => priceRegEx.test(value)
 };
 
--- a/src/components/textField/validators.ts
+++ b/src/components/textField/validators.ts
@@ -16,7 +16,7 @@ const validators = {
     return new EmailValidator().isValid(value);
   },
   url: (value = '') => urlRegEx.test(value),
-  number: (value = '') => _.isEmpty(value) || integerRegEx.test(value) || decimalNumberRegEx.test(value),
+  number: (value = '') =>  integerRegEx.test(value) || decimalNumberRegEx.test(value),
   price: (value = '') => priceRegEx.test(value)
 };
 


brianangulo (approved):

brianangulo:
Thank you! @Inbal-Tish @adids1221 could you please take a peek whenever able 🙏 ? This one has been open for a few months now

src/components/textField/validators.ts
```diff
@@ -8,9 +8,15 @@ const priceRegEx = /^[0-9]{1,9}([.][0-9]{1,2})?$/;
...
+  email: (value = '') => {
+    // Allow empty strings
+    if (_.isEmpty(value)) {
+      return true;
+    }
+    return new EmailValidator().isValid(value);
+  },
   url: (value = '') => urlRegEx.test(value),
-  number: (value = '') => integerRegEx.test(value) || decimalNumberRegEx.test(value),
+  number: (value = '') => _.isEmpty(value) || integerRegEx.test(value) || decimalNumberRegEx.test(value),
```
adids1221, line 19: `integerRegEx.test(value)` allows empty string.
We can remove this test.
```suggestion
  number: (value = '') =>  integerRegEx.test(value) || decimalNumberRegEx.test(value),
```

abdulbasithqb: changes are committed

src/components/textField/validators.ts
```diff
@@ -8,9 +8,15 @@ const priceRegEx = /^[0-9]{1,9}([.][0-9]{1,2})?$/;
...
+  email: (value = '') => {
+    // Allow empty strings
+    if (_.isEmpty(value)) {
+      return true;
+    }
+    return new EmailValidator().isValid(value);
+  },
   url: (value = '') => urlRegEx.test(value),
-  number: (value = '') => integerRegEx.test(value) || decimalNumberRegEx.test(value),
+  number: (value = '') =>  integerRegEx.test(value) || decimalNumberRegEx.test(value),
```
adids1221, line 19:
```suggestion
  number: (value = '') => integerRegEx.test(value) || decimalNumberRegEx.test(value),
```

src/components/textField/validators.ts
```diff
@@ -8,9 +8,15 @@ const priceRegEx = /^[0-9]{1,9}([.][0-9]{1,2})?$/;
 
 const validators = {
   required: (value = '') => !_.isEmpty(value),
-  email: (value = '') => new EmailValidator().isValid(value),
+  email: (value = '') => {
```
adids1221, line 11: I think it could be a breaking change for other users.
I think we should add a new validation `optionalEmail` that will validate this case (empty string) and leave the `email` validator as it is.

adids1221 (commented):
Left another small comment since it could be a breaking change to other users that are using this validation even though the `email` field isn't required.

abdulbasithqb:
> Left another small comment since it could be a breaking change to other users that are using this validation even though the `email` field isn't required.
changes are made as per the suggestion

adids1221 (approved):

PR #2823, opened 2023-11-26
Moved ColorPickerDialog to FC.

NitzanWix:
## Description
Changing the ColorPickerDialog to a FC.

## Changelog
ColorPickerDialog - Changed to FC

### DIFF
--- a/src/components/colorPicker/ColorPickerDialog.tsx
+++ b/src/components/colorPicker/ColorPickerDialog.tsx
@@ -58,13 +58,6 @@ interface Props extends DialogProps {
 }
 export type ColorPickerDialogProps = Props;
 
-interface State {
-  keyboardHeight: number;
-  color: any;
-  text?: string;
-  valid: boolean;
-}
-
 const KEYBOARD_HEIGHT = 216;
 const MODAL_PROPS = {
   supportedOrientations: ['portrait', 'landscape', 'landscape-left', 'landscape-right'] // iOS only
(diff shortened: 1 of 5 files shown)


src/components/colorPicker/ColorPickerDialog.tsx
ethanshar, line 1: General comment on the refactor.. 
We usually keep each component separated in its own file, so I suggest we do the same here so the file won't be too big. 
I suggest renaming the components to something more explicit, for example: instead of Header you can call it PickerDialogHeader and so on.. 

General functions I'd move also to a separate file. 
We usually keep such functions under a presenter file (ColorPickerPresenter) and use these method in the relevant components.

src/components/colorPicker/ColorPickerDialog.tsx
```diff
@@ -58,307 +57,307 @@ interface Props extends DialogProps {
...
-        modalProps={MODAL_PROPS}
-        {...dialogProps}
-      >
-        {this.renderHeader()}
-        {this.renderPreview()}
-        {this.renderSliders()}
-      </Dialog>
-    );
-  }
+  const updateColor = (hex: string) => {
```
ethanshar, line 320: Since you pass this as a prop it's important to wrap it with `useCallback`

ethanshar (commented):
Wrote a few more small comments

ethanshar (approved):

PR #2825, opened 2023-11-26
Revert "Fix/colorpicker debounce (#2777)"

NitzanWix:
This reverts commit af123261a64dd2cbf4f56801f6310119139aa67f.

## Description
Reverting the migration to the new slider in the color picker component.

## Changelog
Removed passing migrate to the ColorPickerDialog.

## Addtional info
I think the other changes beside the migrate prop are just formattings. I get why you said I should do them straight to master.


NitzanWix:
I think the other changes beside the migrate prop are just formattings. I get why you said I should do them straight to master.

PR #2828, opened 2023-11-28
Return old textfield for masked input

ethanshar:
## Description
Return old textfield for masked input until we migrate to the new MaskedInput completely 

## Changelog
Return old textfield for masked input until we migrate to the new MaskedInput completely


ethanshar:
@M-i-k-e-l 
Im working here on returning the old TextField so we won't have to deal with all of this right now
I made this PR, but there's still an issue with ref forwarding, im looking into it
Once it will all work we can hotfix this and not handle all this balagan 🤞

M-i-k-e-l (approved):

PR #2836, opened 2023-12-05
Design Kits issues 

ethanshar:
## Description
I have 2 fixes here
- Our `Colors.generateColorPalette` api is memoized but it didn't handle well the options argument so we got mixed results 
This PR fix it by passing a custom resolver that takes the options into account
- In TabBar I improved the key that rebuilds the components based on orientation, not it will rebuild it also based on labelColor & selectedLabelColor props 
The reason we rebuild it is because this code is heavily based on reanimated and require us to initialize the whole animation

## Changelog
Fix `Colors.generateColorPalette`  memoized result (handle cache properly)
Fix TabController.TabBar not responding to change in labelColor/selectedLabelColor

### DIFF
--- a/src/style/__tests__/colors.spec.js
+++ b/src/style/__tests__/colors.spec.js
@@ -111,7 +111,7 @@ describe('style/Colors', () => {
       uut.getColorTint('#3F88C5', 20);
       uut.getColorTint('#3F88C5', 50);
       uut.getColorTint('#3F88C5', 70);
-      const cachedPalette = uut.generateColorPalette.cache.get('#3F88C5');
+      const cachedPalette = uut.generateColorPalette.cache.get('#3F88C5_');
       expect(cachedPalette).toBeDefined();
       expect(cachedPalette.length).toBe(8);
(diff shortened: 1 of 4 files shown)


src/style/__tests__/colors.spec.js
```diff
@@ -111,7 +111,7 @@ describe('style/Colors', () => {
       uut.getColorTint('#3F88C5', 20);
       uut.getColorTint('#3F88C5', 50);
       uut.getColorTint('#3F88C5', 70);
-      const cachedPalette = uut.generateColorPalette.cache.get('#3F88C5');
+      const cachedPalette = uut.generateColorPalette.cache.get('#3F88C5_');
```
Inbal-Tish, line 114: Why the underscore in the hax string?

Inbal-Tish:
@ethanshar Approved. I left a small comment

Inbal-Tish (approved):

src/style/__tests__/colors.spec.js
ethanshar: Fixed

PR #2838, opened 2023-12-06
Avoid failing rgba API

ethanshar:
## Description
Avoid failing Colors.rgba API on invalid input to avoid app crashes

## Changelog
Avoid failing Colors.rgba API on invalid input to avoid app crashes

### DIFF
--- a/src/style/__tests__/colors.spec.js
+++ b/src/style/__tests__/colors.spec.js
@@ -36,7 +36,7 @@ describe('style/Colors', () => {
     expect(uut.rgba(101, 136, 0.7)).toBe(undefined);
     expect(logServiceSpy).toHaveBeenCalledWith('Colors.rgba fail due to invalid arguments');
     expect(uut.rgba(undefined, 0.2)).toBe(undefined);
-    expect(logServiceSpy).toHaveBeenCalledWith('Colors.rgba fail due to invalid arguments');
+    expect(logServiceSpy).toHaveBeenNthCalledWith(2, 'Colors.rgba fail due to invalid arguments');
   });
 
   it('should handle invalid rgb code', () => {
--- a/src/style/__tests__/colors.spec.js
+++ b/src/style/__tests__/colors.spec.js
@@ -1,8 +1,12 @@
 import uut from '../colors';
+import LogService from '../../services/LogService';
+
 const SYSTEM_COLORS = ['grey', 'white', 'black'];
 const GetColorsByHexOptions = {validColors: SYSTEM_COLORS};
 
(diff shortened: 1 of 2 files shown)


src/style/colors.ts
```diff
@@ -129,7 +130,8 @@ export class Colors {
       blue = validateRGB(p3!);
       opacity = p4;
     } else {
-      throw new Error('rgba can work with either 2 or 4 arguments');
+      LogService.error('Colors.rgba fail due to invalid arguments');
```
M-i-k-e-l, line 133: I think it should be `failed` (also in tests)

src/style/__tests__/colors.spec.js
```diff
@@ -1,8 +1,12 @@
 import uut from '../colors';
+import LogService from '../../services/LogService';
+
 const SYSTEM_COLORS = ['grey', 'white', 'black'];
 const GetColorsByHexOptions = {validColors: SYSTEM_COLORS};
 
 describe('style/Colors', () => {
+  const logServiceSpy = jest.spyOn(LogService, 'error');
+  // jest.mock('../../services/LogService');
```
M-i-k-e-l, line 9: Can be deleted

src/style/__tests__/colors.spec.js
```diff
@@ -29,7 +33,10 @@ describe('style/Colors', () => {
   });
 
   it('should handle wrong number of params', () => {
-    expect(() => uut.rgba(101, 136, 0.7)).toThrow(new Error('rgba can work with either 2 or 4 arguments'));
+    expect(uut.rgba(101, 136, 0.7)).toBe(undefined);
+    expect(logServiceSpy).toHaveBeenCalledWith('Colors.rgba fail due to invalid arguments');
+    expect(uut.rgba(undefined, 0.2)).toBe(undefined);
+    expect(logServiceSpy).toHaveBeenCalledWith('Colors.rgba fail due to invalid arguments');
```
M-i-k-e-l, line 39:
```suggestion
    expect(logServiceSpy).toHaveBeenNthCalledWith(2, 'Colors.rgba fail due to invalid arguments');
```

M-i-k-e-l (approved):

src/style/colors.ts
ethanshar: What do you mean should be failed? 
LogService.error doesn't throw an error

src/style/__tests__/colors.spec.js
ethanshar: true

PR #2839, opened 2023-12-06
Fixing phantom label on the wheel picker.

NitzanWix:
## Description
There were labels near the values of the wheel picker. Because they were white they could only be seen on darker backgrounds. I added opacity = 0 because removing this label was causing the entire layout to change.

## Changelog
WheelPicker - fixed value labels showing near every label.

## Additional info
WOAUILIB-3875

### DIFF
--- a/src/components/WheelPicker/index.tsx
+++ b/src/components/WheelPicker/index.tsx
@@ -222,7 +222,7 @@ const WheelPicker = ({
   const fakeLabelProps = useMemo(() => {
     return {...labelMargins, ...labelProps};
   }, [labelMargins, labelProps]);
-  const itemLabelStyle = useMemo(() => ({...labelStyle, opacity: 0}), [labelStyle]);
+
   const renderItem = useCallback(({item, index}: ListRenderItemInfo<ItemProps>) => {
     return (
       <Item
@@ -235,7 +235,7 @@ const WheelPicker = ({
         {...item}
         disableRTL={shouldDisableRTL}
         fakeLabel={label}
-        fakeLabelStyle={itemLabelStyle}
+        fakeLabelStyle={labelStyle}
         fakeLabelProps={fakeLabelProps}
         centerH={!label}
         onSelect={selectItem}
@@ -249,7 +249,7 @@ const WheelPicker = ({
     fakeLabelProps,
     offset,
     testID,
-    itemLabelStyle,
+    labelStyle,
     label,
     activeTextColor,
     inactiveTextColor,
--- a/src/components/WheelPicker/index.tsx
+++ b/src/components/WheelPicker/index.tsx
@@ -222,7 +222,7 @@ const WheelPicker = ({
   const fakeLabelProps = useMemo(() => {
     return {...labelMargins, ...labelProps};
   }, [labelMargins, labelProps]);
-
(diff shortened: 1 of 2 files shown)


src/components/WheelPicker/index.tsx
```diff
@@ -222,7 +222,7 @@ const WheelPicker = ({
   const fakeLabelProps = useMemo(() => {
     return {...labelMargins, ...labelProps};
   }, [labelMargins, labelProps]);
-
+  const itemLabelStyle = useMemo(() => ({...labelStyle, opacity: 0}), [labelStyle]);
```
M-i-k-e-l, line 225: I'd put it in the `item` itself.
I'd probably write it as `[labelStyle, {opacity}]` but now I've seen [this article](https://medium.com/@fjmorant/styles-and-performance-in-react-native-unlocking-the-best-of-both-worlds-2263632a12bb):
> 4. Best Practices for Styling Performance
...
Flattening Styles: Using arrays for styles can be computationally expensive. Instead, use StyleSheet.flatten to merge style objects.

So maybe more investigation is needed.

NitzanWix:
I moved the entire opacity add to the fake label in the Item component. I used composed instead of flatten. Because its better if the fakeLabelStyle is not passed and because we don't know if the labelStyle is a stylesheet object.

M-i-k-e-l:
> I moved the entire opacity add to the fake label in the Item component. I used composed instead of flatten. Because its better if the fakeLabelStyle is not passed and because we don't know if the labelStyle is a stylesheet object.

Why does it matter?

BTW, `compose` doesn't really do anything really ([link](https://github.com/facebook/react-native/blob/55e4543eea1a2f7a42ed6997c9a692b4fc61550a/packages/react-native/Libraries/StyleSheet/StyleSheet.js#L271-L280)) while `flatten` does ([link](https://github.com/facebook/react-native/blob/55e4543eea1a2f7a42ed6997c9a692b4fc61550a/packages/react-native/Libraries/StyleSheet/flattenStyle.js))

NitzanWix:
Compose returns an array if both are present and only the object if only one is preset. So if only one given then it never creates an array. I didn't see much difference between them other than that

M-i-k-e-l:
> Compose returns an array if both are present and only the object if only one is preset. So if only one given then it never creates an array. I didn't see much difference between them other than that

See the link in [this comment](https://github.com/wix/react-native-ui-lib/pull/2839#discussion_r1436369285) for the reason memoized flatten can be better, and as I stated I am not sure about it and you can search further if this is indeed recommend.

NitzanWix:
I don' really find any major difference between them. Its says that flatten uses something with the id in stylesheet but in some places theres mentions that stylesheet doesn't even use those ids anymore. I'll change to flatten any way.

M-i-k-e-l:
> I don' really find any major difference between them. Its says that flatten uses something with the id in stylesheet but in some places theres mentions that stylesheet doesn't even use those ids anymore. I'll change to flatten any way maybe they'll use it in the future.

I saw somewhere that the docs are not updated to reflect the changed code.
You can see the code is different (I linked it above)

M-i-k-e-l (approved):

PR #2840, opened 2023-12-06
TextField - fix web flexness 2

M-i-k-e-l:
## Description
TextField - fix web flexness 2
Please verify this works for you

## Changelog
TextField - fix web flexness 2

## Additional info
WOAUILIB-3701


AmitShwarts (approved):
💯

PR #2849, opened 2023-12-11
remove static default backgroundColor prop

ethanshar:
## Description
Remove TabBar's default static backgroundColor prop value to avoid race condition when loading design tokens

## Changelog
Remove TabBar's default static backgroundColor prop value to avoid race condition when loading design tokens


M-i-k-e-l (approved):

Inbal-Tish:
@ethanshar Please mark as BREAKING CHANGE

PR #2856, opened 2023-12-17
Maybe fix assets 2

mendyEdri:
## Description
Fix `isSVG` logic on web to ignore base64 assets

## Changelog
Fix `isSVG` logic on web to ignore base64 assets

## Additional info
N/A


src/utils/imageUtils.ts
```diff
@@ -1,14 +1,23 @@
...
-  return typeof source === 'string' || typeof source === 'function' || isSvgUri(source);
+  let isSvg = false;
+  if (typeof source === 'function') {
+    isSvg = true;
+  } else if (isSvgUri(source)) {
+    isSvg = true;
+  } else if (Constants.isWeb && isSvgData(source)) {
+    isSvg = true;
+  }
+  return isSvg;
```
ethanshar, line 20: Why not keep like this? 
```
return (typeof source === 'function') || isSvgUri(source) || (Constants.isWeb && isSvgData(source));
```

src/utils/imageUtils.ts
```diff
@@ -1,14 +1,23 @@
 import get from 'lodash/get';
 import {ImageProps} from 'react-native';
 import Assets from '../assets';
+import {Constants} from 'src/commons/new';
```
ethanshar, line 4: Fix import path to 
`import {Constants} from '../commons/new';`

ethanshar (approved):

PR #2831, opened 2023-11-29
Fix/carousel android rtl jumps

NitzanWix:
## Description
There was a problem with the currentPage calculation. I removed the usage of the getCalcIndex from the goToPage and moved the page calculation in the goToNextPage. On android RTL the children is passed reversed for some reason so I revered so the first item will show first.

## Changelog
Carousel - Fixed Android RTL jumps on random items.

PR #2858, opened 2023-12-18
[HOTFIX] Inverted child inner view on android RTL.

NitzanWix:
## Description
On #2831 I fixed the jumping bugs on android. In order to do that I inverted the ScrollView so we wont have to deal with index calculations. Inverting the ScrollView caused the children to also be inverted so I inverted them back in that case.

## Changelog
Carousel - Fixed inversion on Android RTL.

## Additional info
None

### DIFF
--- a/src/components/carousel/index.tsx
+++ b/src/components/carousel/index.tsx
@@ -381,14 +381,17 @@ class Carousel extends Component<CarouselProps, CarouselState> {
 
       return (
         <View
-          style={{
-            width: pageWidth,
-            height: !horizontal ? pageHeight : undefined,
-            paddingLeft,
-            marginLeft,
-            marginRight,
-            paddingVertical
-          }}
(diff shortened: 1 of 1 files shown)


adids1221:
As we spoke, the Hebrew bug is fixed here but there is another issue.
The carousel seems to invert the initial index in RTL mode on Android ( it will start on index 6 - for example instead of 1 or 0), please check it.
Great work

adids1221 (commented):
As we spoke, the Hebrew bug is fixed here but there is another issue.
The carousel seems to invert the initial index in RTL mode on Android ( it will start on index 6 - for example instead of 1 or 0), please check it.
Great work

adids1221 (approved):
Approved, the Hebrew bug does fixed great work.
Please open a ticket for the index issues in the Carousel on Android.

PR #2862, opened 2023-12-20
Removed initial page calculation

nitzanyiz:
## Description
Initial page on the carousel is not being calculated correctly after the index calculation changes.

## Changelog
Carousel - Fixed initial page not working.

## Additional info
None


adids1221 (commented):
On RTL mode the bug is still reproduced on Android, please check it.

PR #2863, opened 2023-12-21
Web screen size

mendyEdri:
## Description
Support web-device screen size

## Changelog
Changed implementation of Constants screenWidth and screenHeight.

## Additional info
I will remove `useDimension` in another PR since it we have it in mmv and editor and I don't want this fix to be coupled


mendyEdri:
@ethanshar I'm not sure what causes it to fail

src/commons/Constants.ts
```diff
@@ -82,6 +82,12 @@ function setAccessibility() {
 
 setAccessibility();
 
+const WebDeviceFrameSelector = '[data-react-native-web-dimensions="true"]';
```
ethanshar, line 85: Is the selector is something you have in the editor or is it a general selector that web apps have? 

How does it work with both stage and settings panel?

mendyEdri: Yes, it's a check we added to a div wrapping the stage 'device'.
In settings we don't have this problem since the UI is less complex and not involved (as of today) things related to screen size.
Since we don't have a context or something that wraps uilib context, we don't "know" where this will be called from - it's one single uilib for all usages

ethanshar: Let's move the code to private uilib's 
We have Constants overrides

mendyEdri:
Closing in favor of this one:
https://github.com/wix-private/wix-react-native-ui-lib/pull/3415/files

PR #2866, opened 2024-01-01
TextField label issues with enableErrors

ethanshar:
## Description
Avoid hiding TextField label when field is invalid but enableErrors is false

## Changelog
Fix issue with TextField's Label become hidden while field is invalid and enableErrors is false

## Additional info

### DIFF
--- a/demo/src/screens/PlaygroundScreen.tsx
+++ b/demo/src/screens/PlaygroundScreen.tsx
@@ -6,7 +6,7 @@ export default class PlaygroundScreen extends Component {
     return (
       <View bg-grey80 flex padding-20>
         <View marginT-20>
-          <TextField migrate placeholder="Placeholder"/>
+          <TextField placeholder="Placeholder"/>
         </View>
         <Card height={100} center padding-20>
           <Text text50>Playground Screen</Text>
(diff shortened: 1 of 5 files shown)


NitzanWix:
That works. Now when the field is not valid the label become red. Is that what we want? I feel like it should not show any indication about the errors.
Using this:
`
<TextField label="Required" validate={'required'} showMandatoryIndication validationMessage={'required'} enableErrors={false} validateOnChange validationMessagePosition={TextField.validationMessagePositions.TOP}/>
`
Simulator Screen Recording - iPhone 14 Pro Max - 2024-01-02 at 10 01 22

ethanshar:
Let's also confirm this behavior with Diana tomorrow

ethanshar:
@NitzanWix take a look again, I made some fixes

NitzanWix:
Now the label disappears on error. I think its this line

https://github.com/wix/react-native-ui-lib/blob/b193b801ef97c763c594ccc1f080b35962a25b4b/src/components/textField/Label.tsx#L26C3-L26C108


Simulator Screen Recording - iPhone 14 Pro Max - 2024-01-14 at 15 09 05

nitzanyiz:
Snippet to reproduce
```
<TextField
  validate={'required'}
  label="Required"
  showMandatoryIndication
  validateOnChange
  validationMessagePosition="top"
  validationMessage={'this is required'}
  enableErrors={false}
/>
```

PR #2877, opened 2024-01-04
NumberInput - fix floating point error

M-i-k-e-l:
## Description
NumberInput - fix floating point error
If we have another public HF. lets HF this as well (it's small)

## Changelog
NumberInput - fix floating point error

## Additional info
WOAUILIB-3893


adids1221 (approved):
Approved

PR #2885, opened 2024-01-10
Ref Typings

NitzanWix:
## Description
Added RefInterface to the our forwardRef function. Adding this generic sets the type of the ref to the specific interface so when passing ref to a one of our components it will show a ts error when passing an incompatible interface ref.

## Changelog
Added generic typings to the forwardRef and in the asBaseComponent usage.

## Additional info
None

### DIFF
--- a/demo/src/screens/incubatorScreens/IncubatorSliderScreen.tsx
+++ b/demo/src/screens/incubatorScreens/IncubatorSliderScreen.tsx
@@ -23,10 +23,10 @@ const IncubatorSliderScreen = () => {
   const [color, setColor] = useState(COLOR);
   const [alpha, setAlpha] = useState(1);
 
-  const slider = useRef<typeof Incubator.Slider>();
-  const customSlider = useRef<typeof Incubator.Slider>();
-  const negativeSlider = useRef<typeof Incubator.Slider>();
-  const rangeSlider = useRef<typeof Incubator.Slider>();
+  const slider = useRef<Incubator.SliderRef>(null);
+  const customSlider = useRef<Incubator.SliderRef>(null);
+  const negativeSlider = useRef<Incubator.SliderRef>(null);
+  const rangeSlider = useRef<Incubator.SliderRef>(null);
 
   const resetSliders = useCallback(() => {
     slider.current?.reset();
@@ -92,7 +92,6 @@ const IncubatorSliderScreen = () => {
         </Text>
         {renderValuesBox(sliderValue)}
         <Incubator.Slider
-          // @ts-expect-error TODO: need to properly support SliderMethods type to use for ref
           ref={slider}
           onValueChange={onValueChange}
           containerStyle={styles.container}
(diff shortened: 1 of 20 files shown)


demo/src/screens/incubatorScreens/IncubatorSliderScreen.tsx
```diff
@@ -23,10 +23,10 @@ const IncubatorSliderScreen = () => {
   const [color, setColor] = useState(COLOR);
   const [alpha, setAlpha] = useState(1);
 
-  const slider = useRef<typeof Incubator.Slider>();
-  const customSlider = useRef<typeof Incubator.Slider>();
-  const negativeSlider = useRef<typeof Incubator.Slider>();
-  const rangeSlider = useRef<typeof Incubator.Slider>();
+  const slider = useRef<Incubator.SliderRef>(null);
```
ethanshar, line 26: Why do we need to pass null here?

src/components/button/ButtonTypes.tsx
```diff
@@ -27,6 +27,7 @@ export enum ButtonAnimationDirection {
   right = 'right'
 }
 
+
```
ethanshar, line 30: Remove line

src/components/fadedScrollView/index.tsx
```diff
@@ -11,6 +11,7 @@ import useScrollEnabler from '../../hooks/useScrollEnabler';
 import useScrollReached from '../../hooks/useScrollReached';
 import {forwardRef, ForwardRefInjectedProps} from '../../commons/new';
 import {ScrollView as GestureScrollView} from 'react-native-gesture-handler';
+import {ComponentStatics} from 'src/typings/common';
```
ethanshar, line 14: Fix import to be relative

ethanshar: And while here, let's reorder the imports in this file so packages imports are first and internal imports are second

src/components/slider/GradientSlider.tsx
```diff
@@ -213,4 +213,5 @@ GradientSlider.displayName = 'GradientSlider';
 GradientSlider.types = GradientSliderTypes;
 
 // eslint-disable-next-line max-len
-export default asBaseComponent<GradientSliderProps, typeof GradientSlider>(forwardRef(asSliderGroupChild(forwardRef(GradientSlider))));
+// @ts-expect-error
```
ethanshar, line 216: Do you know why we have an error here? 
If it's too complicated we can solve it another time..

src/incubator/Slider/index.tsx
```diff
@@ -15,7 +15,7 @@ import {
 } from './SliderPresenter';
 import Thumb from './Thumb';
 import Track from './Track';
-
+import {ComponentStatics} from 'src/typings/common';
```
ethanshar, line 18: Fix to relative import

ethanshar (commented):
Looks great! 
Wrote few small comments..

demo/src/screens/incubatorScreens/IncubatorSliderScreen.tsx
NitzanWix: When using it without null, the type of the ref is Ref<SliderRef | undefined> which causes an error. This error also occurs with regular react native refs. For example using React Native TextInput.
image

ethanshar (approved):

PR #2890, opened 2024-01-14
TextField - create new driver and refactor tests to it

M-i-k-e-l:
## Description
TextField - create new driver and refactor tests to it

## Changelog
TextField - create new driver and refactor tests to it

## Additional info
None

### DIFF
--- a/src/components/text/Text.driver.new.ts
+++ b/src/components/text/Text.driver.new.ts
@@ -5,7 +5,7 @@ import {usePressableDriver} from '../../testkit/new/usePressable.driver';
 export const TextDriver = (props: ComponentProps) => {
   const driver = usePressableDriver<TextProps>(useComponentDriver(props));
 
-  const getText = (): React.ReactNode | undefined => {
+  const getText = (): React.ReactNode | string | undefined => {
     return driver.getProps().children;
   };
 
--- a/src/components/text/Text.driver.new.ts
+++ b/src/components/text/Text.driver.new.ts
@@ -5,7 +5,7 @@ import {usePressableDriver} from '../../testkit/new/usePressable.driver';
 export const TextDriver = (props: ComponentProps) => {
   const driver = usePressableDriver<TextProps>(useComponentDriver(props));
 
-  const getText = () => {
+  const getText = (): React.ReactNode | undefined => {
     return driver.getProps().children;
   };
 
--- a/src/components/textField/TextField.driver.new.ts
+++ b/src/components/textField/TextField.driver.new.ts
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import {fireEvent} from '@testing-library/react-native';
-import {TextFieldProps} from './index';
+import {TextFieldProps} from './types';
 import {useComponentDriver, ComponentProps} from '../../testkit/new/Component.driver';
 import {TextDriver} from '../text/Text.driver.new';
 
@@ -24,7 +24,7 @@ export const TextFieldDriver = (props: ComponentProps) => {
     testID: `${props.testID}.charCounter`
   });
 
-  const getText = (): string | undefined => {
+  const getValue = (): string | undefined => {
     return driver.getProps().value ?? driver.getProps().defaultValue;
   };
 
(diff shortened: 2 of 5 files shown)


src/components/text/Text.driver.new.ts
```diff
@@ -5,7 +5,7 @@ import {usePressableDriver} from '../../testkit/new/usePressable.driver';
 export const TextDriver = (props: ComponentProps) => {
   const driver = usePressableDriver<TextProps>(useComponentDriver(props));
 
-  const getText = () => {
+  const getText = (): React.ReactNode | undefined => {
```
ethanshar, line 8: Doesn't it suppose to return a string?

src/components/textField/TextField.driver.new.ts
```diff
@@ -0,0 +1,94 @@
+import _ from 'lodash';
+import {fireEvent} from '@testing-library/react-native';
+import {TextFieldProps} from './index';
```
ethanshar, line 3: Can we take typings from TextField types file?

src/components/textField/TextField.driver.new.ts
```diff
@@ -0,0 +1,94 @@
...
+  const validationMsgDriver = TextDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}.validationMessage`
+  });
+  const charCounterDriver = TextDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}.charCounter`
+  });
+
+  const getText = (): string | undefined => {
```
ethanshar, line 27: change to `getValue`

src/components/textField/__tests__/index.driver.spec.tsx
```diff
@@ -1,7 +1,7 @@
 import React, {useState} from 'react';
-import {waitFor} from '@testing-library/react-native';
+import {render} from '@testing-library/react-native';
 import View from '../../../components/view';
-import {TextFieldDriver} from '../TextField.driver';
+import {TextFieldDriver} from '../TextField.driver.new';
```
ethanshar, line 4: Let's remove the old file

src/components/text/Text.driver.new.ts
M-i-k-e-l: I thought so, and then I saw it does not, although I think we do compare it to string in tests, maybe we should return `as string | undefined` so it's more clear?
I am really not sure

src/components/textField/TextField.driver.new.ts
```diff
@@ -0,0 +1,94 @@
+import _ from 'lodash';
+import {fireEvent} from '@testing-library/react-native';
+import {TextFieldProps} from './index';
+import {useComponentDriver, ComponentProps} from '../../testkit/new/Component.driver';
```
NitzanWix, line 4: Is this a good name for the driver arguments? Perhaps ComponentDriverArguments or DriverArguments will be better?

src/components/textField/__tests__/index.driver.spec.tsx
M-i-k-e-l: I think we can't do it yet because we still have other components that depend on the old ones.

src/components/textField/TextField.driver.new.ts
```diff
@@ -0,0 +1,94 @@
+import _ from 'lodash';
+import {fireEvent} from '@testing-library/react-native';
+import {TextFieldProps} from './types';
+import {useComponentDriver, ComponentProps} from '../../testkit/new/Component.driver';
```
NitzanWix, line 4: Is this a good name for the driver arguments? Perhaps ComponentDriverArguments or DriverArguments will be better?

M-i-k-e-l: If you're talking about `ComponentProps` then this is not the PR to change it; feel free to create a PR for it.

src/components/text/Text.driver.new.ts
M-i-k-e-l: As you've mentioned, this can technically be a `Text` child

ethanshar (approved):

PR #2894, opened 2024-01-15
new button driver, image driver, testing

adids1221:
## Description
Button - create new driver and refactor tests to it

## Changelog
Button - create new driver and refactor tests to it

### DIFF
--- a/src/components/button/Button.driver.new.ts
+++ b/src/components/button/Button.driver.new.ts
@@ -1,8 +1,5 @@
 import {ButtonProps} from './ButtonTypes';
-import {useComponentDriver, ComponentProps} from '../../testkit/new/Component.driver';
-import {usePressableDriver} from '../../testkit/new/usePressable.driver';
-import {TextDriver} from '../text/Text.driver.new';
-import {ImageDriver} from '../image/Image.driver.new';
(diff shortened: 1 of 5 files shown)


src/components/button/Button.new.driver.ts
```diff
@@ -0,0 +1,33 @@
+import {ButtonProps} from './ButtonTypes';
```
Inbal-Tish, line 1: The props should be exported from the component IMO

M-i-k-e-l: Ethan commented on my PR tat it should be from the types, not sure why it's named `ButtonTypes` and not `types` but it's not related to this PR

src/components/button/Button.driver.new.ts
```diff
@@ -0,0 +1,34 @@
...
+    testID: `${props.testID}.icon`
+  });
+
+  const getLabel = () => {
+    const exists = (): boolean => {
+      return labelDriver.exists();
+    };
+
+    return {...labelDriver, exists};
+  };
```
M-i-k-e-l, line 23:
```suggestion
  const getLabel = () => {
    return labelDriver;
  };
```

M-i-k-e-l (approved):

PR #2898, opened 2024-01-16
new driver for hint testing

adids1221:
## Description
Hint - create new driver and refactor tests to it

## Additional info


src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,128 @@
...
 
-  it('Test Hint modal is not visible when showHint is false', async () => {
-    const component = <HintTestComponent showHint={false}/>;
-    const driver = new HintDriver({component, testID: 'Hint'});
-    expect(await driver.getHintModal()).toBeFalsy();
+  describe('Test Hint modal visibility:', () => {
+    it('Test Hint modal is not visible when showHint is false', async () => {
+      const renderTree = render(<HintTestComponent showHint={false}/>);
+      const driver = HintDriver({renderTree, testID: TEST_ID});
+      expect(await driver.getModal().exists()).toBeFalsy();
```
M-i-k-e-l, line 89: A note, `exists` is a "more difficult" test than `visible`. you can exist but not be visible (if the user sends `useModal=false` for example the `Modal` will not exist).
It seems to me like if RN's `Modal` will change their implementation this can fail; while testing for `isVisible` should still work.

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,128 @@
...
+      useModal={useModal}
+      icon={useIcon ? settingsIcon : undefined}
+      ref={ref}
+    >
+      <Button round $backgroundDefault/>
+    </Hint>
   );
 };
 
+//TODO: Add test for onPress functionality
```
M-i-k-e-l, line 51: I think you've added this

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,128 @@
...
+    });
+
+    it('should not create touchable overlay driver when onBackgroundPress isn\'t passed', async () => {
+      const renderTree = render(<HintTestComponent showHint/>);
+      const driver = HintDriver({renderTree, testID: TEST_ID});
+      expect(driver.getOverlayTouchable().exists()).toBeFalsy();
+    });
   });
+
+  describe('Test Hint tip', () => {});
```
M-i-k-e-l, line 127: Lets change this to a TODO

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,128 @@
...
+      expect(await driver.getModal().exists()).toBeTruthy();
+    });
   });
 
-  it('Test Hint modal is visible when showHint is true', async () => {
-    const component = <HintTestComponent showHint/>;
-    const driver = new HintDriver({component, testID: 'Hint'});
-    expect(await driver.getElement()).toBeTruthy();
-    expect(await driver.getHintModal()).toBeTruthy();
+  describe('Test Hint onPress & onBackgroundPress', () => {
```
M-i-k-e-l, line 100: These do not have to be together, if you have only a single test for `onBackgroundPress` it can still be on it's own (either as an `it` or as a `describe`)

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,128 @@
...
+    //TODO - add test for onBackgroundTest, should mock measureInWindow functionally to test it
```
M-i-k-e-l, line 101: Didn't you do that?

src/components/hint/index.tsx
```diff
@@ -411,6 +411,7 @@ class Hint extends Component<HintProps, HintState> {
   renderOverlay() {
     const {targetLayoutInWindow} = this.state;
     const {onBackgroundPress, backdropColor, testID} = this.props;
+    console.log(`inside renderOverlay, targetLayoutInWindow:`, targetLayoutInWindow);
```
M-i-k-e-l, line 414: Probably should remove this.

src/components/hint/index.tsx
```diff
@@ -479,8 +483,12 @@ class Hint extends Component<HintProps, HintState> {
         ref={this.hintRef}
       >
         {customContent}
-        {!customContent && icon && <Image source={icon} style={[styles.icon, iconStyle]}/>}
-        {!customContent && <Text recorderTag={'unmask'} style={[styles.hintMessage, messageStyle]}>{message}</Text>}
+        {!customContent && icon && <Image source={icon} style={[styles.icon, iconStyle]} testID={`${testID}.icon`}/>}
+        {!customContent && (
+          <Text recorderTag={'unmask'} style={[styles.hintMessage, messageStyle]}>
```
M-i-k-e-l, line 488: Do we want to add a getter for the message?
The testID is on the container here.

M-i-k-e-l:
@adids1221 I'd also remove the comment from the change log 🙏

M-i-k-e-l:
Also, nice work on the mock, I'm sure that was not easy 👍

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const contentDriver = ViewDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}.message`
+  });
+
+  const hintTouchableDriver = usePressableDriver<TouchableOpacityProps>(useComponentDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}`
+  }));
```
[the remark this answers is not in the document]
adids1221: yes it is, do you think that the naming isn't good ?
It's the `Touchable` part of the component with out the `View` wrapper

jestSetup/jest-setup.js
```diff
@@ -15,6 +16,17 @@ Object.defineProperties = (obj, props) => {
   return obj;
 };
 // =========================================================================
+const OriginalReactComponent = React.Component;
+Object.defineProperty(React, 'Component', {
+  value: class MockedReactComponent extends OriginalReactComponent {
```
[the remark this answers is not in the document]
adids1221: I'll try to ask Niryo about it, but as far as I remember since it's a React component function mock this is what we did.
I'll update.

src/components/hint/__tests__/index.spec.tsx
adids1221: I accomplished it partially, I tested only 1 test case as you can see, we need to add more test coverage here with the press.

adids1221:
@M-i-k-e-l fix most of the comments, about the `onBackgroundPress` I'll try to add more test cases.
And about the mock I'll try to ask Niryo and I'll update.

src/components/hint/Hint.driver.new.ts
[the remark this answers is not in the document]
M-i-k-e-l: Yes, I think it should be exported as `press` and probably something like `pressOnBackground: overlayDriver.press` for the overlay (same as in `Modal`)

jestSetup/jest-setup.js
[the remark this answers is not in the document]
M-i-k-e-l: Ping

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,135 @@
...
+    it('Hint should include message', async () => {
+      const renderTree = render(<HintTestComponent showHint/>);
+      const message = renderTree.getByTestId(`${TEST_ID}.text`).props.children;
+      expect(message).toBe('Hint message to hint things');
+    });
+  });
 
-    const WhiteHint = <HintTestComponent showHint color={Colors.white}/>;
-    const WhiteHintDriver = new HintDriver({component: WhiteHint, testID: 'Hint'});
+  describe('Test Hint modal visibility:', () => {
```
M-i-k-e-l, line 93: You want to test the Hint's visibility not the Modal's

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const getOverlay = () => {
+    return overlayDriver;
+  };
+
+  const hintPress = () => {
+    return hintPressableDriver.press();
+  };
+
+  const hintPressOnBackground = () => {
```
M-i-k-e-l, line 75: This should probably be `pressOnBackground`

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,90 @@
...
+  }));
+
+  const iconDriver = ImageDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}.icon`
+  });
+
+  const getModal = () => {
+    return modalDriver;
+  };
```
M-i-k-e-l, line 49: I think this is an internal implementation detail, you probably want to externalize `isVisible` instead

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const getHintPressable = () => {
+    return hintPressableDriver;
+  };
+
+  const getOverlay = () => {
+    return overlayDriver;
+  };
+
+  const hintPress = () => {
```
M-i-k-e-l, line 71: This should be `press`

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,90 @@
...
+    return StyleSheet.flatten(contentDriver.getProps().style).backgroundColor as ViewStyle;
+  };
+
+  const getMessage = () => {
+    return textDriver.getText();
+  };
+
+  const getHintPressable = () => {
+    return hintPressableDriver;
+  };
```
M-i-k-e-l, line 65: Please remove unused methods

adids1221:
@M-i-k-e-l didn't finished yet, having some issues with the `Modal - isVisble` in the test.
I'll ping you when all will be ready.

adids1221:
@M-i-k-e-l as we spoke we can't use `isVisble` on the Modal since it's not rendering to the screen.

M-i-k-e-l:
Hi @adids1221,
As discussed, please change the `getOverlay` and `modalDriver` (this is not the convention BTW) to specific getters

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const contentDriver = ViewDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}.message`
+  });
+
+  const hintPressableDriver = usePressableDriver(useComponentDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}`
+  }));
```
M-i-k-e-l, line 27: This is simply `driver`, it is a duplicate and should be removed

src/components/hint/Hint.driver.new.ts
M-i-k-e-l: Ping

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,72 @@
...
+    return iconDriver;
+  };
+
+  const getBackgroundColor = (): ViewStyle => {
+    return StyleSheet.flatten(contentDriver.getStyle()).backgroundColor as ViewStyle;
+  };
+
+  const getOnBackgroundPressTouchable = () => {
+    return overlayDriver;
+  };
```
M-i-k-e-l, line 49: This is not the convention I was talking about; if you're using `getOnBackgroundPressTouchable().exists()` the you should export something like `isBackgroundExists()` and not the whole driver.

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const getIsModalVisible = () => {
```
M-i-k-e-l, line 51: IMO there should be an `isVisible` for both `modal` and `overlay`

M-i-k-e-l (changes requested):

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,130 @@
...
+    });
+  });
 
-    const WhiteHintContent = await WhiteHintDriver.getHintContent();
-    color = findStyle('backgroundColor', WhiteHintContent);
-    expect(color).toBe(Colors.white);
+  describe('Test Hint content', () => {
+    it('Hint should include message', async () => {
+      const renderTree = render(<HintTestComponent showHint/>);
+      const message = renderTree.getByTestId(`${TEST_ID}.message.text`).props.children;
```
M-i-k-e-l, line 88: This should be exported from the driver

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,130 @@
...
-    const component = <HintTestComponent showHint={false}/>;
-    const driver = new HintDriver({component, testID: 'Hint'});
-    expect(await driver.getHintModal()).toBeFalsy();
+  describe('Test Hint  visibility:', () => {
+    //Note: Cannot test Hint when showHint is false, the component isn't rendered and return null or the children.
+    it('Test Hint  is visible when showHint is true', async () => {
+      const renderTree = render(<HintTestComponent showHint onBackgroundPress={() => {}}/>);
+      const driver = HintDriver({renderTree, testID: TEST_ID});
+      expect(await driver.exists()).toBeTruthy();
+      expect(await driver.isModalVisible()).toBeTruthy();
```
M-i-k-e-l, line 99: Do we really want this or maybe just a generic `isOpen`? WDYT?

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,67 @@
...
+    return overlayDriver.press();
+  };
+
+  return {
+    getBackgroundColor,
+    getIcon,
+    isBackgroundExists,
+    pressOnBackground,
+    isModalVisible,
+    ...driver,
```
M-i-k-e-l, line 64: I don't like the `...driver` export, maybe export what is actually needed?

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const isBackgroundExists = () => {
+    return overlayDriver.exists();
+  };
+
+  const isModalVisible = () => {
+    return modalDriver.isVisible();
+  };
+
+  const press = () => {
```
M-i-k-e-l, line 50: `press` is `show`? Or can it be both `show` and `hide`?

src/components/hint/__tests__/index.spec.tsx
```diff
@@ -1,69 +1,130 @@
...
-  afterEach(() => {
-    HintDriver.clear();
-  });
+  //TODO: Add test for Hint tip position
+  describe('Test Hint style:', () => {
+    it('Test Hint component background color', async () => {
+      const expectedColor = Colors.$backgroundPrimaryHeavy;
+      const renderTree = render(<HintTestComponent showHint/>);
+      const driver = HintDriver({renderTree, testID: TEST_ID});
+      expect(await driver.getElement()).toBeTruthy();
```
M-i-k-e-l, line 58: I'de change this to something else, what are we really testing here?

src/components/hint/Hint.driver.new.ts
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  const pressOnBackground = () => {
+    return overlayDriver.press();
+  };
+
+  return {
+    getBackgroundColor,
+    getIcon,
+    isBackgroundExists,
```
M-i-k-e-l, line 61: I prefer `isBackgroundVisible`, WDYT?

src/components/hint/Hint.driver.new.ts
M-i-k-e-l: Ping on this

PR #2902, opened 2024-01-17
Automation gap - WOAUILIB-3704

NitzanWix:
## Description
Added automation gap for text alignment on rtl.

## Changelog
N/A

## Additional info
WOAUILIB-3704

### DIFF
--- a/src/components/text/__tests__/index.driver.spec.tsx
+++ b/src/components/text/__tests__/index.driver.spec.tsx
@@ -3,10 +3,13 @@ import {render} from '@testing-library/react-native';
 import View from '../../view';
 import Text from '../index';
 import {TextDriver} from '../Text.driver.new';
-import {StyleSheet, Platform} from 'react-native';
+import {StyleSheet} from 'react-native';
 
 const TEXT_ID = 'text_test_id';
 const TEXT_CONTENT = 'text content';
+beforeEach(() => {
+  jest.resetModules();
+});
 
 function WrapperScreenWithText(textProps: {onPress?: jest.Mock} = {}) {
   const {onPress} = textProps;
@@ -49,16 +52,29 @@ describe('Text', () => {
 jest.mock('react-native/Libraries/ReactNative/I18nManager', () => ({
   isRTL: true
 }));
-jest.mock('react-native/Libraries/Utilities/Platform', () => ({
-  OS: 'android',
-  select: obj => obj.android
-}));
 
 describe('Automation gap - Android', () => {
+  jest.mock('react-native/Libraries/Utilities/Platform', () => ({
+    OS: 'android',
+    select: obj => obj.android
+  }));
   it('Should render text on right on rtl', () => {
     const renderTree = render(<WrapperScreenWithText/>);
     const textDriver = TextDriver({renderTree, testID: TEXT_ID});
     const textStyle = textDriver.getProps().style;
     expect(StyleSheet.flatten(textStyle).textAlign).toEqual('left');
+  });
+});
+
+describe('Automation gap - IOS', () => {
+  jest.mock('react-native/Libraries/Utilities/Platform', () => ({
+    OS: 'ios',
+    select: obj => obj.ios
+  }));
+  it('Should render text on right on rtl', () => {
+    const renderTree = render(<WrapperScreenWithText/>);
+    const textDriver = TextDriver({renderTree, testID: TEXT_ID});
+    const textStyle = textDriver.getProps().style;
+    expect(StyleSheet.flatten(textStyle).writingDirection).toEqual('rtl');
   });
 });
--- a/src/components/text/__tests__/index.driver.spec.tsx
+++ b/src/components/text/__tests__/index.driver.spec.tsx
@@ -7,9 +7,6 @@ import {StyleSheet} from 'react-native';
 
 const TEXT_ID = 'text_test_id';
 const TEXT_CONTENT = 'text content';
-beforeEach(() => {
-  jest.resetModules();
-});
 
 function WrapperScreenWithText(textProps: {onPress?: jest.Mock} = {}) {
   const {onPress} = textProps;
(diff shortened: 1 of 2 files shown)


src/components/text/__tests__/index.driver.spec.tsx
```diff
@@ -44,3 +45,45 @@ describe('Text', () => {
     });
   });
 });
+
+// jest.mock('react-native/Libraries/ReactNative/I18nManager', () => ({
```
ethanshar, line 49: @NitzanWix 
Instead of mocking RN it might be simpler to override/mock our Constants.isRLT value
See here
https://github.com/wix/react-native-ui-lib/blob/master/src/components/sortableGridList/__tests__/usePresenter.spec.tsx

M-i-k-e-l: He's testing the style here, so I don't think that'll work.
Although I'm not sure this is the best way to test this bug, an alternative might be with screenshots and I think we don't want to rely on that here.

NitzanWix: I'm trying to test the following. overriding/mocking in Constants will work for RTL but I want to also mock the platform. Currently this doesn't work and pops a invalid hook usage error.
https://github.com/wix/react-native-ui-lib/blob/068b4d00174bbc13f1ef33dfcb811d9ab7dee369/src/components/text/index.tsx#L195-L203

src/components/text/__tests__/index.driver.spec.tsx
```diff
@@ -1,20 +1,18 @@
...
-import Text from '../index';
 import {TextDriver} from '../Text.driver.new';
+import {StyleSheet} from 'react-native';
 
 const TEXT_ID = 'text_test_id';
 const TEXT_CONTENT = 'text content';
 
 function WrapperScreenWithText(textProps: {onPress?: jest.Mock} = {}) {
   const {onPress} = textProps;
+  const Text = require('../../text').default;
```
M-i-k-e-l, line 11: Why did you move this here?

src/components/text/__tests__/index.driver.spec.tsx
```diff
@@ -44,3 +42,30 @@ describe('Text', () => {
...
+  Constants.isIOS = !isAndroid;
+  Constants.isRTL = isRTL;
+};
+
+describe('Automation gap', () => {
+  it('Should render text on right on rtl - Android', () => {
+    jest.isolateModules(() => {
+      setConstants(true, true);
+      const renderTree = render(<WrapperScreenWithText/>);
+      const textDriver = TextDriver({renderTree, testID: TEXT_ID});
```
M-i-k-e-l, line 57: Consider adding `getDriver`

src/components/text/index.tsx
```diff
@@ -192,14 +191,13 @@ const styles = StyleSheet.create({
   container: {
     backgroundColor: 'transparent',
     color: Colors.$textDefault,
-    ...Platform.select({
-      ios: {
+    ...(Constants.isIOS
```
M-i-k-e-l, line 194: Nice idea!
Did you verify this change works (in the app, not tests)

src/components/text/__tests__/index.driver.spec.tsx
```diff
@@ -44,3 +42,30 @@ describe('Text', () => {
...
   });
 });
+const setConstants = (isAndroid: boolean, isRTL: boolean) => {
+  const {Constants} = require('../../../commons/new');
+  Constants.isAndroid = isAndroid;
+  Constants.isIOS = !isAndroid;
+  Constants.isRTL = isRTL;
+};
+
+describe('Automation gap', () => {
```
M-i-k-e-l, line 52: Consider adding tests for LTR as well

src/components/text/__tests__/index.driver.spec.tsx
NitzanWix: Because the styling is being set when Text is loaded (in StyleSheet.create) I needed a way to make it load again each time I test. Requiring it there and using jest.isolateModules let me do that because its not being cached this way. This is the Same reason for requiring Constants inside setConstants, because Text uses it it loads it before I can change it so this way by loading it before Text does and changing its values Text uses the appropriate values.

src/components/text/index.tsx
NitzanWix: Seems to be working

src/components/text/__tests__/index.driver.spec.tsx
NitzanWix: For text? Or in general?

M-i-k-e-l: The same tests but with LTR

src/components/text/__tests__/index.driver.spec.tsx
```diff
@@ -1,46 +1,71 @@
...
+      {TEXT_CONTENT}
+    </Text>
   );
 }
 
+const getDriver = (textProps: {onPress?: jest.Mock} = {}) => {
+  const {onPress} = textProps;
+  const renderTree = render(<WrapperScreenWithText onPress={onPress}/>);
+  const textDriver = TextDriver({renderTree, testID: TEXT_ID});
+  return {renderTree, textDriver};
```
M-i-k-e-l, line 23: Don't think you need to return `renderTree` here

M-i-k-e-l (approved):
Nice

src/components/text/__tests__/index.driver.spec.tsx
```diff
@@ -1,46 +1,91 @@
...
     });
   });
+  const setConstants = (isAndroid: boolean, isRTL: boolean) => {
+    const {Constants} = require('../../../commons/new');
+    Constants.isAndroid = isAndroid;
+    Constants.isIOS = !isAndroid;
+    Constants.isRTL = isRTL;
+  };
+
+  describe('Automation gap', () => {
```
Inbal-Tish, line 53: Please change the name

PR #2907, opened 2024-01-23
Modal exists even when not visible (automation gap - WOAUILIB-3606)

M-i-k-e-l:
## Description
Modal exists even when not visible (automation gap - WOAUILIB-3606)

## Changelog
None

## Additional info
WOAUILIB-3606

### DIFF
--- a/src/incubator/Dialog/__tests__/index.new.spec.tsx
+++ b/src/incubator/Dialog/__tests__/index.new.spec.tsx
@@ -84,6 +84,7 @@ describe('Incubator.Dialog sanity checks', () => {
     });
     expect(dialogDriver.getModal().isVisible()).toBeFalsy();
   });
+
   it('Should exist only if visible', async () => {
     const onDismiss = jest.fn();
     const component = <TestCase2 onDismiss={onDismiss}/>;
@@ -97,5 +98,12 @@ describe('Incubator.Dialog sanity checks', () => {
     closeButtonDriver.press();
     expect(await dialogDriver.getModal().isVisible()).toBeFalsy();
     expect(onDismiss).toHaveBeenCalledTimes(1);
+  });
+
+  it('Modal exists even when not visible (WOAUILIB-3606)', async () => {
+    const onDismiss = jest.fn();
+    const component = <TestCase2 onDismiss={onDismiss}/>;
+    const {dialogDriver} = getDriver(component);
+    expect(dialogDriver.getModal().exists()).toBeTruthy();
   });
 });
--- a/src/incubator/Dialog/__tests__/index.new.spec.tsx
+++ b/src/incubator/Dialog/__tests__/index.new.spec.tsx
@@ -85,22 +85,22 @@ describe('Incubator.Dialog sanity checks', () => {
     expect(dialogDriver.getModal().isVisible()).toBeFalsy();
   });
 
-  it('Should exist only if visible', async () => {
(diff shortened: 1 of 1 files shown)


src/incubator/Dialog/__tests__/index.new.spec.tsx
```diff
@@ -84,6 +84,7 @@ describe('Incubator.Dialog sanity checks', () => {
     });
     expect(dialogDriver.getModal().isVisible()).toBeFalsy();
   });
+
   it('Should exist only if visible', async () => {
```
NitzanWix, line 88: I probably missed it because I just copied the old tests and change it to the driver.
**I don't think async is needed here**

src/incubator/Dialog/__tests__/index.new.spec.tsx
```diff
@@ -98,4 +99,11 @@ describe('Incubator.Dialog sanity checks', () => {
     expect(await dialogDriver.getModal().isVisible()).toBeFalsy();
     expect(onDismiss).toHaveBeenCalledTimes(1);
   });
+
+  it('Modal exists even when not visible (WOAUILIB-3606)', async () => {
+    const onDismiss = jest.fn();
+    const component = <TestCase2 onDismiss={onDismiss}/>;
+    const {dialogDriver} = getDriver(component);
+    expect(dialogDriver.getModal().exists()).toBeTruthy();
```
NitzanWix, line 107: Maybe we should check that is it not visible first? If some changes the TestCase to be visible for some reason this still passes.
```suggestion
    expect(dialogDriver.getModal().exists()).toBeTruthy();
    expect(dialogDriver.getModal().isVisible()).toBeFalsy())
```

src/incubator/Dialog/__tests__/index.new.spec.tsx
```diff
@@ -98,4 +99,11 @@ describe('Incubator.Dialog sanity checks', () => {
     expect(await dialogDriver.getModal().isVisible()).toBeFalsy();
     expect(onDismiss).toHaveBeenCalledTimes(1);
   });
+
+  it('Modal exists even when not visible (WOAUILIB-3606)', async () => {
```
NitzanWix, line 103: Same here for the async

src/incubator/Dialog/__tests__/index.new.spec.tsx
M-i-k-e-l: Yup, I did the same :)

src/incubator/Dialog/__tests__/index.new.spec.tsx
NitzanWix: Same here for the async
```suggestion
  it('Modal exists even when not visible (WOAUILIB-3606)', () => {
```

src/incubator/Dialog/__tests__/index.new.spec.tsx
NitzanWix: I don't think async is needed here. I probably missed it because I just copied the old tests and change it to the driver.
I also see that this test that I added is not good because I misunderstood the behavior. I'll create a pr to fix.
```suggestion
  it('Should exist only if visible', () => {
```

NitzanWix:
I also see that the test I did doesn't test what I intended to. I'll create a fix.

NitzanWix (approved):

PR #2909, opened 2024-01-23
Fixed dialog test to actually test for existence

nitzanyiz:
## Description
I changed the test so it would test for existence of the dialog instead of testing for the visibility of the modal. I thought they're the same thing.

## Changelog
None

## Additional info
None


M-i-k-e-l (commented):
Technically `exists` and `visible` are not the same.
What triggered you to make this change?
In any case, I'd consider testing for both things

M-i-k-e-l (approved):

PR #2916, opened 2024-01-28
setitem height, number of visible rows as defaults

adids1221:
## Description
WheelPicker driver set default values

## Changelog
WheelPicker driver set default values

### DIFF
--- a/src/components/WheelPicker/WheelPicker.driver.tsx
+++ b/src/components/WheelPicker/WheelPicker.driver.tsx
@@ -1,5 +1,5 @@
 import {FlatListProps} from 'react-native';
-import {WheelPickerProps, WheelPickerItemProps} from './index';
+import {WheelPickerProps, WheelPickerItemProps, ITEM_HEIGHT, NUMBER_OF_VISIBLE_ROWS} from './index';
 import {useComponentDriver, ComponentProps} from '../../testkit/new/Component.driver';
 import {useScrollableDriver} from '../../testkit/new/useScrollable.driver';
 import {TextDriver} from '../../components/Text/Text.driver.new';
@@ -11,10 +11,12 @@ export const WheelPickerDriver = (props: ComponentProps) => {
     renderTree: props.renderTree,
     testID: `${props.testID}.list`
   }));
-  
-  const moveToItem = (index: number, numberOfRows: number, itemHeight: number) => {
+
+  const moveToItem = (index: number,
+    numberOfRows: number = NUMBER_OF_VISIBLE_ROWS,
+    itemHeight: number = ITEM_HEIGHT) => {
     listDriver.triggerEvent('onMomentumScrollEnd', {
-      contentOffset: {x: 0, y: itemHeight * index}, 
+      contentOffset: {x: 0, y: itemHeight * index},
       contentSize: {height: numberOfRows * itemHeight, width: 400},
       layoutMeasurement: {height: 100, width: 400}
     });
--- a/src/components/WheelPicker/index.tsx
+++ b/src/components/WheelPicker/index.tsx
@@ -23,6 +23,8 @@ import {WheelPickerAlign} from './types';
 export {WheelPickerAlign};
 
 const AnimatedFlatList = Animated.createAnimatedComponent<FlatListProps<ItemProps>>(FlatList);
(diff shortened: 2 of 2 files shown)


src/components/WheelPicker/WheelPicker.driver.tsx
```diff
@@ -11,10 +11,12 @@ export const WheelPickerDriver = (props: ComponentProps) => {
     renderTree: props.renderTree,
     testID: `${props.testID}.list`
   }));
-  
-  const moveToItem = (index: number, numberOfRows: number, itemHeight: number) => {
+
+  const moveToItem = (index: number,
+    numberOfRows: number = NUMBER_OF_VISIBLE_ROWS,
```
Inbal-Tish, line 16: This is not the number of visible rows, this is the number of rows i.e. props.items.length

adids1221: fixed it, WDYT about exporting the `ITEMS_HEIGHT` ?
do we need this small export ?
in the WheelPicker testing you have custom values

src/components/WheelPicker/__tests__/index.spec.tsx
```diff
@@ -43,10 +43,10 @@ describe('WheelPicker', () => {
       const renderTree = render(<TestCase/>);
       const driver = WheelPickerDriver({renderTree, testID});
 
-      driver.moveToItem(4, NUM_OF_ROWS, ITEM_HEIGHT);
```
Inbal-Tish, line 46: We can remove the ITEM_HEIGHT and test the defaults or maybe add a test case for the defaults

adids1221: I think adding a test case is the best way

src/components/WheelPicker/WheelPicker.driver.tsx
```diff
@@ -11,10 +11,12 @@ export const WheelPickerDriver = (props: ComponentProps) => {
     renderTree: props.renderTree,
     testID: `${props.testID}.list`
   }));
-  
-  const moveToItem = (index: number, numberOfRows: number, itemHeight: number) => {
+
+  const itemsLength = listDriver.getProps().data?.length ?? 0;
+
+  const moveToItem = (index: number, itemHeight: number, numberOfRows: number = itemsLength) => {
```
Inbal-Tish, line 17: Why did you remove the default ITEM_HEIGHT taken from the component?

src/components/WheelPicker/__tests__/index.spec.tsx
```diff
@@ -43,10 +43,24 @@ describe('WheelPicker', () => {
...
+      driver.moveToItem(4, ITEM_HEIGHT);
       expect(onChange).toHaveBeenCalledWith(4, 4);
 
-      driver.moveToItem(7, NUM_OF_ROWS, ITEM_HEIGHT);
+      driver.moveToItem(7, ITEM_HEIGHT);
+      expect(onChange).toHaveBeenCalledWith(7, 7);
+    });
+
+    it('should call onChange after scrolling ends with default itemHeight and numberOfRows', () => {
+      const itemHeight = 44;
```
Inbal-Tish, line 54: This should not be passed if we want to test the defaults

src/components/WheelPicker/__tests__/index.spec.tsx
```diff
@@ -43,10 +43,22 @@ describe('WheelPicker', () => {
...
-      driver.moveToItem(4, NUM_OF_ROWS, ITEM_HEIGHT);
+      driver.moveToItem(4, ITEM_HEIGHT);
       expect(onChange).toHaveBeenCalledWith(4, 4);
 
-      driver.moveToItem(7, NUM_OF_ROWS, ITEM_HEIGHT);
+      driver.moveToItem(7, ITEM_HEIGHT);
+      expect(onChange).toHaveBeenCalledWith(7, 7);
+    });
+
+    it('should call onChange after scrolling ends with default itemHeight and numberOfRows', () => {
```
Inbal-Tish, line 53: Just move this test before the other one as we test the defaults first

Inbal-Tish (approved):

Inbal-Tish:
@adids1221 Approved with a small comment

PR #2925, opened 2024-01-31
Fix Colors typings

ethanshar:
## Description
Fix Colors.isDark typings to accept undefined argument

## Changelog
Fix Colors.isDark typings to accept undefined argument


adids1221 (approved):
Approved

PR #2929, opened 2024-01-31
new segemntTypography prop

adids1221:
## Description
SegmentControl support in custom typography, new `segmentTypography" prop

## Changelog
SegmentControl support in custom typography

## Additional info

### DIFF
--- a/src/components/segmentedControl/segment.tsx
+++ b/src/components/segmentedControl/segment.tsx
@@ -6,7 +6,7 @@ import {asBaseComponent} from '../../commons/new';
 import TouchableOpacity from '../touchableOpacity';
 import {SegmentedControlProps} from './index';
 
-export type SegmentedControlItemProps = Pick<SegmentedControlProps, 'segmentTypography'> & {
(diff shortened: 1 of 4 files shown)


src/components/segmentedControl/segment.tsx
```diff
@@ -116,7 +118,11 @@ const Segment = React.memo((props: SegmentProps) => {
     >
       {!iconOnRight && renderIcon()}
       {label && (
-        <Reanimated.Text fsTagName={'unmasked'} numberOfLines={1} style={[animatedTextStyle, Typography.text90]}>
+        <Reanimated.Text
+          fsTagName={'unmasked'}
+          numberOfLines={1}
+          style={[animatedTextStyle, Typography.text90, segmentLabelStyle]}
```
ethanshar, line 124: I think it's better to set the animatedTextStyle at the end so it won't get overridden by the user's style..

adids1221:
> Let's add an example for this feature in the example screen as the UX suggested

Iv'e added an example/Switch to change the custom typography.
I'll also go over it with the UX, I'll make some changes if it's needed.

ethanshar (approved):

PR #2941, opened 2024-02-07
icon support in badge

adids1221:
## Description
`Icon` support in badge.
Added 2 new props `badgeProps | hideBadge` to the `Icon` component.

## Changelog
`Icon` support in badge.

### DIFF
--- a/demo/src/screens/componentScreens/IconScreen.tsx
+++ b/demo/src/screens/componentScreens/IconScreen.tsx
@@ -1,13 +1,19 @@
 import React, {useState} from 'react';
-import {Assets, View, Icon, Text, Slider, Switch, GradientSlider, Colors} from 'react-native-ui-lib';
+import {Assets, View, Icon, Text, Slider, GradientSlider, Colors} from 'react-native-ui-lib';
+import {renderBooleanOption} from '../ExampleScreenPresenter';
+
+const DEFAULT_BADGE_SIZE = 20;
+const DEFAULT_PIMPLE_SIZE = 10;
 
 const IconScreen = () => {
   const [size, setSize] = useState(24);
+  const [badgeSize, setBadgeSize] = useState(20);
   const [customSize, setCustomSize] = useState(false);
+  const [customBadgeSize, setCustomBadgeSize] = useState(false);
   const [color, setColor] = useState<string | number>();
   const [customColor, setCustomColor] = useState(false);
   const [useBadge, setUseBadge] = useState(false);
-  const [usePimpleBadge, setUsePimpleBadge] = useState(false);
+  const [usePimple, setUsePimple] = useState(false);
 
   return (
     <View padding-page>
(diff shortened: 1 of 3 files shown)


adids1221:
@Inbal-Tish this PR is mentioned in related PR in the private.
Take a note I changed the `size` type as part of `TODO` I found in the private repo.

src/components/icon/index.tsx
```diff
@@ -18,6 +21,14 @@ export type IconProps = Omit<ImageProps, 'source'> &
...
      */
     assetGroup?: string;
+    /**
+     * Badge props passed down to Badge component
+     */
+    badgeProps?: BadgeProps;
+    /**
+     * Hide badge shown on button
+     */
+    hideBadge?: boolean;
```
Inbal-Tish, line 31: I don't see why this prop is needed... Render badge only if there is a badgeProps prop

demo/src/screens/componentScreens/IconScreen.tsx
```diff
@@ -1,23 +1,25 @@
...
-import {Assets, View, Icon, Text, Slider, Switch, GradientSlider} from 'react-native-ui-lib';
+import {Assets, View, Icon, Text, Slider, Switch, GradientSlider, Colors} from 'react-native-ui-lib';
 
 const IconScreen = () => {
   const [size, setSize] = useState(24);
   const [customSize, setCustomSize] = useState(false);
   const [color, setColor] = useState<string | number>();
   const [customColor, setCustomColor] = useState(false);
+  const [useBadge, setUseBadge] = useState(false);
+  const [usePimpleBadge, setUsePimpleBadge] = useState(false);
```
Inbal-Tish, line 10: The name of the component is 'pimple', not 'pimpleBadge'

demo/src/screens/componentScreens/IconScreen.tsx
```diff
@@ -36,6 +38,20 @@ const IconScreen = () => {
...
       <GradientSlider type={GradientSlider.types.HUE} color={color as string} onValueChange={setColor}/>
       <Text marginT-s2>Custom color: {color || '#000000'}</Text>
+
+      <View marginV-s5>
+        <Text marginR-s2 text80BO>
+          Badge Settings:
+        </Text>
+        <View marginT-s3 row>
+          <Text marginR-s2>Use Badge:</Text>
+          <Switch value={useBadge} onValueChange={setUseBadge}/>
```
Inbal-Tish, line 48: renderBooleanOption function will render both Text and Switch from ExampleScreenPresenter

demo/src/screens/componentScreens/IconScreen.tsx
```diff
@@ -36,6 +38,20 @@ const IconScreen = () => {
       </View>
       <GradientSlider type={GradientSlider.types.HUE} color={color as string} onValueChange={setColor}/>
       <Text marginT-s2>Custom color: {color || '#000000'}</Text>
+
+      <View marginV-s5>
+        <Text marginR-s2 text80BO>
+          Badge Settings:
```
Inbal-Tish, line 44: Don't you want to also display the options to change the icon and badge sizes?

src/components/icon/index.tsx
```diff
@@ -83,7 +130,21 @@ const Icon = forwardRef((props: Props, ref: any) => {
...
 
-  return isSvg(source) ? renderSvg() : renderImage();
+  return (
+    <View>
+      {isSvg(source) ? renderSvg() : renderImage()}
+      {shouldShowBadge && (
+        <Badge
+          pointerEvents={'none'}
+          {...badgeProps}
+          label={getBadgeLabel()}
```
Inbal-Tish, line 140: why not just pass the label from the badgeProps? Why the getter?

src/components/icon/index.tsx
```diff
@@ -54,12 +65,48 @@ const Icon = forwardRef((props: Props, ref: any) => {
...
+    } else {
+      badgePosition.right = -5;
+      badgePosition.top = -5;
+    }
+    return [badgePosition, containerStyle];
+  };
+
+  const getBadgeLabel = () => {
+    const {badgeProps} = props;
+    const label = get(badgeProps, 'label');
```
Inbal-Tish, line 95: why not optional chaining? badgeProps?.label

src/components/icon/index.tsx
```diff
@@ -54,12 +65,48 @@ const Icon = forwardRef((props: Props, ref: any) => {
...
+      badgePosition.top = -2;
+    } else {
+      badgePosition.right = -5;
+      badgePosition.top = -5;
+    }
+    return [badgePosition, containerStyle];
+  };
+
+  const getBadgeLabel = () => {
+    const {badgeProps} = props;
```
Inbal-Tish, line 94: why extract the badgeProps here again? This is a function component, we extract the props at the top

src/components/icon/index.tsx
```diff
@@ -54,12 +65,48 @@ const Icon = forwardRef((props: Props, ref: any) => {
...
   const {margins} = modifiers;
+  const shouldShowBadge = !hideBadge && badgeProps;
   const iconSize = size ? {width: size, height: size} : undefined;
   const shouldFlipRTL = supportRTL && Constants.isRTL;
 
+  const getBadgeStyling = (): StyleProp<ViewStyle> => {
+    const containerStyle = badgeProps?.containerStyle;
+    const badgePosition: StyleProp<ViewStyle> = {
+      position: 'absolute',
+      top: -3
```
Inbal-Tish, line 81: why did you set it to -3 if you're changing it later? I'd set the default here, for the badge, and if it is a pimple override.

src/components/icon/index.tsx
```diff
@@ -54,12 +65,48 @@ const Icon = forwardRef((props: Props, ref: any) => {
...
+    };
+    if (isPimple()) {
+      badgePosition.right = -2;
```
Inbal-Tish, line 84: these values should be a const. Are they not supposed to be dynamic or they should be the same for all icon/badge sizes?

src/components/icon/index.tsx
```diff
@@ -54,12 +65,48 @@ const Icon = forwardRef((props: Props, ref: any) => {
...
+  };
+
+  const getBadgeLabel = () => {
+    const {badgeProps} = props;
+    const label = get(badgeProps, 'label');
+    return label;
+  };
+
+  const getBadgeSize = () => {
+    const size = isPimple() ? 10 : 16;
```
Inbal-Tish, line 100: Why are you setting a default size? The badge and the pimple already have a default, and the user can pass it in the badgeProps if needed

src/components/icon/index.tsx
```diff
@@ -54,12 +65,48 @@ const Icon = forwardRef((props: Props, ref: any) => {
...
+    return badgeProps?.size ? badgeProps?.size : size;
+  };
+
+  const isPimple = () => {
```
Inbal-Tish, line 104: const isPimple = badgeProps && !badgeProps.label

src/components/icon/index.tsx
adids1221: you are right, I used the logic from the private repo - `BadgeButton` logic.

src/components/icon/index.tsx
adids1221: In the last commit I added support for dynamic badge size as we spoke.

Inbal-Tish (approved):

PR #2942, opened 2024-02-07
ColorPicker - fix - hue changes to red after move lightness to white

Inbal-Tish:
## Description
The bug: When moving the lightness slider to white, a #fffff color is set back to the component creating a color of h=0,s=0,l=0 values.
Now when moving the hue slider the color remains grey as the saturation is 0, and when moving the lightness slider the color remains red as the hue is 0.
The solution: Passing the color as HSL instead of converting it to hex allows the new color to keep the values of the hue and saturation in the transition so when moving either slider the color changes from the last values and not from the white color values.

## Changelog
ColorPicker - fix - hue changes to red after moving lightness to white

### DIFF
--- a/src/components/colorPicker/ColorPickerDialog.tsx
+++ b/src/components/colorPicker/ColorPickerDialog.tsx
@@ -1,12 +1,11 @@
 import _ from 'lodash';
 import React, {useCallback, useEffect, useState} from 'react';
 import {LayoutAnimation, StyleSheet, Keyboard, StyleProp, ViewStyle} from 'react-native';
-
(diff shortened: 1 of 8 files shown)


nitzanyiz (approved):
Looks good. Maybe we should create a type for the `string | HSLA` wdyt?

Inbal-Tish:
> Looks good. Maybe we should create a type for the `string | HSLA` wdyt?

Feels forced...

> Looks good. Maybe we should create a type for the `string | HSLA` wdyt?

Feels a bit forced, no?

PR #2947, opened 2024-02-11
Fix prettier fomratter not working

ethanshar:
## Description
Fix prettier fomratter not working

## Changelog
N/A


adids1221 (approved):
Approved

PR #2949, opened 2024-02-12
SectionWheelPicker driver to take testIDs from children

nitzanyiz:
## Description
In case where custom section are passed with custom testID there is no way for us to understand what is the testID without taking it from the element however I fear that if we use that approach if we ever add some component as a child element in of the SectionWheelPicker we will accidentally also give that element as a section driver. So I added a custom testIDs array that will help us handle custom cases. 

## Changelog
None

## Additional info
None


src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
```diff
@@ -1,21 +1,35 @@
...
+  if (customSectionTestIds) {
+    sectionsDrivers = customSectionTestIds.map(testID => {
+      return WheelPickerDriver({
+        renderTree,
+        testID
+      });
+    });
+  } else {
+    // Default sections test ids.
+    sectionsDrivers = React.Children.toArray(driver.getElement().props.children).map((_section, index) => {
```
Inbal-Tish, line 26: Why all of this is necessary if you don't check if the child is of type WheelPicker? What's the difference with iterating over `driver.getProps().children as SectionsWheelPickerProps`?

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
```diff
@@ -1,21 +1,35 @@
...
+};
+
+export const SectionsWheelPickerDriver = (props: ComponentProps & CustomSectionsTestIds) => {
+  const {customSectionTestIds, ...others} = props;
+  const driver = useComponentDriver<SectionsWheelPickerProps>(others);
 
+  let sectionsDrivers: (ReturnType<typeof WheelPickerDriver>)[];
+  const {renderTree} = props;
+  if (customSectionTestIds) {
+    sectionsDrivers = customSectionTestIds.map(testID => {
```
Inbal-Tish, line 18: When iterating over the children we can't get the testID from each prop? The testID is passed to the section (wheel picker) container, no?

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
```diff
@@ -1,21 +1,35 @@
...
+import React from 'react';
 import {useComponentDriver, ComponentProps} from '../../testkit/new/Component.driver';
 import {WheelPickerDriver} from '../WheelPicker/WheelPicker.driver';
 import {SectionsWheelPickerProps} from './index';
 
-export const SectionsWheelPickerDriver = (props: ComponentProps) => {
-  const driver = useComponentDriver<SectionsWheelPickerProps>(props);
-  const sections = driver.getProps().children as SectionsWheelPickerProps;
-  const sectionsDrivers = _.map(sections, (_, index) => {
-    const sectionTestID = `${props.testID}.${index}`;
```
Inbal-Tish, line 10: add `    console.log('sectionWheel: ', section?.props.testID);` here and get the passed testID

Inbal-Tish: I think the only thing we need is to pass the section's testID since the component already sets it if the user didn;t override it

nitzanyiz: This would work. I thought about the case if we ever add some component along side the sections so we will return that other thing as a Wheel driver. Like if we add some label or something.

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
nitzanyiz: We could but for the same reason above.

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
nitzanyiz: Regarding the props. The props that return from getProps are not necessarily SectionsWheelPickerProps they're just the props of the element. (We will probably do something with this function later in the future so I got the props in a different way).
On second look I thinks there might still be a problem here because I'm essentially assuming all the children are WheelPickerSections anyway but thats like the default behavior.
I could change it to always take the testID from the children and hope its a section and not something else we if we ever add anything else there.

nitzanyiz:
I think I'll just change it to take the testID from the child component. This will work...

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
```diff
@@ -1,21 +1,19 @@
...
-    return WheelPickerDriver({
-      renderTree: props.renderTree,
-      testID: sectionTestID
-    });
-  });
+  const sectionsDrivers = React.Children.toArray(driver.getElement().props.children).map((section) => {
+    if (typeof section === 'object' && 'props' in section) {
```
Inbal-Tish, line 9: What is this check 'props' in section?

nitzanyiz: It was needed for TS. React.Children.ToArray return types are unknown are any so It let me use testID and props.
I changed it to be a little more readable because this is kind of weird in second thought.

Inbal-Tish:
@nitzanyiz I am not convinced that this change is necessary. The user is passing the 'section' prop of type WheelPickerProps[] and in SectionsWheelPicker component we are creating the children as WheelPicker elements. Why do we need to check their types?
Seems to me the only change needed is to get the testID from the section instead of creating a new one.

nitzanyiz:
All the if does is to let me take the testID from the children without typescript errors. I could just add some casting to the children but this does the same thing (trying to just cast the children raises other typescript errors because i'm kind of also setting the ground for when we ditch the getProps with the props typing).

Inbal-Tish:
@nitzanyiz Please try this:
`export const SectionsWheelPickerDriver = (props: ComponentProps) => {
  const driver = useComponentDriver<SectionsWheelPickerProps>(props);
  const sections = driver.getElement()?.props?.children as SectionsWheelPickerProps['sections'];
  const sectionsDrivers = _.map(sections, (section, index) => {
    const sectionTestID = `${props.testID}.${index}`;
    return WheelPickerDriver({
      renderTree: props.renderTree,
      testID: section?.testID || sectionTestID
    });
  });

  return {...driver, sections, sectionsDrivers};
};`

nitzanyiz:
This works too. I'll just change to that

Inbal-Tish (approved):

PR #2956, opened 2024-02-19
TextField - fix api file

Inbal-Tish:
## Description
TextField - fix api file

## Changelog
TextField - fix api file


adids1221 (approved):
Approved

PR #2959, opened 2024-02-20
WheelPicker - generic value

Inbal-Tish:
## Description
WheelPicker - generic value

## Changelog
WheelPicker - generic value

## Additional info
ticket 1844


Inbal-Tish:
@M-i-k-e-l Ping

src/components/WheelPicker/index.tsx
```diff
@@ -97,30 +97,34 @@ export interface WheelPickerProps {
...
-  disableRTL,
-  style,
-  children,
-  initialValue = 0,
-  separatorsStyle,
-  testID,
-  faderProps,
-  flatListProps
-}: WheelPickerProps) => {
+const WheelPicker = <T extends string | number = number>(props: WheelPickerProps<T>) => {
```
M-i-k-e-l, line 103: I think we can remove the default here: `<T extends string | number>`
Also, we have this type here as well

src/components/WheelPicker/usePresenter.ts
```diff
@@ -1,59 +1,57 @@
 import _ from 'lodash';
 import React from 'react';
-import {ItemProps} from './Item';
+import {LogService} from '../../services';
 import useMiddleIndex from './helpers/useListMiddleIndex';
+import {ItemProps} from './Item';
 
-export type ItemValueTypes = ItemProps | number | string;
+//TODO: deprecate this type
+export type ItemValueTypes = /* ItemProps |  */number | string;
```
M-i-k-e-l, line 8: We have this type here as well

src/components/WheelPicker/usePresenter.ts
```diff
@@ -1,59 +1,57 @@
...
-interface Presenter {
-  items: ItemProps[];
-  index: number;
-  height: number;
-  getRowItemAtOffset: (offset: number) => RowItem;
-}
-
-const usePresenter = ({
-  initialValue = 0,
+const usePresenter = <T extends string | number = number>({
```
M-i-k-e-l, line 23: We have this type here as well
Don't think we need the default `= number`

src/components/WheelPicker/usePresenter.ts
```diff
@@ -1,59 +1,57 @@
...
   };
-
   const items = children ? extractItemsFromChildren() : propItems || [];
   const middleIndex = useMiddleIndex({itemHeight, listSize: items.length});
 
   const getSelectedValueIndex = () => {
-    if (_.isString(initialValue) || _.isNumber(initialValue)) {
-      return _.findIndex(items, {value: initialValue});
+    if (_.isObject(initialValue)) {
+      LogService.warn('UILib WheelPicker will stop supporting initialValue prop type as an object (ItemProps). Please pass string or number only');
```
M-i-k-e-l, line 43: I'm assuming this is what this refers to: `export type ItemValueTypes = /* ItemProps |  */number | string;`
If we're only starting this migration we should probably only remove the type in the next major when we remove it.

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
```diff
@@ -3,9 +3,9 @@ import {useComponentDriver, ComponentProps} from '../../testkit/new/Component.dr
 import {WheelPickerDriver} from '../WheelPicker/WheelPicker.driver';
 import {SectionsWheelPickerProps} from './index';
 
-export const SectionsWheelPickerDriver = (props: ComponentProps) => {
-  const driver = useComponentDriver<SectionsWheelPickerProps>(props);
-  const sections = driver.getProps().children as SectionsWheelPickerProps;
+export const SectionsWheelPickerDriver = <T extends string | number = number>(props: ComponentProps) => {
```
M-i-k-e-l, line 6: We have this type here as well
Don't think we need the default = number

src/components/sectionsWheelPicker/index.tsx
```diff
@@ -42,9 +43,18 @@ export type SectionsWheelPickerProps = PropsWithChildren<{
  * @gif: https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/SectionsWheelPicker/SectionsWheelPicker.gif?raw=true
  */
 
-const SectionsWheelPicker = (props: SectionsWheelPickerProps) => {
-  const {sections, itemHeight, numberOfVisibleRows, activeTextColor, inactiveTextColor, textStyle, disableRTL, testID} =
-    props;
+const SectionsWheelPicker = <T extends string | number = number>(props: SectionsWheelPickerProps<T>) => {
```
M-i-k-e-l, line 46: We have this type here as well
Don't think we need the default = number

M-i-k-e-l (commented):
Nice work, mostly recurring comments, I did not want to miss them if we make changes

src/components/WheelPicker/usePresenter.ts
Inbal-Tish: We still support it so functionality won't break but they will get TS error that should force them to migrate or add ignore. WDYT?

src/components/WheelPicker/Item.tsx
```diff
@@ -1,23 +1,23 @@
...
+import {useThemeProps} from '../../hooks';
+import Text, {TextProps} from '../text';
+import TouchableOpacity from '../touchableOpacity';
 import {WheelPickerAlign} from './types';
 
 const AnimatedTouchableOpacity = Animated.createAnimatedComponent(TouchableOpacity);
 const AnimatedText = Animated.createAnimatedComponent(Text);
 
-export interface ItemProps {
+export interface ItemProps<T> {
```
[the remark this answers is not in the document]
Inbal-Tish: I'm not sure how to pass the type from the Item either, I'll take a look.
Regarding the DefaultWheelPickerItemType we can use the existing `ItemValueTypes`, no?

src/components/WheelPicker/Item.tsx
```diff
@@ -99,9 +102,9 @@ const WheelPickerItem = memo(({
       )}
     </AnimatedTouchableOpacity>
   );
-});
+};
 
-export default asBaseComponent<InternalProps>(WheelPickerItem);
+export default memo(WheelPickerItem) as typeof WheelPickerItem;
```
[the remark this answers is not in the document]
Inbal-Tish: Don't remember... removed

src/components/WheelPicker/Item.tsx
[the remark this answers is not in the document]
M-i-k-e-l: I guess so (missed that), but not sure about the name, I feel like it should have `default` in it

src/components/WheelPicker/usePresenter.ts
M-i-k-e-l: Generally yes, lets move this to the above thread

src/components/WheelPicker/usePresenter.ts
M-i-k-e-l: We don't usually do that IMO, we don't cause TS errors if we can avoid it

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
M-i-k-e-l: Still have the default

src/components/WheelPicker/usePresenter.ts
Inbal-Tish: I remane it WheelPickerItemValue and defined it in types so I can export it from WheelPicker. I also exported outside

M-i-k-e-l (approved):

PR #2970, opened 2024-03-06
Icon - fix wrapper breaks alignment (center modifier)

Inbal-Tish:
## Description
Icon - fix wrapper breaks alignment (center modifier)

## Changelog
Icon - fix wrapper breaks alignment (center modifier)

## Additional info
ticket 4049


adids1221 (approved):
Approved

PR #2972, opened 2024-03-07
Removed getProps/getElementProps entirely

nitzanyiz:
## Description
Removed the usage of getProps from the basic component driver. To get the element props we now need to use getElement().props.
Changed Text driver getText to return text if it has only one children.

## Changelog
None.

## Additional info
MADS-4058

### DIFF
--- a/src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
+++ b/src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
@@ -4,8 +4,8 @@ import {WheelPickerDriver} from '../WheelPicker/WheelPicker.driver';
 import {SectionsWheelPickerProps} from './index';
 
 export const SectionsWheelPickerDriver = (props: ComponentProps) => {
-  const driver = useComponentDriver<SectionsWheelPickerProps>(props);
-  const sections = driver.getProps().children as SectionsWheelPickerProps;
+  const driver = useComponentDriver(props);
+  const sections = driver.getElementChildren() as SectionsWheelPickerProps;
   const sectionsDrivers = _.map(sections, (_, index) => {
     const sectionTestID = `${props.testID}.${index}`;
     return WheelPickerDriver({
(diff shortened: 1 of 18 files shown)


src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
```diff
@@ -4,8 +4,8 @@ import {WheelPickerDriver} from '../WheelPicker/WheelPicker.driver';
 import {SectionsWheelPickerProps} from './index';
 
 export const SectionsWheelPickerDriver = (props: ComponentProps) => {
-  const driver = useComponentDriver<SectionsWheelPickerProps>(props);
-  const sections = driver.getProps().children as SectionsWheelPickerProps;
+  const driver = useComponentDriver(props);
+  const sections = driver.getElement().children as SectionsWheelPickerProps;
```
M-i-k-e-l, line 8: Why do we need the `SectionsWheelPickerProps`?

M-i-k-e-l (commented):
One small thing

src/components/sectionsWheelPicker/SectionsWheelPicker.driver.tsx
nitzanyiz: This is needed so TS will enable us to get the testID from the children. This is not related to this PR though.

M-i-k-e-l: I don't see a usage of `driver.sections` in private, and I don't see an error in public...

nitzanyiz: Its used here
https://github.com/wix/react-native-ui-lib/blob/43151d5a2e129c3fc9f39bbb8c1fefbb63366536/src/components/sectionsWheelPicker/__tests__/index.spec.tsx#L33-L39

M-i-k-e-l: I still don’t understand why we need the type

M-i-k-e-l (approved):

PR #2973, opened 2024-03-10
Infra/test renderer find style

Inbal-Tish:
## Description
uilib-test-renderer - findStyle - fix TS error: Argument of type 'ReactTestInstance' is not assignable to parameter of type 'Component'

## Changelog
uilib-test-renderer - findStyle - fix TS error: Argument of type 'ReactTestInstance' is not assignable to parameter of type 'Component'


adids1221:
@Inbal-Tish I didn't managed to reproduce the TS error, although `strict` set to true.

adids1221 (approved):

PR #2974, opened 2024-03-10
FloatingButton - add 'fullWidth' prop

Inbal-Tish:
## Description
FloatingButton - add 'fullWidth' prop

## Changelog
FloatingButton - add 'fullWidth' prop

## Additional info
ticket 4057

### DIFF
--- a/src/components/floatingButton/__tests__/index.spec.tsx
+++ b/src/components/floatingButton/__tests__/index.spec.tsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import {render} from '@testing-library/react-native';
+import FloatingButton/* , {FloatingButtonLayouts} */ from '../index';
+import {ButtonDriver} from '../../button/Button.driver.new';
+
+const TEST_ID = 'floating_button';
+const button = {
+  label: 'First'
+};
+const secondaryButton = {
+  label: 'Second'
+};
+// buttonLayout={showVertical ? FloatingButtonLayouts.VERTICAL : FloatingButtonLayouts.HORIZONTAL}
+
+
+const TestCase = (props) => {
+  return <FloatingButton {...props} testID={TEST_ID}/>;
+};
+
+describe('FloatingButton', () => {
+  describe('visible', () => {
+    it('should render a button according to visibility', async () => {
+      const props = {};
+      const renderTree = render(<TestCase {...props}/>);
+      const buttonDriver = ButtonDriver({renderTree, testID: `${TEST_ID}.button`});
+      expect(await buttonDriver.exists()).not.toBeTruthy();
+      
+      renderTree.rerender(<TestCase visible/>);
+      expect(await buttonDriver.exists()).toBeTruthy();
+    });
+  });
+
+  describe('buttons', () => {
+    it('should render a button', async () => {
+      const props = {visible: true};
+      const renderTree = render(<TestCase {...props}/>);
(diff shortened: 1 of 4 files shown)


src/components/floatingButton/floatingButton.api.json
```diff
@@ -16,6 +16,8 @@
       "type": "number",
       "description": "The bottom margin of the button, or secondary button if passed"
     },
+    {"name": "fullWidth", "type": "boolean", "description": "Whether the buttons get the container's full with", "note": "Relevant to a vertical layout with secondary button"},
```
adids1221, line 19:
```suggestion
    {"name": "fullWidth", "type": "boolean", "description": "Whether the buttons get the container's full width", "note": "Relevant to a vertical layout with secondary button"},
```
In the ticket the button in the image seems to be primary, is it relevant only for secondary button?

Inbal-Tish: Right, it's only relevant to vertical layout. I'll fix it

Inbal-Tish: fixed

src/components/floatingButton/index.tsx
```diff
@@ -199,7 +194,8 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
     return (
       <View
         row={!!this.isSecondaryHorizontal}
-        center={!!this.isSecondaryHorizontal}
+        center={!fullWidth}
+        paddingH-16={fullWidth}
```
adids1221, line 198: Why use the `padding` here?
Is it by design?

In the example screen set ti Horizontal layout and play with the `fullWidth` you can see it changes even though it shouldn't.

src/components/floatingButton/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+      expect(await buttonDriver.getLabel().getText()).toEqual(button.label);
+    });
+
+    it('should render secondary button with label', async () => {
+      const props = {visible: true, secondaryButton};
+      const renderTree = render(<TestCase {...props}/>);
+      const buttonDriver = ButtonDriver({renderTree, testID: `${TEST_ID}.secondaryButton`});
+      expect(await buttonDriver.getLabel().getText()).toEqual(secondaryButton.label);
+    });
+  });
```
adids1221, line 61: I think you should test the `fullWidth` prop also, check for the style of the button since this is the feature in this PR eventuality, WDYT ?

Or we should add the test file in separate PR since this PR is for the `fullWidth` feature ?

adids1221 (commented):
Great work, left 2 small comments

src/components/floatingButton/__tests__/index.spec.tsx
Inbal-Tish: The fullWidth prop doesn't effect the Buttons' style, but the container's. How do you suggest to test it?

src/components/floatingButton/index.tsx
Inbal-Tish: I see. Thanks. Fixed

src/components/floatingButton/__tests__/index.spec.tsx
adids1221: We can leave it a side for now.

adids1221 (approved):

PR #2982, opened 2024-03-17
Fix/grid list padding issue

ethanshar:
## Description
Fix GridList issue when it's being rendering in another flat list
The issue is with React Native that for some reason the contentContainerStyle of the inner flatlist (in our case the GridList) doesn't take effect
Also, I fixed some typing issue with GridListItem's itemSize prop

## Changelog
GridList fixes - fix GridListItem type and issue with GridList rended inside a FlatList


adids1221 (approved):
Approved

PR #2983, opened 2024-03-17
Image support in string as source, for svg strings

adids1221:
## Description
Image support strings as type for `source`.
Source now support `svg` image string, requested from the `CTO` and the `web editor` teams.

## Changelog
Image support strings as type for `source`.

### DIFF
--- a/src/components/image/index.tsx
+++ b/src/components/image/index.tsx
@@ -9,8 +9,6 @@ import {
   NativeSyntheticEvent,
   ImageErrorEventData
 } from 'react-native';
-// @ts-expect-error No typings available for 'deprecated-react-native-prop-types'
-import {ImagePropTypes} from 'deprecated-react-native-prop-types';
 import {
   Constants,
   asBaseComponent,
@@ -34,7 +32,7 @@ export type ImageProps = Omit<RNImageProps, 'source'> &
     /**
      * custom source transform handler for manipulating the image source (great for size control)
      */
-    sourceTransformer?: (props: any) => ImagePropTypes.source;
+    sourceTransformer?: (props: any) => ImageSourceType;
     /**
      * if provided image source will be driven from asset name
      */
@@ -126,7 +124,7 @@ class Image extends PureComponent<Props, State> {
   public static overlayTypes = Overlay.overlayTypes;
   public static overlayIntensityType = Overlay.intensityTypes;
 
-  sourceTransformer?: (props: any) => ImagePropTypes;
+  sourceTransformer?: (props: any) => ImageSourceType;
 
   constructor(props: Props) {
     super(props);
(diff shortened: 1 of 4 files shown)


src/utils/imageUtils.ts
```diff
@@ -1,15 +1,14 @@
...
+import type {ImageSourceType} from '../components/image';
 
-export function isSvgUri(source?: ImageProps['source']) {
+export function isSvgUri(source?: ImageSourceType) {
   // @ts-expect-error
   return typeof source === 'object' && source?.uri?.endsWith?.('.svg');
 }
 
-export function isSvg(source?: ImageProps['source']) {
-  return typeof source === 'function' || isSvgUri(source) || (Constants.isWeb && isSvgData(source));
```
AmitShwarts, line 12: SVG data relevant also on mobile 🔥 🔥 🔥

Inbal-Tish:
@adids1221 I can't check out this specific branch... Also, why is that a hotfix?

src/components/image/index.tsx
```diff
@@ -95,13 +98,14 @@ export type ImageProps = RNImageProps &
      * The image height
      */
     height?: string | number;
+    source: ImageSourceType;
```
Inbal-Tish, line 101: What about other props, like `errorSource`, that use the type `ImagePropTypes.source`?

Inbal-Tish: It could have been nice to get rid of the `ImagePropType` altogether...

Inbal-Tish:
@adids1221 Worth adding tests at least to the ImageUtils

src/utils/imageUtils.ts
```diff
@@ -22,7 +21,7 @@ export function getAsset(assetName = '', assetGroup = '') {
   return get(Assets, `${assetGroup}.${assetName}`);
 }
 
-function isSvgData(source?: ImageProps['source']) {
-  const sourceString = (source as string);
+function isSvgData(source?: ImageSourceType) {
+  const sourceString = source as string;
```
Inbal-Tish, line 25: Why cast to a string instead of conditioning it? I think it will be clearer

src/components/image/index.tsx
```diff
@@ -122,7 +126,7 @@ class Image extends PureComponent<Props, State> {
   public static overlayTypes = Overlay.overlayTypes;
   public static overlayIntensityType = Overlay.intensityTypes;
 
-  sourceTransformer?: (props: any) => ImagePropTypes.source;
+  sourceTransformer?: (props: any) => ImagePropTypes;
```
Inbal-Tish, line 129: Did you mean `ImageSourceType`?

src/components/image/index.tsx
Inbal-Tish: You should change the types in the props declaration as well...

Inbal-Tish (approved):

PR #2986, opened 2024-03-18
fix: [MADS-4071] clear picker value on web

AmitShwarts:
## Description

when passing undefined after a value already selected, it's not being cleared on the UI.

## Changelog

fix: clearing picker value on web

## Additional info

MADS-4071

### DIFF
--- a/src/components/picker/__tests__/index.spec.tsx
+++ b/src/components/picker/__tests__/index.spec.tsx
@@ -78,7 +78,7 @@ describe('Picker', () => {
     describe('Test selection', () => {
       it('Should select a single item', () => {
         const driver = getDriver();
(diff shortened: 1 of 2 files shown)


adids1221:
@M-i-k-e-l seems good to me, I also checked it on `TextField`.
 and it's not looking something risky, WDYT ?

M-i-k-e-l:
> @M-i-k-e-l seems good to me, I also checked it on `TextField`. and it's not looking something risky, WDYT ?
> 
> * checking it now on private *

Can't say I like it but not sure there's a different way, I think it's an issue with RN's `TextInput`

AmitShwarts:
can we please hotfix it?

PR #2988, opened 2024-03-19
Incubator Slider - fix step interpolation

Inbal-Tish:
## Description
Incubator Slider - fix step interpolation

to reproduce:
`      
        <Incubator.Slider
          value={sliderValue}
          onValueChange={setSliderValue}
          step={1}
          minimumValue={12}
          maximumValue={50}
        />
`
Slider or press track to see the slider jumps.

## Changelog
Incubator Slider - fix step interpolation

## Additional info
ticket 4070

### DIFF
--- a/src/incubator/Slider/SliderPresenter.ts
+++ b/src/incubator/Slider/SliderPresenter.ts
@@ -70,6 +70,6 @@ export function getStepInterpolated(
   'worklet';
   const outputRange = [0, trackWidth];
   const inputRange =
-    minimumValue < 0 ? [0, Math.abs(minimumValue) - Math.abs(maximumValue)] : [0, maximumValue - minimumValue];
+    minimumValue < 0 ? [0, Math.abs(minimumValue) + maximumValue] : [0, maximumValue - minimumValue];
   return interpolate(stepXValue.value, inputRange, outputRange);
 }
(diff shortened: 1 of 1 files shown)


src/incubator/Slider/SliderPresenter.ts
```diff
@@ -58,7 +58,6 @@ export function validateValues(props: SliderProps) {
 export function getStepInterpolated(trackWidth: number, minimumValue: number, maximumValue: number, stepXValue: SharedValue<number>) {
   'worklet';
   const outputRange = [0, trackWidth];
-  const inputRange = minimumValue < 0 ? 
-    [Math.abs(maximumValue), Math.abs(minimumValue)] : [minimumValue, maximumValue];
+  const inputRange = minimumValue < 0 ? [0, Math.abs(minimumValue)] : [0, maximumValue];
```
ethanshar, line 61: I see that it works, but Im not sure I understand the bug. 
Can you explain what happens here that caused this?

Inbal-Tish: Ok, I added refinement to explain: The interpolation finds a value in one range at a different range. In this case, the value is the `stepXValue`, the x point equal to the step. This value is used to calculate the new location from the current location - where to jump to. Since this is an x value it doesn't care if the initial value is 12, it needs to start from 0 to give a value of x, so instead of defining the range 12 to 50, I'll define it as 0 to 38.

ethanshar: 👍

ethanshar (approved):

PR #2989, opened 2024-03-20
WheelPicker - fix item snap to vertical center

Inbal-Tish:
## Description
WheelPicker - fix item snap to vertical center

## Changelog
WheelPicker - fix item snap to vertical center

## Additional info
ticket 4032, 3661

### DIFF
--- a/src/components/WheelPicker/index.tsx
+++ b/src/components/WheelPicker/index.tsx
@@ -335,6 +335,8 @@ const WheelPicker = ({
     );
   }, []);
 
+  const offsets = items.map((_, i) => i * itemHeight);
+
   return (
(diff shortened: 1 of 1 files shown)


adids1221 (approved):

src/components/WheelPicker/index.tsx
```diff
@@ -335,6 +335,8 @@ const WheelPicker = ({
     );
   }, []);
 
+  const offsets = items.map((_, i) => i * itemHeight);
```
ethanshar, line 338: @Inbal-Tish 
Shouldn't we memoize this? 
This will generate a new list each render and might affect performance

Inbal-Tish: Yes, sure

PR #2994, opened 2024-03-25
Incubator.Slider - fix 'value' updates

Inbal-Tish:
## Description
Incubator.Slider - fix 'value' updates.
When value updates the slider should update position.

## Changelog
Incubator.Slider - fix 'value' updates

## Additional info
ticket 4073

### DIFF
--- a/src/incubator/Slider/index.tsx
+++ b/src/incubator/Slider/index.tsx
@@ -271,7 +271,7 @@ const Slider = React.memo((props: Props) => {
   }, [defaultThumbStyle, thumbStyle]);
 
   const onValueChangeThrottled = useCallback(_.throttle(value => {
-    if (!didValueUpdate.current) { // don't invoke onChange when slider's value changes to prevent updates loop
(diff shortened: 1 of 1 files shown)


src/incubator/Slider/index.tsx
```diff
@@ -235,37 +237,45 @@ const Slider = React.memo((props: Props) => {
...
+  useEffect(() => {
+    if (!thumbStyle) {
+      _thumbStyle.value = StyleUtils.unpackStyle(defaultThumbStyle, {flatten: true});
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [defaultThumbStyle, thumbStyle]);
+
   const onValueChangeThrottled = useCallback(_.throttle(value => {
-    onValueChange?.(value);
+    if (!didValueUpdate.current) { // don't invoke onChange when slider's value changes to prevent updates loop
```
Inbal-Tish, line 274:
```suggestion
    if (!didValueUpdate.current) { // NOTE: fix for GradientSlider (should be removed after fix in the GradientSlider component): don't invoke onChange when slider's value changes to prevent updates loop
```

ethanshar (approved):

PR #2995, opened 2024-03-25
Image\Overlay - support borderRadius

M-i-k-e-l:
## Description
Image\Overlay - support borderRadius (only via the prop, not via style)

## Changelog
Image\Overlay - support borderRadius (only via the prop, not via style)

## Additional info
Ticket 4074


adids1221 (approved):
Approved, you need to fix the build

PR #2997, opened 2024-03-26
removed getProps from FloatingButton tests

nitzanyiz:
## Description
Removed getProps from FloatingButton tests.

## Changelog
None

## Additional info
None

### DIFF
--- a/src/components/floatingButton/__tests__/index.spec.tsx
+++ b/src/components/floatingButton/__tests__/index.spec.tsx
@@ -72,7 +72,7 @@ describe('FloatingButton', () => {
       const renderTree = render(<TestCase {...props}/>);
       const buttonDriver = ButtonDriver({renderTree, testID: `${TEST_ID}.button`});
 
-      expect(buttonDriver.getElement().props?.style?.marginBottom).toBe(Spacings.s8);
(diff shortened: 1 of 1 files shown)


src/components/floatingButton/__tests__/index.spec.tsx
```diff
@@ -19,7 +19,7 @@ const TestCase = (props) => {
 };
 export const FloatingButtonDriver = (props: ComponentProps) => {
   const driver = useComponentDriver(props);
-  const getStyle = () => driver.getProps().style as ViewStyle;
+  const getStyle = () => driver.getElement().props.style as ViewStyle;
```
Inbal-Tish, line 22: Isn't props optional? Don't you need to do `driver.getElement().props**?**.style`?

nitzanyiz: I think they're not. Its always a component that is captured so it has testId so props exists.

nitzanyiz: Plus according to the type `ReactTestInstance` props is not optional so they probably promise us there will be props.

Inbal-Tish: Ok...

Inbal-Tish (approved):

PR #3005, opened 2024-03-28
Assets is now category in the release notes

adids1221:
## Description
Release notes script now accepts `categories` to make it generic for private `categories`.
The script will support any category in the structure of: `{name: 'category', branch: 'cat/', title: 'title for the release notes'}`

## Changelog
none


Inbal-Tish:
@adids1221 Can we do the same to intl?

scripts/prReleaseNotesCommon.js
```diff
@@ -99,12 +100,14 @@ function getPRsByType(PRs) {
       fixes.push(pr);
     } else if (pr.branch.startsWith('infra/')) {
       infra.push(pr);
+    } else if (pr.branch.startsWith('assets/') || pr.branch.startsWith('Assets/')) {
```
ethanshar, line 103: instead of checking both `assets` and `Assets` you can use toLowerCase on the branch name and check only against `assets`

PR #3012, opened 2024-04-03
bottom validation message should render only if validation message sent

adids1221:
## Description
`TextField` bottom view should render the `validationMessage` only if the user passed `validationMessage`.
Currently the `validationMessage` renders most of the times because the default value for `validationMessagePosition` is `bottom` and break the alignment when the user passes `bottomAccessory` by it self.

## Changelog
`TextField` bottom view should render the `validationMessage` only if the user passed `validationMessage`.

### DIFF
--- a/src/components/textField/index.tsx
+++ b/src/components/textField/index.tsx
@@ -198,7 +198,7 @@ const TextField = (props: InternalTextFieldProps) => {
           {/* </View> */}
         </View>
         <View row spread>
-          {validationMessagePosition === ValidationMessagePosition.BOTTOM && (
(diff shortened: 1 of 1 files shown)


Inbal-Tish (approved):

adids1221:
As we spoke it's doesn't break it, it works fine 👍

ethanshar:
@adids1221  @Inbal-Tish 
We are reverting this change. 
The validation message was meant to be rendered in any case because it's keeping the space for the message. Otherwise we'll have "jumps" in the screen

PR #3013, opened 2024-04-03
Upgrade node version in nvmrc

ethanshar:
## Description
Upgrade node version we use in buildkite CI to v16
This is order to use `yarn` in our CI instead of npm 

## Changelog
N/A


M-i-k-e-l (approved):
Why not make the yarn changes together with this one?

PR #3019, opened 2024-04-08
fixed onLayout measurements on web

adids1221:
## Description
`TabBarController` fix item size on web.
`TabBarItem` flex issue, setting `flex to 1` on web only if `spreadItems` is set to true.
Handeling the style of the `TabBarItem` from `TabBar`, the `style` passed via props `override` it.

## Changelog
`TabBarController` fix item size on web.

## Additional info
ticket 4115

### DIFF
--- a/src/components/tabController/TabBar.tsx
+++ b/src/components/tabController/TabBar.tsx
@@ -202,6 +202,7 @@ const TabBar = (props: Props) => {
   });
 
   const tabBarItems = useMemo((): ReactNode => {
+    const flex = Constants.isWeb ? (spreadItems ? 1 : undefined) : 1;
     return _.map(items, (item, index) => {
       return (
         <TabBarItem
@@ -216,6 +217,7 @@ const TabBar = (props: Props) => {
           activeBackgroundColor={activeBackgroundColor}
           {...item}
           {...context}
+          style={[{flex}, item?.style]}
           key={`${index}_${item.label}`}
           index={index}
           onLayout={onItemLayout}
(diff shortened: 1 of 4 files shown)


src/components/tabController/TabBar.tsx
```diff
@@ -216,6 +217,7 @@ const TabBar = (props: Props) => {
           activeBackgroundColor={activeBackgroundColor}
           {...item}
           {...context}
+          style={[{flex}, item?.style]}
```
ethanshar, line 220: Can you memoize this style, it might affect performance if you are passing inline array style

src/components/tabController/TabBar.tsx
```diff
@@ -201,8 +201,16 @@ const TabBar = (props: Props) => {
     }
   });
 
+  const tabBarItemStyle = useMemo(() =>
```
ethanshar, line 204: Is it working? 
You are memoizing a function that returns a value and not the value

adids1221: I had an issue, changed it to `useCallback` and passed the `tabBarItemStyle` as dependency to the `tabBarItems`.
Checked on mobile and web.

adids1221:
@ethanshar as we spoke, passing `spreadItems` to `TabBarItem` and moved the logic there also.

src/components/tabController/TabBarItem.tsx
```diff
@@ -8,11 +8,13 @@ import {Colors, Typography, Spacings} from '../../style';
 import Badge, {BadgeProps} from '../badge';
 import View from '../view';
 import TabBarContext from './TabBarContext';
+import Constants from '../../commons/Constants';
+import {TabControllerBarProps} from './TabBar';
```
ethanshar, line 12: change to `import type {TabControllerBarProps} from './TabBar';`
It might have a cyclic require because TabBar imports TabBarItem and TabBarItem imports TabBar
Maybe just importing with `type` is ok

ethanshar (approved):
Wrote a small comment, anyway approved

PR #3020, opened 2024-04-08
Slider/add throttleTime prop

liatnetach:
## Description

Add throttleTime prop in Slider component

### DIFF
--- a/src/incubator/Slider/index.tsx
+++ b/src/incubator/Slider/index.tsx
@@ -193,7 +193,7 @@ const Slider = React.memo((props: Props) => {
     accessible = true,
     testID,
     enableThumbShadow = true,
-    throttleTime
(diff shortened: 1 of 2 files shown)


src/incubator/Slider/index.tsx
```diff
@@ -278,10 +283,22 @@ const Slider = React.memo((props: Props) => {
     }
   }, 200), [onValueChange]);
 
+  const _onValueChange = useCallback((value: number) => {
```
Inbal-Tish, line 286: Why not pass the throttle wait (200/100 to 0) instead of having code duplications?

liatnetach: added throttleTime prop instead 🙏

ethanshar (approved):

PR #3029, opened 2024-04-16
Fix/wheel picker initial value on change

nitzanyiz:
## Description
The WheelPicker was not invoking `onChange` if the `initialValue` was changed right after `onChange` was called. I added a handling for wether the drag is a user or a programatic scroll.

## Changelog
WheelPicker - Fix onChange not called when initial value is changed.

## Additional info
MADS-4145

### DIFF
--- a/src/components/WheelPicker/index.tsx
+++ b/src/components/WheelPicker/index.tsx
@@ -148,7 +148,7 @@ const WheelPicker = <T extends WheelPickerItemValue>(props: WheelPickerProps<T>)
     preferredNumVisibleRows: numberOfVisibleRows
   });
 
-  const prevInitialValue = useRef(initialValue);
+  const shouldInvokeOnChange = useRef(false);
   const prevIndex = useRef(currentIndex);
   const [flatListWidth, setFlatListWidth] = useState(0);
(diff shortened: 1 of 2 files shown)


src/components/WheelPicker/index.tsx
```diff
@@ -148,7 +148,7 @@ const WheelPicker = <T extends WheelPickerItemValue>(props: WheelPickerProps<T>)
     preferredNumVisibleRows: numberOfVisibleRows
   });
 
-  const prevInitialValue = useRef(initialValue);
+  const shouldInvokeOnChange = useRef(false);
```
ethanshar, line 151: It's better that the flag will represent the edge case and not the common case. 
Meaning that `false` will be the usual value, so it's better to name it something like `shouldSkipNextOnChange` or something similar. 
This way its value will be usually false and only when turning it on set it to true.

ethanshar (commented):
Looks good, wrote small comment on the flag

src/components/WheelPicker/index.tsx
nitzanyiz: Changed

ethanshar (approved):

PR #3030, opened 2024-04-16
changed driver props typing

nitzanyiz:
## Description
Niryo said he need the type here to change so it can integrate with amino

## Changelog
ComponentDriver - Changed props typings.

## Additional info
MADS-4062

### DIFF
--- a/src/testkit/new/Component.driver.ts
+++ b/src/testkit/new/Component.driver.ts
@@ -1,8 +1,8 @@
 import {ReactTestInstance} from 'react-test-renderer';
-import {RenderResult} from '@testing-library/react-native';
+import {within} from '@testing-library/react-native';
 
 export interface ComponentProps {
-  renderTree: RenderResult;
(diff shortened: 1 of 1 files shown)


src/testkit/new/Component.driver.ts
```diff
@@ -1,8 +1,8 @@
 import {ReactTestInstance} from 'react-test-renderer';
-import {RenderResult} from '@testing-library/react-native';
+import {within} from '@testing-library/react-native';
 
 export interface ComponentProps {
-  renderTree: RenderResult;
+  renderTree: ReturnType<typeof within>;
```
ethanshar, line 5: Since it looks strange, I would add a `// Notes:` that explain in short why it was necessary

nitzanyiz: Added.

ethanshar (approved):

PR #3041, opened 2024-04-17
Revert to #3012

nitzanyiz:
## Description
Revert to #3012 . It was causing layouting problems with modules.

## Changelog
TextField - Reverted validation message block not rendered by default.

## Additional info
MADS-4147, MADS-4143


adids1221 (approved):
Approved

PR #3050, opened 2024-05-01
Bug fixed TextField style error: right operand of 'in' is not an object

legend-oe:
## Description
If you use rn-ui-lib with NativeWind (Tailwind CSS), TextField not work with error:
```
ERROR  TypeError: right operand of 'in' is not an object

This error is located at:
    in CssInteropComponent (created by Incubator.TextField)
    in Incubator.TextField (created by TextField)
    in RCTView (created by View)
    in View
    in View (created by TextField)
```

I found and fixed 2 bugs.
1. Problem with style Platform.OS === 'web 
2. Problem with style hasValue and centred.

## Changelog
Fixed TextField + NativeWind style error

### DIFF
--- a/src/components/textField/Input.tsx
+++ b/src/components/textField/Input.tsx
@@ -62,7 +62,7 @@ const Input = ({
   return (
     <TextInput
       fsTagName={recorderTag}
-      style={[styles.input, !!inputColor && {color: inputColor}, style, Constants.isWeb && styles.webStyle]}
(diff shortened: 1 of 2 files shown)


gyanbhartip:
Could someone please explain why does this fix work? Even if short-circuiting is used, it would return a false instead of undefined (after the solution). But Does react-native not accept false as a value in style array? Why would it fail?

PR #3051, opened 2024-05-01
Assets Typing

nitzanyiz:
## Description
Added typing to `loadAssetsGroup`. The function now asserts the type of assets with the object and path passed to the function. **Assets instance no longer return from `loadAssetsGroup`**

## Changelog
Assets - Typing is added to the assets instance after assets are loaded with `loadAssetsGroup`. Asset instance no longer returned from `loadAssetsGroup`.

## Additional info
MADS-4126


ethanshar (approved):

ethanshar:
Approved. Looks good 👌 
Let's see how it all integrated with private and take it from there..

PR #3058, opened 2024-05-06
Overlay component borderRadius fix

adids1221:
## Description
Overlay component fix borderRadius, when using Image component with overlayType and borderRadius the Overlay is covering the Image and leaks out.
Wrapping the Overlay with View with style fix the issue.
Added an example in the screen.

## Changelog
Overlay covering fix when passing borderRadius.

## Additional info
ticket 4162


src/components/overlay/index.tsx
```diff
@@ -105,24 +104,25 @@ class Overlay extends PureComponent<OverlayTypes> {
...
-    if (type === OVERLY_TYPES.VERTICAL) {
-      return (
-        <>
-          {this.renderImage([this.getStyleByType(OVERLY_TYPES.TOP), styles.vertical], imageSource)}
-          {this.renderImage([this.getStyleByType(OVERLY_TYPES.BOTTOM), styles.vertical], imageSource)}
-          {customContent && this.renderCustomContent()}
-        </>
-      );
-    }
+    const isVertical = type === OVERLY_TYPES.VERTICAL;
```
Inbal-Tish, line 109: Do you need to check this one every render?

src/components/overlay/index.tsx
```diff
@@ -105,24 +104,25 @@ class Overlay extends PureComponent<OverlayTypes> {
...
 
     return (
-      <>
-        {type && this.renderImage(this.getStyleByType(), imageSource)}
-        {customContent && this.renderCustomContent()}
-      </>
+      <View style={{overflow: 'hidden', borderRadius, ...StyleSheet.absoluteFillObject}}>
```
Inbal-Tish, line 112: This style can move to the styleSheet while borderRadiuse can be added to an array

Inbal-Tish:
@adids1221 The overlay gradient is messed up in your branch, see screenshots:

src/components/overlay/index.tsx
Inbal-Tish: No need for absolute, pass flex

src/components/overlay/index.tsx
```diff
@@ -134,14 +134,11 @@ const styles = StyleSheet.create({
...
   top: {
     bottom: undefined,
-    top: 0,
     height: '75%'
   },
   bottom: {
-    bottom: 0,
     top: undefined,
-    height: '75%',
-    transform: [{scaleY: -1}]
```
Inbal-Tish, line 144: The transform should not be deleted

src/components/overlay/index.tsx
Inbal-Tish: I don't see you took it out of the render....

Inbal-Tish (approved):

PR #3060, opened 2024-05-07
handled case of assets union

nitzanyiz:
## Description
Fixed typings when assets where loaded to an already loaded path. When loading to an already loaded path the type was overridden and not combined. 

## Changelog
Assets - Fix loading assets typings to preloaded path.

## Additional info
MADS-4126

### DIFF
--- a/src/typings/assets.d.ts
+++ b/src/typings/assets.d.ts
@@ -1,12 +1,31 @@
+
+/**
+ * Adds dot at end of string.
+ */
 type DotExtendedString<T extends string> = `${T}.`;
+
+
+/**
+ * Converts path string to a an array of the path. Acts like split('.')
+ *
+ * Example: PathArray<'path1.path2.path3'> --> ['path1', 'path2', 'path3']
+ */
 type PathArray<S extends string> =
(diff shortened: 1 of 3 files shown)


src/typings/assets.d.ts
```diff
@@ -0,0 +1,22 @@
+type DotExtendedString<T extends string> = `${T}.`;
```
ethanshar, line 1: Ok, I reviewed this again, this time more thoroughly.. 
Since the TS utils are pretty complicated to understand at first glance I'd suggest to make them more readable. 

1. I'd change params names like `T` and `K` to something with more meaning.. 
for example, maybe change this `AssetRecord<T extends string, K extends object = {}, This extends object>` to `AssetRecord<PATH extends string, ASSETS extends object = {}, ASSETS_CLASS extends object>`
2. Let's add a short description above the complicated utils of what exactly they do. If we'll have to investigate an issue in the future it can be very useful 
3. RecRecord util's name is not so clear, can you think of a better name?

ethanshar (approved):

PR #3061, opened 2024-05-07
Feat/segemented control new UI preset

nitzanyiz:
## Description
Segmented Control Form preset added. I created a "hook" that returns new props according to the preset.

## Changelog
SegmentedControl - New `form` preset added.

## Additional info
MADS-4065


Inbal-Tish:
@nitzanyiz Where is the example?

src/components/segmentedControl/index.tsx
```diff
@@ -84,6 +85,10 @@ export type SegmentedControlProps = {
   containerStyle?: StyleProp<ViewStyle>;
   style?: StyleProp<ViewStyle>;
   testID?: string;
+  /**
+   * Preset type
+   */
+  preset?: 'default' | 'form';
```
Inbal-Tish, line 91: Why not use an enum and add 'presets' statically to the SegmentedControl?

src/components/segmentedControl/index.tsx
```diff
@@ -157,11 +165,13 @@ const SegmentedControl = (props: SegmentedControlProps) => {
 
   const animatedStyle = useAnimatedStyle(() => {
     if (segmentsStyle.value.length !== 0) {
-      const isFirstElementSelected = animatedSelectedIndex.value === 0;
-      const isLastElementSelected = animatedSelectedIndex.value === segmentsStyle.value.length - 1;
-      const xOffset = isFirstElementSelected ? -2 : isLastElementSelected ? 2 : 0;
-      const inset = withTiming(segmentsStyle.value[animatedSelectedIndex.value].x + xOffset, TIMING_CONFIG);
-      const width = withTiming(segmentsStyle.value[animatedSelectedIndex.value].width * BORDER_WIDTH, TIMING_CONFIG);
+      const isFirst = animatedSelectedIndex.value === 0;
```
Inbal-Tish, line 168: Why the naming change? 'isFirst' is not the same as 'isFirstSelected'...

src/components/segmentedControl/index.tsx
```diff
@@ -170,39 +180,59 @@ const SegmentedControl = (props: SegmentedControlProps) => {
 
   const renderSegments = () =>
     _.map(segments, (_value, index) => {
+      const isLastSegment = index + 1 === segments?.length;
```
Inbal-Tish, line 183: This is more 'isLast'...

src/components/segmentedControl/index.tsx
```diff
@@ -170,39 +180,59 @@ const SegmentedControl = (props: SegmentedControlProps) => {
...
+            />
+          )}
+        </>
       );
     });
-
   return (
     <View style={containerStyle} testID={testID}>
       <View row center style={[styles.container, style, {borderRadius, backgroundColor}]}>
+        <Reanimated.View
```
Inbal-Tish, line 214: Why separate into two Animated Views instead of one?

src/components/segmentedControl/segment.tsx
```diff
@@ -55,6 +55,7 @@ export type SegmentProps = SegmentedControlItemProps & {
    */
   style?: StyleProp<ViewStyle>;
   testID?: string;
+  iconTintColor?: string;
```
Inbal-Tish, line 58: New prop should be added to api file as well

Inbal-Tish: You have the 'tintColor' in the SegmentedControlItemProps's 'iconStyle'. And even if you want to add this prop I think it makes more sense to add it there

src/components/segmentedControl/useSegmentedControlPreset.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (activeColor && !outlineColor && preset === 'default') {
+    presetProps.outlineColor = activeColor;
+  }
+  if (activeColor || inactiveColor) {
+    delete presetProps.iconTintColor;
+  }
+  return presetProps;
+};
+
+const defaultActiveColor = Colors.$textPrimary;
```
Inbal-Tish, line 27: These two can be defined at the top as consts

src/components/segmentedControl/useSegmentedControlPreset.ts
```diff
@@ -0,0 +1,54 @@
+import {type SegmentedControlProps} from './';
+import {BorderRadiuses, Colors} from '../../style/';
+import {SegmentProps} from './segment';
+import type {ColorValue} from 'react-native';
+
+interface useSegmentedControlPresetProps extends SegmentedControlProps, Partial<SegmentProps> {
```
Inbal-Tish, line 6: Why are you adding "props" at the end? Is "useSegmentedControlPreset' not enough?

nitzanyiz:
Oh sorry. Figma link:
https://www.figma.com/file/IftsoihHtESrm0s7t9LPak/SegmentedControl-New-Preset?type=design&node-id=41-34779&mode=design&t=vX4OLrBC2mRcjDR0-0

Inbal-Tish:
@nitzanyiz Looks great. Left small comments. Also, I have a warning poping-up "Each child in a list should have a quique "key" prop. Check the render method of 'SegmentedControl'". Please take a look

Inbal-Tish:
> Oh sorry. Figma link:

Please don't add private links to the public repo.
I mean the example in the demo screen...

src/components/segmentedControl/index.tsx
nitzanyiz: When the outline and the active background are together the segment divider comes on top of the outline so the outline doesn't cover the divider. Moving the view to be on top of the segments make the active background be on top of the segments label therefore the label is not visible. `In short, moving it causes the outline to cover the divider like we want but the background also covers the segment label.`

src/components/segmentedControl/segment.tsx
nitzanyiz: I guess I can expose it.

src/components/segmentedControl/useSegmentedControlPreset.ts
nitzanyiz: you mean `Partial<SegmentProps>`?

src/components/segmentedControl/index.tsx
nitzanyiz: I created a const with the presets and added it as member to the component.

src/components/segmentedControl/index.tsx
nitzanyiz: Changed.

src/components/segmentedControl/index.tsx
nitzanyiz: What do you mean?

src/components/segmentedControl/index.tsx
```diff
@@ -9,17 +9,26 @@ import Reanimated, {
...
+import {ComponentStatics} from '../../typings/common';
 
-const BORDER_WIDTH = 1;
+const CONTAINER_BORDER_WIDTH = 1;
 const TIMING_CONFIG = {
   duration: 300,
   easing: Easing.bezier(0.33, 1, 0.68, 1)
 };
 
+const presets = {
```
Inbal-Tish, line 25: What's wrong with a simple enum?

src/components/segmentedControl/index.tsx
Inbal-Tish: isLast is a good name to check if the current segment is the last one, while isFirstSelected is a good name to check if the selected segment is the last one

src/components/segmentedControl/index.tsx
```diff
@@ -170,58 +191,75 @@ const SegmentedControl = (props: SegmentedControlProps) => {
...
             },
             animatedStyle
           ]}
         />
-        {renderSegments()}
       </View>
     </View>
   );
-};
+}) as SegmentedControlComponent;
```
Inbal-Tish, line 250: Why is that for?

src/components/segmentedControl/index.tsx
Inbal-Tish: Let's talk tomorrow, I don't understand the issue...

src/components/segmentedControl/index.tsx
nitzanyiz: In this case It doesn't really have any effects. If we would do something like this we get ts errors:
image

src/components/segmentedControl/index.tsx
nitzanyiz: This is so I could put the preset property on the component.

src/components/segmentedControl/index.tsx
nitzanyiz: Oh you just wanted to pin this out. I thought this wasn't good.

src/components/segmentedControl/index.tsx
Inbal-Tish: Let's talk about this point too

src/components/segmentedControl/useSegmentedControlPreset.ts
Inbal-Tish: No, I mean the name of the hook

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
```diff
@@ -66,6 +72,14 @@ const SegmentedControlScreen = () => {
           segments={segments.seventh}
           segmentLabelStyle={styles.customTypography}
         />
+        <Text marginT-s4 center>
```
Inbal-Tish, line 75: Why no simply use the first SegmentedControl to switch btw the default and form presets on all/most of the examples?

nitzanyiz:
* change to enum and set it on the component
* added the preset change to the screen
* removed the extra reanimated view if there is no divider.

src/components/segmentedControl/index.tsx
```diff
@@ -170,39 +185,65 @@ const SegmentedControl = (props: SegmentedControlProps) => {
...
+        <View
+          reanimated
           style={[
             styles.selectedSegment,
             {
-              borderColor: outlineColor,
               borderRadius,
               backgroundColor: activeBackgroundColor,
-              borderWidth: outlineWidth
+              borderWidth: segmentDividerWidth !== 0 ? undefined : outlineWidth,
```
Inbal-Tish, line 226: You can do this once and store it in a const

src/components/segmentedControl/segment.tsx
```diff
@@ -130,5 +135,5 @@ const Segment = React.memo((props: SegmentProps) => {
     </TouchableOpacity>
   );
 });
-
+Segment.displayName = 'SegmentedControl-Segment';
```
Inbal-Tish, line 138:
```suggestion
Segment.displayName = 'SegmentedControl.Segment';
```

Inbal-Tish: or just "Segmnt"

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
```diff
@@ -17,32 +17,56 @@ const segments = {
...
   return (
     <View flex bottom padding-page>
       <Text center text40 $textDefault>
         Segmented Control
       </Text>
-      <View flex marginT-s8>
+      <View
+        flex
+        marginT-s8
+        key={screenPreset /* added here because changing the preset based on state was causing wrong calculations */}
```
Inbal-Tish, line 44: Why it was cousing wrong calculations?

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
```diff
@@ -54,17 +78,19 @@ const SegmentedControlScreen = () => {
             inactiveColor={Colors.$textDisabled}
             style={styles.customStyle}
             segmentsStyle={styles.customSegmentsStyle}
+            preset={screenPreset}
```
Inbal-Tish, line 81: Any meaning for this here when it has a custom style?

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
nitzanyiz: I'm not sure why but when the preset changes based on state the outlines sometimes gets the dimensions and offsets worng.

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
nitzanyiz: I removed the preset from this example and moved it to the bottom.

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
Inbal-Tish: So we should check why. Maybe some props that are not updated and should. This is not a good solution since the probable scenario is using it with a state change

src/components/segmentedControl/segment.tsx
Inbal-Tish: I don't see the change...

nitzanyiz: Fixed

demo/src/screens/componentScreens/SegmentedControlScreen.tsx
ethanshar: Inbal, I discussed this with Nitzan
the component initialize all sizes and keep it as shared values so the easiest way to solve it is by passing a key but because it's not a real use case (there's no reason to change the preset in runtime) I don't think we should invest time on it
We have other examples screen where we use the "key" trick, so it's ok IMO

Inbal-Tish: The real use case will be when the editor will offer the user to select one of the two presets. I think we should try to fix this issue. May be reset the values if the preset changes or at leat do the key hack in the component and not ask the user to do it... just throwing ideas, I didn't check it myself yet

nitzanyiz: @ethanshar Do you think we should add the key thing inside the component?

Inbal-Tish: Very nice fix!

Inbal-Tish (approved):

PR #3064, opened 2024-05-09
Use rowGap and columnGap feature in our GridList and SortableGridList…

ethanshar:
## Description

Use rowGap and columnGap feature in our GridList and SortableGridList components
We had issue with extra item spacing in the last row of the grid 
This fix the issue and also use a better way to add gutter (item spacing) between grid items 

## Changelog
Fix GridList and SortableGridList item spacing issue (using gap feature)


src/components/gridList/useGridLayout.ts
```diff
@@ -42,18 +43,22 @@ const useGridLayout = (props: GridListBaseProps) => {
   }, [numberOfColumns, itemSpacing, keepItemSize ? undefined : _containerWidth]);
 
   const itemContainerStyle = useMemo(() => {
-    return {width: itemWidth /* + itemSpacing */, marginRight: itemSpacing, marginBottom: itemSpacing};
+    return {width: itemWidth /* , marginRight: itemSpacing, marginBottom: itemSpacing */};
```
M-i-k-e-l, line 46: Why keep the comments?

src/components/gridList/useGridLayout.ts
```diff
@@ -42,18 +43,22 @@ const useGridLayout = (props: GridListBaseProps) => {
...
   }, [itemWidth, itemSpacing]);
 
   const listContentStyle = useMemo(() => {
-    return [{paddingHorizontal: listPadding}, contentContainerStyle];
-  }, [listPadding, contentContainerStyle]);
-  
+    return [{/* paddingHorizontal: listPadding, */ rowGap: itemSpacing}, contentContainerStyle];
+  }, [listPadding, itemSpacing, contentContainerStyle]);
```
M-i-k-e-l, line 51: We have extra dependencies now (here and above) from the commented code

src/components/sortableGridList/index.tsx
```diff
@@ -41,14 +42,12 @@ function SortableGridList<T = any>(props: SortableGridListProps<T>) {
...
   const _renderItem = useCallback(({item, index}: ListRenderItemInfo<ItemProps<T>>) => {
-    const lastItemInRow = (index + 1) % numberOfColumns === 0;
-
     return (
       <SortableItem
         key={item.id}
         data={data}
         {...presenter}
-        style={[itemContainerStyle, lastItemInRow && {marginRight: 0}]}
+        style={itemContainerStyle}
```
M-i-k-e-l, line 50: `itemContainerStyle` is not missing (along others) from the dependencies array

M-i-k-e-l (commented):
Nice.
Not sure it's a bug but now we don't have a gap after the last row (visible when there's no `SafeArea`)
image

M-i-k-e-l:
BTW, we currently have a somewhat awkward solution to this in `SortableList` - we've added `itemProps` with `margins` in it.
It does allow for horizontal margins as well, but I'm not sure it's a must in a list.
We can try to change the API in the next major version, WDYT?
(It should still be a dedicated prop because we need this info for the dragging height calculation)

src/components/gridList/useGridLayout.ts
ethanshar: I usually keep comments in case we'll have a weird issue with these change it'll be easier to spot what was before
Usually after a comment "sits" for some time I delete it

ethanshar:
> BTW, we currently have a somewhat awkward solution to this in `SortableList` - we've added `itemProps` with `margins` in it. It does allow for horizontal margins as well, but I'm not sure it's a must in a list. We can try to change the API in the next major version, WDYT? (It should still be a dedicated prop because we need this info for the dragging height calculation)

I'm less familiar with it TBH, if you feel we need to change something there then open a PR

src/components/gridList/useGridLayout.ts
ethanshar: Fixed

src/components/sortableGridList/index.tsx
ethanshar: I added what I think necessary

Inbal-Tish (approved):

PR #3067, opened 2024-05-12
Add flexMigration to SortableList and SortableGridList components

ethanshar:
## Description
When we started wrapping our SortableList and SortableGridList components with `GestureHandlerRootView` we should have pass flex style so the list have a flexed behavior like other RN lists (FlatList, ScrollView, etc..) 
This migration is for transition to this default behavior

## Changelog
Start a migration to add flex to SortableList and SortableGridList

## Additional info
MADS-4163


adids1221 (approved):
Approved, I think we should already open a ticket for the deprecation process.

PR #3069, opened 2024-05-12
deprecate children prop

adids1221:
## Description
Picker children prop deprecation.
As part of the Picker refactor we are deprecating the children prop and moving to use the items prop.
Children cleanup from Picker internal code will be done after merging this [PR](https://github.com/wix/react-native-ui-lib/pull/3070/files).

## Changelog
Picker children prop deprecation.

## Additional info
MADS-4180

### DIFF
--- a/src/components/picker/helpers/usePickerMigrationWarnings.tsx
+++ b/src/components/picker/helpers/usePickerMigrationWarnings.tsx
@@ -4,10 +4,10 @@ import {LogService} from '../../../services';
 import {PickerProps, PickerModes} from '../types';
 
 // @ts-expect-error TODO: Remove this whole file when migration is completed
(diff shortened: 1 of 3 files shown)


Inbal-Tish:
@adids1221 What about all of our inner usages of Picker.Item and the PickerItem API?

adids1221:
@Inbal-Tish I'll do it in another PR before moving to the error phase.
WDYT ?

Inbal-Tish:
> @Inbal-Tish I'll do it in another PR before moving to the error phase. WDYT ?

Why? If you're deprecating something we should stop using it internally. This will also provide an example for the migration, and make sure there are not issues with this deprecations (at least on our side)

adids1221:
Closing this PR.
Starting the deprecation in another PR.

PR #3070, opened 2024-05-13
fix search when passing items

adids1221:
## Description
Picker fix search when using `items` instead of `children`.
usePickerSearch hook returns filteredItems now.

## Changelog
Picker fix search when using `items` instead of `children`.

## Additional info
MADS-4193


src/components/picker/PickerItem.tsx
```diff
@@ -44,7 +44,7 @@ const PickerItem = (props: PickerItemProps) => {
 
   const isItemDisabled = useMemo(() => {
     return !!(disabled || (!isSelected && context.selectionLimit && context.selectionLimit === selectedCounter));
-  }, [selectedCounter]);
+  }, [selectedCounter, disabled]);
```
M-i-k-e-l, line 47: Is this related to the main issue?
Why not add the other missing deps?

src/components/picker/PickerItemsList.tsx
```diff
@@ -35,7 +35,7 @@ const PickerItemsList = (props: PickerItemsListProps) => {
     testID
   } = props;
   const context = useContext(PickerContext);
-  const [wheelPickerValue, setWheelPickerValue] = useState<PickerSingleValue>(context.value ?? items?.[0].value);
+  const [wheelPickerValue, setWheelPickerValue] = useState<PickerSingleValue>(context.value ?? items?.[0]?.value);
```
M-i-k-e-l, line 38: Is this related to the main issue?
Can you provide an example?

src/components/picker/types.tsx
```diff
@@ -141,7 +141,11 @@ export type PickerBaseProps = Omit<NewTextFieldProps, 'value' | 'onChange'> & {
   /**
    * callback for picker modal search input text change (only when passing showSearch)
    */
-  onSearchChange?: (searchValue: string, filteredChildren?: ReactNode | undefined) => void;
+  onSearchChange?: (
+    searchValue: string,
+    filteredChildren?: ReactNode | undefined,
+    filteredItems?: Pick<PickerItemProps, 'label' | 'value' | 'disabled'>[]
```
M-i-k-e-l, line 147: The user should not have to do something like:
```
onSearchChange={(value, _bla, changed) => {
  console.warn('value', value);
  console.warn('changed', changed);
}}
```
in order to get the changes.

You can simply change the name and type of the second value (this should be changed in the API of course)

src/components/picker/helpers/usePickerSearch.tsx
```diff
@@ -23,13 +23,25 @@ const usePickerSearch = (props: UsePickerSearchProps) => {
...
+    if (showSearch && !_.isEmpty(searchValue)) {
+      return _.filter(items, item => {
+        const {label, value} = item;
+        const itemLabel = getItemLabelPresenter(label, value);
+        return !shouldFilterOut(searchValue, itemLabel);
+      });
+    }
+
+    return items;
+  }, [showSearch, searchValue, items]);
```
M-i-k-e-l, line 36: Can you reduce this code duplication?
IMO there should be a single function that uses both `children` and `items`

src/components/picker/index.tsx
```diff
@@ -224,7 +225,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
         useWheelPicker={useWheelPicker}
         mode={mode}
         useDialog={useDialog}
-        items={useItems ? items : undefined}
+        items={useItems ? filteredItems || items : undefined}
```
M-i-k-e-l, line 228: IMO this should be `filteredItems` and not `filteredItems || items`, just like below there's only `{filteredChildren}`

src/components/picker/PickerItem.tsx
adids1221: yes it's related.
If the `disabled` is not in the deps of the `useMemo` then the search return the disabled item as selectable item.

adids1221: about the other missing deps, I'll add `isSelected`.
Not sure if I should add the context we can talk about it.

src/components/picker/PickerItemsList.tsx
adids1221: Yes it related.
When searching and there are no items matching the search it cause an issue for me and crash with error of undefined value.

src/components/picker/helpers/usePickerSearch.tsx
```diff
@@ -3,33 +3,32 @@ import _ from 'lodash';
...
+        const itemLabel = getItemLabelPresenter(label, value, itemGetItemLabel || getItemLabelFunction);
         return !shouldFilterOut(searchValue, itemLabel);
       });
     }
+    return items;
+  };
 
-    return children;
-  }, [showSearch, searchValue, children]);
+  const filteredItems = useMemo(() => filterItems(children || items, getItemLabel), [showSearch, searchValue, items]);
```
M-i-k-e-l, line 23: You probably need `children` as a dependency here

M-i-k-e-l (commented):
Another issue: the `onSearchChange` `value` we get is correct, but the `children`\`items` is from the previous search (it's one search behind).
It does not look like a regression, but you should definitely fix it or add a ticket to fix it.

src/components/picker/PickerItemsList.tsx
adids1221: Go to the PickerScreen, in the first example change it from children to items:
```
  <Picker
            placeholder="Favorite Language"
            floatingPlaceholder
            value={this.state.language}
            enableModalBlur={false}
            onChange={item => this.setState({language: item})}
            topBarProps={{title: 'Languages'}}
            // style={{color: Colors.red20}}
            showSearch
            searchPlaceholder={'Search a language'}
            searchStyle={{color: Colors.blue30, placeholderTextColor: Colors.grey50}}
            // onSearchChange={value => console.warn('value', value)}
            items={longOptions}
          />
 ```
 Enter some `bla-bla` phrase in the search box and see the difference.

File: src/components/picker/helpers/usePickerSearch.tsx [hunk 7]
```diff
@@ -3,33 +3,32 @@ import _ from 'lodash';
...
+        const itemLabel = getItemLabelPresenter(label, value, getItemLabel || getItemLabelFunction);
         return !shouldFilterOut(searchValue, itemLabel);
       });
     }
+    return items;
+  };
 
-    return children;
-  }, [showSearch, searchValue, children]);
+  const filteredItems = useMemo(() => filterItems(children || items, getItemLabel), [showSearch, searchValue, items, children]);
```
M-i-k-e-l, line 23: Lets move everything into the `useMemo`?

adids1221: @M-i-k-e-l After going over it few times, the are many type errors when moving the `filterItems` function into `filteredItems` and sending items as `children || items` the type we are getting back from the function is problemtic.
I'm not sure if it's worth time to go into it since it will be solved after the children migration finish.
WDYT ?

src/components/picker/helpers/usePickerSearch.tsx
```diff
@@ -3,33 +3,32 @@ import _ from 'lodash';
...
-  const filteredChildren = useMemo(() => {
+  const filterItems = (items: any, getItemLabelFunction: any) => {
     if (showSearch && !_.isEmpty(searchValue)) {
-      // @ts-expect-error need to fix children type
-      return _.filter(children, child => {
-        // @ts-expect-error need to fix children type to be based on PickerItemProps
-        const {label, value, getItemLabel: childGetItemLabel} = child.props;
-        const itemLabel = getItemLabelPresenter(label, value, childGetItemLabel || getItemLabel);
+      return _.filter(items, item => {
+        const {label, value, getItemLabel: itemGetItemLabel} = item.props || item;
```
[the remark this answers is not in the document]
M-i-k-e-l: I'd add a comment that `item.props` is from `children`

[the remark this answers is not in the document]
M-i-k-e-l: You've resolved half the issue.
```
const filterItems = (items: any, getItemLabelFunction: any)
```
Should be
```
const filterItems = (items: any)
```
The second argument should be gotten from this:
```
const {showSearch, onSearchChange, children, getItemLabel, items} = props;
```
Which is why I suggested renaming this `getItemLabel`

src/components/picker/helpers/usePickerSearch.tsx
M-i-k-e-l: Lets take a look after you've resolved the other issues.

src/components/picker/helpers/usePickerSearch.tsx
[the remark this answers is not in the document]
adids1221: fixed

src/components/picker/helpers/usePickerSearch.tsx
adids1221: I wrapped the filterItems with `useCallback`

src/components/picker/helpers/usePickerSearch.tsx
```diff
@@ -3,33 +3,35 @@ import _ from 'lodash';
...
+  const filterItems = useCallback((items: any) => {
     if (showSearch && !_.isEmpty(searchValue)) {
-      // @ts-expect-error need to fix children type
-      return _.filter(children, child => {
-        // @ts-expect-error need to fix children type to be based on PickerItemProps
-        const {label, value, getItemLabel: childGetItemLabel} = child.props;
-        const itemLabel = getItemLabelPresenter(label, value, childGetItemLabel || getItemLabel);
+      return _.filter(items, item => {
+        const {label, value, getItemLabel} = item.props || item;
+        const itemLabel = getItemLabelPresenter(label, value, getItemLabel || getItemLabelPickerProp);
```
M-i-k-e-l, line 16: I think you've reversed the order (I think it should be `getItemLabelPickerProp || getItemLabel`)
I'd also consider keeping the old naming convention:
```
const {label, value, getItemLabel: childGetItemLabel} = item.props || item;
const itemLabel = getItemLabelPresenter(label, value, childGetItemLabel || getItemLabel);
```
This way only the local prop is renamed, which does not affect the whole file.

M-i-k-e-l: I’ve suggested returning the `getItemLabelPickerProp` to the original `getItemLabel`, not sure if you’ve missed it or it was on purpose

src/components/picker/helpers/usePickerSearch.tsx
```diff
@@ -3,33 +3,35 @@ import _ from 'lodash';
...
+  const filterItems = useCallback((items: any) => {
     if (showSearch && !_.isEmpty(searchValue)) {
-      // @ts-expect-error need to fix children type
-      return _.filter(children, child => {
-        // @ts-expect-error need to fix children type to be based on PickerItemProps
-        const {label, value, getItemLabel: childGetItemLabel} = child.props;
-        const itemLabel = getItemLabelPresenter(label, value, childGetItemLabel || getItemLabel);
+      return _.filter(items, item => {
+        const {label, value, getItemLabel: childGetItemLabel} = item.props || item;
+        const itemLabel = getItemLabelPresenter(label, value, getItemLabel || childGetItemLabel);
```
M-i-k-e-l, line 16:
```suggestion
        const itemLabel = getItemLabelPresenter(label, value, childGetItemLabel || getItemLabel);
```

M-i-k-e-l (approved):

PR #3080, opened 2024-05-21
Picker screens aligned to the new props migration

adids1221:
## Description
Picker screens aligned to the new picker props deprecation.
Note: the only example didn't fixed yet is the `Custom Modal` which has an issue with the `renderItem` function.
I'm working on a fix separately. 

## Changelog
Picker screens refactor according to the Picker refactor.

## Additional info
None

### DIFF
--- a/demo/src/screens/componentScreens/PickerScreen.tsx
+++ b/demo/src/screens/componentScreens/PickerScreen.tsx
@@ -24,14 +24,39 @@ const tagIcon = require('../../assets/icons/tags.png');
 const dropdown = require('../../assets/icons/chevronDown.png');
 const dropdownIcon = <Icon source={dropdown} tintColor={Colors.$iconDefault}/>;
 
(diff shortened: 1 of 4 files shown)


demo/src/screens/componentScreens/PickerScreen.tsx
```diff
@@ -256,78 +253,36 @@ export default class PickerScreen extends Component {
                 </View>
               );
             }}
-          >
-            {_.map(contacts, contact => (
-              <Picker.Item
-                key={contact.name}
-                value={contact.value}
-                label={contact.label}
-                renderItem={(contactValue, props) => {
```
Inbal-Tish, line 265: Why to remove the renderItem example? We can move it to the Picker, no?

Inbal-Tish (approved):

Inbal-Tish:
@adids1221 There is another usage in the Picker tests. What about the WebDemo Picker example?

PR #3081, opened 2024-05-21
Picker render custom modal support items

adids1221:
## Description
Picker support passing items when passing renderCustomModal.
When passing items the Picker index file will return array of PickerItems instead of Children to the modal props.

## Changelog
Picker support passing items when passing renderCustomModal.

## Additional info
MADS-4228

### DIFF
--- a/src/components/picker/index.tsx
+++ b/src/components/picker/index.tsx
@@ -96,8 +96,8 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
   } = themeProps;
   const {preset} = others;
 
-  const [selectedItemPosition, setSelectedItemPosition] = useState(0);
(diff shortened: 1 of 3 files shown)


src/components/picker/index.tsx
```diff
@@ -206,7 +210,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
         closeModal: closeExpandable,
         toggleModal: toggleExpandable,
         onSearchChange: _onSearchChange,
-        children,
+        children: children || (items && _.map(items, item => renderPickerItem(item))),
```
Inbal-Tish, line 213: Let's move the items part to a 'renderItems' method

Inbal-Tish (approved):

PR #3086, opened 2024-05-23
fixed outline height not covering all container when segments have di…

nitzanyiz:
## Description
fixed outline height not covering all container when segments have different heights. The previous onLayout was just setting the outline height based of the segments. When the segments doesn't have the same height, for example when one has an Icon and the others don't sometimes the final outline height doesn't get the height of the container so it doesn't fully wrap some of the segments.

## Changelog
SegmentedControl - Fixed outline doesn't wrap the entire segment.

## Additional info
Code snippet to reproduce:
```JSX
<SegmentedControl segments={[{label: 'Fill', iconSource: Assets.icons.checkSmall}, {label: 'Fit'}]}/>
```
Its flaky because of the shared value so if it doesn't reproduce on the first time try to go back and into the screen.

### DIFF
--- a/src/components/segmentedControl/index.tsx
+++ b/src/components/segmentedControl/index.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, {useRef, useCallback, useEffect} from 'react';
+import React, {useRef, useCallback, useEffect, useState} from 'react';
 import {StyleSheet, StyleProp, ViewStyle, TextStyle, LayoutChangeEvent} from 'react-native';
 import {
   Easing,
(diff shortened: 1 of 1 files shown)


nitzanyiz:
There is a simple solution where we can just use
```JSX
    runOnUI((value: number) => {
      segmentedControlHeight.value = Math.max(value, segmentedControlHeight.value);
    })(height + 2 * CONTAINER_BORDER_WIDTH);
```
inside the onLayout which then I believe always gets the most updated value.

Inbal-Tish (approved):

Inbal-Tish:
> There is a simple solution where we can just use
> 
> ```js
> runOnUI((value: number) => {
>   segmentedControlHeight.value = Math.max(value, segmentedControlHeight.value);
> })(height + 2 * CONTAINER_BORDER_WIDTH);
> ```
> 
> inside the onLayout which then I believe always gets the most updated value. So onLayout will be
> 
> ```js
> const onLayout = useCallback((index: number, event: LayoutChangeEvent) => {
>   // if (shouldResetOnDimensionsOnNextLayout.current) {
>   //   shouldResetOnDimensionsOnNextLayout.current = false;
>   //   // segmentsDimensions.current = getInitialSegmentsDimensionsArray(segments?.length || 0);
>   // }
>   const {x, width, height} = event.nativeEvent.layout;
>   segmentsDimensions.current[index] = {x, width};
>   runOnUI((value: number) => {
>     segmentedControlHeight.value = Math.max(value, segmentedControlHeight.value);
>   })(height + 2 * CONTAINER_BORDER_WIDTH);
>   if (nonAreUndefined(segmentsDimensions.current)) {
>     segmentsStyle.value = [...segmentsDimensions.current];
>     // shouldResetOnDimensionsOnNextLayout.current = true;// in case onLayout will be called again (orientation change etc.)
>   }
> },
> // eslint-disable-next-line react-hooks/exhaustive-deps
> [initialIndex, segments?.length]);
> ```

Since the container gets the height of the highest segment it seems simpler to get it from it's onLayout like you did

PR #3092, opened 2024-05-26
changed the type of pickerModalProps

adids1221:
## Description
Picker `pickerModalProps` type refactor.

## Changelog
Picker `pickerModalProps` type refactor.

## Additional info
None

### DIFF
--- a/src/components/picker/types.tsx
+++ b/src/components/picker/types.tsx
@@ -166,7 +166,7 @@ export type PickerBaseProps = Omit<NewTextFieldProps, 'value' | 'onChange'> & {
   /**
    * Pass props to the picker modal
    */
-  pickerModalProps?: object;
+  pickerModalProps?: ExpandableOverlayProps['modalProps'];
   /**
    * Custom container style
    */


nitzanyiz:
Looks good. If we're already refactoring the usage of the modal props. What do you think if we make it a little more conditional. If `useDialog` is passed with true than were ignoring the `pickerModalProps` and with false, the picker is rendered with a modal and then we use the `pickerModalProps`. We can do something like this and it will not compile when using the Picker with the dialog and passing pickerModalProps. WDYT?
```ts
interface PickerDialogProps {
  useDialog: true;
}
interface PickerModalProps {
  useDialog?: false;
  /**
   * Render custom picker modal (e.g ({visible, children, toggleModal}) => {...})
   */
  renderCustomModal?: (modalProps: RenderCustomModalProps) => React.ReactElement;
    /**
   * Pass props to the picker modal
   */
  pickerModalProps?: ModalProps;
  /**
   * The picker modal top bar props
   */
  topBarProps?: ModalTopBarProps;
  /**
   * Adds blur effect to picker modal (iOS only)
   */
  enableModalBlur?: boolean;
}
export type PickerBaseProps = (PickerDialogProps | PickerModalProps) 
```

adids1221:
Closing this PR the types are changed in the `pickerType` PR.

PR #3094, opened 2024-05-26
ChipsInput - fix validation

M-i-k-e-l:
## Description
ChipsInput - fix validation

## Changelog
ChipsInput - fix validation

## Additional info
None


src/components/chipsInput/index.tsx
```diff
@@ -146,6 +147,24 @@ const ChipsInput = forwardRef((props: ChipsInputProps, refToForward: React.Ref<a
...
+    return inputValidators.includes('required');
+  }, [validate]);
+
+  const validateChipsInput = useCallback(() => {
+    return !isRequired || (chips?.length ?? 0) > 0;
+  }, [isRequired, chips]);
+
+  useDidUpdate(() => {
+    // @ts-expect-error
+    fieldRef.current?.validate();
```
Inbal-Tish, line 165: Need to check 'validateOnChange' prop or move it to the TextField's onChangeText

src/components/chipsInput/index.tsx
```diff
@@ -146,6 +147,24 @@ const ChipsInput = forwardRef((props: ChipsInputProps, refToForward: React.Ref<a
...
+  const isRequired = useMemo(() => {
+    if (!validate) {
+      return false;
+    }
+
+    const inputValidators = Array.isArray(validate) ? validate : [validate];
+    return inputValidators.includes('required');
+  }, [validate]);
+
+  const validateChipsInput = useCallback(() => {
```
Inbal-Tish, line 159: The only validity here is 'required' but the user might pass other validations, mainly a custom validation function

M-i-k-e-l: Fixed, can be tested by adding the following to the screen:
```
validate={['required', () => {
    return this.state.chips2[0]?.label !== 'Bla';
  }]}
validateOnChange
```

The code handling this is a little complicated, but I'm not sure it can be simplified.

src/components/chipsInput/index.tsx
```diff
@@ -146,6 +147,37 @@ const ChipsInput = forwardRef((props: ChipsInputProps, refToForward: React.Ref<a
...
+    } else if (isRequired) {
+      const inputValidators = Array.isArray(validate) ? validate : [validate];
+      return inputValidators.map(validator => (validator === 'required' ? requiredValidator : validator));
+    } else {
+      return validate;
+    }
+  }, [validate, isRequired, requiredValidator]);
+
+  useDidUpdate(() => {
+    if (others.validateOnChange) {
```
Inbal-Tish, line 175: What about 'validateOnStart' and 'validateOnBlur' props? Are we not supporting them here?

M-i-k-e-l: They're working as far as I've tested

Inbal-Tish (approved):

PR #3096, opened 2024-05-27
Fix/segmented control icon spacings

nitzanyiz:
## Description
Added spacing between icon and label in the `SegmentedControl` segments.

## Changelog
SegmentedControl - Added S1 spacings between icon and label.

## Additional info
MADS-4239

### DIFF
--- a/src/components/segmentedControl/segment.tsx
+++ b/src/components/segmentedControl/segment.tsx
@@ -112,7 +112,7 @@ const Segment = React.memo((props: SegmentProps) => {
     onLayout?.(index, event);
   },
   [onLayout, index]);
-  const labelAndIconSpacings = !!iconSource && (iconOnRight ? styles.rightMargin : styles.leftMargin);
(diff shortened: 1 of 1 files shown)


src/components/segmentedControl/segment.tsx
```diff
@@ -112,7 +112,7 @@ const Segment = React.memo((props: SegmentProps) => {
     onLayout?.(index, event);
   },
   [onLayout, index]);
-
+  const labelAndIconSpacings = !!iconSource && (iconOnRight ? styles.rightMargin : styles.leftMargin);
```
adids1221, line 115: I think you should rename it since the Spacing is going to the Text component.
`labelMargins` or something like that WDYT ?

adids1221 (commented):
Left a small comment

src/components/segmentedControl/segment.tsx
nitzanyiz: Hmm, I dont see much difference. Your is shorter so we can use that.

adids1221 (approved):
Approved

PR #3097, opened 2024-05-27
TextField - new features: helperText and validationIcon

Inbal-Tish:
## Description
TextField - new features: helperText and validationIcon

## Changelog
TextField - new features: helperText and validationIcon

## Additional info
ticket 4173


demo/src/screens/componentScreens/TextFieldScreen.tsx
```diff
@@ -16,6 +16,7 @@ import {
 import Assets from '../../assets/Assets';
 const {KeyboardAwareInsetsView} = Keyboard;
 const priceFormatter = Intl.NumberFormat('en-US');
+const validationIcon = require('../../assets/icons/star.png');
```
ethanshar, line 19: Let's use a better icon for this, we want to make the example to make sense.. 
You can copy the icon from private for that matter..

src/components/textField/ValidationMessage.tsx
```diff
@@ -24,11 +28,24 @@ const ValidationMessage = ({
...
-  );
+  const renderMessage = () => {
+    return (
+      <Text testID={testID} $textDangerLight style={style}>
+        {showValidationMessage ? relevantValidationMessage : ''}
+      </Text>
+    );
+  };
+
+  if (validationIcon) {
```
ethanshar, line 39: Check the example screen with long validation messages.. 
There's on example when the layout breaks (It might be the example fault)

src/components/textField/index.tsx
```diff
@@ -74,7 +76,7 @@ const TextField = (props: InternalTextFieldProps) => {
     enableErrors, // TODO: rename to enableValidation
     validationMessageStyle,
     validationMessagePosition = ValidationMessagePosition.BOTTOM,
-    retainValidationSpace = true,
+    retainValidationSpace = !helperText,
```
ethanshar, line 79: Just to make sure I understand
because helperText now renders under the input then we don't need to retain space if it exists?

src/components/textField/index.tsx
```diff
@@ -107,6 +109,8 @@ const TextField = (props: InternalTextFieldProps) => {
     }
   }, [leadingAccessory]);
 
+  const _helperText = <Text $textNeutralHeavy subtext>{helperText}</Text>;
```
ethanshar, line 112: Why not render it directly without the const, it's not necessarily more readable like this, it's a single line anyway and it also kinda weird you create a JSX code before all the hooks and the props destruction

src/components/textField/ValidationMessage.tsx
Inbal-Tish: Long validation message breaks the layout even without the icon or helperText. I'll fix it

src/components/textField/index.tsx
Inbal-Tish: Yes. Otherwise there will be a gap btw the input and the helperText

src/components/textField/index.tsx
Inbal-Tish: Done

demo/src/screens/componentScreens/TextFieldScreen.tsx
Inbal-Tish: Done

demo/src/screens/componentScreens/TextFieldScreen.tsx
```diff
@@ -193,15 +194,17 @@ export default class TextFieldScreen extends Component {
...
+        <View row spread centerV>
           <TextField
             ref={this.inputWithValidation}
             label="Name"
             placeholder="Enter full name"
             validate="required"
-            validationMessage="This field is required"
-            containerStyle={{flexGrow: 1}}
+            validationMessage="This field is required. That means you have to enter some value"
+            containerStyle={{width: 300}}
```
ethanshar, line 204: It's not a good fix, we can't expect the user to do this, so we need to understand better the issue maybe it'll lead us to a problem in the TextField implmentation
We can do it separately since it's not related to the PR as you said

ethanshar (approved):
Approved. But see my comment about the fix in the example screen 
Anyway the build fails

demo/src/screens/componentScreens/TextFieldScreen.tsx
Inbal-Tish: fixed

PR #3102, opened 2024-05-28
PickerDriver support is validation message

adids1221:
## Description
PickerDriver support isValidationMessage - to indicated if there is validation message in testing.

## Changelog
PickerDriver support isValidationMessage

## Additional info
None


nitzanyiz:
Can you add a test that uses it that check scenarios where when it returns true and false?

adids1221:
> Can you add a test that uses it that check scenarios where when it returns true and false?

yes the test is in the private repo, it's part of the refactor testing.

nitzanyiz:
Shouldn’t it be here because the feature is here? If we break it somehow,
we will only know it until we update private. WDYT?

PR #3107, opened 2024-05-29
Infra/picker testing coverage

adids1221:
## Description
Picker testing coverage.
Testing picker modes (modal and dialog) - currently there is an issue with WheelPicker testing.
Testing picker field types by style.
Search functionality testing have the same issue as the WheelPicker.

## Changelog
Picker testing coverage

## Additional info
MADS-4181

PR #3122, opened 2024-06-04
Infra/picker new picker type prop

adids1221:
## Description
New Picker prop `pickerType` to choose the "render" mode of the picker - 'modal, dialog, wheelPicker, custom" and according to that to pass the right type of header props.
Default value 'Modal'.
Deprecation of `useDialog, useWheelPicker, pickerModalProps, topBarProps, renderCustomModal`.

Note: I moved the Picker prop that are deprecated to `PickerDeprecationsProps` type so in `v8` removing the props from `PickerBaseProps` will be much easier. 

[Testing PR](https://github.com/wix/react-native-ui-lib/pull/3107)

## Changelog
New Picker prop `pickerType` and 'headerProps' props to choose the "render" mode of the picker.

## Additional info
MADS-4187


src/components/picker/helpers/usePickerMigrationWarnings.tsx
```diff
@@ -1,10 +1,10 @@
 import {useEffect} from 'react';
 import {LogService} from '../../../services';
-import {PickerProps} from '../types';
+import {PickerPropsDeprecation} from '../types';
 
 // TODO: Remove this whole file when migration is completed
 type UsePickerMigrationWarnings = Pick<
-  PickerProps,
+PickerPropsDeprecation,
```
Inbal-Tish, line 7: Why didn't you add warnings to the new deprecated props?

src/components/picker/index.tsx
```diff
@@ -93,6 +96,19 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
   } = themeProps;
   const {preset} = others;
 
+  const {
+    type,
+    headerProps,
+    renderCustomModal: typeRenderCustomModal,
```
Inbal-Tish, line 102: I don't like 'typeRenderCustomModal', how about 'customModal'

src/components/picker/helpers/usePickerMode.ts
```diff
@@ -0,0 +1,47 @@
+import {ExpandableOverlayProps} from 'src/incubator';
+import {PickerProps, PickerModeTypes} from '../index';
+import {CustomPickerProps} from '../types';
+
+type PickerType = {
```
Inbal-Tish, line 5: Why declaring the type again and not using PickerModeTypes?

src/components/picker/helpers/usePickerMode.ts
```diff
@@ -0,0 +1,47 @@
...
+  custom?: boolean;
+};
+
+const usePickerMode = (props: PickerProps) => {
+  let type: PickerType = {dialog: false, wheelPicker: false, modal: false, custom: false};
+  let headerProps: any;
+  let renderCustomModal: CustomPickerProps['renderCustomModal'];
+  let dialogProps: ExpandableOverlayProps['dialogProps'];
+  let pickerModalProps: ExpandableOverlayProps['modalProps'];
+  if ('pickerType' in props) {
```
Inbal-Tish, line 18: What this check is for?

src/components/picker/helpers/usePickerMode.ts
```diff
@@ -0,0 +1,47 @@
...
+        dialogProps = headerProps && {...props?.customPickerProps?.dialogProps, headerProps: props.headerProps};
+        break;
+      case PickerModeTypes.Custom:
+        renderCustomModal = props.renderCustomModal;
+        break;
+    }
+    return {type, headerProps, renderCustomModal, dialogProps, pickerModalProps};
+  } else {
+    const {useDialog, useWheelPicker} = props;
+    type = {dialog: !!useDialog, wheelPicker: !!useWheelPicker, modal: !useDialog && !useWheelPicker, custom: false};
```
Inbal-Tish, line 41: If there's no 'pickerType' you're not using the renderCustomModal, dialogProps or pickerModalProps?

src/components/picker/helpers/usePickerMode.ts
```diff
@@ -0,0 +1,47 @@
...
+import {CustomPickerProps} from '../types';
+
+type PickerType = {
+  dialog?: boolean;
+  wheelPicker?: boolean;
+  modal?: boolean;
+  custom?: boolean;
+};
+
+const usePickerMode = (props: PickerProps) => {
```
Inbal-Tish, line 12: Why is it named "usePickerMode" and not "usePickerType"?

src/components/picker/index.tsx
```diff
@@ -93,6 +96,19 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
+    dialogProps,
+    pickerModalProps
+    //@ts-expect-error
+  } = usePickerMode({
+    pickerType: 'modal',
```
Inbal-Tish, line 107: If you're passing a default here it will never go to the else in the usePickerMode...

src/components/picker/index.tsx
```diff
@@ -93,6 +96,19 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
+    ...themeProps
+  });
+  const isDialog = useDialog || type.dialog;
```
Inbal-Tish, line 110: Type wheelPicker is also a Dialog, no?

src/components/picker/helpers/usePickerSelection.tsx
```diff
@@ -1,12 +1,21 @@
...
+import {ModalTopBarProps} from '../../modal/TopBar';
 
-interface UsePickerSelectionProps
-  extends Pick<PickerProps, 'migrate' | 'value' | 'onChange' | 'getItemValue' | 'topBarProps' | 'mode'> {
-  pickerExpandableRef: RefObject<any>;
-  setSearchValue: (searchValue: string) => void;
-}
+type UsePickerSelectionProps = Pick<PickerProps, 'value' | 'onChange' | 'mode'> &
+  Pick<PickerPropsDeprecation, 'migrate' | 'getItemValue' | 'topBarProps'> & {
+    headerProps?: ModalTopBarProps;
```
Inbal-Tish, line 15: Why do you need the headerProps here?

src/components/picker/index.tsx
```diff
@@ -120,6 +136,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
     pickerExpandableRef: pickerExpandable,
     getItemValue,
     topBarProps,
+    headerProps: type.modal && headerProps,
```
Inbal-Tish, line 139: Why you need to pass the headerProps?

src/components/picker/index.tsx
```diff
@@ -218,21 +235,22 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
 
   const expandableModalContent = useMemo(() => {
-    const useItems = useWheelPicker || propItems;
+    const useItems = isWheelPicker || propItems;
+    const listTopBarProps = type.modal ? headerProps : topBarProps;
     return (
       <PickerItemsList
         testID={`${testID}.modal`}
-        useWheelPicker={useWheelPicker}
+        useWheelPicker={isWheelPicker}
```
Inbal-Tish, line 248: Why PickerItemsList is not moving to use the picker type?

src/components/picker/types.tsx
```diff
@@ -220,11 +307,11 @@ export interface PickerItemProps extends Pick<TouchableOpacityProps, 'customValu
   /**
    * Custom function for the item label (e.g (value) => customLabel)
    */
-  getItemLabel?: PickerProps['getItemLabel'];
+  getItemLabel?: PickerPropsDeprecation['getItemLabel'];
```
Inbal-Tish, line 310: If this prop is not deprecated why its type is on the PickerPropsDeprecation?

src/components/picker/types.tsx
```diff
@@ -48,9 +57,68 @@ export interface PickerSearchStyle {
...
+
+export interface CustomPickerProps {
   /**
+   * Type of picker to render
+   */
+  pickerType: PickerModeTypes.Custom | `${PickerModeTypes.Custom}`;
+  renderCustomModal?: (modalProps: RenderCustomModalProps) => React.ReactElement;
+}
+
+export type PickerModeProps = Partial<ModalPickerProps | DialogPickerProps | WheelPickerProps | CustomPickerProps>;
```
Inbal-Tish, line 112: Why it is named 'PickerModeProps' and not 'PickerTypeProps'?

src/components/picker/index.tsx
```diff
@@ -298,14 +318,14 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
         /* @ts-expect-error */
         <ExpandableOverlay
           ref={pickerExpandable}
-          useDialog={useDialog || useWheelPicker}
+          useDialog={isDialog || isWheelPicker}
```
Inbal-Tish, line 321: This should be returned from the hook in the form of type. What I mean is that the hook should be the one source that determines the type of the picker and returns it. Here you should only check if type == dialog.
There are more cases like this in this file

src/components/picker/helpers/usePickerMigrationWarnings.tsx
adids1221: We had some issue with some of the props that are still in use in our internal code, I'll do it in separate PR.

src/components/picker/index.tsx
adids1221: `WheelPicker` does use Dialog eventually (from `ExpandableOverlay`), but the difference is that `WheelPicker` renders inside `PickerItemList` as `<WheelPicker/>` component and handeling the items by itself so it should be separated.

src/components/picker/helpers/usePickerMode.ts
adids1221: `PickerModeTypes` is an enum, if I want to use the enum I need to add interface to convert the keys/values into type that holds boolean value.
Am I wrong ?

src/components/picker/helpers/usePickerMode.ts
adids1221: We can change it and extract the `pickerType` outside since it will always be passed.
I tried to make some using of conditional typing that didn't worked.

src/components/picker/helpers/usePickerMode.ts
adids1221: Yes you are right.
I expect that `headerProps` and `renderCustomModal` will be passed for a specific `pickerType` so if the user doesn't passing `pickerType` I'm not using those props.

src/components/picker/helpers/usePickerMode.ts
adids1221: renamed.

src/components/picker/index.tsx
adids1221: yes I know, for know we won't set it as `default prop` since typing issues ( I spoke with Ethan about it).
passing it to the hook will return the right type.

src/components/picker/helpers/usePickerSelection.tsx
adids1221: It's redundant removed, I forgot it.

src/components/picker/index.tsx
adids1221: using the `headerProps` `onCancel` prop.

src/components/picker/types.tsx
adids1221: I'll add this new type in separate PR I have a task about it.

src/components/picker/index.tsx
adids1221: Created new `type` prop that will have the same type as the `usePickerType` returned type value, and will use it to determine the picker type.

src/components/picker/types.tsx
Inbal-Tish: Please merge master, I've merged this PR

src/components/picker/index.tsx
```diff
@@ -316,6 +312,8 @@ Picker.modes = PickerModes;
 Picker.fieldTypes = PickerFieldTypes;
 // @ts-expect-error
 Picker.extractPickerItems = extractPickerItems;
+// @ts-expect-error
+Picker.pickerType = PickerModeTypes;
```
Inbal-Tish, line 316: pickerTypes - plural

src/components/picker/helpers/usePickerType.ts
```diff
@@ -0,0 +1,67 @@
...
+          headerProps: props.headerProps
+        };
+        break;
+      case PickerModeTypes.Custom:
+        componentProps.customModal = props.renderCustomModal;
+        break;
+    }
+  } else {
+    const {useDialog, useWheelPicker, topBarProps, pickerModalProps, customPickerProps} = props;
+    type = {dialog: !!useDialog, wheelPicker: !!useWheelPicker, modal: !useDialog && !useWheelPicker, custom: false};
```
Inbal-Tish, line 56: If you extracting the type from the old props in case pickerType was not passed, why not do it at the top and then assign the type and pass it in the switch case? It will be more readable and clean

src/components/picker/helpers/usePickerSelection.tsx
Inbal-Tish: Can't it be of type DialogHeaderProps?

src/components/picker/index.tsx
```diff
@@ -173,14 +175,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
 
-  const modalProps: ExpandableOverlayProps['modalProps'] = {
-    animationType: 'slide',
-    transparent: Constants.isIOS && enableModalBlur,
-    enableModalBlur: Constants.isIOS && enableModalBlur,
-    onRequestClose: topBarProps?.onCancel,
-    onShow,
-    ...pickerModalProps
-  };
+  const modalProps: ExpandableOverlayProps['modalProps'] = componentProps.pickerModalProps;
```
Inbal-Tish, line 178: Why here you assign component.xxx to a new const? It's hard to keep track where this is from

src/components/picker/index.tsx
```diff
@@ -114,7 +116,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
     onChange,
     pickerExpandableRef: pickerExpandable,
     getItemValue,
-    topBarProps,
+    headerProps: type.modal && componentProps.headerProps,
```
Inbal-Tish, line 119: headerProps is a dialog prop as well, why it not passed in that case?

src/components/picker/index.tsx
```diff
@@ -208,21 +203,21 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
         onCancel: cancelSelect
       };
 
-      return renderCustomModal(modalProps);
+      return componentProps?.customModal && componentProps.customModal?.(modalProps);
```
Inbal-Tish, line 206: Why checking again componentProps?.customModal? You have an if statement line 194

src/components/picker/index.tsx
```diff
@@ -208,21 +203,21 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
     }
   };
 
   const expandableModalContent = useMemo(() => {
-    const useItems = useWheelPicker || propItems;
+    const useItems = type.wheelPicker || propItems;
+    const listTopBarProps = componentProps.headerProps;
```
Inbal-Tish, line 212: Why assign to another const if you use it once? Also, it is difficult to keep track if your code is not consistent...

src/components/picker/index.tsx
```diff
@@ -196,7 +191,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
     closeExpandable,
     toggleExpandable
   }) => {
-    if (renderCustomModal) {
+    if (componentProps.customModal) {
```
Inbal-Tish, line 194: Don't you need to check if type === 'custom' as well?

src/components/picker/index.tsx
```diff
@@ -267,14 +263,14 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
+          useDialog={type.dialog || type.wheelPicker}
           modalProps={modalProps}
-          dialogProps={customPickerProps?.dialogProps || DIALOG_PROPS}
           expandableContent={expandableModalContent}
-          renderCustomOverlay={renderCustomModal ? _renderCustomModal : undefined}
+          renderCustomOverlay={componentProps.customModal ? _renderCustomModal : undefined}
           onPress={onPress}
           testID={testID}
           {...customPickerProps}
+          dialogProps={componentProps.dialogProps || DIALOG_PROPS}
```
Inbal-Tish, line 273: Why DIALOG_PROPS does not set to default in the hook?

src/components/picker/index.tsx
Inbal-Tish: Also you have a const under the same name (different scope sure) on line 195

src/components/picker/helpers/usePickerType.ts
```diff
@@ -0,0 +1,67 @@
...
+          ...props?.customPickerProps?.dialogProps,
+          headerProps: props.headerProps
+        };
+        break;
+      case PickerModeTypes.Custom:
+        componentProps.customModal = props.renderCustomModal;
+        break;
+    }
+  } else {
+    const {useDialog, useWheelPicker, topBarProps, pickerModalProps, customPickerProps} = props;
```
Inbal-Tish, line 55: I would add a TODO here that this else should be deleted once the migration has completed

src/components/picker/index.tsx
adids1221: Dialog doesn't uses `cancelSelect` when dismiss, Modal does.
We can pass `headerProps` in each case it won't make any change.

src/components/picker/index.tsx
adids1221: removed, passing `componentProps.pickerModalProps` directly to the `ExpandableOverlay` as `modalProps`

src/components/picker/index.tsx
adids1221: I'm not sure, the main idea was when the picker is set to type `custom` only then the user can pass `customModal`.
I'll add the type check for now.

src/components/picker/index.tsx
Inbal-Tish: So how about we pass 'onCancel' instead of the whole headerProps? This way you won't need to use the deprecated topBarProps in the hook. The component will pass the proper callback or undefined

src/components/picker/index.tsx
Inbal-Tish: Yes. That's what I mean... If the user is not passing type='custom' don't run this. If you don't check this type it will render if there's no type (once you take out the fallback in the hook). Now it works "properly" with other types, like "modal", since the modal case in the switch is placed before the custom, otherwise you'll run this even if the type passed is not "custom"

src/components/picker/index.tsx
Inbal-Tish: So how about we pass 'onCancel' instead of the whole headerProps? This way you won't need to use the deprecated topBarProps in the hook. The component will pass the proper callback or undefined

src/components/picker/index.tsx
Inbal-Tish: See my comment below - why not simply have a onCancel prop?

src/components/picker/index.tsx
Inbal-Tish: Not sure I follow... I see you fixed it by checking in the hook if the pickerType prop was passed else you determine the type using deprecated props and have it defaults to 'modal' - that's good

src/components/picker/helpers/usePickerMode.ts
Inbal-Tish: Why custom is always false? If the user passed renderCustomModal can't we make the type be 'custom' and not default to 'modal'?

src/components/picker/helpers/usePickerMode.ts
Inbal-Tish: I see you're checking if(pickerType) instead so this is good

src/components/picker/helpers/usePickerMode.ts
Inbal-Tish: I think the booleans are redundant. You're only using them in the else statement to extract type from deprecated props. You can simply do this for example (better in a dedicated function): 
`if (useWheelPicker) type = PickerModeTypes.WheelPicker`

src/components/picker/types.tsx
```diff
@@ -59,6 +131,47 @@ export type PickerBaseProps = Omit<NewTextFieldProps, 'value' | 'onChange'> & {
...
+   */
+  renderCustomModal?: (modalProps: RenderCustomModalProps) => React.ReactElement;
+  /**
+   * @deprecated
+   * Pass props to the picker modal
+   */
+  pickerModalProps?: object;
+};
+
+export type PickerTypes = PickerTypeProps | PickerPropsDeprecation;
```
Inbal-Tish, line 170: Where do you use 'PickerTypes'?

Inbal-Tish: I think you can remove this and rename 'PickerModeTypes' -> 'PickerTypes'/'PickerOverlayTypes"/'PickerExpandableTypes" etc... 
The Picker "mode" refers to single/multi and there's already a "PickerModes" type for that. 
It can be confusing...

src/components/picker/helpers/usePickerType.ts
```diff
@@ -42,10 +48,12 @@ const usePickerType = (props: PickerProps) => {
       case PickerModeTypes.Dialog:
       case PickerModeTypes.WheelPicker:
         componentProps.headerProps = props.headerProps;
-        componentProps.dialogProps = componentProps.headerProps && {
-          ...props?.customPickerProps?.dialogProps,
-          headerProps: props.headerProps
-        };
+        componentProps.dialogProps =
+          (componentProps.headerProps && {
```
Inbal-Tish, line 52: This is very unclear... Why pass customPickerProps? Why only if there are headerProps?

src/components/picker/helpers/usePickerMode.ts
adids1221: fixed 
```
custom: !!renderCustomModal
```

src/components/picker/helpers/usePickerSelection.tsx
adids1221: No it can't be `DialogHeaderProps` since the `onCancel` is `ModalTopBarProps` type

src/components/picker/index.tsx
```diff
@@ -267,14 +263,14 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
         /* @ts-expect-error */
         <ExpandableOverlay
           ref={pickerExpandable}
-          useDialog={useDialog || useWheelPicker}
+          useDialog={type.dialog || type.wheelPicker}
           modalProps={modalProps}
-          dialogProps={customPickerProps?.dialogProps || DIALOG_PROPS}
           expandableContent={expandableModalContent}
-          renderCustomOverlay={renderCustomModal ? _renderCustomModal : undefined}
+          renderCustomOverlay={componentProps.customModal ? _renderCustomModal : undefined}
```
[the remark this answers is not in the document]
adids1221: I did a small change, checking only once when rendering `ExpandableOverlay`

src/components/picker/types.tsx
adids1221: Actually I'll remove it (`PickerTypes`), there is not a use in this type.
I used it before for the conditional typings I think.

I'll rename `PickerModeTypes ` - `PickerTypes`

Inbal-Tish:
@adids1221 I notice a couple of bugs with pickerType="dialog":
1) If you're not passing anything but the picker type the dialog is cut at the bottom (I'll send you a screenshot).
2) if you only pass 'headerProps' - the dialog is not rendering at all, only the overlay

src/components/picker/index.tsx
```diff
@@ -267,14 +252,14 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
...
-          dialogProps={customPickerProps?.dialogProps || DIALOG_PROPS}
+          useDialog={type.dialog || type.wheelPicker}
+          modalProps={componentProps.pickerModalProps}
           expandableContent={expandableModalContent}
-          renderCustomOverlay={renderCustomModal ? _renderCustomModal : undefined}
+          renderCustomOverlay={type.custom ? _renderCustomModal : undefined}
           onPress={onPress}
           testID={testID}
           {...customPickerProps}
+          dialogProps={componentProps.dialogProps}
```
Inbal-Tish, line 262: The dialogProps are overriding the customPickerProps. Question: is the customPickerProps prop should be applied only when pickerType=custom? What is this prop for? The description is not clear

src/components/picker/index.tsx
adids1221: I'll add the `onCancel` prop, great idea.

src/components/picker/helpers/usePickerSelection.tsx
adids1221: done in the new `onCancel` commit

src/components/picker/index.tsx
adids1221: mark as resolved

src/components/picker/helpers/usePickerMode.ts
adids1221: as we spoke I'll leave it for now

src/components/picker/helpers/usePickerSelection.tsx
```diff
@@ -45,8 +53,12 @@ const usePickerSelection = (props: UsePickerSelectionProps) => {
     setSearchValue('');
     setMultiDraftValue(multiFinalValue);
     pickerExpandableRef.current?.closeExpandable?.();
-    topBarProps?.onCancel?.();
-  }, [multiFinalValue, topBarProps]);
+    if (onCancel) {
```
Inbal-Tish, line 56: Why not: `onCancel?.() || topBarProps?.onCancel?.();`

src/components/picker/helpers/usePickerSelection.tsx
```diff
@@ -1,15 +1,23 @@
...
+  PickerModes,
+  PickerPropsDeprecation
+} from '../types';
 
-interface UsePickerSelectionProps
-  extends Pick<PickerProps, 'migrate' | 'value' | 'onChange' | 'getItemValue' | 'topBarProps' | 'mode'> {
-  pickerExpandableRef: RefObject<any>;
-  setSearchValue: (searchValue: string) => void;
-}
+type UsePickerSelectionProps = Pick<PickerProps, 'value' | 'onChange' | 'mode' | 'onCancel'> &
```
Inbal-Tish, line 12: I didn't mean to add 'onCancel' prop to picker prop, but to have in usePickerSelection the 'onCancel' prop instead of 'headerProps'. The picker component will pass the hook it from the headerProps

src/components/picker/helpers/usePickerMode.ts
Inbal-Tish: For the sack of clearity I wounld change the boolead type to be `isModal`|`isDialog` etc. and in Picker component the check will be `type.isDialog` etc. It will be cleared.
`PickerModeBooleans` I would change to `PickerTypesBooleans`

PR #3130, opened 2024-06-10
useFieldType hook

adids1221:
## Description
Picker useFieldType hook

## Changelog
Picker useFieldType hook

## Additional info
MADS-4188


src/components/picker/index.tsx
```diff
@@ -326,7 +295,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
               value={label}
               selection={Constants.isAndroid ? {start: 0} : undefined}
             >
-              {renderPickerInnerInput()}
+              {renderPickerInnerInput}
```
nitzanyiz, line 298: Looks good. Maybe we should rename this because its not a render function anymore? 
Approved anyway.

nitzanyiz (approved):

PR #3131, opened 2024-06-10
Support passing imageStyle when rendering internally ImageBackground

ethanshar:
## Description
Pass `imageStyle` to Image component when internally it renders ImageBackground (whenever we render children inside the Image) 
This fix some issues where image styling is not being passed to the Image that rendered by ImageBackground component 

See reference in [React Native docs](https://reactnative.dev/docs/imagebackground#imagestyle)

## Changelog
Fix nesting image style props in Image component with children (overlay, content).. 

## Additional info
ticket 4290

### DIFF
--- a/src/components/image/index.tsx
+++ b/src/components/image/index.tsx
@@ -6,6 +6,7 @@ import {
   Image as RNImage,
   ImageProps as RNImageProps,
   ImageBackground,
+  ImageBackgroundProps,
   NativeSyntheticEvent,
   ImageErrorEventData
 } from 'react-native';
@@ -27,6 +28,7 @@ import {ComponentStatics} from 'src/typings/common';
 export type ImageSourceType = string | RNImageProps['source'];
 
 export type ImageProps = Omit<RNImageProps, 'source'> &
+  Pick<ImageBackgroundProps, 'imageStyle'> &
   MarginModifiers &
   RecorderProps & {
     /**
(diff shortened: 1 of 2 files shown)


src/components/image/index.tsx
```diff
@@ -183,6 +185,13 @@ class Image extends PureComponent<Props, State> {
     return this.getVerifiedSource(source);
   }
 
+  getImageStyle = () => {
```
adids1221, line 188: I think we should wrap this with `useMemo`, also `this.shouldUseImageBackground()` should be memoized since we are using it's value more than once and store it's value into variable.
WDYT ?

adids1221 (commented):
Left a small comment

src/components/image/index.tsx
ethanshar: WDYM? 
First of all, notice this is a class component, we can't use `useMemo` here. 
Also there's no point in memoizing, notice I return this `return borderRadius ? [{borderRadius}, imageStyle] : imageStyle;` so if a `borderRadius` was passed the style will be dynamic anyway and if it wasn't passed, I am just passing the `imageStyle` prop that the user has passed. 

as for `this.shouldUseImageBackground()` we don't store its value anywhere.. it's just a util method. where do we store its value in a variable?

adids1221: I had some mistake here my bad.
About the `getImageStyle` you are right since the return statement will return `imageStyle`.

adids1221 (approved):

PR #3133, opened 2024-06-12
[interface] Allow array in HintProps.message

andykog:
## Description
Allow passing `(string | ReactElement)[]` in HintProps.message. This is handy when we want to pass a string with formatting (array of strings / <Text /> elements) 

## Changelog
Improved HintProps typing

## Additional info
N/A

### DIFF
--- a/src/components/hint/index.tsx
+++ b/src/components/hint/index.tsx
@@ -67,7 +67,7 @@ export interface HintProps {
   /**
    * The hint message
    */
-  message?: string | ReactElement;
+  message?: string | ReactElement | (string | ReactElement)[];
   /**
    * The hint message custom style
    */
--- a/src/components/hint/index.tsx
+++ b/src/components/hint/index.tsx
@@ -67,7 +67,7 @@ export interface HintProps {
   /**
    * The hint message
    */
-  message?: string | ReactElement | (string | ReactElement)[];
+  message?: string | ReactElement | (string | ReactElement | null)[];
   /**
    * The hint message custom style
    */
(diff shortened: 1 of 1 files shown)


src/components/hint/index.tsx
```diff
@@ -67,7 +67,7 @@ export interface HintProps {
   /**
    * The hint message
    */
-  message?: string | ReactElement;
+  message?: string | ReactElement | (string | ReactElement | null)[];
```
adids1221, line 70:
```suggestion
  message?: string | ReactElement | (string | ReactElement)[];
```
Hi Andrew, is there a reason you are using `null` as type ?

adids1221 (commented):
Left a small comment

src/components/hint/index.tsx
ipanasenko: I'd use ReactNode instead of all of those, TBH. Includes string, ReactElement and null. Less restrictive
```suggestion
  message?: ReactNode | ReactNode[];
```

andykog: @adids1221 it allows optional parts of the string like `['No items', null]` or `['No items', <Text onPress={}>Add</Text>]`

adids1221: I agree with @ipanasenko.
@andykog change it to `ReactNode `

andykog: @adids1221 updated, note that now it will allow passing `message={<View>...</View>}` which is not valid

ipanasenko: It was already allowed before by ReactElement type, no?

andykog: It was indeed, probably the idea was to allow passing fragments

adids1221 (approved):
Approved

PR #3136, opened 2024-06-16
TextField - clearButton - fix animation

Inbal-Tish:
## Description
TextField - clearButton - fix animation

## Changelog
TextField - clearButton - fix animation


adids1221 (approved):
Approved

PR #3143, opened 2024-06-20
TextField - screen for snapshot tests

Inbal-Tish:
## Description
TextField - screen for snapshot tests

## Changelog
TextField - screen for snapshot tests


adids1221 (approved):
Approved

PR #3145, opened 2024-06-20
TextField - fix screen cases

Inbal-Tish:
## Description
TextField - fix screen cases

## Changelog
TextField - fix screen cases for snapshot tests


adids1221 (approved):
Approved

PR #3146, opened 2024-06-20
Fixed Hint message type

adids1221:
## Description
After changing `Hint` `message` prop type in this [PR](https://github.com/wix/react-native-ui-lib/pull/3133) which became a bit lossy type, changing the `message` type to be `ContentType` more strict type.

## Changelog
Hint message prop type fix.

## Additional info
None

### DIFF
--- a/src/components/hint/index.tsx
+++ b/src/components/hint/index.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, {Component, ReactNode, isValidElement, ElementRef} from 'react';
+import React, {Component, ReactElement, isValidElement, ElementRef} from 'react';
 import {
   Animated,
   StyleSheet,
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
> @Inbal-Tish we can discuss about the comment in the old PR when the user passing `null` as Andrew mentioned there.

Ok. Why do we want to support null?

Inbal-Tish (approved):

adids1221:
I think the only case using null is when the user has some state with the `message` and sometime the state is set to `null`

I'll check it with Andrew.

Inbal-Tish:
> I think the only case using null is when the user has some state with the `message` and sometime the state is set to `null`
> 
> I can think of 2 use cases:
> 
> 1. there is a state that the user set to `null` sometime which isn't right.
> 2. the user is passing `Text` element which is and `ReactElement` with null value inside - Andrew case.

Not our business to fix. If the user doesn't want to render a message don't pass the prop, if using  react element it's the user's responsibility to pass it correctly

PR #3149, opened 2024-06-26
Infra/picker types arrangement

adids1221:
## Description
Picker types arrangement, new `useNewPickerProps` to convert old props into new props.
Picker types divide into type of the component that uses these props.
Also some props are renamed according to the component that uses them.

## Changelog
Picker types arrangement - new props, props deprecation.

## Additional info
None

PR #3150, opened 2024-06-27
Infra/picker modal top bar usage

adids1221:
## Description
Picker `Modal.TopBar` usage removed from `PickerItemList`, it will be render from `ExpandableOverlay`.
Based on [related PR](https://github.com/wix/react-native-ui-lib/pull/3149)

## Changelog
Picker `Modal.TopBar` usage refactor.

## Additional info
MDAS-4314


src/components/picker/index.tsx
```diff
@@ -242,11 +254,6 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
         mode={mode}
         useDialog={useDialog}
         items={useItems ? items : undefined}
-        topBarProps={{
```
Inbal-Tish, line 245: remove 'topBarProps' from the PickerItemsList  types as well

src/components/picker/index.tsx
```diff
@@ -114,6 +113,19 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
     mode
   });
 
+  const overlayTopBarProps = useMemo(() => {
+    return {
+      ...topBarProps,
+      onCancel: cancelSelect,
+      onDone: mode === PickerModes.MULTI ? () => onDoneSelecting(multiDraftValue) : undefined
```
Inbal-Tish, line 120: You can insert the testID here as well

src/incubator/expandableOverlay/index.tsx
```diff
@@ -86,6 +86,9 @@ const ExpandableOverlay = (props: ExpandableOverlayProps, ref: any) => {
     toggleExpandable
   }));
 
+  const modalTopBar = useMemo(() => showTopBar && <Modal.TopBar onDone={closeExpandable} {...topBarProps}/>,
```
Inbal-Tish, line 89: Why no simply call it "topBar"? It is passed to both Modal and Dialog

src/components/picker/helpers/useNewPickerProps.tsx
```diff
@@ -24,11 +27,19 @@ const useNewPickerProps = (props: PickerProps) => {
     onRequestClose: topBarProps?.onCancel
   };
 
+  const showTopBar = (!useDialog || mode === PickerModes.MULTI) && !renderHeader;
+  const overlayTopBarProps = showTopBar ? {testID: `${testID}.modal.topBar`, ...topBarProps} : undefined;
```
Inbal-Tish, line 31: The testID can be passed from the Picker. 
Why is it called "...modal.topBar' and not "..overlay.topBar"? Are you not passing it to the Dialog as well?

Inbal-Tish:
@adids1221 Very nice. Left few small comments

src/incubator/expandableOverlay/index.tsx
adids1221: Yes it is passed in both cases if `showTopBar` is set to `true`

src/components/picker/helpers/useNewPickerProps.tsx
adids1221: This is the testID that used in the PickerItemList, changing it might break some users test's.

src/components/picker/index.tsx
adids1221: about that, I notice 1 issue when using `WheelPicker`.
when using wheel picker the `renderCancel` function getting the `cancelButtonProps || cancelLabel` from the `topBarProps`.
And also `renderPickerHeader` uses `topBarProps` to get the `onDone, onCancel` functionalities.
Let's sync about that

src/components/picker/helpers/useNewPickerProps.tsx
Inbal-Tish: This is what I see: `testID={`${props.testID}.topBar`}`

src/components/picker/index.tsx
Inbal-Tish: I think the problem is that the WheelPicker has it own "header" instead of using the Dialog's header...

adids1221:
@Inbal-Tish I moved the renderHeader outside of the PickerItemList as we discussed.

About the useWheelPicker header we can move it outside the PickerItemList but we will need to pass new prop for the `WheelPicker` `onChange` function to pass the `setState` from outside the PickerItemList component.

I think it's redundant for now, WDYT ?

PR #3153, opened 2024-06-27
SortableList - Pass from and to indexes of changed item

nitzanyiz:
## Description
Adding option for onChange to send the indexes the item moved from and to.

## Changelog
SortableList - onOrderChange now sends the old and new index of the changed item.

## Additional info
MADS-4312


src/components/sortableList/SortableList.api.json
```diff
@@ -16,8 +16,8 @@
     },
     {
       "name": "onOrderChange",
-      "type": "(data: ItemT[]) => void",
-      "description": "A callback to get the new order (or swapped items).",
+      "type": "(data: ItemT[], from: number, to: number) => void",
```
ethanshar, line 19: consider wrapping `from` and `to` in an object, just in case we'll have to pass more info in the future, we don't want to have lots of arguments

src/components/sortableList/types.ts
```diff
@@ -21,7 +21,7 @@ export interface SortableListProps<ItemT extends SortableListItemProps>
   /**
    * A callback to get the new order (or swapped items).
    */
-  onOrderChange: (data: ItemT[] /* TODO: add more data? */) => void;
+  onOrderChange: (data: ItemT[], from: number, to: number /* TODO: add more data? */) => void;
```
ethanshar, line 24: see my comment on the API

ethanshar (commented):
Looks good, added a small comment

src/components/sortableList/SortableList.api.json
nitzanyiz: I would have wanted to combine all the arguments to one object (like they do in [RN draggable](https://github.com/computerjazz/react-native-draggable-flatlist?tab=readme-ov-file#props)) but that would create a migration. WDYT?

ethanshar: Yes it would require a migration and also combining all not necessarily the right solution
I personally like to have a main argument (the data) separated from the extra arguments that do grouped as an object
anyway there's no right or wrong, but we don't want to go into a migration just for that, so I think we can settle an adding the new info/data as an object

nitzanyiz: Alright. I've added an info object to the callback that contains the from to values.

ethanshar (approved):

PR #3158, opened 2024-07-01
Fix/button icon

ethanshar:
## Description
Fix icon coloring issue on Web 
It appear passing `tintColor` in style does not work on Web while passing it as prop works

Also remove our usage of Image component for icon in Button and unify the code for web and mobile

(I kept previous code in a comment cause we weren't sure before about using Icon on mobile, so In case an issue will arise we'll quickly know where it comes from) 

## Changelog
Fix button's icon coloring issue on Web


adids1221 (approved):
Approved

adids1221:
I found 1 issue that doesn't relate to this PR but worth mention.
Image component doesn't render `base64` source, for example take Image component and pass some Asset as source the Image doesn't render.
Reproduce on Picker component in web when selected Item isn't marked as selected (btw I didn't done anything in the refactor in this area).

PR #3159, opened 2024-07-02
TextField - clearButton tests - fir snapshot and driver

Inbal-Tish:
## Description
TextField - clearButton tests - fir snapshot and driver

## Changelog
TextField - clearButton tests - fir snapshot and driver


adids1221 (approved):
Approved

PR #3160, opened 2024-07-02
fix types for TextField `validate` props

ipanasenko:
## Description
Type of TextField's `validate` prop is too generic. We know it will be invoked with a string and needs to return a boolean value

## Changelog
- `TextField` - update type for `validate` prop


ethanshar (approved):

ethanshar:
@ipanasenko Fixed some TS error should be ok now

PR #3162, opened 2024-07-03
Picker - fix usePickerLabel

Inbal-Tish:
## Description
Picker - usePickerLabel - fix picker label is empty if the 'value' prop is a string and no 'items' passed.

usacase: `<Picker value={'Picker Value'}/>`

## Changelog
⚠️ Picker - usePickerLabel - fix picker label is empty if the 'value' prop is a string and no 'items' passed.


adids1221 (commented):
Great work, I think we should add test coverage for this case.

adids1221 (approved):

PR #3163, opened 2024-07-03
Emojis type export

nitzanyiz:
## Description
Export emojis type so it can be added to the assets typings.

## Changelog
Export assets emojis type

## Additional info
MADS-4306


src/index.ts
```diff
@@ -1,5 +1,6 @@
 /* ===== General Stuff ===== */
 export {default as Assets} from './assets';
+export type {emojis as EmojisAssetsType} from './assets/emojis';
```
M-i-k-e-l, line 3: I'm assuming you'll use this in private, how?

nitzanyiz: We have the casting on the assets object in the private repo. When i'll have this type I can add it to the case and it will be added to the type.

M-i-k-e-l (approved):

PR #3171, opened 2024-07-10
Feat/sortabe grid list re order by index

nitzanyiz:

### DIFF
--- a/src/components/sortableGridList/types.ts
+++ b/src/components/sortableGridList/types.ts
@@ -20,7 +20,7 @@ export interface SortableGridListProps<T = any> extends GridListBaseProps, Scrol
    * Wether to reorder the items by index instead of by replacing locations.
    * Items will move to the new index by pushing other items a ahead or aback instead of swapping places with the item at the new index.
    */
-  shouldOrderItemsByIndex?: boolean
+  orderItemsByIndex?: boolean
 }
 
 export interface SortableItemProps {
@@ -29,5 +29,5 @@ export interface SortableItemProps {
   itemsOrder: Animated.SharedValue<ItemsOrder>;
   onChange: () => void;
   style: StyleProp<ViewStyle>;
-  reorderByIndex: boolean;
+  orderByIndex: boolean;
 }
--- a/src/components/sortableGridList/types.ts
+++ b/src/components/sortableGridList/types.ts
@@ -16,6 +16,11 @@ export interface SortableGridListProps<T = any> extends GridListBaseProps, Scrol
    * Temporary migration flag for enabling flex on the container of the list (like it should be by default)
    */
   flexMigration?: boolean;
+  /**
+   * Wether to reorder the items by index instead of by replacing locations.
+   * Items will move to the new index by pushing other items a ahead or aback instead of swapping places with the item at the new index.
+   */
+  shouldOrderItemsByIndex?: boolean
 }
 
 export interface SortableItemProps {
@@ -24,4 +29,5 @@ export interface SortableItemProps {
   itemsOrder: Animated.SharedValue<ItemsOrder>;
   onChange: () => void;
   style: StyleProp<ViewStyle>;
+  reorderByIndex: boolean;
 }
--- a/demo/src/screens/componentScreens/SortableGridListScreen.tsx
+++ b/demo/src/screens/componentScreens/SortableGridListScreen.tsx
@@ -10,20 +10,22 @@ import {
   BorderRadiuses,
   GridListProps,
   SortableGridListProps,
-  Button
+  Button,
+  Switch
 } from 'react-native-ui-lib';
 import _ from 'lodash';
 import products from '../../data/products';
 
-const productsWithIds = products.map((product) => ({...product}));
-type Item = typeof productsWithIds[0];
+const productsWithIds = products.map(product => ({...product}));
+type Item = (typeof productsWithIds)[0];
 
 class SortableGridListScreen extends Component {
   state = {
     orientation: Constants.orientation,
     selectedItemId: undefined,
     items: productsWithIds,
-    removedItems: [] as Item[]
+    removedItems: [] as Item[],
+    shouldOrderByIndex: false
   };
   itemsOrdered = this.state.items;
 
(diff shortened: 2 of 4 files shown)


src/components/sortableGridList/sortableGridList.api.json
```diff
@@ -27,6 +27,11 @@
       "name": "flexMigration",
       "type": "boolean",
       "description": "A temporary migration flag for enabling flex on the list's container (like it should be by default)"
+    },
+    {
+      "name": "orderItemsByIndex",
+      "type": "boolean",
+      "description": "Wether to reorder the items by index instead of by replacing locations."
```
M-i-k-e-l, line 34: Maybe `swapping` instead of `replacing`?

src/components/sortableGridList/types.ts
```diff
@@ -16,6 +16,11 @@ export interface SortableGridListProps<T = any> extends GridListBaseProps, Scrol
    * Temporary migration flag for enabling flex on the container of the list (like it should be by default)
    */
   flexMigration?: boolean;
+  /**
+   * Wether to reorder the items by index instead of by replacing locations.
+   * Items will move to the new index by pushing other items ahead or aback instead of swapping places with the item at the new index.
```
M-i-k-e-l, line 21: Why does this has added description, don't we want it in the API?

nitzanyiz: I felt like the docs should be more concise. Do you think we should add it to the docs?

src/components/sortableGridList/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+describe('SortableGridlist', () => {
+  it('should render a sortable grid list', () => {
+    render(<Testcase/>);
+  });
+  it('should reorder by index', () => {
+    const onOrderChange = jest.fn();
+    const renderTree = render(<Testcase onOrderChange={onOrderChange} orderItemsByIndex/>);
+    const driverItem = useDraggableDriver(useComponentDriver({renderTree, testID: itemsTestId('0')}));
+    driverItem.drag(150); // Items height is 50 but dragging 100 doesn't work for some reason. 150 works and drags one row down.
```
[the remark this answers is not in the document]
nitzanyiz: I'm not sure what the calculation for the drag is. I used 150 because it worked but I don't know If I can make a driver that will be correct for all cases.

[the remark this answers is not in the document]
M-i-k-e-l: You can look at `SortableListItem.driver.new.ts`; even if it's not perfect, it's better than nothing

src/components/sortableGridList/sortableGridList.api.json
```diff
@@ -31,7 +31,7 @@
     {
       "name": "orderItemsByIndex",
       "type": "boolean",
-      "description": "Wether to reorder the items by index instead of by swapping locations."
+      "description": "Wether to reorder the items by index instead of by swapping locations.\nItems will move to the new index by pushing other items ahead or aback instead of swapping places with the item at the new index."
```
M-i-k-e-l, line 34: Not sure `\n` works here

src/components/sortableGridList/__tests__/index.spec.tsx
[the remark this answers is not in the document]
nitzanyiz: In the `SortableListItem` we have a default height for the items so its easy to calculate. Also the widths are kind of calculated based on the props. I don't really think we need to invest much time in this currently tbh. We can come back to this some other time.

src/components/sortableGridList/sortableGridList.api.json
M-i-k-e-l: Were? We can test it :)

src/components/sortableGridList/__tests__/index.spec.tsx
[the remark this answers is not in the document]
M-i-k-e-l: You've added tests, so it seems logical to add a driver...
We will probably never have time to refactor this...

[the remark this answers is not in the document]
nitzanyiz: We'll probably get to this if the component is used more

src/components/sortableGridList/sortableGridList.api.json
nitzanyiz: https://wix-private.github.io/wix-react-native-ui-lib/docs/components/incubator/Dialog#testid-1
https://github.com/wix/react-native-ui-lib/blob/a8b433ccf1d8cd659c4b7c833e4ded4bb9618033/src/incubator/Dialog/dialog.api.json#L34-L38

M-i-k-e-l: Seems to work, hopefully the `space` is not required

src/components/sortableGridList/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+describe('SortableGridlist', () => {
+  it('should render a sortable grid list', () => {
+    render(<TestCase/>);
+  });
+  it('should reorder by index', () => {
+    const onOrderChange = jest.fn();
+    const renderTree = render(<TestCase onOrderChange={onOrderChange} orderByIndex/>);
+    const driverItem = useDraggableDriver(useComponentDriver({renderTree, testID: itemsTestId('0')}));
+    driverItem.drag(150); // Items height is 50 but dragging 100 doesn't work for some reason. 150 works and drags one row down.
```
M-i-k-e-l, line 34: `drag(300)` also "works" - any idea why?

src/components/sortableGridList/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return <SortableGridList {...props} numColumns={3} testID={testID} data={TEST_DATA} renderItem={testRenderItem}/>;
+};
+
+describe('SortableGridlist', () => {
+  it('should render a sortable grid list', () => {
+    render(<TestCase/>);
+  });
+  it('should reorder by index', () => {
+    const onOrderChange = jest.fn();
+    const renderTree = render(<TestCase onOrderChange={onOrderChange} orderByIndex/>);
```
M-i-k-e-l, line 32: Lets add a test for `orderByIndex={false}`

src/components/sortableGridList/__tests__/index.spec.tsx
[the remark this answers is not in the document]
M-i-k-e-l: How do you intent to measure usage?

src/components/sortableGridList/__tests__/index.spec.tsx
nitzanyiz: No 🤷🏻‍♂️

M-i-k-e-l (approved):

Issue #306, opened 2018-11-21
Text doesn't work with Cambodian Language


parinha:
When I simply do:

```
<Text text80 dark10>សួស្តី</Text>
```
<img width="228" alt="screen shot 2018-11-21 at 1 37 35 pm" src="https://user-images.githubusercontent.com/12582359/48823460-7f81f980-ed93-11e8-9592-3b570c3350e3.png">

i can solve this problem by adjusting padding margin, but i think its not the right way to do it. 
So any ideas or solution for fix this problem?
Thanks

BunHouth:
don't adjust padding or margin. better try to change `lineHeight` 

`<Text text80 dark10 style={{ lineHeight: 0 }}>សួស្តី</Text>`

PR #3180, opened 2024-07-17
Separate icon color passed in Button from icon style

ethanshar:
## Description
Separate the icon color we pass in Button from the icon style object. 
This allows up to pass iconColor to the `tintColor` prop and allow the user to override it color properly 

## Changelog
Fix issue with Button's icon color not being overridden when passing `iconProps`/`iconStyle`

### DIFF
--- a/demo/src/screens/componentScreens/ButtonsScreen.tsx
+++ b/demo/src/screens/componentScreens/ButtonsScreen.tsx
@@ -234,15 +234,16 @@ export default class ButtonsScreen extends Component {
               style={{marginBottom: ButtonSpace}}
               iconSource={iconStyle => (
                 <View
-                  style={{
-                    width: 20,
-                    height: 20,
-                    // @ts-expect-error
-                    backgroundColor: iconStyle[0]?.tintColor,
-                    borderRadius: 10,
-                    // @ts-expect-error
-                    marginRight: iconStyle[0]?.marginRight
-                  }}
+                  style={[
+                    iconStyle,
+                    {
+                      width: 20,
+                      height: 20,
+                      // @ts-expect-error
+                      backgroundColor: iconStyle[0]?.tintColor,
+                      borderRadius: 10
+                    }
+                  ]}
                 />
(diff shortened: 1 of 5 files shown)


demo/src/screens/componentScreens/ButtonsScreen.tsx
```diff
@@ -234,15 +234,16 @@ export default class ButtonsScreen extends Component {
...
-                  style={{
-                    width: 20,
-                    height: 20,
-                    // @ts-expect-error
-                    backgroundColor: iconStyle[0]?.tintColor,
-                    borderRadius: 10,
-                    // @ts-expect-error
-                    marginRight: iconStyle[0]?.marginRight
-                  }}
+                  style={[
```
Inbal-Tish, line 237: Let's add all the usages we found for passing tintColor (in iconStyle, in iconProps) so we'll have senety check. What do you think?

src/components/button/index.tsx
```diff
@@ -124,6 +124,17 @@ class Button extends PureComponent<Props, ButtonState> {
     return color;
   }
 
+  getIconColor(): string | undefined {
+    const {disabled} = this.props;
+    let tintColor = this.getLabelColor();
```
Inbal-Tish, line 129: That ignores the tintColor passed in iconStyle and iconProps, no?

ethanshar: no, passing iconStyle is always prioritized and it's include in the `getIconStyle` method

Inbal-Tish: Ok. I thought it was the props.iconStyle that is passed, I missed line #306 (which means the naming is confusing :) Maybe change to _iconStyle?)

Inbal-Tish (approved):

src/components/button/index.tsx
Inbal-Tish: Wait. You mean it is prioritized in the Icon component because you now override the style with props.style, right?

PR #3182, opened 2024-07-18
TextField - fix for centered input

Inbal-Tish:
## Description
TextField - fix for centered input
Fix for this PR: https://github.com/wix/react-native-ui-lib/pull/3178

## Changelog
TextField - fix for centered input

### DIFF
--- a/demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
+++ b/demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
@@ -3398,39 +3398,6 @@ exports[`TextField Screen renders screen 1`] = `
             }
           >
             <Text
-              $textDangerLight={true}
-              style={
-                [
-                  {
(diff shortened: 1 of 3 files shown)


ethanshar (commented):
I rendered a few scenarios with `centered` mode that looks a little weird
I think it's better we ask the UX to prepare guidelines for centered text field and how they want all the surrounding elements to behave 
For example 
- a centered text field with (only) character counter looks weird
- When I render the following it also looks strange, the validation is centered but the helper text is not
```
<TextField
  placeholder="Placeholder"
  validationMessage="This is an error message, a really long validation message"
  helperText="This is a helper text"
  bottomAccessory={<Text>This is a bottom accessory</Text>}
  showCharCounter
   maxLength={20}
   centered
/>
```


Maybe those are edge cases but I don't want us to create a too complex layout that will give us issues in the future

ethanshar (approved):

PR #3185, opened 2024-07-24
TextField - add margin btw bottom elements and char counter

Inbal-Tish:
## Description
TextField - add margin btw bottom elements and char counter, adding tests to validationMessage change

* Will need to update private snapshot tests after merge

## Changelog
TextField - add margin btw bottom elements and char counter, adding tests to validationMessage change

### DIFF
--- a/demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
+++ b/demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
@@ -226,6 +226,7 @@ exports[`TextField Screen renders screen 1`] = `
           }
         >
           <View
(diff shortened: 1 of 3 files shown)


adids1221:
@Inbal-Tish any related ticket ?
Why do we add the margin ?

Inbal-Tish:
> @Inbal-Tish any related ticket ? Why do we add the margin ?
> 
> The code looks good, I'm just not sure where the margin is coming from.

We need the margin since it's defined in the guidelines and so the validationMassage / helperTest won't stick to the charCounter - I fixed the condition to reflect that. Thanks

adids1221 (approved):
Approved

PR #3189, opened 2024-07-30
TextField - show validation icon only with validation massage

Inbal-Tish:
## Description
TextField - show validation icon only with validation massage

## Changelog
TextField - show validation icon only with validation massage

## Additional info
ticket 4391

### DIFF
--- a/src/components/textField/ValidationMessage.tsx
+++ b/src/components/textField/ValidationMessage.tsx
@@ -39,7 +39,7 @@ const ValidationMessage = ({
   if (validationIcon?.source) {
     return (
       <View row>
-        {showValidationMessage &&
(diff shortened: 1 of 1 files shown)


src/components/textField/ValidationMessage.tsx
```diff
@@ -39,7 +39,7 @@ const ValidationMessage = ({
   if (validationIcon?.source) {
     return (
       <View row>
-        {showValidationMessage &&
+        {showValidationMessage && validationMessage &&
```
M-i-k-e-l, line 42: Isn't it better to change `showValidationMessage`?
Either way, it should be `!!validationMessage`.

M-i-k-e-l (approved):

PR #3192, opened 2024-07-31
Dialog - add showClose

M-i-k-e-l:
## Description
Dialog - add showClose

## Changelog
⚠️ Dialog - add showClose

## Additional info
None


src/incubator/Dialog/index.tsx
```diff
@@ -42,8 +43,10 @@ const Dialog = (props: DialogProps, ref: ForwardedRef<DialogImperativeMethods>)
   const {
     visible = false,
     headerProps,
-    containerStyle,
-    containerProps,
+    showClose,
+    closeProps,
+    containerStyle: propsContainerStyle,
```
Inbal-Tish, line 48: Why the rename?

src/incubator/Dialog/types.ts
```diff
@@ -86,6 +94,32 @@ export interface _DialogProps extends AlignmentModifiers, Pick<ViewProps, 'useSa
...
    */
   containerProps?: Omit<ViewProps, 'reanimated' | 'animated' | 'style' | 'onLayout' | 'ref' | 'testID'>;
+  /**
+   * Whether to show the close button or not
+   */
+  showClose?: boolean;
+  /**
+   * The close button props
+   */
+  closeProps?: DialogCloseButton;
```
Inbal-Tish, line 104: 'buttonProps' is a better name IMO

src/incubator/Dialog/types.ts
```diff
@@ -69,15 +70,22 @@ export interface DialogHeaderProps extends ViewProps {
...
 }
 
-export interface _DialogProps extends AlignmentModifiers, Pick<ViewProps, 'useSafeArea'> {
+export interface DialogCloseButton {
   /**
-   * The visibility of the dialog.
+   * Change the default close button text
    */
-  visible?: boolean;
+  closeText?: string;
```
Inbal-Tish, line 77: 'label' is a better name IMO

src/incubator/Dialog/types.ts
```diff
@@ -69,15 +70,22 @@ export interface DialogHeaderProps extends ViewProps {
...
   /**
-   * The Dialog's header (title, subtitle etc)
+   * The text props
    */
-  headerProps?: DialogHeaderProps;
+  textProps?: TextProps;
```
Inbal-Tish, line 81: 'labelProps' is a better name IMO

src/incubator/Dialog/types.ts
```diff
@@ -86,6 +94,32 @@ export interface _DialogProps extends AlignmentModifiers, Pick<ViewProps, 'useSa
    * Extra props for the container
    */
   containerProps?: Omit<ViewProps, 'reanimated' | 'animated' | 'style' | 'onLayout' | 'ref' | 'testID'>;
+  /**
+   * Whether to show the close button or not
+   */
+  showClose?: boolean;
```
Inbal-Tish, line 100: 'visible' (or 'showCloseButton') is a better name IMO

src/incubator/Dialog/useDialogCloseButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+    if (!showClose) {
+      return null;
+    }
+
+    return (
+      <View left centerV pointerEvents={'box-none'}>
+        <TouchableOpacity paddingB-s2 row onPress={close}>
+          <Icon source={Assets.icons.xMedium} tintColor={Colors.white} {...closeProps?.iconProps}/>
+          <Text recorderTag={'unmask'} text70BO white {...closeProps?.textProps}>
+            {closeProps?.closeText || 'Close'}
```
Inbal-Tish, line 35: This will need to be configured in private to pass an i18n string

src/incubator/Dialog/useDialogCloseButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <View style={styles.dialogContentContainer}>
+            {headerProps && <DialogHeader {...headerProps}/>}
+            {children}
+          </View>
+        </>
+      );
+    } else {
+      return (
+        <>
+          {headerProps && <DialogHeader {...headerProps}/>}
```
Inbal-Tish, line 65: This part can move to a separate render to remove duplications

src/incubator/Dialog/useDialogCloseButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const containerProps = useMemo((): DialogCloseButtonProps['containerProps'] => {
+    return showClose ? {...propsContainerProps, pointerEvents: 'box-none'} : propsContainerProps;
+  }, [showClose, propsContainerProps]);
+
+  const containerStyle = useMemo(() => {
+    return showClose ? [propsContainerStyle, styles.transparent] : propsContainerStyle;
+  }, [showClose, propsContainerStyle]);
+
+  const DialogContent = (props: PropsWithChildren<DialogContentProps>) => {
```
Inbal-Tish, line 50: Why name this 'DialogContent' and not 'CloseButton'?

src/incubator/Dialog/index.tsx
```diff
@@ -202,8 +213,7 @@ const Dialog = (props: DialogProps, ref: ForwardedRef<DialogImperativeMethods>)
     <GestureDetector gesture={panGesture}>
       {/* @ts-expect-error should be fixed in version 3.5 (https://github.com/software-mansion/react-native-reanimated/pull/4881) */}
       <View {...containerProps} reanimated style={style} onLayout={onLayout} ref={setRef} testID={testID}>
-        {headerProps && <DialogHeader {...headerProps}/>}
-        {children}
+        <DialogContent headerProps={headerProps} children={children}/>
```
Inbal-Tish, line 216: This is confusing... You're passing 'headerProps' as a prop to the DialogContent that you got from the hook that sets 'headerProps' in the 'DialogContent' it returns... Why not just pass it to the hook with the other props and set it there, only rendering the returned component in the Dialog?
Same as 'children'...

src/incubator/Dialog/types.ts
M-i-k-e-l: `closeLabel`?

src/incubator/Dialog/types.ts
M-i-k-e-l: `closeLabelProps`?

src/incubator/Dialog/types.ts
Inbal-Tish: Just 'label' this is the type of the Dialog's 'closeButtonProps' prop (at the same time it can be ButtonProps type)...

src/incubator/Dialog/types.ts
Inbal-Tish: just 'labelProps'

src/incubator/Dialog/useDialogCloseButton.tsx
M-i-k-e-l: Yes, that's why I added the props (text + text style + icon)

M-i-k-e-l:
@Inbal-Tish ping 🙏

src/incubator/Dialog/useDialogCloseButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+import Icon from '../../components/icon';
+import Text from '../../components/text';
+import TouchableOpacity from '../../components/touchableOpacity';
+import View from '../../components/view';
+
+interface InternalDialogCloseButtonProps extends PropsWithChildren<DialogCloseButtonProps> {
+  close: () => void;
+}
+
+const useDialogCloseButton = (props: InternalDialogCloseButtonProps) => {
```
Inbal-Tish, line 16: Please rename it to useDialogContent as you return the whole content and not just the Button

PR #3196, opened 2024-08-04
Fixed lib/package.js main

evgeni-wix:
## Description
Fixed lib/package.js main. It should be done in order to prevent require.resolve("uilib-native") crashed.

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -3,7 +3,7 @@
   "version": "4.1.3",
   "homepage": "https://github.com/wix/react-native-ui-lib",
   "description": "uilib native components (separated from js components)",
-  "main": "components/index.js",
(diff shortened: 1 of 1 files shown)


evgeni-wix:
@M-i-k-e-l  There is [this dependency](https://github.com/wix-private/mobile-apps-engine/blob/664ce55b22d6bada39ee3986176c858af39eca6e/packages/native/mobile-apps-engine-native/package.json#L94) in the engine, that causes the issue. The package.json of uilib-native defines `"main": "components/index.js",` while actually index.ts is in the components folder.

M-i-k-e-l:
@evgeni-wix please try to use version `4.3.0` instead - it has the transpiled code

PR #3204, opened 2024-08-08
fix: picker items list is error when array is empty

crazyshakalaka:
## Description
When passing items as empty array to the Picker, it will report an error
error message:  Cannot read property 'value' of undefined

image

## Changelog
When passing items as empty array to the Picker, it will report an error.


adids1221 (commented):
Approved, thank you for contributing.

adids1221 (approved):

Issue #3111, opened 2024-05-31
Can't set the picker's modal background color.


king960:
## Description
It is not possible to change the picker's modal colour, because in PickerItemsList you use
'$backgroundDefault' that always falls back into white colour.
`<View bg-$backgroundDefault style={wrapperContainerStyle} useSafeArea={useSafeArea}>`

### Related to

- [x] Components
- [ ] Demo
- [ ] Docs
- [ ] Typings

### Steps to reproduce
1. Create a picker
2. Open the picker and see that the background is white.

#### Expected behavior
If there's an option to specifically set the overlay color, then it should work.
#### Actual behavior
`pickerModalProps={{overlayBackgroundColor: 'red'}}` has no effect

## More Info

You can confirm that the overlayBackgroundColor will work if you do:
`Colors.loadSchemes({
    light: {
        $backgroundDefault: 'transparent',
    },
    dark: {
        $backgroundDefault: 'transparent'
    }
});`

### Code snippet

` <Picker
            pickerModalProps={{overlayBackgroundColor: 'red'}}
            label={'test'}
            value={'test}
            onChange={(value) => {}}
            items={{label: 'test', value: ''}}
        />`

### Environment

- React Native version: 0.74.1
- React Native UI Lib version: 7.23.2

### Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

adids1221:
Hi @king960 ,
First of all it's isn't a bug, the `Modal` overlay background is visible when using Dialog mode.
When using `Picker` as `Modal` there is no overlay background since the `Modal` covers all the screen.
What you asked for is a feature, Iv'e created this [PR](https://github.com/wix/react-native-ui-lib/pull/3214) to add `background` color prop to the Picker which will override the `PickerItemList` background color.

PR #3214, opened 2024-08-18
picker list support custom backgroundColor

adids1221:
## Description
Picker - PickerItemList support custom backgroundColor.

## Changelog
Picker - PickerItemList support custom backgroundColor.

## Additional info
Issue #3111

### DIFF
--- a/src/components/picker/PickerItemsList.tsx
+++ b/src/components/picker/PickerItemsList.tsx
@@ -32,6 +32,7 @@ const PickerItemsList = (props: PickerItemsListProps) => {
     useSafeArea,
     useDialog,
     mode,
+    backgroundColor = Colors.$backgroundDefault,
     testID
   } = props;
   const context = useContext(PickerContext);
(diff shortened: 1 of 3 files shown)


adids1221:
> @adids1221 Why don't we let the user change the background color through `customPickerProps.modalProps.overlayBackgroundColor` (Like the user mentioned). (Btw it seems to be breaking the modal behavior when you pass it (any modalProps)

Ok, in the issue the user mentioned that he can't change the modal bg color since the PickerItemList has bgdefault token.
The issue is coming from the picker item list, the custom overlay bg color will be seen only when the modal have overlay background which is only Modal mode.

adids1221:
Closing the PR, all the issue was only the bg color in the PickerItemList.

theobouwman:
@adids1221 where can you set it in the PickerItemList?

adids1221:
> @adids1221 where can you set it in the PickerItemList?

Hi @theobouwman  (related [PR](https://github.com/wix/react-native-ui-lib/pull/3233)),
You can pass `customPickerProps={{modalProps: {overlayBackgroundColor: 'red'}}}` to the `Picker` and override the `PickerItemList` `overlayBackgroundColor`.

PR #3001, opened 2024-03-27
GridListItem - horizontal alignment support

adids1221:
## Description
Grid List Item support in horizontal alignment.
New `horizontalAlignment` prop to control the horizontal alignment of grid list item in all aspects - text, custom element, image and etc.

## Changelog
Grid List Item support in horizontal alignment

## Additional info
ticket 4082

PR #3215, opened 2024-08-18
GridView - fix after gridListIItem change (#3001)

M-i-k-e-l:
## Description
GridView - fix after gridListIItem change (#3001)

## Changelog
GridView - fix after gridListIItem change (#3001)

## Additional info
None

### DIFF
--- a/src/components/gridListItem/index.tsx
+++ b/src/components/gridListItem/index.tsx
@@ -213,7 +213,12 @@ class GridListItem extends Component<GridListItemProps> {
         accessible={renderCustomItem ? true : undefined}
         {...Modifiers.extractAccessibilityProps(this.props)}
       >
-        {imageProps && <Image style={itemSize} {...imageProps} customOverlayContent={children}/>}
+        {imageProps && (
+          <View style={[{borderRadius: imageProps?.borderRadius}, itemSize]}>
+            <Image {...imageProps} style={[itemSize, imageProps?.style]}/>
+            {children}
+          </View>
+        )}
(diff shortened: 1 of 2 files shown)


src/components/gridListItem/index.tsx
```diff
@@ -213,7 +213,12 @@ class GridListItem extends Component<GridListItemProps> {
         accessible={renderCustomItem ? true : undefined}
         {...Modifiers.extractAccessibilityProps(this.props)}
       >
-        {imageProps && <Image style={itemSize} {...imageProps} customOverlayContent={children}/>}
+        {imageProps && (
+          <View style={[{borderRadius: imageProps?.borderRadius}, itemSize]}>
+            <Image {...imageProps} style={[itemSize, imageProps?.style]}/>
```
adids1221, line 218: I found that the issue is coming from the style override.
```
style={itemSize} {...imageProps}
```
So when the user pass some style via imageProps the size getting override.

I think the better solution will be to fix the style override.

```
{imageProps && <Image {...imageProps} style={[itemSize, imageProps?.style]} customOverlayContent={children}/>}
```

adids1221 (commented):
Left a small comment, the issue coming from style override.

src/components/gridListItem/index.tsx
M-i-k-e-l: So create a PR, if it works I'll close this one

adids1221: [related fix](https://github.com/wix/react-native-ui-lib/pull/3222)

PR #3220, opened 2024-08-21
ESLint - Require capture and custom message

nitzanyiz:
## Description
* Added handling for require in our eslint no-direct-import rule.
* Added custom message to the rules.

## Changelog
ESLint - Added require capture to the no-direct-import rule and custom message functionality.

## Additional info
MADS-4434


eslint-rules/lib/rules/no-direct-import.js
```diff
@@ -85,8 +87,45 @@ module.exports = {
       }
     }
 
+    function reportDirectRequire(node, rule) {
```
M-i-k-e-l, line 90: Couldn't you use the utils as we've discussed? Did it not work?

nitzanyiz: I think the utils we saw do a little more than I need. They aggregate the components and from everywhere and all I need to do is just prohibit the usage of a library.

M-i-k-e-l: It does not sound like a good enough reason to duplicate the code.
Using the util should remove a lot of code.

nitzanyiz: I could use the utils some how but the code will be much uglier. It aggregates all the imports and returns an object. Checking the imports after that is kind of ugly because I will need to go over all the keys of the object and check if its in the rules. I guess I could do that but its not clear. I guess thats why it wasn't used in the import declaration before. I'm also not sure it will exactly do what I need and will not capture other cases as well.

nitzanyiz:
@M-i-k-e-l  as we talked, I've added a more complex example that I couldn't capture with the util.

eslint-rules/tests/lib/rules/no-direct-import.js
```diff
@@ -69,6 +93,44 @@ ruleTester.run('no-direct-import', rule, {
...
+      ]
+    },
+    {
+      options: ruleWithCustomMessage,
+      code: invalidExample3,
+      errors: [
+        {message: customErrorMessage}
+      ]
+    },
+      {
```
M-i-k-e-l, line 127: Please indent this

eslint-rules/lib/rules/no-direct-import.js
```diff
@@ -85,8 +87,45 @@ module.exports = {
...
+      try {
+        const {applyAutofix, destination} = rule;
+        const message = getErrorMessage(rule, 'require');
+        const moduleImported = (node.init.object ? node.init.object.arguments : node.init.arguments)[0];
+        context.report({
```
M-i-k-e-l, line 95: I don't like that we have two different `context.report` but I see we already have a precedent so... 🤷‍♂️

eslint-rules/lib/rules/no-direct-import.js
```diff
@@ -85,8 +87,45 @@ module.exports = {
...
+      return collection;
+    }
+    function checkRequires(node) {
+      const imports = [];
+      addToImports(node, imports);
+      const modules = new Set(collectModulesFromImports(imports));
+      const rule = getRules().find((rule) => modules.has(rule.origin));
+      if (rule) {
+        rule.applyAutofix = false;
+        reportDirectRequire(node, {...rule, applyAutofix: true});
```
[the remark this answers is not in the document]
nitzanyiz: Forgot to remove.

eslint-rules/lib/rules/no-direct-import.js
nitzanyiz: I could change the implementation a little so there would be just one report. With something like this:
```javascript
function reportError(node, rule, textToReplace) {
  try {
    const {applyAutofix, destination} = rule;
    const message = getErrorMessage(rule, 'import');
    context.report({
      node,
      message,
      fix(fixer) {
        if (node && applyAutofix && destination) {
          return fixer.replaceText(textToReplace, `'${destination}'`);
        }
      }
    });
  } catch (err) {
    handleError(RULE_ID, err, context.getFilename());
}
```

nitzanyiz: I combined the implementation

M-i-k-e-l: Excelent!

M-i-k-e-l (approved):

PR #3227, opened 2024-08-25
DateTimePicker - invoke value change only on real change

nitzanyiz:
## Description
Added handling to invoke onChange only when the value really change.

## Changelog
DateTimePicker - onChange fixes. Only calls onChange when value really changes.

## Additional info
MADS-4209
MADS-4210
MADS-4211

### DIFF
--- a/src/components/dateTimePicker/DateTimePicker.driver.ts
+++ b/src/components/dateTimePicker/DateTimePicker.driver.ts
@@ -1,0 +1,31 @@
+import {fireEvent} from '@testing-library/react-native';
+import {
+  ComponentProps,
+  ExpandableOverlayDriver,
+  ButtonDriver
+} from '../../testkit';
+
+export const DateTimePickerDriver = (props: ComponentProps) => {
+  const {renderTree, testID} = props;
+  const expandableOverlayDriver = ExpandableOverlayDriver({...props, testID: `${testID}.overlay`});
+  const openPicker = () => {
+    expandableOverlayDriver.open();
+  };
+  const discardPicker = () => {
+    ButtonDriver({...props, testID: `${testID}.cancel`}).press();
+  };
+  const submitSelection = () => {
+    ButtonDriver({...props, testID: `${testID}.done`}).press();
+  };
+  const changeDateTo = (date: Date) => {
+    const pickerElement = renderTree.getByTestId(`${testID}.picker`);
+    fireEvent(pickerElement, 'onChange', {type: 'set'}, date);
+  };
+
+  return {
+    openPicker,
+    discardPicker,
+    submitSelection,
+    changeDateTo
+  };
+};
--- a/src/components/dateTimePicker/DateTimePicker.driver.ts
+++ b/src/components/dateTimePicker/DateTimePicker.driver.ts
@@ -8,13 +8,13 @@ import {
 export const DateTimePickerDriver = (props: ComponentProps) => {
   const {renderTree, testID} = props;
   const expandableOverlayDriver = ExpandableOverlayDriver({...props, testID: `${testID}.overlay`});
-  const openPicker = () => {
+  const open = () => {
     expandableOverlayDriver.open();
   };
-  const discardPicker = () => {
+  const close = () => {
     ButtonDriver({...props, testID: `${testID}.cancel`}).press();
   };
-  const submitSelection = () => {
+  const select = () => {
     ButtonDriver({...props, testID: `${testID}.done`}).press();
   };
   const changeDateTo = (date: Date) => {
@@ -23,9 +23,9 @@ export const DateTimePickerDriver = (props: ComponentProps) => {
   };
 
   return {
-    openPicker,
-    discardPicker,
-    submitSelection,
+    open,
+    close,
+    select,
     changeDateTo
   };
 };
--- a/src/components/dateTimePicker/__tests__/index.spec.tsx
+++ b/src/components/dateTimePicker/__tests__/index.spec.tsx
@@ -31,10 +31,10 @@ describe('DateTimePicker', () => {
     const renderTree = render(<TestCase onChange={onChange} mode={mode} value={someDate}/>);
     expect(onChange).not.toHaveBeenCalled();
     const driver = DateTimePickerDriver({renderTree, testID});
-    driver.openPicker();
+    driver.open();
     driver.changeDateTo(mode === 'time' ? someDateNextHour : someDateNextDay);
     expect(onChange).not.toHaveBeenCalled();
-    driver.submitSelection();
+    driver.select();
     expect(onChange).toHaveBeenCalled();
   });
   test.each(['time', 'date'] as const)('should not invoke onChange when value is not changed - mode %s', mode => {
@@ -42,8 +42,8 @@ describe('DateTimePicker', () => {
     const renderTree = render(<TestCase onChange={onChange} mode={mode}/>);
     expect(onChange).not.toHaveBeenCalled();
     const driver = DateTimePickerDriver({renderTree, testID});
-    driver.openPicker();
-    driver.submitSelection();
+    driver.open();
+    driver.select();
     expect(onChange).not.toHaveBeenCalled();
   });
 });
--- a/src/components/dateTimePicker/__tests__/index.spec.tsx
+++ b/src/components/dateTimePicker/__tests__/index.spec.tsx
@@ -14,17 +14,9 @@ const TestCase = props => {
 };
 
 describe('DateTimePicker', () => {
-  it('should not invoke onChange when value is not changed - mode time', () => {
+  test.each(['time', 'date'])('should not invoke onChange when value is not changed - mode %s', mode => {
     const onChange = jest.fn();
-    const renderTree = render(<TestCase onChange={onChange} mode="time"/>);
-    expect(onChange).not.toHaveBeenCalled();
-    fireEvent.press(renderTree.getByTestId(testID));
-    fireEvent.press(renderTree.getByTestId(`${testID}.done`));
-    expect(onChange).not.toHaveBeenCalled();
-  });
-  it('should not invoke onChange when value is not changed - mode date', () => {
-    const onChange = jest.fn();
-    const renderTree = render(<TestCase onChange={onChange} mode="date"/>);
+    const renderTree = render(<TestCase onChange={onChange} mode={mode}/>);
     expect(onChange).not.toHaveBeenCalled();
     fireEvent.press(renderTree.getByTestId(testID));
     fireEvent.press(renderTree.getByTestId(`${testID}.done`));
--- a/src/components/dateTimePicker/__tests__/index.spec.tsx
+++ b/src/components/dateTimePicker/__tests__/index.spec.tsx
@@ -26,15 +26,6 @@ jest.mock('../../../optionalDependencies', () => {
 });
 
 describe('DateTimePicker', () => {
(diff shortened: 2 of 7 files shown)


src/components/dateTimePicker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const onChange = jest.fn();
+    const renderTree = render(<TestCase onChange={onChange} mode={mode} value={someDate}/>);
+    expect(onChange).not.toHaveBeenCalled();
+    act(() => {
+      fireEvent.press(renderTree.getByTestId(`${testID}.overlay`));
+    });
+    fireEvent(renderTree.getByTestId(`${testID}.picker`),
+      'onChange',
+      {type: 'set'},
+      mode === 'time' ? someDateNextHour : someDateNextDay);
```
M-i-k-e-l, line 47: Nice code, consider adding a driver

src/components/dateTimePicker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+describe('DateTimePicker', () => {
+  test.each(['time', 'date'] as const)('should not invoke onChange when value is not changed - mode %s', mode => {
+    const onChange = jest.fn();
+    const renderTree = render(<TestCase onChange={onChange} mode={mode}/>);
+    expect(onChange).not.toHaveBeenCalled();
+    fireEvent.press(renderTree.getByTestId(`${testID}.overlay`));
+    fireEvent.press(renderTree.getByTestId(`${testID}.done`));
+    expect(onChange).not.toHaveBeenCalled();
+  });
+  test.each(['time', 'date'] as const)('should invoke onChange when value is changed - mode %s', async mode => {
```
M-i-k-e-l, line 37: I did not know this, nice!

src/components/dateTimePicker/index.tsx
```diff
@@ -203,16 +203,30 @@ const DateTimePicker = forwardRef((props: DateTimePickerPropsInternal, ref: Forw
...
+
   const onDonePressed = useCallback(() => {
     toggleExpandableOverlay();
     if (Constants.isIOS && !chosenDate.current) {
       // since handleChange() is not called on iOS when there is no actual change
       chosenDate.current = new Date();
     }
-
-    onChange?.(chosenDate.current!);
+    if (isValueChanged() && chosenDate.current) {
```
M-i-k-e-l, line 225: I'd change the order of the tests

src/components/dateTimePicker/index.tsx
```diff
@@ -203,16 +203,30 @@ const DateTimePicker = forwardRef((props: DateTimePickerPropsInternal, ref: Forw
...
+  const isValueChanged = useCallback(() => {
+    if (mode === 'time') {
+      return (
+        chosenDate.current?.getHours() !== value?.getHours() || chosenDate.current?.getMinutes() !== value?.getMinutes()
+      );
+    }
+    return (
+      chosenDate.current?.getFullYear() !== value?.getFullYear() ||
+      chosenDate.current?.getMonth() !== value?.getMonth() ||
+      chosenDate.current?.getDate() !== value?.getDate()
```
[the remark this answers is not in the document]
nitzanyiz: You're right, its probably better. I created a date utils in the util folder and exported it. I'll use these functions in private after we merge.

src/components/dateTimePicker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    value: new Date('2021-04-04T00:00:00Z'),
+    migrateDialog: true,
+    testID
+  };
+  return <DateTimePicker {...defaultProps} {...props}/>;
+};
+
+const someDate = new Date('2021-04-04T00:00:00Z');
+const someDateNextDay = new Date(someDate.getTime() + 24 * 60 * 60 * 1000);
+const someDateNextHour = new Date(someDate.getTime() + 60 * 60 * 1000);
```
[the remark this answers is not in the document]
nitzanyiz: Wdym here?
Do you think `date`, `dateNextHour`, `dateNextDay` is better?

src/components/dateTimePicker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+jest.mock('../../../optionalDependencies', () => {
```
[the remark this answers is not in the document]
nitzanyiz: Not sure. jest-setup will mock this for every test file we have. I only see external modules in the jest-mock file i think its more organized this way.
Adding it to the jest-setup file also works. I guess we could do that. Do you think it will be clearer to put it there?

src/components/dateTimePicker/__tests__/index.spec.tsx
nitzanyiz: Added.

nitzanyiz: Added. Let me know what you think about it.

src/components/dateTimePicker/__tests__/index.spec.tsx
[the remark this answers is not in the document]
M-i-k-e-l: Yes, sorry I was not clear - this is about naming.
Personally I think `sameTime`, `nextDay` and `nextHour`.

src/components/dateTimePicker/__tests__/index.spec.tsx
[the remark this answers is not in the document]
M-i-k-e-l: Yes, I think that's the proper place for it because if we'll have an issue in private we'll look there first for a mock.

src/utils/dateUtils.ts
```diff
@@ -0,0 +1,10 @@
+export function isSameDate(date1?: Date, date2?: Date) {
+  return (
+    date1?.getFullYear() === date2?.getFullYear() &&
+    date1?.getMonth() === date2?.getMonth() &&
+    date1?.getDate() === date2?.getDate()
+  );
+}
+export function isSameHourAndMinute(date1?: Date, date2?: Date) {
```
M-i-k-e-l, line 8: I think `isSameTime` is better, WDYT?

src/components/dateTimePicker/DateTimePicker.driver.ts
```diff
@@ -0,0 +1,31 @@
...
+import {
+  ComponentProps,
+  ExpandableOverlayDriver,
+  ButtonDriver
+} from '../../testkit';
+
+export const DateTimePickerDriver = (props: ComponentProps) => {
+  const {renderTree, testID} = props;
+  const expandableOverlayDriver = ExpandableOverlayDriver({...props, testID: `${testID}.overlay`});
+  const openPicker = () => {
```
M-i-k-e-l, line 11: Maybe rename to `open`? Or is there another open here?

src/components/dateTimePicker/DateTimePicker.driver.ts
```diff
@@ -0,0 +1,31 @@
...
+    expandableOverlayDriver.open();
+  };
+  const discardPicker = () => {
```
M-i-k-e-l, line 14: Three options for names:
`discardAndClose` \ `cancel` \ `cancelAndClose`
In any case I don't think `picker` should be part of the name.

src/components/dateTimePicker/DateTimePicker.driver.ts
```diff
@@ -0,0 +1,31 @@
...
+    ButtonDriver({...props, testID: `${testID}.cancel`}).press();
+  };
+  const submitSelection = () => {
```
M-i-k-e-l, line 17: Although `submitSelection` is not bad, don't we usually go with `done` or maybe `select`?

src/components/dateTimePicker/DateTimePicker.driver.ts
```diff
@@ -0,0 +1,31 @@
...
+    ButtonDriver({...props, testID: `${testID}.done`}).press();
+  };
+  const changeDateTo = (date: Date) => {
```
M-i-k-e-l, line 20: Maybe `selectDate`?

src/components/dateTimePicker/index.tsx
M-i-k-e-l: I meant `chosenDate.current && isValueChanged()`

src/components/dateTimePicker/__tests__/index.spec.tsx
```diff
@@ -26,24 +26,24 @@ jest.mock('../../../optionalDependencies', () => {
 });
 
 describe('DateTimePicker', () => {
-  test.each(['time', 'date'] as const)('should not invoke onChange when value is not changed - mode %s', mode => {
+  test.each(['time', 'date'] as const)('should invoke onChange when value is changed - mode %s', async mode => {
```
M-i-k-e-l, line 29: Why is there an `async`?

src/components/dateTimePicker/__tests__/index.spec.tsx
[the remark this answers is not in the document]
nitzanyiz: Why would you call it sameDate? What is it similar to?
I feel like its pretty clear, it also states that the date its self is not important.

src/components/dateTimePicker/__tests__/index.spec.tsx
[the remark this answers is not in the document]
nitzanyiz: What kind of issue will possibly have? When I you the jest mock here I don't think it has any effect on other test files. If I put it in the jest-setup one it does.

src/utils/dateUtils.ts
nitzanyiz: I think same hour and time is clearer because the check is different than just checking with `getTime` which also checks for the milliseconds equality.

src/components/dateTimePicker/DateTimePicker.driver.ts
nitzanyiz: But it doesn't really `select` the value. It only invokes the onChange of the picker. The user has to submit this date (by clicking the done button).

src/components/dateTimePicker/DateTimePicker.driver.ts
nitzanyiz: Hmm. Not sure which one I think is better. I'll change it to close just to be aligned with the other drivers we have.

src/components/dateTimePicker/DateTimePicker.driver.ts
nitzanyiz: same here. I'll change it to select.

src/components/dateTimePicker/__tests__/index.spec.tsx
nitzanyiz: Missed that. Its not needed anymore after I've used the driver.

src/components/dateTimePicker/__tests__/index.spec.tsx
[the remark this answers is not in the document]
M-i-k-e-l: If we'll need to mock the `DateTimePickerPackage` we'll probably look there and not here

src/components/dateTimePicker/__tests__/index.spec.tsx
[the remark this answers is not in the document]
nitzanyiz: I moved the date objects to the top of the file and used them in the default props.

M-i-k-e-l (approved):

Issue #3110, opened 2024-05-31
Picker prop 'migrate' warning despite not using it.


king960:
## Description

When creating a Picker and not using the 'migrate' flag, there's a warning about using the 'migration' prop.
Probably wrong check, as passing migrate=false does not trigger the warning.

### Related to

- [x] Components
- [ ] Demo
- [ ] Docs
- [ ] Typings

### Steps to reproduce
Copy the code snippet and run. :)

#### Expected behavior
Using the 'migrate' flag with value either 'false' or 'true' triggers a warning.

#### Actual behavior
Not using the 'migrate' flag triggers the warning.
Using the 'migrate' flag with value 'false' does not trigger the warning.

## More Info
"react-native-ui-lib": "^7.23.2"
### Code snippet
<Picker
          value={''}
          onChange={(value) => setTeam(value)}
          items={[{value: '1', label: 'one'}]}
      />

```


```

### Environment

- React Native version:
- React Native UI Lib version:

### Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

janoslc:
I can confirm this. I was searching for this usage, then realised they have this defaulted as `true`.
Take a look at `node_modules/react-native-ui-lib/src/components/picker/index.js` line 70.

If you set it for false manually Picker won't work properly. Kind of annoying, but I guess they will get it done if the new version if fully prepared.

adids1221:
Hi @king960 and @janoslc ,
The `migrate` prop will be deprecated in the next major version `v8`.
I think we can close this issue, make sure to migrate your code and remove the usage of the `migrate` prop.

janoslc:
Hey @adids1221! We understood that `migrate` prop will be removed from the next major version. Our problem is that we don't use it, yet we get a warning as if we did. 
Therefore the issue is valid, of course it's not a blocker, but it shouldn't warn us if we don't use `migrate`.

janoslc:
Thank you @adids1221 for your work! The ticket can be closed.

PR #3229, opened 2024-08-26
removed picker migrate rule

adids1221:
## Description
Removed `Picker` `migrate` prop rule from `eslint-rules`.
The `migrate` prop set as default to `true` the rule isn't necessary due the `Picker` refactor effort.

## Changelog
Removed `Picker` `migrate` prop rule from `eslint-rules`.

## Additional info
#3110


ethanshar:
@adids1221 
These files are just for testing, they don't really affect anything or anyone
They are tests files for our eslint rules

PR #3231, opened 2024-08-26
Incubator.Dialog - Fixed content component being created on every render.

nitzanyiz:
## Description
The new useDialogContent hook was creating a component (DialogContent) on every render and then rendering it. This component is a "new" component every time therefore causing the component to render nonstop.
This could also be solved using useMemo but I don't see any use to it since DialogContent has no state.

## Changelog
Incubator.Dialog - Fixed content renders.

## Additional info
MADS-4444

### DIFF
--- a/src/incubator/Dialog/useDialogContent.tsx
+++ b/src/incubator/Dialog/useDialogContent.tsx
@@ -49,7 +49,7 @@ const useDialogContent = (props: InternalDialogCloseButtonProps) => {
     return showCloseButton ? [propsContainerStyle, styles.transparent] : propsContainerStyle;
   }, [showCloseButton, propsContainerStyle]);
 
-  const getDialogContent = () => {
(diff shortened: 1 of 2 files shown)


src/incubator/Dialog/useDialogContent.tsx
```diff
@@ -49,7 +49,7 @@ const useDialogContent = (props: InternalDialogCloseButtonProps) => {
     return showCloseButton ? [propsContainerStyle, styles.transparent] : propsContainerStyle;
   }, [showCloseButton, propsContainerStyle]);
 
-  const DialogContent = () => {
+  const getDialogContent = () => {
```
ethanshar, line 52: let's rename it to`renderDialogContent` because it's a render method (it returns an element)

ethanshar (approved):

PR #3233, opened 2024-08-27
removed default background from picker item list

adids1221:
## Description
Remove default background color from `PickerItemList` to let the user pass `overlayBackgroundColor` to the Modal.
Please check this in dark mode also.

## Changelog
Picker support for Modal `overlayBackgroundColor`, removed default color from PickerItemList.

## Additional info
#3111


M-i-k-e-l (approved):
I'm assuming you've tested this in private (merge if yes)

PR #3235, opened 2024-08-27
Support rendering TabController inside a ScrollView (with a header)

ethanshar:
## Description
When rendering TabController layout inside a ScrollView, especially when having a header (see new example screen) 
the layout doesn't work well
In order to support that I added the `nestedInScrollView` prop that handles the TabPages layout differently 

## Changelog
Support rendering TabController inside a ScrollView (with a header) by passing `nestedInScrollView`


demo/src/screens/componentScreens/index.js
```diff
@@ -55,6 +55,7 @@ export function registerScreens(registrar) {
   registrar('unicorn.components.StepperScreen', () => require('./StepperScreen').default);
   registrar('unicorn.components.SwitchScreen', () => require('./SwitchScreen').default);
   registrar('unicorn.components.TabControllerScreen', () => require('./TabControllerScreen').default);
+  registrar('unicorn.components.TabControllerWithStickyHeaderScreen', () => require('./TabControllerWithStickyHeaderScreen').default);
```
Inbal-Tish, line 58: You need to wrap screen with gestureHandlerRootHOC

Inbal-Tish:
@ethanshar Looks great. Left a small comment

demo/src/screens/componentScreens/TabControllerWithStickyHeaderScreen.tsx
```diff
@@ -0,0 +1,68 @@
...
+      </View>
+    );
+  };
+
+  render() {
+    return (
+      <TabController items={items} nestedInScrollView>
+        <ScrollView
+          // stickyHeaderHiddenOnScroll
+          stickyHeaderIndices={[0, 1]}
```
Inbal-Tish, line 55: Child 0 won't be sticky here so you can also pass only '[1]' as index

ethanshar: Fixed

Inbal-Tish (approved):

Issue #3202, opened 2024-08-08
Color Style props should be typed as ColorValue, not string


sn-cuong:
## Description

Currently in components such as `Button` and `TouchableOpacity`, the `activeBackgroundColor` and `backgroundColor` among other props used for setting the color of a visual UI element should be typed as the `ColorValue` type from `react-native`. Currently, they are typed as `string` types.

### Related to

- [x] Components
- [ ] Demo
- [ ] Docs
- [x] Typings

### Steps to reproduce

1. Pass a `ColorValue` typed prop used for styling UI elements into activeBackgroundColor. 

#### Expected behavior

- `backgroundColor`
- `activeBackground` 
props should accept "ColorValue" types from `react-native`.

#### Actual behavior

See type error `Type 'ColorValue' is not assignable to type 'string | undefined'.`

### Screenshots/Video

![Screenshot 2024-08-07 at 6 38 16 PM](https://github.com/user-attachments/assets/a94e3a35-45d9-4dc5-aa49-9dafe67cfe97)

### Environment

- React Native version: latest stable
- React Native UI Lib version: latest stable

### Affected platforms

- [x] Android
- [x] iOS
- [x] Web

PR #3238, opened 2024-09-02
refcator color props type

adids1221:
## Description
Changed `color` props type from` `string` to `react-native - ColorValue` type.
The `ColorValue` support `OpaqueColorValue` which support platform colors and dynamic colors.
Currently with `string` type we don't have support for that, this was request in the community open issues.
Example to reproduce type error:
```
 const dynamicColor: ColorValue = PlatformColor('systemBlueColor');
 <Button marginV-20 label="Button" backgroundColor={dynamicColor}/>
```

## Changelog
Components `colors` props type change from `string` to `ColorValue`.

## Additional info
#3202

### DIFF
--- a/src/components/actionBar/index.tsx
+++ b/src/components/actionBar/index.tsx
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import React, {Component} from 'react';
-import {StyleSheet, ViewStyle} from 'react-native';
+import {StyleSheet, ViewStyle, ColorValue} from 'react-native';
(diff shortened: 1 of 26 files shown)


adids1221:
TODO:
After installing `snapshot` version on the private repo I found few issues to fix
- ColorPicker - value prop should be `ColorValue`
- ColorPalette & ColorSwatch value props also should be `ColorValue`
- Fix the TS errors for this components

PR #3244, opened 2024-09-05
GestureHandlerRoot - wrap components instead of screens

Inbal-Tish:
## Description
⚠️ GestureHandlerRoot - wrap components instead of screens

## Changelog
GestureHandlerRoot - wrap components instead of screens

### DIFF
--- a/src/components/WheelPicker/index.tsx
+++ b/src/components/WheelPicker/index.tsx
@@ -351,7 +351,7 @@ const WheelPicker = <T extends WheelPickerItemValue>(props: WheelPickerProps<T>)
     <View testID={testID} bg-$backgroundDefault style={style}>
       {separators}
       <View row centerH>
-        <GestureHandlerRootView style={{flexGrow: 1}}>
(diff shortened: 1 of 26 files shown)


src/components/WheelPicker/index.tsx
```diff
@@ -351,7 +351,7 @@ const WheelPicker = <T extends WheelPickerItemValue>(props: WheelPickerProps<T>)
     <View testID={testID} bg-$backgroundDefault style={style}>
       {separators}
       <View row centerH>
-        <View flexG>
+        <GestureHandlerRootView style={{flexGrow: 1}}>
```
ethanshar, line 354: avoid passing inline style, let's move to stylesheet

ethanshar (commented):
Looks great. 
Wrote a small comment

src/components/WheelPicker/index.tsx
Inbal-Tish: Done

PR #3247, opened 2024-09-05
Add new type for Stepper component - floating

ethanshar:
## Description
Add new type for Stepper component - floating

## Changelog
Add new type for Stepper component - floating

## Additional info
MADS-4319


adids1221 (approved):
Approved.
A small comment, since the usage of the floating Stepper is on another component let's add at least one example to the screen, WDYT ?
Of course this could be in another PR.

ethanshar:
> Approved. A small comment, since the usage of the floating Stepper is on another component let's add at least one example to the screen, WDYT ? Of course this could be in another PR.

Done ✅

PR #3255, opened 2024-09-12
Support overriding backgroundColor for SortableListItem

ethanshar:
## Description
Support overriding backgroundColor for SortableListItem
Also fixed several TS issues

## Changelog
Support overriding backgroundColor for SortableListItem

### DIFF
--- a/src/components/sortableList/index.tsx
+++ b/src/components/sortableList/index.tsx
@@ -84,7 +84,8 @@ const SortableList = <ItemT extends SortableListItemProps>(props: SortableListPr
       enableHaptic,
       scale
     };
-  }, [data]);
+  }, [data, onChange]);
+
   return (
     <GestureHandlerRootView style={flexMigration ? styles.container : undefined}>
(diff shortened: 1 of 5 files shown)


src/components/sortableList/index.tsx
```diff
@@ -84,7 +84,8 @@ const SortableList = <ItemT extends SortableListItemProps>(props: SortableListPr
       enableHaptic,
       scale
     };
-  }, [data]);
+  }, [data, onChange]);
```
M-i-k-e-l, line 87: Did this cause a bug (it looks like this is only affected by the user changing `onOrderChange`)?

M-i-k-e-l (approved):
Approved with a question

src/components/sortableList/index.tsx
ethanshar: You mean it's on purpose? 
But if the user changes the onOrderChange, it's not implausible

PR #3264, opened 2024-09-19
Infra/rn73

ethanshar:


src/incubator/Dialog/__tests__/index.new.spec.tsx
```diff
@@ -104,7 +104,7 @@ describe('Incubator.Dialog sanity checks', () => {
     const onDismiss = jest.fn();
     const component = <TestCase2 onDismiss={onDismiss}/>;
     const {dialogDriver} = getDriver(component);
-    expect(dialogDriver.exists()).toBeTruthy();
+    expect(dialogDriver.exists()).toBeFalsy();
```
Inbal-Tish, line 107: You need to change the description of the test - Modal should not exist...

src/incubator/expandableOverlay/ExpandableOverlay.driver.ts
```diff
@@ -3,11 +3,12 @@ import {DialogDriver} from '../Dialog/Dialog.driver.new';
...
 import {usePressableDriver} from '../../testkit/new/usePressable.driver';
 
-export const ExpandableOverlayDriver = (props: ComponentProps) => {
+export const ExpandableOverlayDriver = (props: ComponentProps, useDialog: boolean) => {
   const {renderTree, testID} = props;
 
   const driver = usePressableDriver(useComponentDriver({renderTree, testID}));
-  const isUsingDialog = !!renderTree.queryByTestId(`${testID}.overlay.modal`);
+  // const isUsingDialog = !!renderTree.queryByTestId(`${testID}.overlay.modal`);
+  const isUsingDialog = useDialog;
```
Inbal-Tish, line 11: Why we need `isUsingDialog` const now?

src/testkit/new/Component.driver.ts
```diff
@@ -1,13 +1,13 @@
...
 import {within} from '@testing-library/react-native';
-
 export interface ComponentProps {
   renderTree: ReturnType<typeof within>; // Note: This changed was asked for integration with amino. This still gives all querying functionality.
   testID: string | RegExp;
 }
 
 export interface ComponentDriverResult {
   getElement: () => ReactTestInstance;
+  queryElement:() => ReactTestInstance | undefined;
```
Inbal-Tish, line 10: Missing a space after the colon

Inbal-Tish:
@ethanshar Looks good. Left super small comments

src/components/card/index.tsx
```diff
@@ -42,11 +42,11 @@ export type CardProps = ViewProps &
     /**
      * card custom width
      */
-    width?: number | string;
+    width?: DimensionValue;
```
adids1221, line 45: `DimensionValue` doesn't support `string` value without `%`.
Do we want to remove the `string` support  from these props?

Relevant for all the components with `DimensionValue` type.

M-i-k-e-l:
Managed to install Android after updating Android Studio and jdk to version 17.
Still cannot install pods: `CocoaPods could not find compatible versions for pod "RCT-Folly"`

src/components/picker/Picker.driver.new.ts
```diff
@@ -3,8 +3,14 @@ import {TextFieldDriver} from '../textField/TextField.driver.new';
 import {ModalDriver} from '../modal/Modal.driver.new';
 import {DialogDriver} from '../../incubator/Dialog/Dialog.driver.new';
 import {ButtonDriver} from '../button/Button.driver.new';
+import {ExpandableOverlayDriver} from '../../incubator/expandableOverlay/ExpandableOverlay.driver';
+
+export const PickerDriver = (props: ComponentProps, useDialog: boolean) => {
```
M-i-k-e-l, line 8: I don't like this addition of special input prop `useDialog` - I think we had that added somewhere before and you objected as well...

src/components/picker/Picker.driver.new.ts
```diff
@@ -3,8 +3,14 @@ import {TextFieldDriver} from '../textField/TextField.driver.new';
...
+  const mainDriver = ExpandableOverlayDriver({
```
M-i-k-e-l, line 9: I feel like `expandableOverlayDriver` is more readable

src/components/card/index.tsx
ethanshar: What do you mean by "remove `string` support"? 
Do we do anything special with the strings? we just pass it on to style props..

babel.config.js
```diff
@@ -1,10 +1,23 @@
...
-  plugins: ['react-native-reanimated/plugin']
+  presets: ['module:@react-native/babel-preset'],
+  plugins: [
+    'react-native-reanimated/plugin',
+    [
+      'module-resolver',
+      {
+        extensions: ['.js', '.jsx', '.ts', '.tsx'],
+        root: ['.'],
+        alias: {
```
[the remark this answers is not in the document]
ethanshar: I'm not sure TBH, I forgot about this file... now I see it did the same thing in the main babel

src/components/dialog/index.tsx
```diff
@@ -42,11 +42,11 @@ export interface DialogProps extends AlignmentModifiers, RNPartialProps {
   /**
    * The dialog width (default: 90%)
    */
-  width?: string | number;
+  width?: DimensionValue;
   /**
    * The dialog height (default: undefined)
    */
-  height?: string | number | null;
+  height?: DimensionValue | null;
```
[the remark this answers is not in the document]
ethanshar: I think the `null` is on purpose 
That's the use way to make the height be content based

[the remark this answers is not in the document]
M-i-k-e-l: I think that `DimensionValue` already has `null` (can't verify since it's not installed ATM)

metro.config.js
```diff
@@ -1,28 +1,60 @@
...
+//   } = await getDefaultConfig();
+//   return {
+//     transformer: {
+//       getTransformOptions: async () => ({
+//         transform: {
+//           experimentalImportSupport: false,
+//           inlineRequires: false
+//         }
+//       }),
+//       babelTransformerPath: require.resolve('react-native-svg-transformer')
```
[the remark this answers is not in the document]
M-i-k-e-l: It's working on master

src/components/dialog/index.tsx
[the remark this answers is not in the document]
ethanshar: You are right, fixed

metro.config.js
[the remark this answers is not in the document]
ethanshar: Yes, I put it aside during upgrade. Fixed now
Thanks 🙏

adids1221 (approved):

PR #3268, opened 2024-09-24
Support live code for components

M-i-k-e-l:
## Description
Support live code for components .

* Need to add components to `ReactLiveScope` and test all relevant components.
* Some snippets need to be fixed.
* Looks like the `$1` (etc) are ignored (at least in the components I've tested).
* In `View`: `<View style={{width: 50, height: 50}} bg-orange50/>` is working while `<View width={50} height={50} bg-orange50/>` is not working (I did not test in the webDemo or dig deeper, it is also working in `Image` for example).

## Changelog
None

## Additional info
Ticket 4470

### DIFF
--- a/src/components/view/view.api.json
+++ b/src/components/view/view.api.json
@@ -25,5 +25,14 @@
     {"name": "recorderTag", "type": "'mask' | 'unmask'", "description": "Recorder Tag"},
     {"name": "style", "type": "ViewStyle", "description": "Custom style"}
   ],
-  "snippet": ["<View style={{width: 50, height: 50}} bg-orange50/>"]
+  "snippet": [
+   "<View row gap-s5 centerV>",
+   "  <View style={{width: 200, height: 200}} bg-purple40 centerH>",
+   "  <View style={{width: 60, height: 60}} bg-green20/>",
+   "  </View>",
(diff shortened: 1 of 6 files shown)


src/components/view/view.api.json
```diff
@@ -24,5 +24,6 @@
     {"name": "backgroundColor", "type": "string", "description": "Set background color"},
     {"name": "recorderTag", "type": "'mask' | 'unmask'", "description": "Recorder Tag"},
     {"name": "style", "type": "ViewStyle", "description": "Custom style"}
-  ]
+  ],
+  "snippet": ["<View style={{width: 50, height: 50}} bg-orange50/>"]
```
ethanshar, line 28: I played with and it's really fun, I make a more interesting snippet, let's use it (: 
```
<View row gap-s5 centerV>
  <View style={{width: 200, height: 200}} bg-purple40 centerH>
   <View style={{width: 60, height: 60}} bg-green20/>
  </View>
  <View style={{width: 150, height: 150}} bg-orange30 bottom right>
   <View style={{width: 50, height: 50}} bg-yellow40 br100 margin-s2/>
  </View>
</View>
```

ethanshar (approved):
Approved. 
Added a suggestion for the View snippet (:

PR #3269, opened 2024-09-24
Live Code - support more components 1

M-i-k-e-l:
## Description
Live Code - support more components 1
Go over each commit alone (or don't)

ActionBar - floating over the result (added a margin to the example to solve it)
Checkbox - the checked v is not showing

## Changelog
None

## Additional info
None

### DIFF
--- a/src/components/checkbox/checkbox.api.json
+++ b/src/components/checkbox/checkbox.api.json
@@ -42,6 +42,13 @@
     }
   ],
   "snippet": [
-    "<Checkbox value={false$1} onValueChange={() => console.log('value changed')$2}/>"
+    "function Example(props) {",
+    "  const [value, setValue] = useState(false$1);",
+    "  return (",
+    "    <div>",
+    "      <Checkbox value={value$2} onValueChange={setValue$3}/>",
+    "    </div>",
+    "  );",
+    "}"
   ]
 }
--- a/src/components/slider/slider.api.json
+++ b/src/components/slider/slider.api.json
@@ -85,13 +85,5 @@
     },
     {"name": "testID", "type": "string", "description": "The component test id"},
     {"name": "migrate", "type": "boolean", "description": "Temporary prop required for migration to the Slider's new implementation"}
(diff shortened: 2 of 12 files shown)


src/components/checkbox/checkbox.api.json
```diff
@@ -42,6 +42,13 @@
     }
   ],
   "snippet": [
-    "<Checkbox value={false$1} onValueChange={() => console.log('value changed')$2}/>"
+    "function Example(props) {",
+    "  const [value, setValue] = useState(false$1);",
+    "  return (",
+    "    <div>",
+    "      <Checkbox value={value$2} onValueChange={setValue$3}/>",
```
adids1221, line 49: When selecting the check box there is an issue with the V icon, in the webDemo it works.

src/components/slider/slider.api.json
```diff
@@ -85,13 +85,5 @@
     },
     {"name": "testID", "type": "string", "description": "The component test id"},
     {"name": "migrate", "type": "boolean", "description": "Temporary prop required for migration to the Slider's new implementation"}
-  ],
-  "snippet": [
-    "<Slider",
```
adids1221, line 90: Why removing the Slider snippet ?

adids1221 (commented):
Left a small comment about the Slider snippet.
Also I notice a general issue when trying to change some value is throw an error (let's sync about it).

src/components/slider/slider.api.json
M-i-k-e-l: We could leave it but IMO we should not support components we want to deprecate (we have the `Incubator.Slider` that will be added in another PR).

M-i-k-e-l:
> Left a small comment about the Slider snippet. Also I notice a general issue when trying to change some value is throw an error (let's sync about it).

I assume the errors we see when we edit (big annoying overlay) will not be there in production (or we'll have to fix it :)

adids1221 (approved):

PR #3272, opened 2024-09-25
Live Code - support more components 2

M-i-k-e-l:
## Description
Live Code - support more components 1
Go over each commit alone (or don't)

General issue: `react-native`’s `Animated` is not working properly on web ([see docs](https://necolas.github.io/react-native-web/docs/animated/))

Carousel - cannot drag (probably because of `Animated`) and initial page is bugged (starts on first page but then scrolls to the correct page, probably fixable)
Drawer - not working (probably because of `Animated`)

Assets: can we use web assets somehow so we don’t need to set icon’s sizes?

## Changelog
None

## Additional info
None


scripts/buildDocs.js
```diff
@@ -100,6 +100,13 @@ components.forEach(component => {
   /* Snippet */
   if (component.snippet) {
     content += `### Usage\n`;
+    if (component.usesAnimated) {
```
ethanshar, line 103: I prefer to avoid adding warnings at this point
I rather add specials notes after we understand all of the complexities of rendering components on web and then decide how to tackle it - fix them, add warnings, etc.. 

I personally ok that during the development some stuff won't work 

Maybe instead we can have a list somewhere that document issues with each component. WDYT?

src/components/sortableList/SortableList.api.json
```diff
@@ -43,11 +43,55 @@
...
   ],
   "snippet": [
-    "<SortableList",
-    "  data={data$1}",
-    "  onOrderChange={onOrderChange$2}",
-    "  renderItem={renderItem$3}",
-    "  keyExtractor={keyExtractor$4}",
-    "/>"
+    "function Example(props) {",
+    "  const data = Array.from({length: 30}, (_, index) => {",
```
ethanshar, line 47: Let's change the length to something smaller so it won't take up the whole screen space

src/components/sortableGridList/sortableGridList.api.json
```diff
@@ -35,12 +35,43 @@
...
+    "      <View style={{width: 800}}>",
+    "        <SortableGridList",
+    "          data={data$1}",
+    "          flexMigration",
+    "          // onOrderChange={onOrderChange$2}",
+    "          maxItemWidth={140$3}",
+    "          renderItem={renderItem$4}",
+    "          itemSpacing={Spacings.s3$5}",
+    "          listPadding={Spacings.s5$6}",
+    "          containerWidth={800$7}",
```
ethanshar, line 70: Let's add the `orderByIndex` prop to the snippet, I think it's important to show

ethanshar (commented):
About what you wrote
> General issue: react-native’s Animated is not working properly on web ([see docs](https://necolas.github.io/react-native-web/docs/animated/))

We'll have to figure it out, let's create a list with problematic components and the reason to tackle later

>Carousel - cannot drag (probably because of Animated) and initial page is bugged (starts on first page but then scrolls to the correct page, probably fixable)

It doesn't drag, you scroll to move the carousel.. it makes sense because in mobile you drag scrollviews and in web you just scroll with the mouse 



> Assets: can we use web assets somehow so we don’t need to set icon’s sizes?

Maybe create a config for Icon component to have a size, I think we do the same in private uilib

scripts/buildDocs.js
M-i-k-e-l: Maybe we can add a more general `snippetNote` \ `snippetCaution` \ `usageNote` (etc) that will have the component's specific error\bug?
This way we have the issue and a way to find all issues without an outside list.
WDYT?

M-i-k-e-l:
> We'll have to figure it out, let's create a list with problematic components and the reason to tackle later

Wrote my suggestion in another comment (lets continue there).

> It doesn't drag, you scroll to move the carousel.. it makes sense because in mobile you drag scrollviews and in web you just scroll with the mouse

Makes sense 🤦‍♂️ 

 
> Maybe create a config for Icon component to have a size, I think we do the same in private uilib

I'll take a look

M-i-k-e-l:
Something else, do you like the `Assets.data = {products};` or maybe we should add another object `const Data = {products};`?

M-i-k-e-l:
> Maybe create a config for Icon component to have a size, I think we do the same in private uilib

I can make this work (override `Icon` and add some logic there) however I am not sure if we want to hide this from the users so they don't get confused if they encounter this issue, not sure how to do that.

scripts/buildDocs.js
ethanshar: In theory if the component live code isn't working, Im not sure we want to show it.. 
For now there are only several components that don't work, right? let's document it somewhere, not sure I want to use the json for tracking stuff, someone might misuse it for actual notes

ethanshar: As discussed, let's remove snippets from components that don't work 
and add a warning in the build script about these components

docuilib/src/theme/ReactLiveScope/index.js
```diff
@@ -1,33 +1,78 @@
...
+  // plus: require('../../assets/icons/plus.png').default,
+  // refresh: require('../../assets/icons/refresh.png').default,
+  // search: require('../../assets/icons/search.png').default,
+  // settings: require('../../assets/icons/settings.png').default,
+  // share: require('../../assets/icons/share.png').default,
+  // info: require('../../assets/icons/info.png').default,
+  // exclamation: require('../../assets/icons/exclamationFillSmall.png').default
+});
+
+Assets.data = {products};
```
ethanshar, line 43: Let's store it under `Data` namespace

docuilib/src/theme/ReactLiveScope/index.js
```diff
@@ -1,33 +1,78 @@
...
   RadioGroup,
   SegmentedControl,
+  SortableGridList,
+  SortableList,
+  Spacings,
   Switch,
   Text,
   TextField,
   TouchableOpacity,
+  Utils,
```
ethanshar, line 75: Let's change the Utils namespace to `Playground` and then we can use its components like `Playground.ToggleControl`...

docuilib/src/theme/ReactLiveScope/index.js
```diff
@@ -1,33 +1,78 @@
...
+  // settings: require('../../assets/icons/settings.png').default,
+  // share: require('../../assets/icons/share.png').default,
+  // info: require('../../assets/icons/info.png').default,
+  // exclamation: require('../../assets/icons/exclamationFillSmall.png').default
+});
+
+Assets.data = {products};
 
 // Add react-live imports you need here
 const ReactLiveScope = {
```
ethanshar, line 46: organize the exports, export first the general namespaces like Data, Playground and then the list of components

M-i-k-e-l:
@ethanshar 
I don't think we decided about the icon's configuration, this is what I came up with:
`Icon.js`:
```
import React from 'react';
import Icon from 'react-native-ui-lib/icon';

export default props => {
  if (props.source?.source) {
    const {source, style, ...others} = props;
    return <Icon {...others} source={source.source} style={[{width: source.width, height: source.height}, style]} />;
  } else {
    return <Icon {...props} />;
  }
};
```

Icon example:
```
chevronRight: {
  source: require('../../assets/icons/chevronRight.png').default,
  width: 8,
  height: 14
}
```

ethanshar (approved):
Approved, I see you didn't set any fixed size in the Icon config, I assume you didn't need it?

M-i-k-e-l:
> Approved, I see you didn't set any fixed size in the Icon config, I assume you didn't need it?

There's a default for web (16), the code is old - around a year and a half

PR #3293, opened 2024-10-06
Infra/oss packages publishing

M-i-k-e-l:
## Description
Publish packages automatically with build.
This will publish if the version in the `package.json` is changed and merged to `master`.

Look at each commit separately 

## Changelog
None

## Additional info
None


ethanshar (approved):
Approved, and looking at it now I would definitely consider create a shared release script or something that the other will use

M-i-k-e-l:
> Approved, and looking at it now I would definitely consider create a shared release script or something that the other will use

I agree, there are also other improvements to be made, like running tests\build were relevant.
Added ticket 4518

PR #3299, opened 2024-10-08
Add Docs to showcase with image thumbnails for different layouts

Inbal-Tish:
## Description
Add Docs to showcase with image thumbnails for different layouts

## Changelog
Add Docs to showcase with image thumbnails for different layouts

## Additional info
Add Docs to showcase with image thumbnails for different layouts


Inbal-Tish:
#skipreview

PR #3304, opened 2024-10-09
Separate uilib-native package

ethanshar:
## Description
Decouple uilib-native (lib) components from uilib src components 
This is in order to support mini-bundles and eventually code push 
The fact that uilib-native's native part is in one package but the JS components is in anohter (react-native-ui-lib) was causing an issue 
This PR separates the two without a breaking change to the community (they won't be required to install `uilib-native` package) 

## Changelog
Decouple uilib's native components from its JS library


M-i-k-e-l:
@ethanshar, this change does potentially break public users that use our native components, an error will appear only when such component is used (tested with the fix below*). This happens because they do not have the `native-uilib` package installed.
Moreover, in our project the Android build is failing: `java.lang.StackOverflowError (no error message)` not sure what's exactly is the issue there, however it works on `master` (I have a general issue with iOS ATM).

* Our native components are broken ATM - PR to fix it: #3319

ethanshar:
@M-i-k-e-l 
I merged your other PR #3319 and merge it to this PR 
Both Android and iOS builds are working for me locally in this PR, can you double check? 
I will try installing this version on a random project to see the issue you talked about

ethanshar:
@M-i-k-e-l 
This is ready for review and verified by Niryo

PR #3305, opened 2024-10-10
Docs - page sections

Inbal-Tish:
## Description
Docs - page sections

IMPORTANT:
The sections are implemented on the second tab ("Guidelines") in Text component only - this is for demo only.
This is an initial POC and I will need to clean the code before merging.
Please see Text.api.json for usage demo

## Changelog
Docs - page sections

## Additional info
ticket #4469


src/components/text/text.api.json
```diff
@@ -12,16 +12,182 @@
...
+            "title": "Table",
+            "description": "Examples of using the Text component",
+            "name": "Table Name",
+            "columns": ["Item", "Content[0]", "Content[1]"],
+            "content": [
+              {
+                "title": "Item #1",
+                "description": "Item #1 Subtitle",
+                "content": [
+                  "https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/Docs/thumbnail-horizontal.png?raw=true",
```
ethanshar, line 58: Following our discussion this morning 
I suggest to at least have each item in the content array as an object, something like this 
```
"content": [
  {value: "https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/Docs/thumbnail-horizontal.png?raw=true"},
  {value: "https://github.com/wix/react-native-ui-lib/blob/master/demo/showcase/Docs/thumbnail-horizontal.png?raw=true",}
]

```

This will allow it to easily adapt to changes if we'll need more info (like type, or anything else) without changing the structure

src/components/text/text.api.json
```diff
@@ -12,16 +12,182 @@
...
+          {
+            "title": "Usage",
+            "type": "usage"
+          },
+          {
+            "type": "info"
+          },
+          {
+            "type": "tip"
+          },
```
ethanshar, line 183: In the case of these sections (info and tip) It's better if we'll move the content here as well instead of taking it from the component basic info
This will allow us to add multiple tips/infos in different areas in the component's page

scripts/buildDocs.js
ethanshar, line 1: I think this file is getting too big, let's break it into separate files (main, utils, sections, etc...) and wrap it all under a scripts/docs folder

ethanshar (commented):
Looks great, wrote a few comments..

src/components/text/text.api.json
Inbal-Tish: done

scripts/buildDocs.js
Inbal-Tish: I'll separate to component when I'll get the new ComponentPage component to work

docuilib/src/components/ComponentPage.tsx
ethanshar, line 1: We have a lot of repeating styles like: text styles, row/column blocks, etc..  
Consider moving them to a dedicate `ComponentPage.module.scss` file (see how we use it in other components) you can have a single shared scss file for all sub components 
In any case it will make this code more readable

docuilib/src/components/ComponentPage.tsx
```diff
@@ -0,0 +1,417 @@
...
+    switch (type) {
+      case 'usage':
+        return (
+          <div
+            style={{
+              display: 'flex',
+              flexDirection: 'row',
+              alignItems: 'center',
+              justifyContent: 'space-between',
+              margin: '0 0 12px 0'
```
ethanshar, line 76: I might have confused you when we talked on margins, while there are no marginHorizontal/Vertical in css you do have independent margin prop for Top/Left/Right/Bottom, you can use them when setting margin on one side only

ethanshar (commented):
Looks great! I mostly wrote comments on clean ups

docuilib/src/components/ComponentPage.tsx
Inbal-Tish: I gave ti to the default (line 85), this is the Usage section's header

docuilib/src/components/ComponentPage.tsx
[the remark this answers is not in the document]
Inbal-Tish: Yes. That was the intention. Will do

docuilib/src/components/ComponentPage.tsx
```diff
@@ -0,0 +1,417 @@
...
+        } else {
+          return getImage(value);
+        }
+      }
+    }
+  };
+
+  const getSectionContent = content => {
+    return (
+      <div style={{display: 'flex', flexDirection: 'column'}}>
```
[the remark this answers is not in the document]
Inbal-Tish: I added the flex: 1 to the header, are you still getting it? If I'll add the flex here as well I'll get smaller images for the horizontal layouts (section and list)...

docuilib/src/components/ComponentPage.tsx
Inbal-Tish: I moved the flex styles to scc file but the text style didn't work using className. I'll investigate later

ethanshar: Ok, if something not working easily, add a TODO and let's keep it for another PR

docuilib/src/components/ComponentPage.tsx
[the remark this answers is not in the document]
ethanshar: Technically you can play with the proportions by setting the right flex ratio between Header and Content 
For example: you can set flex: 2 to Header and flex: 3 to Content so images appears bigger then the text

docuilib/src/components/ComponentPage.tsx
```diff
@@ -0,0 +1,417 @@
...
+  };
+
+  const getHeader = (title, description, type) => {
+    const desColor = getDescriptionColor(type);
+
+    if (!title && !description) {
+      return;
+    }
+
+    switch (type) {
```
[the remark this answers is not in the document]
Inbal-Tish: Why? Isn't it more readable with the switch?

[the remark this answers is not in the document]
ethanshar: it's a switch with with a single case and a default it's basically an if-else

docuilib/src/css/custom.css
```diff
@@ -29,3 +29,16 @@
 html[data-theme='dark'] .docusaurus-highlight-code-line {
   background-color: rgba(0, 0, 0, 0.3);
 }
+
+.main-tabs {
```
ethanshar, line 33: maybe move this style to a scss files of the ComponentPage

ethanshar (approved):
Wrote a small comment, anyway approved

docuilib/src/css/custom.css
Inbal-Tish: It's not working...

docuilib/src/components/ComponentPage.tsx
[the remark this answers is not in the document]
Inbal-Tish: oh this one. Done

PR #3308, opened 2024-10-13
Fix sortable list components flexMigration

ethanshar:
## Description
Fix sortable list components flexMigration  to use a container with flexGrow
Using flex caused issue when rendering the list in a specific layout with horizontal direction

```
<View flex>
  <View>
    <SortableList horizontal ...../>
  </View>
<View>

```

## Changelog
Fix sortable list components flexMigration


adids1221 (approved):
Approved

PR #3328, opened 2024-10-29
Fix TextField usage in Usage.md page

ethanshar:
## Description
Fix TextField usage in Usage.md page

## Changelog
Fix TextField usage in Usage.md page


adids1221 (commented):
The `jsx` and the `jsx live` sections are duplicates.
Let's remove the `jsx` section and change the `preset` in the `jsx live` section, so it will change in the `live code` playground.

adids1221 (approved):

PR #3331, opened 2024-10-29
Fix Picker to use new dialog and also fix height issues

ethanshar:
## Description
I fixed here couple of things
- First, I'm always using the new dialog (Incubator.Dialog) in Picker since `useDialog` is a new feature there's no need to go thru migration
- There were some issues when the dialog height. When the user don't pass any height it should based the dialog height on its content. I fixed how the height is set and the layout inside PickerItemsList component

## Changelog
Picker - Fix dialog UI when passing useDialog


adids1221 (approved):

hachem-benamor:
@ethanshar , It seems this issue resolves the problem on iOS and not Android? It works well on iOS.

PR #3388, opened 2024-11-13
Docs/add default devs tab

ethanshar:
## Description
**Depends on https://github.com/wix/react-native-ui-lib/pull/3387**

To avoid having the UX adding the same default Devs tab I added it in our ComponentPage component 
I also fixed an issue I found with rendering extending components link 

## Changelog
N/A


docuilib/src/components/pageComponents/DefaultTabs.ts
```diff
@@ -0,0 +1,32 @@
...
+    //   type: 'banner',
+    //   title: 'Banner',
+    //   description: 'This is some massage to the user',
+    //   color: 'pink',
+    //   icon: ''
+    // },
+    {
+      type: 'props',
+      title: 'API',
+      description: 'This is the list of additional props for the Text component'
```
Inbal-Tish, line 29:
```suggestion
      description: 'This is the list of additional props for the component'
```

docuilib/src/components/pageComponents/DefaultTabs.ts
```diff
@@ -0,0 +1,32 @@
...
+      title: 'Usage',
+      type: 'usage'
+    },
+    {
+      type: 'info'
+    },
+    {
+      type: 'tip'
+    },
+    // {
```
Inbal-Tish, line 14: You can remove the commented code here, this section can be added in noci

docuilib/src/components/pageComponents/DefaultTabs.ts
ethanshar: Fixed manually

docuilib/src/components/pageComponents/DefaultTabs.ts
ethanshar: Fixed

Inbal-Tish (approved):

PR #3394, opened 2024-11-16
Fix/react native svg cssuri path

nitzanyiz:


M-i-k-e-l (commented):
Let's update our usage of `react-native-svg`

M-i-k-e-l (approved):

PR #3434, opened 2024-12-02
Share docs between compound component of the same parent

ethanshar:
## Description
In order to share docs between compound components of the same parent I copy the docs entry from the parent to children 
for example components like (TabController and its children TabController.TabBar, TabController.TabPage, etc...) 

## Changelog
Share docs between compound components


docuilib/src/components/ComponentPage.tsx
```diff
@@ -79,6 +79,7 @@ export default function ComponentPage({component}) {
         // title: name,
         layout: 'horizontal',
         ...hero,
+        title: component.name,
```
Inbal-Tish, line 82: Look at the commented out code. We don't to add the 'incubator' prefix to the incubator components?

Inbal-Tish:
@ethanshar I'm not sure I understand this PR... How will the TabController docs will look like if we want to add docs to both TabBar (exists) and TabPage components?

docuilib/src/components/ComponentPage.tsx
ethanshar: I don't see why not
We want developers to know how to access Incubator components, especially if we still have incubator and non-incubator version (like Dialog and Slider)

ethanshar:
> @ethanshar I'm not sure I understand this PR... How will the TabController docs will look like if we want to add docs to both TabBar (exists) and TabPage components?

I discussed with Yulia, she also preferred to remove 🙄 so I removed

scripts/docs/buildDocsCommon.js
```diff
@@ -60,6 +69,11 @@ function processComponents(components) {
     const isParentComponent = parentComponents.includes(componentName);
     const isIncubatorComponent = component.category === 'incubator';
 
+    if (isCompoundComponent(component.name)) {
+      const parentComponent = getParentComponent(component.name, components);
+      component.docs = parentComponent.docs;
```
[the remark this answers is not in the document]
ethanshar: I removed this logic

docuilib/src/components/ComponentPage.tsx
Inbal-Tish: So use the commented 'name' in the title filed, no?

Inbal-Tish:
> > @ethanshar I'm not sure I understand this PR... How will the TabController docs will look like if we want to add docs to both TabBar (exists) and TabPage components?
> 
> I discussed with Yulia, she also preferred to remove 🙄 so I removed

Then why you did remove the docs for the TabBar after moving them to the TabController? Now there is a duplication...

Inbal-Tish (approved):

PR #3452, opened 2024-12-09
usage section should show LivePreview or CodeBlock

adids1221:
## Description
Docs - `Usage` section, should show live playground or code block.
If the component is part of the `ReactLiveScope` we should show live playground else code block.

## Changelog
Docs - `Usage` section, should show live playground or code block.

## Additional info
None


docuilib/src/components/pageComponents/Usage.tsx
```diff
@@ -2,14 +2,18 @@ import _ from 'lodash';
...
 import '../ComponentPage.module.scss';
+import ReactLiveScope from '../../theme/ReactLiveScope';
+import UILivePreview from '../UILivePreview';
 
 export const Usage = ({component}) => {
+  const componentLivePlaygroundSupport = !!ReactLiveScope[component.name];
+  const code = component.snippet?.map(item => _.replace(item, new RegExp(/\$[1-9]/, 'g'), '')).join('\n');
   if (component.snippet) {
-    return (
+    return componentLivePlaygroundSupport ? (
```
Inbal-Tish, line 12: You can simplify: `return componentLivePlaygroundSupport ? <UILivePreview code={code}/> : <CodeBlock language="jsx">{code}</CodeBlock>;`

Inbal-Tish:
@adids1221 Is this PR ready for review?

adids1221:
@Inbal-Tish I'll ask Miki first about it, after that I'll the PR will be ready for review.

@M-i-k-e-l WDYT about this PR?
Since we've wrapped some snippet's with `function` and added some `states` according to the props.
Instead of wrapping it in a function let the user add the props.
Then we can this PR, if the `ReactLiveScope` has the component then the live play ground should be supported.

PR #3456, opened 2024-12-11
refactor: simplify RTL handling in tab controller components

adids1221:
## Description
`TabController` support RTL on IOS and Android.
After the RN 73 upgrade there isn't and RTL issue on Android but on IOS.
Fixing the issue in `TabController` by reversing the selected index when using RTL on IOS.

## Changelog
`TabController` support RTL on IOS and Android.

## Additional info
MADS-4571

### DIFF
--- a/src/components/tabController/useScrollToItem.ts
+++ b/src/components/tabController/useScrollToItem.ts
@@ -5,6 +5,8 @@ import {useSharedValue} from 'react-native-reanimated';
 import {useScrollTo, ScrollToSupportedViews, ScrollToResultProps} from 'hooks';
 import {Constants} from '../../commons/new';
 
+const FIX_RTL = Constants.isRTL && Constants.isIOS;
+
 export enum OffsetType {
   CENTER = 'CENTER',
   DYNAMIC = 'DYNAMIC',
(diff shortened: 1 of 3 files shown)


adids1221:
@ethanshar Iv'e checked it both on Android and IOS, also on the private repo (`7.34.5-snapshot.5922`).

src/components/tabController/useScrollToItem.ts
```diff
@@ -177,12 +178,13 @@ const useScrollToItem = <T extends ScrollToSupportedViews>(props: ScrollToItemPr
 
   const focusIndex = useCallback((index: number, animated = true) => {
     if (index >= 0 && offsets.CENTER.length > index) {
+      const rtlIndex = Constants.isRTL && Constants.isIOS ? itemsCount - index - 1 : index;
```
ethanshar, line 181: Let's move `Constants.isRTL && Constants.isIOS` to a const at the top of the file like we did before so it's easier to spot there's an RTL fix

ethanshar (commented):
Looks good, wrote a small comment

M-i-k-e-l (approved):

PR #3458, opened 2024-12-12
Docs - add content component tooltip

Inbal-Tish:
## Description
Docs - add content component tooltip

## Changelog
Docs - add content component tooltip

## Additional info
https://wix.atlassian.net/browse/MADS-4558


ethanshar:
https://github.com/user-attachments/assets/0fc7a203-00c8-44da-ba71-58b2c12f50d0

The experience is not great IMO. 
The tooltip doesn't show the whole code but also I think that showing it when hovering the example is a little annoying. 

Let's instead add a button at the right-top of the cell that user can click on. It will show the tooltip and maybe clicking on the tooltip will just copy the code

Inbal-Tish:
@ethanshar I replaced the hover with a button. Copy was already implemented on a tooltip click

ethanshar:
@Inbal-Tish 
Can we position the show code with absolute position? right now it pushes the components example to the side (instead of the center) and it doesn't look good

Also, is there a way to render the code in a code block that formats it nicely?

Issue #3396, opened 2024-11-18
Not Scrollview inside renderSheet


jpcastro4:
## Description

Inside ActionSheet on Android not possible scrollview for much items. 

### Related to

- [x ] Components

### Steps to reproduce
 
Add more 50 items in options

#### Expected behavior

Scroll items

#### Suggested
I am suggest add ScrollView inside renderSheet for resolve bug

```
renderSheet() {
    const {
      renderTitle
    } = this.props;
    const {
      containerStyle
    } = this.props;
    return <ScrollView style={[styles.sheet, containerStyle]}>
        {_isFunction(renderTitle) ? renderTitle() : this.renderTitle()}
        {this.renderActions()}
      </ScrollView>;
  }
```
 

### Affected platforms

- [x ] Android

adids1221:
Hi @jpcastro4 we are working on a solution, we will update.

adids1221:
Duplicate of #2873.

PR #3462, opened 2024-12-17
add scroll container option to ActionSheet component

adids1221:
## Description
ActionSheet support for `ScrollView`.
New `scrollContainer` prop to enable usage of `ScrollView` inside of ActionSheet.

## Changelog
ActionSheet support rendering list inside `ScrollView`.

## Additional info
#3396

### DIFF
--- a/src/components/actionSheet/index.tsx
+++ b/src/components/actionSheet/index.tsx
@@ -1,6 +1,14 @@
 import _ from 'lodash';
 import React, {Component} from 'react';
-import {ActionSheetIOS, StyleSheet, StyleProp, ViewStyle, ImageProps, ImageSourcePropType} from 'react-native';
+import {
+  ActionSheetIOS,
+  StyleSheet,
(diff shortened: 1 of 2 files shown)


src/components/actionSheet/index.tsx
```diff
@@ -101,6 +109,10 @@ type ActionSheetProps = {
    * testID for e2e tests
    */
   testID?: string;
+  /**
+   * Whether to use ScrollView as the container
+   */
+  scrollContainer?: boolean;
```
Inbal-Tish, line 115: Naming convention in this case is more like 'useScrollView' or 'useScrollContainer' since it's a boolean and not element.
But anyways I wonder if we should add this prop at all or figure out in the component when the children height is greater then the container's and use the scrollview instead of a view.

adids1221:
Closing this PR, working on `Incubator.ActionSheet`.

PR #3465, opened 2024-12-23
Feat/picker custom top element

adids1221:
## Description
Picker new `RenderCustomTopElement` prop.
`renderCustomTopElement` will render `CustomTopElement` above the `PickerItemList`.

## Changelog
Picker support for new `customTopElement` and `onItemSelection` callback function.

## Additional info
MADS-4575


src/components/picker/PickerItemsList.tsx
```diff
@@ -161,12 +162,26 @@ const PickerItemsList = (props: PickerItemsListProps) => {
...
+  const renderCustomTopElement = () => {
+    const {isMultiMode, value, setValue} = context;
+    if (customTopElement) {
+      if (isMultiMode) {
+        console.log(`renderCustomTopElement, isMultiMode!, value:`, value);
+        //@ts-expect-error - PickerWithMultiValue props need to pass the correct props
+        return customTopElement({value, setValue});
+      }
+      //@ts-expect-error - PickerWithSingleValue props
+      return customTopElement();
```
M-i-k-e-l, line 174: I don't like the difference in API

src/components/picker/types.ts
```diff
@@ -237,12 +252,20 @@ export type PickerBaseProps = Omit<TextFieldProps, 'value' | 'onChange'> &
   };
 
 export type PickerPropsWithSingle = PickerBaseProps & {
+  /**
+   * Custom top element
+   */
+  customTopElement?: () => React.ReactElement;
```
M-i-k-e-l, line 258: You're calling the prop `customTopElement` but it's actually a `renderCustomTopElement` (it's a function)

src/components/picker/PickerItemsList.tsx
```diff
@@ -161,12 +162,26 @@ const PickerItemsList = (props: PickerItemsListProps) => {
     );
   };
 
+  const renderCustomTopElement = () => {
+    const {isMultiMode, value, setValue} = context;
+    if (customTopElement) {
+      if (isMultiMode) {
+        console.log(`renderCustomTopElement, isMultiMode!, value:`, value);
+        //@ts-expect-error - PickerWithMultiValue props need to pass the correct props
+        return customTopElement({value, setValue});
```
M-i-k-e-l, line 171: We should avoid exposing internal workings on the components.
What if we want to change it in the future?

src/components/picker/types.ts
```diff
@@ -5,6 +5,17 @@ import {ModalTopBarProps} from '../modal/TopBar';
...
 
+export interface CustomTopElementProps {
+  /**
+   * The current multi draft value
+   */
+  value: PickerMultiValue;
+  /*
+   * Set the new multi draft value
+   */
+  setValue: (value: PickerMultiValue) => void;
```
M-i-k-e-l, line 16: Let's remove the `setValue` and only expose the `value`, the user can set their internal state thus affecting the component.

src/components/picker/PickerItemsList.tsx
adids1221: You are right, fixed.

src/components/picker/types.ts
```diff
@@ -206,6 +210,10 @@ export type PickerBaseProps = Omit<TextFieldProps, 'value' | 'onChange'> &
       itemProps: PickerItemProps & {isSelected: boolean; isItemDisabled: boolean},
       label?: string
     ) => React.ReactElement;
+    /**
+     * Custom top element, value prop is relevant for multi picker use cases
```
M-i-k-e-l, line 214: Please fix the docs (it's not only for multi picker)

src/components/picker/types.ts
```diff
@@ -180,6 +180,10 @@ export type PickerBaseProps = Omit<TextFieldProps, 'value' | 'onChange'> &
      * Callback for when picker value change
      */
     onChange?: (value: PickerValue) => void;
+    /**
+     * Callback for when picker item is selected (only for multi mode)
+     */
+    onItemSelection?: (value: PickerMultiValue) => void;
```
M-i-k-e-l, line 186: Is this still needed?

src/components/picker/index.tsx
```diff
@@ -145,6 +148,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
     return {
       migrate,
       value: mode === PickerModes.MULTI ? multiDraftValue : pickerValue,
+      setValue: mode === PickerModes.MULTI ? setMultiDraftValue : undefined,
```
M-i-k-e-l, line 151: Is this change needed?

src/components/picker/PickerItemsList.tsx
```diff
@@ -167,6 +168,7 @@ const PickerItemsList = (props: PickerItemsListProps) => {
     ) : (
       <>
         {renderSearchInput()}
+        {!!renderCustomTopElement && renderCustomTopElement(context.value)}
```
M-i-k-e-l, line 171:
```suggestion
        {renderCustomTopElement?.(context.value)}
```

src/components/picker/index.tsx
adids1221: Not in use anymore ,removed

M-i-k-e-l (approved):
Please add `renderCustomTopElement` to the API
Also fix the PR's description (it only adds `renderCustomTopElement`, name change etc)

PR #3470, opened 2024-12-26
New component - PieChart

nitzanyiz:
## Description
Added a new component PieChart. The component gets segments which is an array of percentages to that each segment of the pie should cover. Each segment is a percentage representing the percentage of the pie to cover and a color for the pie segment.

## Changelog
⭐  PieChart - New Component ⭐ 

## Additional info
MADS-3418


ethanshar:
@nitzanyiz 
When creating a new component, always create an example screen with it
It also makes it easier for the reviewer to check the new component

demo/src/screens/MenuStructure.js
```diff
@@ -44,7 +44,8 @@ export const navigationData = {
         screen: 'unicorn.components.SharedTransitionScreen'
       },
       {title: 'Stack Aggregator', tags: 'stack aggregator', screen: 'unicorn.components.StackAggregatorScreen'},
-      {title: 'Marquee', tags: 'sliding text', screen: 'unicorn.components.MarqueeScreen'}
+      {title: 'Marquee', tags: 'sliding text', screen: 'unicorn.components.MarqueeScreen'},
+      {title: 'PieChart', tags: 'pie chart data', screen: 'unicorn.components.PieChartScreen'}
```
ethanshar, line 48: Let's put this new component under Charts category, we don't have this category in public yet but it does exist in private

src/index.ts
```diff
@@ -82,6 +82,7 @@ export {default as HapticService, HapticType} from './services/HapticService';
 export {default as Hint, HintProps} from './components/hint';
 export {default as Icon, IconProps} from './components/icon';
 export {default as Image, ImageProps} from './components/image';
+export {default as PieChart, PieChartSegmentProps} from './components/piechart';
```
ethanshar, line 85: try keeping the exports order alphabetically

src/components/piechart/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const DEFAULT_SIZE = 144;
+
+const PieChart = (props: PieChartProps) => {
+  const {segments, size = DEFAULT_SIZE} = props;
+
+  const total = useMemo(() => {
+    return _.sum(segments.map(s => s.percentage));
+  }, [segments]);
+  if (total !== 100) {
```
ethanshar, line 21: I wonder if we should allow a total smaller than 100 
I saw partial pie charts before, maybe users would want that. We can have the missing part as blank

src/components/piechart/PartialCircle.tsx
```diff
@@ -0,0 +1,78 @@
...
+  color: string;
+  startAngle?: number;
+  padding?: number;
+};
+
+const DEFAULT_DIVIDER_COLOR = '#FFFFFF';
+const DEFAULT_DIVIDER_WIDTH = 4;
+const DEFAULT_PADDING = 0;
+
+const PartialCircle = (props: PartialCircleProps) => {
```
ethanshar, line 19: I find Segment or PieSegment a better name for this component, WDYT?

src/components/piechart/PartialCircle.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export type PartialCircleProps = {
+  percentage: number;
+  radius: number;
+  color: string;
+  startAngle?: number;
+  padding?: number;
+};
+
+const DEFAULT_DIVIDER_COLOR = '#FFFFFF';
```
ethanshar, line 15: I suggest using our Colors.$backgroundDefault here so in dark mode it won't stay white but use the default background color

src/components/piechart/index.tsx
```diff
@@ -0,0 +1,41 @@
+import React, {useMemo} from 'react';
+import _ from 'lodash';
+import View from '../view';
+import PartialCircle, {PartialCircleProps} from './PartialCircle';
+
+export type PieChartSegmentProps = Pick<PartialCircleProps, 'percentage' | 'color'>;
+
+export type PieChartProps = {
+  segments: PieChartSegmentProps[];
+  size?: number;
```
ethanshar, line 10: consider adding a `defaultSegmentProps` where the user can control default UI for all segments like the divider color and width

src/components/piechart/index.tsx
ethanshar, line 1: Fix folder name to `pieChart`

src/components/piechart/PartialCircle.tsx
ethanshar, line 1: Since you are using an optional dependency it's possible the user won't have it installed, what we usually do is warn the user so they'll know to install it when using this component 
moreover don't render the component if it's not installed, otherwise it will throw the user an error

You can see a reference for this behavior in our Card component that depends on blurView dep

ethanshar (commented):
Looks great! 
Wrote a few comment

src/components/piechart/PartialCircle.tsx
nitzanyiz: I think PieSegment is more clear. I'll change.

src/components/piechart/index.tsx
nitzanyiz: This actually works just by removing the error. Changed.

src/components/piechart/PartialCircle.tsx
nitzanyiz: Is it ok to se tit in the global scope of the file or should it be at the component render? my meaning is using `const DEFAULT_DIVIDER_COLOR = Colors.$backgroundDefault;` inside the component and not out side of it.

nitzanyiz:
@ethanshar , I made all the changes and added some to the api (border width and color). Also I added the api json.

src/components/piechart/PartialCircle.tsx
ethanshar: For Branded and Editor stage purposes it's better if you'll set it directly when destructing the prop here
```
dividerColor = DEFAULT_DIVIDER_COLOR
```

src/components/piechart/index.tsx
ethanshar: Cool, don't forget to add them also to the api.json

src/components/piechart/PartialCircle.tsx
ethanshar: I suggest moving the validation to the main component so the user won't get an error per segment but only once

ethanshar:
@nitzanyiz looks good, see my replies - left few small things. Ping me on slack when it's ready and I'll approve

src/components/pieChart/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+} & Pick<PieSegmentProps, 'dividerWidth' | 'dividerColor'>;
+
+const DEFAULT_DIAMETER = 144;
+
+const PieChart = (props: PieChartProps) => {
+  const {segments, diameter = DEFAULT_DIAMETER, ...others} = props;
+
+  if (!Svg || !Path) {
+    console.error(`RNUILib PieChart requires installing "@react-native-svg" dependency`);
+  }
```
ethanshar, line 27: Let's return here null after showing the error to avoid an actual exception that might crash the whole screen or tests

ethanshar (approved):
Wrote one last comment, you can fix and merge. I approved

PR #3482, opened 2025-01-12
TextField - ClearButton - fix right margin

M-i-k-e-l:
## Description
TextField - ClearButton - fix right margin with `preset={'outline'}`

To test use:
```
<TextField value={'Test '} showClearButton preset={'underline'}/>
<TextField value={'Test '} showClearButton preset={'outline'}/>
```

## Changelog
TextField - ClearButton - fix right margin with `preset={'outline'}`

## Additional info
See internal channel

### DIFF
--- a/src/components/textField/index.tsx
+++ b/src/components/textField/index.tsx
@@ -208,7 +208,12 @@ const TextField = (props: InternalTextFieldProps) => {
             </View>
           )}
           {showClearButton && (
-            <ClearButton onClear={onClear} testID={`${props.testID}.clearButton`} onChangeText={onChangeText}/>
+            <ClearButton
+              onClear={onClear}
+              testID={`${props.testID}.clearButton`}
+              onChangeText={onChangeText}
+              preset={props.preset}
+            />
           )}
           {trailingAccessory}
           {/* </View> */}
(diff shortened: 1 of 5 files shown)


src/components/textField/index.tsx
```diff
@@ -208,7 +208,12 @@ const TextField = (props: InternalTextFieldProps) => {
             </View>
           )}
           {showClearButton && (
-            <ClearButton onClear={onClear} testID={`${props.testID}.clearButton`} onChangeText={onChangeText}/>
+            <ClearButton
+              onClear={onClear}
+              testID={`${props.testID}.clearButton`}
+              onChangeText={onChangeText}
+              preset={props.preset}
```
Inbal-Tish, line 215: The ClearButton component shouldn't be aware of the preset. I think we should pass 'style' instead

M-i-k-e-l: Fixed

src/components/textField/presets/outline.ts
```diff
@@ -14,5 +15,6 @@ const styles = StyleSheet.create({
 
 export default {
   ...underline,
+  clearButtonStyle: styles.clearButtonStyle,
```
Inbal-Tish, line 18: You get this from the underline speared, so if you don't change it here you can remove this line

Inbal-Tish (approved):

PR #3483, opened 2025-01-14
Update ButtonProps to use IconProps for iconProps type

nitzanyiz:
## Description
Changed `ButtonProps.iconProps` to use `IconProps` instead of `ImageProps`.

## Changelog
Button - Fixed `iconProps` to use `IconProps`;

## Additional info
MADS-4593


adids1221 (approved):
Approved

PR #3486, opened 2025-01-21
Hint Refactor

ethanshar:
## Description
Refactor Hint component to make it more readable and maintainable 
- I changed implementation for the hint positioning so it cover more random edge cases 
- I broke into smaller components
- I add separated logic to hooks 
- I reduced a lot of code redundancy  
- Implemented a new driver and added minimal tests (there are some limitations with testing the Hint component)

**How to review**
I started with breaking this refactor into several different PRs, but TBH I think it's best to just review it as new code from scratch so I submitted only the last PR 

## Changelog
Refactor and fix Hint component edge case position issues


Inbal-Tish:
@ethanshar The hint with right targetPosition and custom content is broken:
Hint-Right-CustomContent-Reactions

ethanshar:
> @ethanshar The hint with right targetPosition and custom content is broken: ![Hint-Right-CustomContent-Reactions]

The issue is in the HintScreen, It didn't update the key when using the custom content feature 
Now it does and the calculation works properly

ethanshar:
@Inbal-Tish @nitzanyiz 
Please review this one, we have few GA blockers pending on this

nitzanyiz:
@ethanshar , I was wondering if we can keep all the visibility of the hint inside the `useHintAnimation` hook (maybe rename it to `useHintVisibility` and that it would return something like `{isVisible, visibilityProgress (/AnimatedValue)`). We could also keep the
```
  useDidUpdate(() => {
    animateHint();
  }, [visible]);
```
inside this hook

nitzanyiz:
Not sure if if we want to handle this in this pr. When rendering hint with a short message in RTL the hint is disconnected from the hint bubble.

| Hebrew message | English message|
| -- | -- |
| |  |

ethanshar:
@nitzanyiz 
Fixed and added useHintVisibility as suggested

demo/src/screens/componentScreens/HintsScreen.tsx
```diff
@@ -215,6 +223,19 @@ export default class HintsScreen extends Component<HintScreenProps> {
               </View>
             </>
           )}
+
+          <View marginT-100 row center>
+            {targetPosition !== 'flex-start' && <Text marginH-s3>Text pushing button</Text>}
+            <Hint
+              message={'Hint'}
```
Inbal-Tish, line 230: We should use 'message' so we can text the long massage here as well

demo/src/screens/componentScreens/HintsScreen.tsx
```diff
@@ -215,6 +223,19 @@ export default class HintsScreen extends Component<HintScreenProps> {
...
+              visible={showSecondHint}
+              onBackgroundPress={this.toggleSecondHint}
+              useSideTip={false}
```
Inbal-Tish, line 233: useSideTip must be false or we can play with this from the switches?

demo/src/screens/componentScreens/HintsScreen.tsx
ethanshar: The whole idea of this use case was to test short message, that's why it's hard coded
I can change it, I just need the short variation of message to be extra short

demo/src/screens/componentScreens/HintsScreen.tsx
ethanshar: Passing true will result in a wrong UI that looks bad, it's not a bug, it's just not meant for that so I rather pass false

demo/src/screens/componentScreens/HintsScreen.tsx
ethanshar: I looked at it again, I prefer to keep it as it is

nitzanyiz (approved):

PR #3494, opened 2025-01-30
V8

M-i-k-e-l:


ethanshar:
@M-i-k-e-l 
Last time we opened a dedicate branch for the next major version, we had to maintain it for a while and keep fixing conflicts so we decided to create it only on the day of the migration. So consider what you want to do here

M-i-k-e-l:
> @M-i-k-e-l Last time we opened a dedicate branch for the next major version, we had to maintain it for a while and keep fixing conflicts so we decided to create it only on the day of the migration. So consider what you want to do here

Yes, I'm aware, I hope it won't take too long and it'll work out this time...

PR #3242, opened 2024-09-04
Picker dialog validation

adids1221:
## Description
Picker Dialog validation feature.
New `UsePickerDialogProps` hook to determine the default props of the dialog instead of the `index` file.
`UsePickerSelection` hook handles the validation and trigger the `selectionValidation` function.

## Changelog
Picker Dialog validation feature.

## Additional info
MADS-4430

PR #3497, opened 2025-02-04
Add selection validation support

adids1221:
## Description
Picker Dialog validation feature (supported only in the new Dialog).
New `UsePickerDialogProps` hook to determine the default props of the dialog instead of the `index` file.
`UsePickerSelection` hook handles the validation and trigger the `selectionValidation` function.
Also in this PR there is a fix, Picker Dialog shouldn't render `Modal.TopBar` (at least when the user pass `migrateDialog`).
(old PR #3242)

## Changelog
Picker Dialog validation feature.

## Additional info
MADS-4430


M-i-k-e-l (commented):
I don't understand why this feature is only for Dialog.
Maybe `shouldDisableDoneButton` should be `shouldDisableSelection`?

M-i-k-e-l:
Lets separate the `Modal.TopBar` fix to another PR, even if it's very small.

adids1221:
> Lets separate the `Modal.TopBar` fix to another PR, even if it's very small.

The `Modal.TopBar` fix is relevant to this Pr it shouldn't be in separate PR.

adids1221:
@M-i-k-e-l about supporting Modal, I don't see this right now.
We don't have the relevant ux for that.
If we'll want to support Modal in the future will do it in separate PR.

adids1221:
Closing this feature should be handled in private

PR #3501, opened 2025-02-12
TabBar - Better accessibility roles and instructions

nitzanyiz:
## Description
- Added `tab` accessibility role to the tab items and `tablist` role to the tab controller.
- Added selected accessibility state - It was reading that everything was not selected, even when the tab was selected.
- [x] Remove the state and test the instructions it gives on a real device. When it is not passed it doesn't give feedback that the tab was selected and doesn't state if the tab is selected when you focus it.

**We can give each tab item an isSelected state and change it accordingly so a state change will only effect the specific TabItem instead of handling it in the context scope and passing it down. wdyt?**
## Changelog
TabBar - Added better accessibility roles.

## Additional info
MADS-4590


src/components/tabController/TabBarItem.tsx
```diff
@@ -226,6 +226,9 @@ export default function TabBarItem({
         style={_style}
         onLayout={onLayout}
         testID={testID}
+        accessible
+        accessibilityRole="tab"
+        accessibilityState={{selected: selectedIndex === index}}
```
M-i-k-e-l, line 231: I think you can use `currentPage.value` to remove most of the code in the PR
```suggestion
        accessibilityState={{selected: currentPage.value === index}}
```

M-i-k-e-l (commented):
Not sure why but it's not working for me

nitzanyiz:
Did you try on a real device? It not working in the inspector but does on a real device.

M-i-k-e-l:
Yes, I've tested on a real device withe the public demo

src/components/tabController/TabBarItem.tsx
M-i-k-e-l: Maybe it's the re-selection, I'll re-test

nitzanyiz: Changed. I added `useAnimatedReaction` in the `TabBarItem` that will change its state only when become selected or unselected.

nitzanyiz:
@M-i-k-e-l I changed the state to be controlled in the `TabBarItem`. Wdyt?

src/components/tabController/TabBarItem.tsx
```diff
@@ -226,6 +233,9 @@ export default function TabBarItem({
         style={_style}
         onLayout={onLayout}
         testID={testID}
+        accessible
+        accessibilityRole="tab"
+        accessibilityState={{selected: isSelected}}
```
M-i-k-e-l, line 238: Let's memoize this

M-i-k-e-l (approved):

src/components/tabController/TabBarItem.tsx
nitzanyiz: Done 👌

PR #3502, opened 2025-02-12
Refactor HintDriver to allow optional style flattening in tests

ethanshar:
## Description
Add support to flatten style in View driver, this helps with testing specific cases when styles are a complex, nested object

## Changelog
Support flatten style in View driver


adids1221 (approved):
Approved

PR #3504, opened 2025-02-13
TextField - Increase hitslop

nitzanyiz:
## Description
Increased the hit slop of the text field.

## Changelog
TextField - Hit target increased

## Additional info
MADS-4597


ethanshar (approved):
@nitzanyiz 
Approved. 
This might break tests in private so make sure to fix private after merging this one

PR #3507, opened 2025-02-13
Fix build on real device

M-i-k-e-l:
## Description
Fix build on real device

## Changelog
Fix build on real device

## Additional info
None


ios/rnuilib/AppDelegate.mm
```diff
@@ -27,7 +27,7 @@ - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
 - (NSURL *)getBundleURL
 {
 #if DEBUG
-  return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"demo"];
+  return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index"];
```
Inbal-Tish, line 30: Which index are you pointing to here?

PR #3509, opened 2025-02-16
Incubator.Toast - fix voice over

M-i-k-e-l:
## Description
Incubator.Toast - fix voice over

## Changelog
Incubator.Toast - fix voice over

## Additional info
Ticket 4618


Inbal-Tish:
@M-i-k-e-l I don't get how I can test this, the Toast in the DialogScreen is not auto dismissing so I can tell if the voice finishes the reading before that. What am I missing?

Inbal-Tish (approved):

PR #3529, opened 2025-02-25
Picker - Larger hit target for accessibility

nitzanyiz:
## Description
When the Picker is in the regular form type, with the text field, the picker hit target is large enough to fix the 48x48 hit box. When the picker is in the other field types - settings and filter its smaller - no label on top so the hit target is small. I made the hit target a little larger when its in these special cases.
## Changelog
Picker - Larger hit target for settings and filter pickers

## Additional info
MADS-4591


nitzanyiz:
@adids1221 I added a little more just to reach the 48x48 with the regular font.

PR #3531, opened 2025-02-26
Slider - Increased hit target

nitzanyiz:
## Description
Added accessible hitSlop to the slider thumb.

## Changelog
Incubator.Slider - Increased thumb hitSlop.

## Additional info
MADS-4591

### DIFF
--- a/src/incubator/slider/Thumb.tsx
+++ b/src/incubator/slider/Thumb.tsx
@@ -28,7 +28,13 @@ interface ThumbProps extends ViewProps {
 
 const SHADOW_RADIUS = 4;
 const THUMB_SIZE = 24;
-const DEFAULT_THUMB_HIT_SLOP = {top: 12, bottom: 12, left: 12, right: 12} as const;
+const THUMB_ACCESSIBLE_HITSLOP = Math.max(0, 48 - THUMB_SIZE) / 2;
+const DEFAULT_THUMB_HIT_SLOP = {
+  top: THUMB_ACCESSIBLE_HITSLOP,
+  bottom: THUMB_ACCESSIBLE_HITSLOP,
+  left: THUMB_ACCESSIBLE_HITSLOP,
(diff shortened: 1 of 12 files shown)


src/incubator/slider/Thumb.tsx
```diff
@@ -28,14 +28,21 @@ interface ThumbProps extends ViewProps {
 
 const SHADOW_RADIUS = 4;
 const THUMB_SIZE = 24;
+const THUMB_ACCESSIBLE_HITSLOP = Math.max(0, 48 - THUMB_SIZE) / 2;
```
adids1221, line 31: As we discussed before, `hitSlop` can be `Rect or number`.
Since all the values are the same let's pass the `THUMB_ACCESSIBLE_HITSLOP` number as default.
Please check it and confirm.

nitzanyiz: The thumb is view and not a pressable. For some reason view's hit slop doesn't receive number :/
https://reactnative.dev/docs/0.73/view#hitslop

adids1221 (approved):

adids1221:
@nitzanyiz Approved as we discussed about it.
Please update int he description that the hitSlop grows when pressing on the Slider thumb.

src/incubator/slider/Thumb.tsx
```diff
@@ -28,14 +28,21 @@ interface ThumbProps extends ViewProps {
...
+const DEFAULT_THUMB_HIT_SLOP = {
```
Inbal-Tish, line 32: You don't need the DEFAULT_THUMB_HIT_SLOP. If all sides are equal you can pass one value - THUMB_ACCESSIBLE_HITSLOP

nitzanyiz: https://github.com/wix/react-native-ui-lib/pull/3531#discussion_r1977553272

PR #3545, opened 2025-03-02
fix badge should be pimple when label undefined

adids1221:
## Description
Badge should be pimple size when label undefined

## Changelog
Fix Badge should be pimple size when label undefined

## Additional info
MADS-4577

### DIFF
--- a/src/components/badge/index.tsx
+++ b/src/components/badge/index.tsx
@@ -18,8 +18,9 @@ import Image from '../image';
 import View from '../view';
 import Text from '../text';
 
-
 const LABEL_FORMATTER_VALUES = [1, 2, 3, 4] as const;
+const DEFAULT_PIMPLE_SIZE = 10;
+const DEFAULT_BADGE_SIZE = 20;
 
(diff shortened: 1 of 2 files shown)


ethanshar:
@adids1221 some tests fail

src/components/badge/index.tsx
```diff
@@ -121,7 +121,8 @@ class Badge extends PureComponent<BadgeProps> {
   }
 
   get size() {
-    return this.props.size || 20;
+    const {size, label} = this.props;
+    return label === undefined ? 10 : size !== undefined ? size : 20;
```
ethanshar, line 125: You can make this last line more readable by setting a default value to size when destructing it

ethanshar: That's an improvement and way more readable
Just double checking on what I wrote, can't we set the default size (`DEFAULT_BADGE_SIZE`)  when destructing the size prop?

ethanshar: 👍 right

ethanshar (approved):

PR #3548, opened 2025-03-03
add customizable header button props and deprecate doneButtonColor

adids1221:
## Description
ColorPickerDialog new `colorPickerHeaderProps` prop to customize the header button icons.
`doneButtonColor` prop deprecation, moved into the `colorPickerHeaderProps`.
ColorSwatch new `selectedIcon` prop to customize the selected button.

## Changelog
ColorPickerDialog and ColorSwatch buttons and icons customize props.

## Additional info
MADS-4531


M-i-k-e-l:
Let's try to move the assets to `Assets.components` like we've discussed

PR #3554, opened 2025-03-05
Avatar animate mode not supported on web

adids1221:
## Description
AnimatedImage doesn't work on web platform.
When using animated Avatar we render AnimatedImage which doesn't work, for now when using web we will set animate to false to prevent Avatar not to render as expected.
snippet:
```
<Avatar source={src} animate /> 
```

## Changelog
Avatar with animated not supported on web.

## Additional info
MADS-4577

### DIFF
--- a/src/components/avatar/index.tsx
+++ b/src/components/avatar/index.tsx
@@ -21,6 +21,7 @@ import AnimatedImage, {AnimatedImageProps} from '../animatedImage';
 import * as AvatarHelper from '../../helpers/AvatarHelper';
 import {useThemeProps} from '../../hooks';
 import {isSvg} from '../../utils/imageUtils';
+import Constants from '../../commons/Constants';
 
 export enum BadgePosition {
   TOP_RIGHT = 'TOP_RIGHT',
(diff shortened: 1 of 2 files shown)


src/components/avatar/index.tsx
```diff
@@ -287,7 +288,7 @@ const Avatar = forwardRef<any, AvatarProps>((props: AvatarProps, ref: React.Forw
   const renderImage = () => {
     if (source !== undefined) {
       // Looks like reanimated does not support SVG
-      const ImageContainer = animate && !isSvg(source) ? AnimatedImage : Image;
+      const ImageContainer = animate && !isSvg(source) && !Constants.isWeb ? AnimatedImage : Image;
```
ethanshar, line 291: Let's also update in api.json about this prop that it's not supported on Web

ethanshar (approved):
Approved, but see my comment

src/components/avatar/index.tsx
adids1221: Fixed, missed that.

PR #3559, opened 2025-03-06
Documentation - Direct link to props in props list

nitzanyiz:
## Description
Added direct link to props in the props list. Changing the prop from a div to a heading and passing an id added a direct link by default. Currently this doesn't work in dev. I'm wondering if this will work in the production build.

## Changelog
N/A

## Additional info
MADS-4530


M-i-k-e-l:
How can I reproduce this? I did not see it

nitzanyiz:
Go to `react-native-ui-lib/docs/next/components/status/Badge` and then the dev section. There should be a `#` next to each prop.

docuilib/src/components/ComponentPage.tsx
```diff
@@ -63,7 +63,7 @@ export default function ComponentPage({component}) {
 
     // TODO: align Tabs bottom border with TabItem's selected indication line
     if (tabs) {
-      return <Tabs className="main-tabs">{getTabItems([...tabs, devTab])}</Tabs>;
+      return <Tabs queryString="tab" className="main-tabs">{getTabItems([...tabs, devTab])}</Tabs>;
```
M-i-k-e-l, line 66: What does the `queryString` do?

docuilib/src/components/ComponentPage.tsx
```diff
@@ -49,9 +49,9 @@ export default function ComponentPage({component}) {
   /** Tabs */
 
   const getTabItems = tabs => {
-    return _.map(tabs, (tab, index) => {
+    return _.map(tabs, (tab) => {
       return (
-        <TabItem value={index} label={tab.title} attributes={{className: 'single-tab'}}>
+        <TabItem value={tab.title.toLowerCase()} label={tab.title} attributes={{className: 'single-tab'}}>
```
M-i-k-e-l, line 54: Does it matter it's `lowerCase`?

docuilib/src/components/pageComponents/PropsList.tsx
```diff
@@ -47,24 +47,35 @@ export const PropsList = ({props}) => {
...
+        <>
+          <Heading key={`${prop.name}-heading`} as="h4" style={{display: 'flex', flexDirection: 'row', alignItems: 'center'}} id={prop.name}>
+            {prop.name}
             {getTag(prop.type, getTypeColor(prop.type))}
             {prop.required && getTag('Required', getTypeColor())}
             {prop.platform && getTag(prop.platform, getTypeColor())}
-          </div>
-          <span style={{display: 'block', marginBottom: prop.note ? 12 : 28, fontSize: '16px', fontWeight: '400', color: descColor}}>
+          </Heading>
+          <span
```
M-i-k-e-l, line 62: The prettiying part is making the rest of the PR harder than it should.
In the future, please move to a different commit or even better: push to master without a review.

M-i-k-e-l (commented):
Your description reads: `Currently this doesn't work in dev` - what doesn't work?

docuilib/src/components/ComponentPage.tsx
nitzanyiz: This adds the currently selected tab to the url so now when you'll copy this link and go to it in a new tab it will open the tab that was selected. This is needed that if we get a link to some header it will open the appropriate tab.

docuilib/src/components/ComponentPage.tsx
nitzanyiz: No really. the query string I just felt that `dev`, `overview`... are better as query string than `Dev`, `Overview`...

docuilib/src/components/pageComponents/PropsList.tsx
nitzanyiz: Sorry 😅

M-i-k-e-l (approved):

M-i-k-e-l:
@nitzanyiz you have a conflict

PR #3579, opened 2025-03-09
Added jsx rendering in docs

nitzanyiz:
## Description
Adding option to render jsx in the component item.

## Changelog
N/A

## Additional info
N/A


docuilib/src/components/pageComponents/ContentItem.tsx
```diff
@@ -76,21 +79,32 @@ type Item = {
...
+  if (!snippet.startsWith('<')) {
+    return;
+  }
+  const firstWord = snippet.split(' ')[0];
+  return firstWord.slice(1);
+};
+
 export const ContentItem = ({item, componentName, showCodeButton}: ContentItemProps) => {
-  const getFigmaEmbed = item => {
+  const getFigmaEmbed = (item: {value: string; height?: number}) => {
```
Inbal-Tish, line 100: You can change the item to value: string; height?: number (like in the getImage func)

docuilib/src/components/pageComponents/ContentItem.tsx
```diff
@@ -76,21 +79,32 @@ type Item = {
...
     const value = item.value;
     const height = item.height || 450;
 
     return <iframe width={'100%'} height={height} src={value}/>;
   };
 
-  const getImage = (value, style = undefined) => {
+  const getImage = (value: string, style: ComponentProps<'img'>['style'] = undefined) => {
```
Inbal-Tish, line 107: If you added the types you can make style optional instead of init with undefined

docuilib/src/components/pageComponents/ContentItem.tsx
```diff
@@ -8,11 +8,12 @@ import CodeIcon from '../../assets/icons/code';
 
 type ComponentItemProps = {
   componentName: string;
-  props: Record<string, unknown> | Record<string, unknown>[];
+  props?: Record<string, unknown> | Record<string, unknown>[];
+  jsx?: string
```
Inbal-Tish, line 12: Why not just call it 'snippet' or 'element'? We're talking about a string containing an element, like `<Button label={'Press'}/>`, right?

docuilib/src/components/pageComponents/ContentItem.tsx
nitzanyiz: Changed.

docuilib/src/components/pageComponents/ContentItem.tsx
nitzanyiz: Changed.

docuilib/src/components/pageComponents/ContentItem.tsx
nitzanyiz: I feel like jsx is much more informative but I guess we can change it to snippet.

Inbal-Tish (approved):

PR #3599, opened 2025-03-11
Add support for foundation category in documentation build process

ethanshar:
## Description
Add support for foundation category in documentation build process
Now it will create a foundation dir at root level of the docs instead of creating it inside the components folder

## Changelog
NA


nitzanyiz (approved):
👌 
We also need to bump the version 🙏

PR #3600, opened 2025-03-11
Dev tab - Don't render empty sections or the tab

Inbal-Tish:
## Description
Dev tab - Don't render empty sections or the tab:
No usage without snipper, no banner without description and no Dev tab without props

## Changelog
Dev tab - Don't render empty sections or the tab

### DIFF
--- a/docuilib/src/components/ComponentPage.tsx
+++ b/docuilib/src/components/ComponentPage.tsx
@@ -60,10 +60,12 @@ export default function ComponentPage({component}) {
 
   const buildTabs = () => {
     const tabs = component.docs?.tabs;
-
+    const api = component.props;
+    const tabsArray = api ? [...tabs, devTab] : tabs;
+    
     // TODO: align Tabs bottom border with TabItem's selected indication line
     if (tabs) {
-      return <Tabs className="main-tabs">{getTabItems([...tabs, devTab])}</Tabs>;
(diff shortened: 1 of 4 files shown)


docuilib/src/components/ComponentPage.tsx
```diff
@@ -60,10 +60,12 @@ export default function ComponentPage({component}) {
...
   const buildTabs = () => {
     const tabs = component.docs?.tabs;
-
+    const api = component.props;
+    const tabsArray = api ? [...tabs, devTab] : tabs;
+    
     // TODO: align Tabs bottom border with TabItem's selected indication line
     if (tabs) {
-      return <Tabs className="main-tabs">{getTabItems([...tabs, devTab])}</Tabs>;
+      return <Tabs className="main-tabs">{getTabItems(tabsArray)}</Tabs>;
```
ethanshar, line 68: If we remove Dev tab based on props, it won't render things like notes and tips, no? 
Which component/page has this issue (has no api)?

Inbal-Tish: That's why I based it on the props - that the most important feature that without it I don't think we need the tab. 
Can you think of a case where we don't have props but need to display the extend/modifiers/snippet?

ethanshar: Not really, approving, worse case we'll change it

ethanshar (approved):

PR #3609, opened 2025-03-12
Component Utils for docs

adids1221:
## Description
New componentUtils file to include common logic, in this case `isComponentSupported` which is used in `UILivePreview` and `Usage` components.

## Changelog
New componentUtils file to include common logic.

## Additional info
None


nitzanyiz (approved):
🚀

PR #3610, opened 2025-03-12
Docs - LiveCode in table cell - block interaction

Inbal-Tish:
## Description
Docs - LiveCode in table cell - block interaction

## Changelog
Docs - LiveCode in table cell - block interaction

### DIFF
--- a/docuilib/src/components/pageComponents/ContentItem.tsx
+++ b/docuilib/src/components/pageComponents/ContentItem.tsx
@@ -62,9 +62,12 @@ const ComponentItem = (props: ComponentItemProps) => {
 
   const content = showCode ? codePreview : componentPreview;
 
+  const interactionBlocker = <div style={{position: 'absolute', top: 0, left: 0, height: '100%', width: '100%'}}/>;
+
   return (
     <div className={`${styles.componentItemContainer} ${!showCode ? styles.componentSpotlightStyle : ''}`}>
       {content}
+      {interactionBlocker}
       {showCodeButton && (
         <button onClick={toggleCode} className={styles.showCodeButton}>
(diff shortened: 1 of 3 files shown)


docuilib/src/components/pageComponents/ContentItem.tsx
```diff
@@ -62,9 +62,12 @@ const ComponentItem = (props: ComponentItemProps) => {
 
   const content = showCode ? codePreview : componentPreview;
 
+  const interactionBlocker = <div style={{position: 'absolute', top: 0, left: 0, height: '100%', width: '100%'}}/>;
+
   return (
     <div className={`${styles.componentItemContainer} ${!showCode ? styles.componentSpotlightStyle : ''}`}>
       {content}
+      {interactionBlocker}
```
nitzanyiz, line 70: Because this div is over the content when the content is the codePreview the user can select the code or copy it. It blocks the interactions with it. Perhaps we can just wrap the `componentPreview` with a div that has `pointerEvents: none`.

docuilib/src/components/pageComponents/ContentItem.tsx
```diff
@@ -53,9 +53,11 @@ const ComponentItem = (props: ComponentItemProps) => {
   }, []);
 
   const componentPreview = (
-    <LiveProvider code={code} scope={ReactLiveScope}>
-      <LivePreview/>
-    </LiveProvider>
+    <div style={{pointerEvents: 'none'}}>
```
nitzanyiz, line 56: You could add this as a class in the styles file. Wdyt?

nitzanyiz (approved):

PR #3612, opened 2025-03-12
Fix not showing dev tab when props is an empty array

ethanshar:
## Description
Fix not showing dev tab when props is an empty array

## Changelog
NA


adids1221 (approved):
Approved, but it you have any example it will be great.

PR #3614, opened 2025-03-13
Docs - Toast Live Preview

nitzanyiz:
## Description
Added live preview of incubator toast. Local assets doesn't work.
## Changelog
N/A

## Additional info
N/A


adids1221:
@nitzanyiz as we spoke, take a look at the first Toast example the icon is missing there.

nitzanyiz:
Fixed. Apparently the `general` preset is not the default.

PR #3602, opened 2025-03-11
Add new icon assets for various components and sizes

adids1221:
## Description
Assets - moved local components assets to be under `Assets.components/icons`.
There are 2 components that still using local assets: `Overaly, ScrollBar`.
Component which are part of this PR:
`ChipsInput, ColorSwatch, Picker - FieldType, StackAggregator, Hint, Stepper, Wizard`
This is part of the web effort to support assets.
Old PR #3552 closed

## Changelog
Assets - moved local components assets to be under `Assets.components/icons`.

## Additional info
None

PR #3618, opened 2025-03-16
Infra/assets internal

adids1221:
## Description
Assets new `internal` path, moved local components assets to be under `Assets.internal.icons`.
Component which are part of this PR:
`ChipsInput, ColorSwatch, Picker - FieldType, Hint,  Wizard, Overaly, ScrollBar, Stepper`
This is part of the web effort to support assets.
`TODO: add StackAggregator`.
Old PR #3602 closed

## Changelog
Assets new `internal` path, moved local components assets to be under `Assets.internal.icons`.

## Additional info
MADS-4531


Inbal-Tish:
@adids1221 If we're using Assets.internal.icons in our demo screens our users will assume they can use them too. Can we find a solution to mask this usage?

src/assets/internal/icons/index.js
```diff
@@ -2,9 +2,24 @@ export const icons = {
   get check() {
     return require('./check.png');
   },
+  get checkMarkSmall() {
+    return require('./checkMarkSmall.png');
+  },
   get checkSmall() {
     return require('./check-small.png');
```
Inbal-Tish, line 9: I think we should rename the source to checkSmall

src/assets/internal/icons/index.js
```diff
@@ -14,13 +29,19 @@ export const icons = {
   get search() {
     return require('./search.png');
   },
+  get transparentSwatch() {
```
Inbal-Tish, line 32: This one is an image, not an icon

demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
```diff
@@ -3933,7 +3933,7 @@ exports[`TextField Screen renders screen 1`] = `
                 accessible={false}
                 source={
                   {
-                    "testUri": "../../../src/assets/icons/xFlat.png",
```
Inbal-Tish, line 3936: I think we should move the snap tests to the component's dir, they shouldn't be in the demo

src/assets/internal/icons/index.js
```diff
@@ -2,9 +2,24 @@ export const icons = {
   get check() {
     return require('./check.png');
   },
+  get checkMarkSmall() {
+    return require('./checkMarkSmall.png');
+  },
   get checkSmall() {
     return require('./check-small.png');
   },
+  get dropdown() {
```
Inbal-Tish, line 11: I think we should rename it chevronDown

demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
ethanshar: But then you'll be referencing the demo from the src code 
Also, focus on the assets change, if you decide to do other changes, do it in another PR please

ethanshar (commented):
@adids1221 
I see you use Assets.internal.icons in both src code and demo example screens
IMO, we should it only in the our src code.
For the demo screens (mobile or web) we should demonstrate how the user can load their own assets and use it, we are already doing it, you can find the `demo/src/configurations.js` file where we configure the demo assets. 
Let's use these assets for our example screens instead 

@Inbal-Tish @M-i-k-e-l @nitzanyiz

demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
Inbal-Tish: Yes, we do it already in other snaps and in private as well. I think we better keep tests in src instead of the demo

src/assets/internal/icons/index.js
adids1221: Agreed, changed.

adids1221:
Fixed the demo screens Assets usage, changed some file names.
`TODO: Test it on docs locally`.

expoDemo/MainScreen.tsx
```diff
@@ -83,7 +83,7 @@ export default function MainScreen({navigation}) {
             preset={null}
             text70
             fieldStyle={styles.fieldStyle}
-            leadingAccessory={<Icon source={Assets.icons.search} marginH-s2/>}
```
Inbal-Tish, line 86: This one still points to internal...

demo/src/screens/__tests__/__snapshots__/TextFieldScreen.spec.js.snap
```diff
@@ -3933,7 +3933,7 @@ exports[`TextField Screen renders screen 1`] = `
...
+                    "testUri": "../../../src/assets/internal/icons/xFlat.png",
                   }
                 }
```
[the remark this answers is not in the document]
adids1221: This is the usage of `xFlat` in `ClearButton` - the component usage, there is no usage of `xFlat` in the screen.

Inbal-Tish (approved):

PR #3628, opened 2025-03-17
Docs - Fix development main entry

nitzanyiz:
## Description
Fixed view docs button going to `next` version of docs when in localhost. This was breaking private docs because we don't use next in it.

## Changelog
N/A

## Additional info
N/A

### DIFF
--- a/docuilib/docusaurus.config.js
+++ b/docuilib/docusaurus.config.js
@@ -18,7 +18,7 @@ const darkCodeTheme = themes.dracula;
     trailingSlash: false,
     customFields: {
       docsMainEntry: 'getting-started/setup',
-      docsDevelopmentMainEntry: 'next/getting-started/setup',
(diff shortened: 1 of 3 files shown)


docuilib/docusaurus.config.js
```diff
@@ -18,6 +18,7 @@ const darkCodeTheme = themes.dracula;
     trailingSlash: false,
     customFields: {
       docsMainEntry: 'getting-started/setup',
+      docsDevelopmentMainEntry: 'next/getting-started/setup',
```
ethanshar, line 21: since we already have `docsMainEntry`, no reason to have a duplicate 
instead you can have something like `docsDevelopmentVersionPrefix` (or something similar) and only have `next` to prefix the `docsMainEntry` field
WDYT?

ethanshar (approved):

PR #3633, opened 2025-03-18
Enhance ContentItem component with flexed layout option and update Picker API

ethanshar:
## Description
- Updated Picker components docs (from private), I removed features that are not relevant in public
- Add support for a `flexed` behavior for ContentItem (relevant for code snippets) - it allows to stretch the components (default is centered)   
- Fix some styling in ContentItem related to the previous feature
- Fix and reimplement the support for Incubator component (see two last commits in this PR)

## Changelog
N/A

### DIFF
--- a/src/components/picker/api/picker.api.json
+++ b/src/components/picker/api/picker.api.json
@@ -308,7 +308,7 @@
                   {
                     "background": "#E8ECF0",
                     "flexed": true,
-                    "snippet": "<View bg-$backgroundDefault padding-20><Picker label='Label' placeholder='Select' fieldType='settings'/></View>"
+                    "snippet": "<Picker label='Label' placeholder='Select' fieldType='settings' containerStyle={{padding: 20, backgroundColor: Colors.$backgroundDefault}}/>"
                   }
                 ]
               },
@@ -318,7 +318,7 @@
                   {
                     "background": "#E8ECF0",
                     "flexed": true,
-                    "snippet": "<View bg-$backgroundDefault padding-20><Picker label='Label' placeholder='Select' value={1} fieldType='settings' items={[{label: 'Selected value', value: 1}]}/></View>",
+                    "snippet": "<Picker label='Label' placeholder='Select' value={1} fieldType='settings' items={[{label: 'Selected value', value: 1}]} containerStyle={{padding: 20, backgroundColor: Colors.$backgroundDefault}}/>",
                     "props": {
                       "label": "Label",
                       "placeholder": "Select",


src/components/picker/api/picker.api.json
```diff
@@ -108,13 +108,403 @@
...
+            "type": "table",
+            "columns": ["State/Type/Preset", "Preview"],
+            "items": [
+              {
+                "title": "Default",
+                "content": [
+                  {
+                    "background": "#E8ECF0",
+                    "flexed": true,
+                    "snippet": "<View bg-$backgroundDefault padding-20><Picker label='Label' placeholder='Select' fieldType='settings'/></View>"
```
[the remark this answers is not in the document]
Inbal-Tish: Isn't `<Picker padding-20 label='Label' placeholder='Select' fieldType='settings'/>` better than adding the wrapper View?

docuilib/src/components/pageComponents/ContentItem.tsx
```diff
@@ -84,22 +85,16 @@ type Item = {
   value?: any;
   snippet?: string;
   height?: number;
+  flexed?: boolean;
 };
 type ContentItemProps = {
   item: Item;
   componentName: string;
   showCodeButton?: boolean;
+  isIncubator: boolean;
```
Inbal-Tish, line 94: Why not pass category instead and check isIncubator here?

src/components/picker/api/picker.api.json
```diff
@@ -108,13 +108,403 @@
...
+                    }
+                  }
+                ]
+              },
+              {
+                "title": "Value selected + icon prefix ",
+                "content": [
+                  {
+                    "flexed": true,
+                    "snippet": "<Picker preset='outline' placeholder='select' leadingAccessory={<Icon source={Assets.icons.general.favorite}/>} items={[{label: 'Selected Value', value: 1}]} value={1}/>"
```
Inbal-Tish, line 210: This snippet not rendering...

src/components/picker/api/picker.api.json
```diff
@@ -108,13 +108,403 @@
...
+            "title": "Usage Example",
+            "description": "Pickers allow selection of an option or multiple options from a list usually displayed using an actionSheet. Pickers can be used in forms, settings screens or as part of an interactive flow.",
+            "content": [
+              {
+                "value": "https://embed.figma.com/design/xFjvYNkGTmYTGYMLrmz9Ir/Guidelines-to-Docs?node-id=4415-185562&embed-host=share"
+              }
+            ]
+          },
+          {
+            "type": "list",
```
Inbal-Tish, line 145: I think we're missing a preset example since we have the default, underline and outline presets here and may consider removing preset=outline from snippets to keep default. WDYT?

src/components/picker/api/picker.api.json
```diff
@@ -108,13 +108,403 @@
...
+            "type": "table",
+            "columns": ["State/Type/Preset", "Preview"],
+            "items": [
+              {
+                "title": "Default",
+                "content": [
+                  {
+                    "flexed": true,
+                    "props": {
+                      "preset": "outline",
```
Inbal-Tish, line 198: The default can't have a preset=outline... Also, we might want to add examples for the underline and outline presets?

docuilib/src/components/pageComponents/ContentItem.module.scss
```diff
@@ -12,6 +12,16 @@
...
 .blocker {
   pointer-events: none;
+
+  &.flexed {
+    width: 100%;
+    height: 100%;
+    padding: 20px;
+  }
+
+  button {
```
Inbal-Tish, line 22: What is this style for?

src/components/picker/api/picker.api.json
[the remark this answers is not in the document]
ethanshar: I doesn't look good when it's without the view, the View represent some sort of a list item the user can wrap it with 
I think it's important to show a usage that make sense and visually more appealing 
At the end the user can understand what to do with it (if they don't want to the wrapping view)

src/components/picker/api/picker.api.json
ethanshar: Fixed

ethanshar:
I also updated the examples now to add the chevronDown trailing accessory because I see it wasn't included in the component itself

src/components/picker/api/picker.api.json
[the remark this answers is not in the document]
Inbal-Tish: Since you flexed the container, removing the wrapper View and the component rendered fine (minus the padding, that can be added from the Picker), no? It did to me...
If the wrapper view is not the one flexing the component this example seems misleading and just adding a strange bg color that is not a part of the picker component...

docuilib/src/components/pageComponents/ContentItem.module.scss
Inbal-Tish: Can you please leave a note here? I'm not sure it will be clear in the future

src/components/picker/api/picker.api.json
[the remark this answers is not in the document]
ethanshar: Im adding the View for the white background not the flex
Unlike the other Picker example that looks good, in this case, without the background it just looks like two texts elements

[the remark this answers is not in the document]
ethanshar: nvm I removed the wrapping View and used Picker's containerStyle instead

Inbal-Tish (approved):

PR #3635, opened 2025-03-18
Infra/assets internal new structure

adids1221:
## Description
Based on #3618 changes, implemented new structure for local assets on web.
Assets in `.web` files now represent in object of: `{uri, width, height}`, this object is valid source for RN Image.
In the `uri` we will set `require(...).default` as value to support `webDemo` and `docuilib`. 
`webDemo` set `eslint` in `imageLoader` to `true` (true by default [link](https://v4.webpack.js.org/loaders/url-loader/#esmodule))

⚠️ Removed `ReactLiveScope` configurations file since it's not necessary anymore, it was a fix for images with `.default` - all the assets files are with `.default`  (related [PR](https://github.com/wix/react-native-ui-lib/pull/3620)).

Snapshot version for testing: `7.39.0 - 6582`

## Changelog
Local Assets new structure, support web and mobile devices.

## Additional info
MADS-4531


src/utils/imageUtils.ts
```diff
@@ -2,6 +2,23 @@ import get from 'lodash/get';
 import Assets from '../assets';
 import type {ImageSourceType} from '../components/image';
 
+export function extractImageSource(source?: ImageSourceType) {
+  const sourceUri = get(source, 'uri');
+
+  if (sourceUri) {
+    // Case when uri is a number (Local Assets)
+    if (typeof sourceUri === 'number') {
```
ethanshar, line 10: I personally don't like this solution, it changes the behavior of a local asset. You are changing the way local assets are passed which is not a good idea (for instance, you can see that ColorSwatch is broken now because it's not using our Icon component)

docuilib/src/theme/ReactLiveScope/index.js
```diff
@@ -1,6 +1,5 @@
 import React from 'react';
 import products from '../../assets/data/products';
-require('./configurations');
```
M-i-k-e-l, line 3: See the icons here `chevronRight` and `drag` - they are probably broken now (see `SortableList`)

package.json
```diff
@@ -39,7 +39,8 @@
     "releaseDemo": "./scripts/release/releaseDemo.sh",
     "releaseDocs": "./scripts/release/releaseDocs.sh",
     "releaseEslint": "./scripts/release/releaseEslint.sh",
-    "releaseNative": "./scripts/release/releaseNative.sh"
+    "releaseNative": "./scripts/release/releaseNative.sh",
+    "update-web-assets": "node scripts/assets/extract-dimensions.js"
```
Inbal-Tish, line 43: updateWebAssets - let's follow the other scripts' naming convention

scripts/assets/extract-dimensions.js
```diff
@@ -0,0 +1,86 @@
...
+
+    try {
+      const buffer = fs.readFileSync(imagePath);
+      const dimensions = sizeOf(buffer);
+      return {
+        width: dimensions.width,
+        height: dimensions.height
+      };
+    } catch (sizeError) {
+      console.log(`Error getting dimensions for ${imagePath}:`, sizeError);
```
M-i-k-e-l, line 32: I'd put this as error

scripts/assets/extract-dimensions.js
```diff
@@ -0,0 +1,86 @@
...
+      // Default dimensions if sizeOf fails
+      return {width: 24, height: 24};
+    }
+  } catch (error) {
+    console.log(`Error getting dimensions for ${imagePath}:`, error);
```
M-i-k-e-l, line 37: This should be error

scripts/assets/extract-dimensions.js
```diff
@@ -0,0 +1,86 @@
...
+    return {width: 0, height: 0};
+  }
+}
+
+// Function to create index files with dimensions
+function createIndexFile(sourcePath, targetPath, fileType) {
+  const files = fs.readdirSync(sourcePath).filter(file => !file.includes('@') && !file.startsWith('.'));
```
M-i-k-e-l, line 44: Do you know what the `!file.startsWith('.')` is about?

scripts/assets/extract-dimensions.js
```diff
@@ -0,0 +1,86 @@
+const fs = require('fs');
```
M-i-k-e-l, line 1: Let's rename the file to `extractDimensions.js`

scripts/assets/extract-dimensions.js
adids1221: Nope, this is Devin script.
Removed, for now we don't want to include the "sized" files `@1.5x...`

M-i-k-e-l (approved):

PR #3641, opened 2025-03-23
Internal components - move display name to 'ignore'

Inbal-Tish:
## Description
Internal components - move display name to 'ignore'

## Changelog
Internal components - move display name to 'ignore'


Inbal-Tish:
@M-i-k-e-l Do I need to ignore the Pan components as well? (PanDismissibleView, PanGestureView, PanListenerView, PanningContext, PanResponderView, asPanViewConsumer)

M-i-k-e-l (commented):
Why are we ignoring these components? We're exporting them via `src` or `Incubator`.
Regarding the components you've mentioned, I think they'll all be removed in v8, so they can be ignored.

M-i-k-e-l (approved):

PR #3645, opened 2025-03-23
Docs - fix deploy

Inbal-Tish:
## Description
Docs - fix deploy - not rendring Tabs with empty tabs array

## Changelog
Docs - fix deploy


Inbal-Tish:
#skipreview

PR #3646, opened 2025-03-23
Fix missing dev page in some components

ethanshar:
## Description
Fixing missing Dev page in some components, and wrap ComponentPage with BrowserOnly

## Changelog
NA

### DIFF
--- a/docuilib/src/components/ComponentPage.tsx
+++ b/docuilib/src/components/ComponentPage.tsx
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import React from 'react';
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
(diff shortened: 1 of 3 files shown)


docuilib/src/components/ComponentPage.tsx
```diff
@@ -87,9 +93,15 @@ export default function ComponentPage({component}) {
   };
 
   return (
-    <div>
-      {buildHero()}
-      {buildTabs()}
-    </div>
+    <BrowserOnly>
```
Inbal-Tish, line 96: What's the BrowserOnly for? Does it fixes an issue?

ethanshar: I've just double checked on private, it didn't fix. 
It did fix in public, but I guess it wasn't the same.

Im removing it

Inbal-Tish (approved):

PR #3651, opened 2025-03-24
Docs - Change favicon and add logo

Inbal-Tish:
## Description
Docs - Change favicon and add logo

## Changelog
Docs - Change favicon and add logo


Inbal-Tish:
#skipreview

PR #3653, opened 2025-03-25
Fix docuilib assets by formatting our base64 is generated

ethanshar:
## Description
Fix docuilib assets by formatting how our base64 is generated using asset loader in our uilib webpack

## Changelog
N/A


adids1221 (approved):
Approved

PR #3654, opened 2025-03-26
Improve live code visuals

ethanshar:
## Description
Improve live code visuals
- Move to css styling 
- Add theme to the live code
- Fix preview container flex to behave like it does in mobile
- Increase page max width 

Unrelated, fix minor issue in TableSection to avoid error when missing cell (item) data - this will show empty cell instead of throwing a vague error 

## Changelog
N/A

### DIFF
--- a/docuilib/src/pages/livePreview.module.css
+++ b/docuilib/src/pages/livePreview.module.css
@@ -1,0 +1,6 @@
+.mobileFlexContainer {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  padding: 20px;
+}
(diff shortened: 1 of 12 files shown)


docuilib/src/pages/livePreview.module.css
```diff
@@ -0,0 +1,6 @@
+.mobileFlexContainer {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  padding: 20px;
```
M-i-k-e-l, line 5: Don't we want the "user" to decide on the `padding`? Could we have a screen that wants no `padding`?

M-i-k-e-l (approved):
I feel like there's not enough contrast between the "device" and the screen

PR #3657, opened 2025-03-27
Fix transition of navbar tabs

ethanshar:
## Description
Fix for a UX QA - Fix transition of navbar tabs

## Changelog
NA


nitzanyiz (approved):
🚀 
(We need to bump the version with this)

ethanshar:
I didn't bump on purpose because we have more changes coming

PR #3658, opened 2025-03-27
Fix component support check to handle nested component names

adids1221:
## Description
Usage component support check to handle nested component names

## Changelog
N/A

## Additional info
N/A

### DIFF
--- a/docuilib/src/utils/componentUtils.ts
+++ b/docuilib/src/utils/componentUtils.ts
@@ -3,5 +3,6 @@ import ReactLiveScope from '../theme/ReactLiveScope';
 const supportedComponentNames = Object.keys(ReactLiveScope);
 
 export const isComponentSupported = componentName => {
-  return supportedComponentNames.includes(componentName);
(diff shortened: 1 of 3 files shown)


docuilib/src/utils/componentUtils.ts
```diff
@@ -3,5 +3,6 @@ import ReactLiveScope from '../theme/ReactLiveScope';
 const supportedComponentNames = Object.keys(ReactLiveScope);
 
 export const isComponentSupported = componentName => {
-  return supportedComponentNames.includes(componentName);
```
ethanshar, line 6: I now see we have this check in several places and each time we use something different 
When I fixed the issue for Incubator components [here](https://github.com/wix/react-native-ui-lib/blob/master/docuilib/src/components/pageComponents/ContentItem.tsx#L119) I used lodash `_.get` for that

Can we unify the solution and use the util where I used `_.get` directly

ethanshar (approved):

PR #3668, opened 2025-03-31
Update button docs

ethanshar:
## Description
Update button docs, remove stuff that are not relevant to public docs 
I did keep some UX guidelines that are still possible to achieve 

## Changelog
N/A

### DIFF
--- a/src/components/button/button.api.json
+++ b/src/components/button/button.api.json
@@ -135,7 +135,7 @@
     {
       "name": "enableShadow",
       "type": "boolean",
-      "description": "Control shadow visibility (iOS-only)"
+      "description": "Control shadow visibility"
     },
     {
       "name": "avoidInnerPadding",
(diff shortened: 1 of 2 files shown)


src/components/button/button.api.json
```diff
@@ -217,15 +222,6 @@
                   }
                 ]
               },
-              {
```
M-i-k-e-l, line 220: You kept the `Full-Width Vs. Floating Button` section in the `UX Guidelines`

src/components/button/button.api.json
```diff
@@ -297,19 +284,7 @@
                 ]
               },
               {
-                "title": "Inverted",
```
M-i-k-e-l, line 300: I feel like you can leave the `Inverted`.
If you do return it then there's an `Inverted` section under `Icon Button` that was removed.
If not, there's a section below about `Inverted` that's still there.

src/components/button/button.api.json
ethanshar: I think we can keep it, FloatingButton is available in public

PR #3670, opened 2025-04-01
Refactor Button component to determine filled state by background color

adids1221:
## Description
Refactor Button component to determine filled state by background color instead of checking `outline` and `link` props.
This is fixing an issue when button has `backgroundColor` and `icon` in disabled mode, the Icon should be in the same color of the label if there is a background.

## Changelog
Refactor Button component to determine filled state by background color

## Additional info
None

### DIFF
--- a/src/components/button/index.tsx
+++ b/src/components/button/index.tsx
@@ -72,7 +72,7 @@ class Button extends PureComponent<Props, ButtonState> {
   }
 
   get isFilled() {
-    return !this.isOutline && !this.isLink;
+    return this.getBackgroundColor() !== 'transparent';
   }
 
   get isIconButton() {


src/components/button/index.tsx
```diff
@@ -72,7 +72,7 @@ class Button extends PureComponent<Props, ButtonState> {
   }
 
   get isFilled() {
-    return !this.isOutline && !this.isLink;
```
Inbal-Tish, line 75: Check if isOutline method is still used anywhere else...

Inbal-Tish: Please check if condition in line 119 (label color) should not be change to !isFilled as well

PR #3672, opened 2025-04-02
Checkbox - validate to return validity

nitzanyiz:
## Description
Fixed `CheckBox` validate function to return validity.

## Changelog
Checkbox - Fixed validate function not returning validity

## Additional info
Action item of MADS-4643


src/components/checkbox/index.tsx
```diff
@@ -309,6 +309,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     const error = required && !value;
     this.validationState = true; 
     this.setState({showError: error, isValid: !error});
+    return !error;
```
M-i-k-e-l, line 312: Personally I'd add `const isValid = !error;` above and change these two, leaving it up to you

M-i-k-e-l (approved):

src/components/checkbox/index.tsx
nitzanyiz: I changed. Can you take a look and approve again? 🙏

PR #3677, opened 2025-04-07
Docs/landing page

ethanshar:
## Description
- Inject options for how to render the landing page
- Update CodeSection to use our new LivePreview

## Changelog
NA


docuilib/src/components/MainSection.tsx
```diff
@@ -22,7 +24,6 @@ export default () => {
     const data = await response.json();
     const _starsCount = (data.stargazers_count / 1000).toFixed(1);
 
-
     localStorage.setItem(STARS_COUNT_KEY, _starsCount.toString());
     setStarsCount(_starsCount);
```
adids1221, line 28: Stars still showing `NanK`, the value should come form the config file if _starsCount doesn't have any value.

adids1221 (approved):
Approved, left a small comment to fix.

docuilib/src/components/MainSection.tsx
ethanshar: I added a fallback to handle this use case

PR #3682, opened 2025-04-09
Fix issue with missing ellipsis

ethanshar:
## Description
It seems like setting `left` alignment for the textfield input cause issue with displaying ellipsis 
changing it to undefined (basically the default behavior) fix it 
This fix the issue only iOS, on Android the issue remains

## Changelog
Fix missing ellipsis when text is overflowing in TextField

## Additional info
https://wix.slack.com/archives/C3B1EMVF0/p1744121944667609


ethanshar:
> Is it even an issue on android? With long text it shows the end of it

You are right, it's not an issue on Android

PR #3685, opened 2025-04-14
Feat/calendar revival

ethanshar:
## Description
I went over the code we worked on for the Calendar component
I revive it and improve event block visuals 
I fixed some syncing issue between Calendar and Agenda

There are still bugs of course, we will tackle everything during the upcoming milestone

## Changelog
NA


ethanshar:
@Inbal-Tish @M-i-k-e-l 
Please find some time to review this, I'd like to merge it while im here

src/incubator/calendar/types.ts
```diff
@@ -129,7 +129,7 @@ export interface CalendarProps {
 export interface AgendaProps {
   renderEvent?: (event: Event) => React.ReactElement | null;
   renderHeader?: (header: DateSectionHeader) => React.ReactElement | null;
-  itemHeight?: number;
+  itemHeight?: number | null;
```
Inbal-Tish, line 132: Why adding null and not removing the itemHeight default value if we want to let it take each item's height? Does it has to do with the scrollToIndex?

Inbal-Tish (approved):

src/incubator/calendar/index.tsx
```diff
@@ -31,23 +31,23 @@ function Calendar(props: PropsWithChildren<CalendarProps>) {
...
       }
     }
     return -1;
   },
-  [items]);
+  [monthItems]);
 
   const flashListRef = useRef();
-  const current = useSharedValue<number>(initialDate);
+  const current = useSharedValue<number>(new Date(initialDate).setHours(0, 0, 0, 0));
```
M-i-k-e-l, line 50: You're changing the `initialDate` the user provided, We can probably set 0 to `sec` \ `ms` but I'm not sure about `min` and `hours`

yarn.lock
```diff
@@ -2616,13 +2616,13 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@react-native-community/blur@npm:4.4.1":
-  version: 4.4.1
-  resolution: "@react-native-community/blur@npm:4.4.1"
+"@react-native-community/blur@npm:4.3.0":
```
M-i-k-e-l, line 2619: You should yarn again after the revert so this is removed.

src/incubator/calendar/types.ts
ethanshar: I decided to remove this prop entirely (I also found a TODO for that), we don't seem to use or need it really

src/incubator/calendar/Agenda.tsx
```diff
@@ -12,18 +13,33 @@ import CalendarContext from './CalendarContext';
...
 function Agenda(props: AgendaProps) {
-  const {renderEvent, renderHeader, itemHeight = 50, onEndReached, showLoader} = props;
+  // TODO: Consider removing itemHeight if it's not needed
+  const {renderEvent, renderHeader, itemHeight, onEndReached, showLoader} = props;
   const {data, selectedDate, setDate, updateSource} = useContext(CalendarContext);
   const flashList = useRef<FlashListType<InternalEvent>>(null);
   const closestSectionHeader = useSharedValue<DateSectionHeader | null>(null);
   const scrolledByUser = useSharedValue<boolean>(false);
+  const [stickyHeaderIndices, setStickyHeaderIndices] = useState<number[]>([]);
+  const lastDateBeforeLoadingNewEvents = useSharedValue<number>(selectedDate.value);
```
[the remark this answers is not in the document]
ethanshar: WDYM? can you elaborate?

src/incubator/calendar/Agenda.tsx
```diff
@@ -114,6 +130,10 @@ function Agenda(props: AgendaProps) {
             const _isSameMonth = isSameMonth(selected, previous);
             runOnJS(scrollToIndex)(index, _isSameMonth);
           }
+        } else {
+          // Note: We got here because we are missing future agenda events to scroll to.
+          // therefor we should expect and new events data load
+          lastDateBeforeLoadingNewEvents.value = selectedDate.value;
```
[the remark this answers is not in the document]
ethanshar: Well, isn't it? The last selected date is the last date the user was focused on before loading new events, so when we reload the data we can scroll back to that date. 
What am I missing here?

yarn.lock
ethanshar: Done

src/incubator/calendar/Agenda.tsx
```diff
@@ -12,18 +13,33 @@ import CalendarContext from './CalendarContext';
...
 
   /* const keyExtractor = useCallback((item: InternalEvent) => {
     return item.type === 'Event' ? item.id : item.header;
   }, []); */
 
+  useDidUpdate(() => {
+    const result = findClosestDateAfter(lastDateBeforeLoadingNewEvents.value);
+    if (result?.index) {
+      setTimeout(() => scrollToIndex(result?.index, false), 200);
+    }
```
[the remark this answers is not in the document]
ethanshar: I tried it, but I see now it breaks the initial scroll when the calendar mounts 
I agree it can be improved and we probably will, the sync between agenda and calendar is not perfect yet anyway

src/incubator/calendar/Header.tsx
```diff
@@ -10,7 +10,10 @@ import {HeaderProps, DayNamesFormat, UpdateSource} from './types';
 import CalendarContext from './CalendarContext';
 import WeekDaysNames from './WeekDaysNames';
 
+// Note: this fixes the updates on the header month title
+Reanimated.addWhitelistedNativeProps({text: true});
```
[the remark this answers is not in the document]
ethanshar: If I remove this line of code the issue returns, this was suggested in one of their issues. 
Unless you have a better solution we'll have to keep it for now

src/incubator/calendar/index.tsx
ethanshar: The time of the date is not important as we are working with days, we need the timestamp to react to day changes and not to hours/mins/secs..

PR #3698, opened 2025-04-27
Docs - Added figma embed links for missing pages

nitzanyiz:
## Description
Add missing pages to with figma embed

## Changelog
N/A

## Additional info
MADS-4671

### DIFF
--- a/src/components/WheelPicker/wheelPicker.api.json
+++ b/src/components/WheelPicker/wheelPicker.api.json
@@ -100,8 +100,7 @@
   ],
   "docs": {
     "hero": {
-      "title": "WheelPicker",
-      "description": ""
+      "title": "WheelPicker"
     },
     "tabs": [
       {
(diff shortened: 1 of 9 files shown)


src/components/WheelPicker/wheelPicker.api.json
```diff
@@ -47,5 +97,27 @@
     "  initialValue={'yes'$2}",
     "  onChange={() => console.log('changed')$3}",
     "/>"
-  ]
+  ],
+  "docs": {
+    "hero": {
+      "title": "WheelPicker",
+      "description": ""
```
Inbal-Tish, line 104: Can we take the description from this file as a default?

Inbal-Tish (approved):

src/components/WheelPicker/wheelPicker.api.json
nitzanyiz: You mean to remove the "description"?

Inbal-Tish: No. I mean to set the description from this file. Every api json has a "description" field that describs the component. If we don't have one from the UX we can add our in as default. WDYT?

nitzanyiz: Oh, got you.

Inbal-Tish (approved):

PR #3706, opened 2025-04-30
Timeline - Add testIDs

adids1221:
## Description
TimeLine component `testID` prop set to use for testing improvements, passed to the sub components: `Line, Point`.

## Changelog
TimeLine `testID` prop.

## Additional info
None

### DIFF
--- a/src/components/timeline/Line.tsx
+++ b/src/components/timeline/Line.tsx
@@ -29,7 +29,7 @@ const Line = React.memo((props: LinePropsInternal) => {
       return (
         <View
           style={[styles.entryPoint, {backgroundColor: color}]}
-          testID={`${testID}.${top ? 'startPoint' : 'endPoint'}`}
+          testID={`${testID}.entryPoint`}
         />
       );
     }
@@ -37,9 +37,9 @@ const Line = React.memo((props: LinePropsInternal) => {
 
   const renderLine = () => {
     if (type === LineTypes.DASHED) {
-      return <Dash vertical color={color} containerStyle={dashedLineStyle} testID={`${testID}.dashedLine`}/>;
+      return <Dash vertical color={color} containerStyle={dashedLineStyle} testID={`${testID}`}/>;
     }
-    return <View style={solidLineStyle} testID={`${testID}.solidLine`}/>;
+    return <View style={solidLineStyle} testID={`${testID}`}/>;
   };
 
   return (
--- a/src/components/timeline/Line.tsx
+++ b/src/components/timeline/Line.tsx
@@ -37,9 +37,9 @@ const Line = React.memo((props: LinePropsInternal) => {
 
   const renderLine = () => {
     if (type === LineTypes.DASHED) {
-      return <Dash vertical color={color} containerStyle={dashedLineStyle} testID={`${testID}`}/>;
+      return <Dash vertical color={color} containerStyle={dashedLineStyle} testID={testID}/>;
     }
-    return <View style={solidLineStyle} testID={`${testID}`}/>;
+    return <View style={solidLineStyle} testID={testID}/>;
   };
 
   return (
--- a/src/components/timeline/Line.tsx
+++ b/src/components/timeline/Line.tsx
@@ -10,6 +10,7 @@ const ENTRY_POINT_HEIGHT = 2;
 type LinePropsInternal = LineProps & {
   top?: boolean;
   style?: ViewStyle;
+  testID?: string;
 };
 
 const Line = React.memo((props: LinePropsInternal) => {
--- a/src/components/timeline/Line.tsx
+++ b/src/components/timeline/Line.tsx
@@ -13,7 +13,7 @@ type LinePropsInternal = LineProps & {
 };
 
 const Line = React.memo((props: LinePropsInternal) => {
-  const {type, color = 'transparent', entry, top, style, width = LINE_WIDTH} = props;
+  const {type, color = 'transparent', entry, top, style, width = LINE_WIDTH, testID} = props;
 
   const solidLineStyle = useMemo(() => {
     return [style, styles.line, {width, backgroundColor: color}];
@@ -25,15 +25,20 @@ const Line = React.memo((props: LinePropsInternal) => {
 
   const renderStartPoint = () => {
     if (entry) {
-      return <View style={[styles.entryPoint, {backgroundColor: color}]}/>;
+      return (
+        <View
+          style={[styles.entryPoint, {backgroundColor: color}]}
+          testID={`${testID}.${top ? 'startPoint' : 'endPoint'}`}
+        />
+      );
     }
   };
 
   const renderLine = () => {
     if (type === LineTypes.DASHED) {
-      return <Dash vertical color={color} containerStyle={dashedLineStyle}/>;
+      return <Dash vertical color={color} containerStyle={dashedLineStyle} testID={`${testID}.dashedLine`}/>;
     }
-    return <View style={solidLineStyle}/>;
+    return <View style={solidLineStyle} testID={`${testID}.solidLine`}/>;
   };
 
   return (
--- a/src/components/timeline/Point.tsx
+++ b/src/components/timeline/Point.tsx
@@ -59,7 +59,7 @@ const Point = (props: PointPropsInternal) => {
   };
 
   return (
-    <View center style={pointStyle} onLayout={onLayout} testID={`${testID}.container`}>
+    <View center style={pointStyle} onLayout={onLayout} testID={`${testID}`}>
       {renderPointContent()}
     </View>
   );
--- a/src/components/timeline/Point.tsx
+++ b/src/components/timeline/Point.tsx
@@ -16,6 +16,7 @@ const ICON_SIZE = 16;
 
 type PointPropsInternal = PointProps & {
   onLayout?: (event: LayoutChangeEvent) => void;
+  testID?: string;
 };
 
 const Point = (props: PointPropsInternal) => {
--- a/src/components/timeline/Point.tsx
+++ b/src/components/timeline/Point.tsx
@@ -19,7 +19,7 @@ type PointPropsInternal = PointProps & {
 };
 
 const Point = (props: PointPropsInternal) => {
-  const {icon, iconProps, removeIconBackground, label, type, color, onLayout} = props;
+  const {icon, iconProps, removeIconBackground, label, type, color, onLayout, testID} = props;
 
   const pointStyle = useMemo(() => {
     const hasOutline = type === PointTypes.OUTLINE;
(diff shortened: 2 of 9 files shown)


src/components/timeline/index.tsx
```diff
@@ -117,10 +119,15 @@ const Timeline = (props: TimelineProps) => {
   };
 
   return (
-    <View row style={containerStyle}>
-      <View style={styles.timelineContainer}>
+    <View row style={containerStyle} testID={testID}>
+      <View style={styles.timelineContainer} testID={`${testID}.timelineContainer`}>
```
Inbal-Tish, line 123: Why we need this testID as well?

adids1221: I thought that could be helpful in some cases if we will want to test the entire `timeline` container.
But since the `lines` and the `point` as their own test id's can remove it.

src/components/timeline/Line.tsx
```diff
@@ -25,15 +25,20 @@ const Line = React.memo((props: LinePropsInternal) => {
 
   const renderStartPoint = () => {
     if (entry) {
-      return <View style={[styles.entryPoint, {backgroundColor: color}]}/>;
+      return (
+        <View
+          style={[styles.entryPoint, {backgroundColor: color}]}
+          testID={`${testID}.${top ? 'startPoint' : 'endPoint'}`}
```
Inbal-Tish, line 31: Why do we need the testID here? What are we testing?

src/components/timeline/Point.tsx
```diff
@@ -45,18 +45,20 @@ const Point = (props: PointPropsInternal) => {
     const tintColor = removeIconBackground ? Colors.$iconDefault : Colors.$iconDefaultLight;
     const iconSize = removeIconBackground ? undefined : ICON_SIZE;
     if (icon) {
-      return <Icon tintColor={tintColor} {...iconProps} size={iconSize} source={icon}/>;
+      return (
+        <Icon testID={`${testID}.icon`} tintColor={tintColor} {...iconProps} size={iconSize} source={icon}/>
```
Inbal-Tish, line 49: Can we have both icon and label at the same point?

Inbal-Tish:
@adids1221 What I miss most in this PR is a test driver and maybe some tests... WDYT?

src/components/timeline/Point.tsx
adids1221: No we can't according to the guidelines it's Icon or Label not both of them together.
The point is small so Icon or Label sounds right.

src/components/timeline/Line.tsx
adids1221: This could be useful when testing the line and the end/start point colors.

src/components/timeline/types.ts
```diff
@@ -27,7 +27,8 @@ export type LineProps = {
   /** to mark as entry point */
   entry?: boolean;
   width?: number;
-}
+  testID?: string;
```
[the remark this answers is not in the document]
adids1221: No this is should be internal usage only.
Fixing it.

adids1221:
@Inbal-Tish I agree, not sure if it should be in this PR.
I'll create new one based on this branch.

src/components/timeline/Point.tsx
Inbal-Tish: So why we need two different testIDs? What are we testing with these ids?

Inbal-Tish:
> @Inbal-Tish I agree, not sure if it should be in this PR. I'll create new one based on this branch.

I think the driver should be on the same PR so we can understand the use of these testIDs.

src/components/timeline/Point.tsx
adids1221: I think it's better since those are different type of components.
Since there are only two cases I think it's ok to have testIDs, when using label you expect text, the driver is different and etc.

adids1221:
@Inbal-Tish I already started new PR, I'll finish the work there and I'll move the file here.

adids1221:
@Inbal-Tish about `getPoint` in the driver, do you think it should include `getIcon & getLabel` under Point?

src/components/timeline/timeline.driver.ts
```diff
@@ -0,0 +1,131 @@
...
+    renderTree: props.renderTree,
+    testID: `${props.testID}.point.label`
+  });
+
+  const iconDriver = ImageDriver({
+    renderTree: props.renderTree,
+    testID: `${props.testID}.point.icon`
+  });
+
+  const solidTopLineDriver = ViewDriver({
```
adids1221, line 25: @Inbal-Tish Since solid and dashed line has different testID's I created these drivers, I don't want it to become messy but we get more clarity when they are separated, WDYT should we merge them under the same View driver ?

src/components/timeline/Line.tsx
```diff
@@ -25,15 +26,20 @@ const Line = React.memo((props: LinePropsInternal) => {
...
     }
   };
 
   const renderLine = () => {
     if (type === LineTypes.DASHED) {
-      return <Dash vertical color={color} containerStyle={dashedLineStyle}/>;
+      return <Dash vertical color={color} containerStyle={dashedLineStyle} testID={`${testID}`}/>;
     }
-    return <View style={solidLineStyle}/>;
+    return <View style={solidLineStyle} testID={`${testID}`}/>;
```
Inbal-Tish, line 42: Why the string formatting around the testID? Line 40, same question

src/components/timeline/Point.tsx
```diff
@@ -45,18 +46,20 @@ const Point = (props: PointPropsInternal) => {
...
+        <Text testID={`${testID}.label`} recorderTag={'unmask'} $textDefaultLight subtextBold color={labelColor}>
           {label}
         </Text>
       );
     }
   };
 
   return (
-    <View center style={pointStyle} onLayout={onLayout}>
+    <View center style={pointStyle} onLayout={onLayout} testID={`${testID}`}>
```
Inbal-Tish, line 62: Why string formatting around testID?

src/components/timeline/Line.tsx
adids1221: Must be copy paste mistake, removed

src/components/timeline/__tests__/driver.index.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+const testID = 'test-timeline';
+const labelContent = 2;
+const defaultIcon = Assets.internal.icons.check;
+
+const getDriver = (props?: TimelineProps) => {
+  const renderTree = render(<Timeline testID={testID} {...props}>
+    <Text>Timeline</Text>
+  </Timeline>);
+  const timelineDriver = TimelineDriver({renderTree, testID});
+  return {timelineDriver};
```
Inbal-Tish, line 18: Why the driver in an object?

src/components/timeline/__tests__/driver.index.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+      expect(timelineDriver.exists()).toBeTruthy();
+    });
+
+    it('should render Point', () => {
+      const {timelineDriver} = getDriver();
+      expect(timelineDriver.getPoint().exists()).toBeTruthy();
+    });
+
+    it('should render TopLine', () => {
+      const props = {topLine: {}};
```
Inbal-Tish, line 33: Senaty means you don't send any prop add see that it doesn't exists

src/components/timeline/__tests__/driver.index.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+    it('should render BottomLine', () => {
+      const props = {bottomLine: {}};
+      const {timelineDriver} = getDriver(props);
+      const bottomLine = timelineDriver.getBottomLine();
+      expect(bottomLine.exists()).toBeTruthy();
+    });
+  });
+
+  describe('Point', () => {
+    describe('with Icon', () => {
```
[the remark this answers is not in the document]
adids1221: Moved to BottomLine test cases

src/components/timeline/__tests__/driver.index.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    it('should render TopLine', () => {
+      const props = {topLine: {}};
+      const {timelineDriver} = getDriver(props);
+      const topLine = timelineDriver.getTopLine();
+      expect(topLine.exists()).toBeTruthy();
+    });
+
+    it('should render BottomLine', () => {
+      const props = {bottomLine: {}};
```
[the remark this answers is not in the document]
adids1221: Moved to BottomLine test cases

src/components/timeline/__tests__/driver.index.spec.tsx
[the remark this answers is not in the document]
Inbal-Tish: You can add it to the 'should render Point' test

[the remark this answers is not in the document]
Inbal-Tish: Senaty test:
`
it('should render Timeline', () => {
      const timelineDriver = getDriver();
      expect(timelineDriver.exists()).toBeTruthy();
      
expect(timelineDriver.getPoint().exists()).toBeTruthy();
      expect(contentStyle.color).toBeUndefined();
      expect(contentStyle.tintColor).toBeUndefined();
     
expect(topLine.exists()).toBeFalsy();
     expect(bottomLine.exists()).toBeFalsy();
});
`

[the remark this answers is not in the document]
Inbal-Tish: You can check all in one go since this is the default that you expect when you render a clean Timeline component (no props passed)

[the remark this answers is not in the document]
adids1221: fixed, note that TopLine is rendered any way with default/min height of `0`.
[github link](https://github.com/wix/react-native-ui-lib/blob/master/src/components/timeline/index.tsx#L96-L109)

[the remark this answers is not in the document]
Inbal-Tish: Then add expect for it's style to have height=0

src/components/timeline/__tests__/driver.index.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+      const timelineDriver = getDriver();
+      const contentStyle = timelineDriver.getPoint().getContentStyle();
+      expect(contentStyle.tintColor).toBeUndefined();
+      expect(contentStyle.color).toBeUndefined();
+    });
+  });
+
+  describe('TopLine', () => {
+    it('should render TopLine', () => {
+      const props = {topLine: {}};
```
Inbal-Tish, line 70: So what's the point of this test if the topLine is rendered by default?

src/components/timeline/__tests__/driver.index.spec.tsx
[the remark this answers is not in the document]
adids1221: Iv'e add `isVisible` to the `Line` driver.
Currently Iv'e add it to the `sainty` test, we can set it in another `it` like that also: (but I think it's redundant)
```
    it('top line should have correct initial line visibility', () => {
      const timelineDriver = getDriver();
      const topLine = timelineDriver.getTopLine();
      expect(topLine.exists()).toBeTruthy();
      expect(topLine.isVisible()).toBeFalsy();
    });
```

src/components/timeline/line.driver.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const getLine = () => {
+    const exists = (): boolean => {
+      return lineDriver.exists();
+    };
+    const getStyle = () => {
+      return StyleSheet.flatten(lineDriver.getElement().props.style);
+    };
+    const isVisible = (): boolean => {
+      const height = getStyle().height;
```
Inbal-Tish, line 24: I think getStyle() can return undefined

adids1221: StyleSheet.flatten can handle `undefined`, if `undefined` was passed it will return empty object.
```
const flattenedStyle = StyleSheet.flatten(undefined);
// flattenedStyle will be: {}
```

adids1221:
```suggestion
    const isVisible = (): boolean => {
      if (exists()) {
        const height = getStyle()?.height;
        return (height ?? 0) > 0;
      }
      return false;
    };
```

Inbal-Tish (approved):

PR #3724, opened 2025-05-12
Picker - test items exists (Automation gap)

Inbal-Tish:
## Description
Picker - test items exists (Automation gap)

## Changelog
Picker - test items exists

## Additional info
ticket 4716

### DIFF
--- a/src/components/picker/__tests__/index.spec.tsx
+++ b/src/components/picker/__tests__/index.spec.tsx
@@ -37,6 +37,58 @@ describe('Picker', () => {
     jest.clearAllMocks();
   });
 
+  describe('Items', () => {
+    const testItems = [{key: 'one', value: 1, label: 'One', testID: 'item_one'}, {key: 'two', value: 2, label: 'Two', testID: 'item_two'}];
+    const testItemExists = (TestCase: () => React.JSX.Element) => {
+      const renderTree = render(<TestCase/>);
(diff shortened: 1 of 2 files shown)


src/components/picker/__tests__/index.spec.tsx
```diff
@@ -37,6 +37,58 @@ describe('Picker', () => {
...
+    it('should render picker items when passing children', () => {
+      const TestCase = () => {
+        return (
+          <Picker testID={testID}>
+            <Picker.Item {...testItems[0]}/>
+            <Picker.Item {...testItems[1]}/>
+          </Picker>
+        );
+      };
+      testItemExists(TestCase);
```
adids1221, line 59: In this test you are using 2 items as the Picker children.
But in the `testItemExists` function you are checking only if the first item exist.
The function should support array of the relevant items.

src/components/picker/__tests__/index.spec.tsx
```diff
@@ -37,6 +37,58 @@ describe('Picker', () => {
...
+    });
+
+    it('should render picker items when passing items', () => {
+      const TestCase = () => {
+        return (
+          <Picker testID={testID} items={testItems}/>
```
adids1221, line 65: Also here, you are passing all the items but check for the first one only.

adids1221 (commented):
Left a small comment

PR #3730, opened 2025-05-14
Docs - Release new docs

nitzanyiz:
## Description
Removed versioning from docs. This will remove the next version and the main version will be the only version (With the update docs structure and content).

## Changelog
⭐ Docs - Release new docs content and structure.
## Additional info
MADS-4469


adids1221 (approved):
🚀

PR #3739, opened 2025-05-25
Dialog - useHiddenLocation, added reduce motion listener for testing on Android.

adids1221:
## Description
Dialog component only showed overlay (no content) when isReduceMotionEnabled was set to true on Android, content was positioned off-screen with no animation to bring it into view.

**Fix:**
Centralized accessibility state: Added isReduceMotionEnabled to Constants.accessibility following the existing pattern.
Fixed Dialog animations: Modified Dialog's open() and close() functions to respect reduce motion preference

**Changes:**
- Constants - Added isReduceMotionEnabled property and initialization
- Dialog - Added reduce motion checks in animation functions
- JestSetup - Fixed AccessibilityInfo mocking
- Fixed multiple test files affected by AccessibilityInfo mocking issues

[React native docs](https://reactnative.dev/docs/accessibilityinfo#isreducemotionenabled)

## Changelog
Fix Dialog component only showed overlay (no content) when isReduceMotionEnabled was set to true on Android.

## Additional info
Slack thread in support channel

### DIFF
--- a/jestSetup/jest-setup.js
+++ b/jestSetup/jest-setup.js
@@ -100,7 +100,6 @@ jest.mock('react-native', () => {
   reactNative.NativeModules.KeyboardTrackingViewTempManager = {};
   reactNative.NativeModules.StatusBarManager = {getHeight: jest.fn()};
   
-  // Mock AccessibilityInfo methods properly
   reactNative.AccessibilityInfo = {
     ...reactNative.AccessibilityInfo,
     isScreenReaderEnabled: jest.fn(() => Promise.resolve(false)),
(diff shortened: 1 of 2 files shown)


adids1221:
After changing the `useHIddenLocation` the issue still reproduce.
The fix is in Dialog `open, close` functions, the `visibility` is now not Animated when `isReduceMotionEnabled` set to true.

M-i-k-e-l:
> @M-i-k-e-l After changing the `useHiddenLocation` the issue still reproduce. The fix is in Dialog `open, close` functions, the `visibility` is now not Animated when `isReduceMotionEnabled` set to true.

@adids1221 
1. WDYM still reprodices? We've tested it, didn't we?
2. The changes to include the test seem to be too big. Furthermore, if you'll add the test to master (without the `jest.spyOn(AccessibilityInfo, 'isReduceMotionEnabled').mockResolvedValue(true);`) it'll still pass, so I don't think it actually tests anything. If you really want to test it you can probably add a Detox test in the private repo.

PR #3746, opened 2025-06-03
Chip - Increased hit slop for accessibility

nitzanyiz:
## Description
Increased chip hit slop to fix 48x48 hit target.

## Changelog
Chips - Increased chip hit slop to hit 48x48 hit target for accessibility

## Additional info
N/A

### DIFF
--- a/src/components/chip/index.tsx
+++ b/src/components/chip/index.tsx
@@ -313,11 +313,13 @@ const Chip = ({
     const containerSize = getContainerSize();
     const height = ('height' in containerSize ? containerSize.height : containerSize?.minHeight) ?? 0;
     const width = containerSize?.width ?? 48;
+    const verticalHitSlop = Math.max(0, (48 - height) / 2);
+    const horizontalHitSlop = Math.max(0, (48 - width) / 2);
     return {
-      top: Math.max(0, (48 - height) / 2),
-      bottom: Math.max(0, (48 - height) / 2),
-      left: Math.max(0, (48 - width) / 2),
-      right: Math.max(0, (48 - width) / 2)
+      top: verticalHitSlop,
+      bottom: verticalHitSlop,
+      left: horizontalHitSlop,
+      right: horizontalHitSlop
     };
   }, [getContainerSize]);
   return (
--- a/src/components/chip/index.tsx
+++ b/src/components/chip/index.tsx
@@ -328,7 +328,7 @@ const Chip = ({
       onPress={onPress}
       style={[styles.container, {backgroundColor}, {borderRadius}, containerStyle, getContainerSize()]}
       testID={testID}
-      hitSlop={onPress ? hitSlop : undefined}
+      hitSlop={hitSlop}
       {...others}
     >
       {avatarProps && renderAvatar()}
--- a/src/components/chip/index.tsx
+++ b/src/components/chip/index.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, {useCallback} from 'react';
+import React, {useCallback, useMemo} from 'react';
 import {StyleSheet, StyleProp, ViewStyle, ViewProps, ImageStyle, TextStyle, ImageSourcePropType} from 'react-native';
 import Assets from '../../assets';
 import {asBaseComponent} from '../../commons/new';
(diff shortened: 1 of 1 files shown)


src/components/chip/index.tsx
```diff
@@ -303,22 +303,30 @@ const Chip = ({
...
   }, [size]);
 
   const Container = onPress ? TouchableOpacity : View;
-
+  const hitSlop = useMemo(() => {
+    const containerSize = getContainerSize();
+    const height = ('height' in containerSize ? containerSize.height : containerSize?.minHeight) ?? 0;
+    const width = containerSize?.width ?? 48;
+    return {
+      top: Math.max(0, (48 - height) / 2),
```
Inbal-Tish: You can calc each once

src/components/chip/index.tsx
```diff
@@ -303,22 +303,30 @@ const Chip = ({
...
   const getContainerSize = useCallback(() => {
-    const width = useSizeAsMinimum ? 'minWidth' : 'width';
-    const height = useSizeAsMinimum ? 'minHeight' : 'height';
-
-    return typeof size === 'object'
-      ? {[width]: _.get(size, 'width'), [height]: _.get(size, 'height')}
-      : {[width]: size, [height]: size};
+    const width = typeof size === 'object' ? _.get(size, 'width') : size;
+    const height = typeof size === 'object' ? _.get(size, 'height') : size;
+    return useSizeAsMinimum ? {minWidth: width, minHeight: height} : {width, height};
```
Inbal-Tish: I would separate the return to a different method so getContainerSize will only return the size, i.e. height and width so you can use it in the hitSlop calc and wont have to extract it from the style object.

src/components/chip/index.tsx
```diff
@@ -303,22 +303,30 @@ const Chip = ({
...
+      right: Math.max(0, (48 - width) / 2)
+    };
+  }, [getContainerSize]);
   return (
     <Container
       activeOpacity={1}
       onPress={onPress}
       style={[styles.container, {backgroundColor}, {borderRadius}, containerStyle, getContainerSize()]}
       testID={testID}
+      hitSlop={onPress ? hitSlop : undefined}
```
Inbal-Tish: Why do we care if the View gets the hitSlop as well?

src/components/chip/index.tsx
nitzanyiz: Not sure I understand. To something like this?
```
  const chipSize = useMemo(() => {
    const width = typeof size === 'object' ? _.get(size, 'width') : size;
    const height = typeof size === 'object' ? _.get(size, 'height') : size;
    return {width, height};
  }, [size]);

  const containerSizeStyle = useMemo(() => {
    const {width, height} = chipSize;
    return useSizeAsMinimum ? {minWidth: width, minHeight: height} : {width, height};
  }, [chipSize]);

  const hitSlop = useMemo(() => {
    const {width = 48, height = 48} = chipSize;
    const verticalHitSlop = Math.max(0, (48 - height) / 2);
    const horizontalHitSlop = Math.max(0, (48 - width) / 2);
    return {
      top: verticalHitSlop,
      bottom: verticalHitSlop,
      left: horizontalHitSlop,
      right: horizontalHitSlop
    };
  }, [chipSize]);

  return (
    <Container
      activeOpacity={1}
      onPress={onPress}
      style={[styles.container, {backgroundColor}, {borderRadius}, containerStyle, containerSizeStyle]}
      .
      .
      .
```

src/components/chip/index.tsx
nitzanyiz: I guess we don't

nitzanyiz: I guess I can pass it...

src/components/chip/index.tsx
Inbal-Tish: Yes

src/components/chip/index.tsx
```diff
@@ -302,23 +302,37 @@ const Chip = ({
...
 
-  const getContainerSize = useCallback(() => {
-    const width = useSizeAsMinimum ? 'minWidth' : 'width';
-    const height = useSizeAsMinimum ? 'minHeight' : 'height';
-
-    return typeof size === 'object'
-      ? {[width]: _.get(size, 'width'), [height]: _.get(size, 'height')}
-      : {[width]: size, [height]: size};
+  const chipSize = useMemo(() => {
+    const width = typeof size === 'object' ? _.get(size, 'width') : size;
```
Inbal-Tish: Let's get rid of the _.get

PR #3763, opened 2025-06-12
Carousel - page control hit slop fix a11y

adids1221:
## Description
Carousel - pageControl set hitSlop as part of a11y fix.
Moved `getAccessibleHitSlop` function into `StyleUtils` to make reuse instead of duplications.

## Changelog
Carousel - pageControl set hitSlop as part of a11y fix.

## Additional info
MADS-4757

### DIFF
--- a/src/components/pageControl/index.tsx
+++ b/src/components/pageControl/index.tsx
@@ -3,6 +3,7 @@ import React, {PureComponent} from 'react';
 import {StyleSheet, LayoutAnimation, StyleProp, ViewStyle} from 'react-native';
 import {asBaseComponent} from '../../commons/new';
 import {Colors} from '../../style';
+import {StyleUtils} from '../../utils';
 import TouchableOpacity, {TouchableOpacityProps} from '../touchableOpacity';
 import View from '../view';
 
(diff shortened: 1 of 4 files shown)


src/components/pageControl/index.tsx
```diff
@@ -201,6 +202,10 @@ class PageControl extends PureComponent<PageControlProps, State> {
           getColorStyle(index === currentPage, color, inactiveColor),
           getSizeStyle(size, index, currentPage, enlargeActive)
         ]}
+        hitSlop={{
+          top: hitSlopValue,
```
ethanshar, line 206: Did you on purpose didn't assign the left/right hit slop?

ethanshar (approved):
Asked a question.. 
it's possible they detect a small hit slop and automatically report it, so I hope this would be good enough 
Anyway approved

src/components/pageControl/index.tsx
adids1221: Yes, Alexey approved it also.
When assigning left/right they are overlapping each other, we don't want to change the UX specs of the PageControl for now.

PR #3766, opened 2025-06-25
Bug Fix : UI inputbox and textbox inputs fix for building the lib with rn 79+

sridhar-mani:
I was unable to use the library with search and text box input in the android device but it worked in the emulator.

## Description
Issue with the Text box and input box when building with react native 0.79.2 on real device(Android 14).

## Changelog
Fixed the ui module check.


sridhar-mani:
why is no one reviewing the pull request

PR #3781, opened 2025-07-14
Carousel - Flip page control in rtl.

nitzanyiz:
## Description
Fixed page control not aligned with scrolling direction in carousel.

## Changelog
Carousel - Fixed PageControl not aligned with scrolling direction in carousel.

## Additional info
MADS-4775

### DIFF
--- a/src/components/carousel/index.tsx
+++ b/src/components/carousel/index.tsx
@@ -447,7 +447,7 @@ class Carousel extends Component<CarouselProps, CarouselState> {
         <PageControl
           size={size}
           spacing={spacing}
-          containerStyle={containerStyle}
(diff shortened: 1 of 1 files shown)


src/components/carousel/index.tsx
```diff
@@ -447,7 +447,7 @@ class Carousel extends Component<CarouselProps, CarouselState> {
         <PageControl
           size={size}
           spacing={spacing}
-          containerStyle={containerStyle}
+          containerStyle={[containerStyle, Constants.isRTL && styles.flip]}
```
M-i-k-e-l: Looks like it's only on iOS

nitzanyiz: Fixed. Thanks

PR #3782, opened 2025-07-17
Pass on avatar accessibility label

ethanshar:
## Description
When the user pass a custom accessibility label (when avatar is NOT clickable) we are ignoring it because the reader just read the avatar text. 
This pass on the label to the Avatar text to give the user more control

## Changelog
Fix Avatar custom accessibility label when Avatar is not clickable


adids1221 (approved):
Approved 🚀

PR #3795, opened 2025-08-21
Incubator.Gradient - new component

M-i-k-e-l:
## Description
Incubator.Gradient - new component
Ideally this would be using something other than `react-native-linear-gradient`, however I did not manage to get all the features to work consistently.

## Changelog
Incubator.Gradient - new component

## Additional info
None

### DIFF
--- a/demo/src/screens/ExampleScreenPresenter.tsx
+++ b/demo/src/screens/ExampleScreenPresenter.tsx
@@ -16,22 +16,27 @@ import {
   View
 } from 'react-native-ui-lib';
 
-interface RadioGroupOptions {
+interface StateOptions {
+  state?: string;
+  setState?: React.Dispatch<React.SetStateAction<any /** no suitable solution for enum */>>;
+}
+
+interface RadioGroupBaseOptions {
   isRow?: boolean;
-  afterValueChanged?: () => void;
   useValueAsLabel?: boolean;
 }
+
+type RadioGroupOptions =
+  | (RadioGroupBaseOptions & {
+      afterValueChanged?: () => void;
+    })
+  | (RadioGroupBaseOptions & StateOptions);
 
 interface BooleanGroupOptions {
   spread?: boolean;
   afterValueChanged?: () => void;
   state?: boolean;
   setState?: React.Dispatch<React.SetStateAction<boolean>>;
-}
-
-interface SegmentsExtraOptions {
-  state?: string;
-  setState?: React.Dispatch<React.SetStateAction<any /** no suitable solution for enum */>>;
 }
 
 export function renderHeader(title: string, others?: TextProps) {
@@ -103,9 +108,10 @@ export function renderBooleanGroup(title: string, options: string[]) {
 export function renderRadioGroup(title: string,
   key: string,
   options: object,
-  {isRow, afterValueChanged, useValueAsLabel}: RadioGroupOptions = {}) {
   // @ts-ignore
-  const value = this.state[key];
+  {isRow, afterValueChanged, useValueAsLabel, state, setState}: RadioGroupOptions = {}) {
+  // @ts-ignore
+  const value = state ?? this.state[key];
   return (
     <View marginB-s2>
       {!_.isUndefined(title) && (
@@ -118,7 +124,18 @@ export function renderRadioGroup(title: string,
         style={isRow && styles.rowWrap}
         initialValue={value}
         // @ts-ignore
-        onValueChange={value => this.setState({[key]: value}, afterValueChanged)}
+        onValueChange={value => {
+          if (setState) {
+            setState(value);
+            if (afterValueChanged) {
+              // eslint-disable-next-line no-restricted-syntax
+              console.error('afterValueChanged is not supported together with the state option');
+            }
+          } else {
+            // @ts-ignore
+            this.setState({[key]: value}, afterValueChanged);
+          }
+        }}
       >
         {_.map(options, (value, key) => {
           return (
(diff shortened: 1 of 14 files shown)


src/incubator/gradient/types.ts
```diff
@@ -0,0 +1,35 @@
+import type {LinearGradientProps} from 'react-native-linear-gradient';
+
+type CommonGradientProps = Pick<LinearGradientProps, 'colors' | 'children'> & {
+  angle?: number;
+  center?: boolean;
+  centerH?: boolean;
+  centerV?: boolean;
+};
+
+// type GradientType = 'rectangle' | 'circle' | 'border';
```
adids1221, line 10: This can be removed

src/incubator/gradient/types.ts
```diff
@@ -0,0 +1,35 @@
+import type {LinearGradientProps} from 'react-native-linear-gradient';
+
+type CommonGradientProps = Pick<LinearGradientProps, 'colors' | 'children'> & {
```
adids1221, line 3: LinearGradientProps has `start & end` props ([link](https://github.com/react-native-linear-gradient/react-native-linear-gradient?tab=readme-ov-file#start)), I think the user should be able to control them.
Nice work with the `angle` prop but playing with `start & end` can give you more results (what I asked about if the gradient can start from another point)

src/incubator/gradient/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  angle?: number;
+  center?: boolean;
```
adids1221, line 5: The center alignment props only affect the children content, not the gradient.
These should be moved to a separate prop group like `ContentAlignmentProps` with boolean values, WDYT ?

src/incubator/gradient/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+import CircleGradient from './CircleGradient';
+import BorderGradient from './BorderGradient';
+
+export {GradientProps};
+
+const Gradient = (props: GradientProps) => {
+  const {type = 'rectangle', ...others} = props;
+
+  useEffect(() => {
+    if (LinearGradient === undefined) {
```
adids1221, line 16: This `useEffect` isn't relevant because the fallback is `{default: () => null}`, not `undefined`.
Am I missing something ?

src/incubator/gradient/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+      LogService.error(`RNUILib Gradient requires installing "react-native-linear-gradient" dependency`);
+    }
+  }, []);
+
+  if (!LinearGradient) {
```
adids1221: This if `(!LinearGradient)` check is irrelevant because `LinearGradient` cannot be falsy (it's either a component or a function).
The fallback function `() => null` will pass this check consider checking `if (LinearGradient === null)` instead.

demo/src/screens/ExampleScreenPresenter.tsx
```diff
@@ -16,24 +16,29 @@ import {
   View
 } from 'react-native-ui-lib';
 
-interface RadioGroupOptions {
+interface StateOptions {
```
adids1221, line 19: This file includes changes that belong in a different PR.
Have you verified that these changes don't break other screens ?

src/incubator/gradient/types.ts
M-i-k-e-l: Do you have an example where this is useful? More to the point, if we ever move to [react-native-gradients](https://github.com/FriendsOfReact/react-native-gradients) - it supports `angle` and not `start` & `end`

src/incubator/gradient/types.ts
M-i-k-e-l: I hoped to fix the expect-error in the screen, kinda still want to keep this

src/incubator/gradient/types.ts
M-i-k-e-l: Doesn't `center` always affects the children?

src/incubator/gradient/index.tsx
M-i-k-e-l: You're right, I'm reverting the change to `LinearGradientPackage` since it affects `SkeletonView` as well

src/incubator/gradient/index.tsx
M-i-k-e-l: Changed a little

demo/src/screens/ExampleScreenPresenter.tsx
M-i-k-e-l: The changes are required for this PR, if I made these changed in another (clean) PR, how would we test them?
Tested a few screens

src/incubator/gradient/types.ts
adids1221: it does, for me when I pass `center` or `centerV/H` to the gradient I expect the color of the gradient to be aligned also.
I leave it up to you.

src/incubator/gradient/types.ts
adids1221: example:
```
        <View width={80} height={80}>
          <LinearGradient
            colors={[Colors.$backgroundPrimaryHeavy, Colors.$backgroundPrimaryMedium, Colors.$backgroundPrimaryLight]}
            start={{x: 0.01, y: 0.02}}
            end={{x: 0.9, y: 0.8}}
          >
            <View width={80} height={80} center>
              <Text white>Hello</Text>
            </View>
          </LinearGradient>
        </View>
 ```
 
As you can see the gradient as more of the `backgroundPrimaryLight` since the start coordinates changed.

Does the [react-native-gradients](https://github.com/FriendsOfReact/react-native-gradients) migration will be soon ?

M-i-k-e-l: Technically we can get something very similar using a different `colors` array (for example: `const COLORS = [Colors.$backgroundPrimaryHeavy, Colors.$backgroundPrimaryMedium, Colors.$backgroundPrimaryMedium];`)

Either way I don't want to add a general support for `start` and `end`, if it'll be required we can add a specific `type` for it (so we can deprecate it more easily).

Not sure we'll do it, it does't seem like it's very active, but it's only a few JS files, so we can more easily handle changes if we need to.

src/incubator/gradient/types.ts
M-i-k-e-l: I've changed the screen's text but I think the modifier is ok

PR #3799, opened 2025-09-03
feat(Button): add customBackground prop for custom background elements

adids1221:
## Description
New `customBackground` prop to the Button component that allows users to render custom React elements as button backgrounds.
This prop takes precedence over the `backgroundColor`.

## Changelog
Button - Added `customBackground` prop to enable custom background elements

## Additional info
N/A

### DIFF
--- a/demo/src/screens/componentScreens/ButtonsScreen.tsx
+++ b/demo/src/screens/componentScreens/ButtonsScreen.tsx
@@ -296,12 +296,15 @@ export default class ButtonsScreen extends Component {
               label="Image Background"
               customBackground={
                 <Image
-                  source={{uri: 'https://picsum.photos/300'}}
+                  source={{
+                    uri: 'https://images.pexels.com/photos/748837/pexels-photo-748837.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=750&w=1260'
+                  }}
                   style={{width: 200, height: 50}}
                   resizeMode="cover"
                 />
               }
               style={{marginBottom: ButtonSpace}}
+              $textNeutralHeavy
             />
           </View>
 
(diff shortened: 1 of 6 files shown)


demo/src/screens/componentScreens/ButtonsScreen.tsx
```diff
@@ -267,6 +278,31 @@ export default class ButtonsScreen extends Component {
...
+                  height={50}
+                />
+              }
+              style={{marginBottom: ButtonSpace}}
+            />
+            <Button
+              label="Image Background"
+              customBackground={
+                <Image
+                  source={{uri: 'https://picsum.photos/300'}}
```
M-i-k-e-l: This seems to generate a random picture every time, it can cause issues in tests (if we have or will have them)

src/components/button/index.tsx
```diff
@@ -440,6 +460,14 @@ function createStyles() {
       backgroundColor: 'transparent',
       flexDirection: 'row',
       ...Typography.text70
+    },
+    backgroundElement: {
+      position: 'absolute',
+      top: 0,
+      left: 0,
+      right: 0,
+      bottom: 0,
```
M-i-k-e-l: If you use our `View`, you can use the `absF` modifier to avoid these (except the `overflow`).
Up to you.

src/components/button/index.tsx
```diff
@@ -372,11 +390,12 @@ class Button extends PureComponent<Props, ButtonState> {
       modifiers,
       forwardedRef,
       hitSlop: hitSlopProp,
+      customBackground,
       ...others
     } = this.props;
     const shadowStyle = this.getShadowStyle();
     const {margins, paddings} = modifiers;
-    const backgroundColor = this.getBackgroundColor();
+    const backgroundColor = customBackground ? 'transparent' : this.getBackgroundColor();
```
M-i-k-e-l:
```suggestion
    const backgroundColor = customBackground ? undefined : this.getBackgroundColor();
```

M-i-k-e-l (approved):
A few small comments, feel free to merge after they are handled

PR #3831, opened 2025-11-04
Fix warning

NikitaShavit:
## Description
Fixed warnings in the Hint component for React Native 77 compatibility. Moved default props from the deprecated `defaultProps` pattern to function parameter defaults, and suppressed false-positive exhaustive-deps warnings.

## Changelog
Hint - Fixed React Native 77 compatibility warnings by migrating from `defaultProps` to default parameter values.

## Additional info
None

### DIFF
--- a/src/components/hint/index.tsx
+++ b/src/components/hint/index.tsx
@@ -305,4 +305,5 @@ Hint.positions = HintPositions;
 
 export {HintProps, Hint};
 
+// @ts-expect-error
 export default asBaseComponent<HintProps, typeof Hint>(Hint);


src/components/hint/index.tsx
adids1221, line 211: There is a ts error here that failed the build.
`asBaseComponent` expect `ReactElement` while `children` is `ReactNode`.

```suggestion
    // Type assertion needed: children is React.ReactNode but asBaseComponent expects ReactElement | null
    return (children || null) as React.ReactElement | null;
```

adids1221 (commented):
Left a small comment

src/components/hint/index.tsx
NikitaShavit: Type assertion is not a valid solution IMO

adids1221: Let's return the `// @ts-expect-error` over the export line

PR #3833, opened 2025-11-05
Allow passing counter style to adjust margins of carousel component

alwaysintune:
## Description

Adds a new `counterContainerStyle` prop to the Carousel component to allow customization of the counter container's styling. This complements the existing `counterTextStyle` prop by enabling developers to adjust the counter container's appearance (margins, padding, positioning, background, etc.) while `counterTextStyle` continues to control the text styling inside the counter.

## Changelog

**Carousel** - Added `counterContainerStyle` prop (StyleProp<ViewStyle>) to customize the counter container's style. This allows adjusting margins, padding, positioning, and other visual properties of the counter container when `showCounter` is enabled.

## Additional info

This enhancement provides more flexibility for customizing the counter appearance, particularly useful for adjusting margins and positioning to match different design requirements or layouts.

### DIFF
--- a/src/components/carousel/types.ts
+++ b/src/components/carousel/types.ts
@@ -67,9 +67,6 @@ export interface CarouselProps extends ScrollViewProps {
    */
   showCounter?: boolean;
   /**
-   * the counter's style
-   */
-  /**
    * the counter's container style
    */
  counterContainerStyle?: StyleProp<ViewStyle>;


src/components/carousel/index.tsx
```diff
@@ -459,13 +459,13 @@ class Carousel extends Component<CarouselProps, CarouselState> {
   }
 
   renderCounter() {
-    const {pageWidth, showCounter, counterTextStyle} = this.props;
+    const {pageWidth, showCounter, counterStyle, counterTextStyle} = this.props;
```
adids1221:
```suggestion
    const {pageWidth, showCounter, counterContainerStyle, counterTextStyle} = this.props;
```

src/components/carousel/index.tsx
```diff
@@ -459,13 +459,13 @@ class Carousel extends Component<CarouselProps, CarouselState> {
...
     const {currentStandingPage} = this.state;
     const pagesCount = presenter.getChildrenLength(this.props);
 
     if (showCounter && !pageWidth) {
       return (
-        <View center style={styles.counter}>
+        <View center style={[styles.counter, counterStyle]}>
```
adids1221:
```suggestion
        <View center style={[styles.counter, counterContainerStyle]}>
```

src/components/carousel/types.ts
```diff
@@ -66,6 +66,10 @@ export interface CarouselProps extends ScrollViewProps {
    * whether to show a page counter (will not work with 'pageWidth' prop)
    */
   showCounter?: boolean;
+  /**
+   * the counter's style
+   */
+  counterStyle?: StyleProp<ViewStyle>;
```
adids1221:
```suggestion
  /**
   * the counter's container style
   */
 counterContainerStyle?: StyleProp<ViewStyle>;
```

adids1221 (commented):
Hi @alwaysintune, I left a small comment about the naming convenience of the prop name.

Since it's only for the counter container, I think `counterContainerStyle` is better naming here and creates a clear separation between the container and the text styling.

src/components/carousel/types.ts
```diff
@@ -66,6 +66,13 @@ export interface CarouselProps extends ScrollViewProps {
    * whether to show a page counter (will not work with 'pageWidth' prop)
    */
   showCounter?: boolean;
+  /**
+   * the counter's style
+   */
+  /**
+   * the counter's container style
+   */
+ counterContainerStyle?: StyleProp<ViewStyle>;
```
M-i-k-e-l, line 72: @alwaysintune @adids1221
Generally, we should avoid adding styles for each view in the hierarchy.
Let's try to find a solution that solves the issue without an additional prop.
Please open a ticket if needed.

src/components/carousel/types.ts
```diff
@@ -66,6 +66,13 @@ export interface CarouselProps extends ScrollViewProps {
    * whether to show a page counter (will not work with 'pageWidth' prop)
    */
   showCounter?: boolean;
+  /**
+   * the counter's style
+   */
+  /**
+   * the counter's container style
+   */
```
M-i-k-e-l: It looks like the a duplicate

src/components/carousel/types.ts
alwaysintune: That's the only solution to add margin to the right

src/components/carousel/types.ts
alwaysintune: fixed

src/components/carousel/types.ts
alwaysintune: To elaborate more - currently counter has an absolute position. Absolute position forces item design to conform, fill the full width, whilst carousel is written with use cases in mind, so a lot of props can't be used together (i. e. loop and item spacings).

alwaysintune: Mixing with banners leads to cases where it would be best to have a different position for counter because it conflicts with dismiss x action

PR #3835, opened 2025-11-12
Changed boarder color of ColorSwatch. to $outlineNeutralHeavy

ZivSa1:
## Description
borderColor: Colors.rgba(Colors.$outlineDisabledHeavy, 0.8)
wasn't enough so final fix is
borderColor: Colors.rgba(Colors.$outlineNeutralHeavy, 1)

## Changelog
Changed Color Swatch border from 0.2 opacity to 1 for accessibility. 

## Additional info
Ticket 4867


M-i-k-e-l (approved):
@ZivSa1 

If Alexey approved then it's ok.
Please add the assignee as well, my filters reliy on it.

I would put more details in the description rather than in the change log, the change log should be concise and readable to the public user.
I did change the `Additional info` - we don't put links in the public repo.

```
## Description
borderColor: Colors.rgba(Colors.$outlineDisabledHeavy, 0.8)
wasn't enough so final fix is
borderColor: Colors.rgba(Colors.$outlineNeutralHeavy, 1)

## Changelog
Changed Color Swatch border from 0.2 opacity to 1 for accessibility. 

## Additional info
Ticket 4867
```

PR #3844, opened 2025-11-20
fix: add web-specific styles for overlay component rendering

adids1221:
## Description
Fixes overlay rendering issues on web platform for the `Image` component.
Added web-specific styles to the `Overlay` component to ensure proper width handling (`width: '100%'`) when rendering on web.

## Changelog
Image - Fixed overlay rendering on web platform.

## Additional info
MADS-4848

### DIFF
--- a/src/components/overlay/index.tsx
+++ b/src/components/overlay/index.tsx
@@ -86,7 +86,7 @@ class Overlay extends PureComponent<OverlayTypes> {
   };
 
   renderImage = (style: any, source: ImageSourcePropType) => {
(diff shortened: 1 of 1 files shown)


src/components/overlay/index.tsx
```diff
@@ -128,6 +129,9 @@ const styles = StyleSheet.create({
     ...StyleSheet.absoluteFillObject,
     width: undefined
```
M-i-k-e-l: Can we try to move the `Constants.isWeb` here?

adids1221: Changed, working fine.

PR #3856, opened 2025-12-10
reduced scrollEventThrottle on android (Ticket 4885)

ZivSa1:
## Description
On Android, the scrollEventThrottle of 200ms was too slow to reliably detect when the scroll landed on a cloned page during fast swipes. This caused the carousel to get stuck on clone pages instead of jumping back to the real page, breaking the infinite loop behavior.
The fix reduces scrollEventThrottle to 16ms on Android to ensure the onScroll handler catches the out-of-bounds position and triggers the loop jump-back.

## Changelog
Carousel - Android fixed loop behaviour.

## Additional info
Ticket 4885

### DIFF
--- a/packages/react-native-ui-lib/src/components/carousel/index.tsx
+++ b/packages/react-native-ui-lib/src/components/carousel/index.tsx
@@ -308,9 +308,10 @@ class Carousel extends Component<CarouselProps, CarouselState> {
     // finished full page scroll
     const {currentStandingPage, currentPage} = this.state;
     const pagesCount = presenter.getChildrenLength(this.props);
-  
+
     if (currentPage < pagesCount) {
       this.setState({currentStandingPage: currentPage});
+
       if (currentStandingPage !== currentPage) {
         this.props.onChangePage?.(currentPage, currentStandingPage, {isAutoScrolled: this.isAutoScrolled});
         this.isAutoScrolled = false;


src/components/carousel/index.tsx
```diff
@@ -511,7 +510,7 @@ class Carousel extends Component<CarouselProps, CarouselState> {
           showsHorizontalScrollIndicator={false}
           showsVerticalScrollIndicator={false}
           decelerationRate="fast"
-          scrollEventThrottle={200}
+          scrollEventThrottle={Constants.isAndroid ? 16 : 200} // Android needs 16ms throttle to reliably catch loop boundary during fast swipes
```
M-i-k-e-l: Could you please check with the user if this is a regression from RN77? If yes, please start the comment with `// TODO: RN 77`

I would also add `(Ticket 4885)` at the end

Is this easy to reproduce in the public demo?

ZivSa1: so I asked (again) the user if it's' a regression and stil got not answer. regarding reproducing: yes it's easy to reproduce, at first the loop is just not working (going to the clone cells but not continuing from there) and then the fix enables this continuos scroll.

ZivSa1: @M-i-k-e-l I didn't get a response from the user. Can I merge the fix?

M-i-k-e-l (approved):

PR #3857, opened 2025-12-10
Touchables 48x48 hitslop support

yotam-wix:
## Description
Added better hitslop calculation for buttons to handle edge cases i.e. link/icon buttons.
Added better hitslop calculation for slider thumb.
all touchable component should have a 48x48 hitslop now.

## Changelog
Button and Slider Thumb Fixed hit target to be at least of 48x48.

## Additional info
MADS-4840

### DIFF
--- a/packages/react-native-ui-lib/src/components/numberInput/numberInput.api.json
+++ b/packages/react-native-ui-lib/src/components/numberInput/numberInput.api.json
@@ -60,9 +60,7 @@
       "description": "Most of TextField's props can be applied, except for ones that are passed directly via named props."
     }
   ],
-  "snippet": [
-    "<NumberInput initialValue={1506$1} onChange={onChange$2} placeholder={'Price$3'}/>"
-  ],
+  "snippet": ["<NumberInput initialValue={1506$1} onChange={onChange$2} placeholder={'Price$3'}/>"],
   "docs": {
     "hero": {
       "title": "NumberInput",
@@ -97,22 +95,14 @@
           },
           {
             "type": "table",
-            "columns": [
-              "State",
-              "Component"
-            ],
+            "columns": ["Type", "Component"],
             "items": [
               {
                 "title": "Currency",
                 "description": "",
                 "content": [
                   {
-                    "props": {
-                      "textFieldProps": {
-                        "label": "Enter Amount"
-                      },
-                      "leadingText": "$"
-                    }
+                    "value": "https://wixmp-1d257fba8470f1b562a0f5f2.wixmp.com/mads-docs-assets/assets/Components%20Docs/NumberInput/NumberInput_Overview_Type_Currency.png"
                   }
                 ]
               },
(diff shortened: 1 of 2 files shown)


packages/react-native-ui-lib/src/components/button/index.tsx
```diff
@@ -36,7 +36,7 @@ class Button extends PureComponent<Props, ButtonState> {
     super(props);
   }
 
-  state: Record<'size', undefined | number> = {
+  state: ButtonState = {
```
adids1221, line 39: You switched to using `ButtonState` type (which is great!), but `measuredSize` and `borderRadius` aren't initialized in the state object. They're set later in `onLayout`, but it's better to initialize them upfront for clarity.

packages/react-native-ui-lib/src/components/button/index.tsx
```diff
@@ -49,19 +49,26 @@ class Button extends PureComponent<Props, ButtonState> {
...
+    }
 
     if (Constants.isAndroid && (Platform.Version as number) <= 17) {
       this.setState({borderRadius: height / 2});
     }
   };
 
+  needsHitslopMeasurement() {
+    const {avoidMinWidth, avoidInnerPadding, round} = this.props;
+    return this.isLink || (this.isIconButton && !round) || avoidMinWidth || avoidInnerPadding;
```
adids1221, line 71: Why this  `(this.isIconButton && !round)` checking is related here ?

yotam-wix: for round buttons in general we have a separate measurement happening already in the code, and icon buttons have an issue where the button size depends on the icon size (there's a separate check in the code where if an icon button is non-round no padding is applied and no minWidth is set), so we have to measure its dimensions in this case.

packages/react-native-ui-lib/src/components/button/index.tsx
yotam-wix: so should I just initialize them both to be undefined here?

packages/react-native-ui-lib/src/components/slider/index.tsx
```diff
@@ -561,14 +561,24 @@ class Slider extends PureComponent<InternalSliderProps, State> {
 
   getThumbProps = () => {
     const {thumbStyle, activeThumbStyle, disableActiveStyling, disabled, thumbTintColor, thumbHitSlop} = this.props;
+    const {thumbSize} = this.state;
+
+    const calculatedHitSlop = thumbSize.width > 0
```
adids1221, line 568: Extract repeated calculations to constants to improves readability and avoids duplicate calculations.
```
const verticalHitslop = Math.max(0, (48 - thumbSize.height) / 2);
const horizontalHitslop = Math.max(0, (48 - thumbSize.width) / 2);
```

adids1221 (commented):
Nice work, left few small comments.

packages/react-native-ui-lib/src/components/button/index.tsx
adids1221: I think it's better for readability, WDYT ?

adids1221: @yotam-wix did you fixed that ?

yotam-wix:
#rebuild

yotam-wix:
#rebuild

PR #3858, opened 2025-12-11
Picker dialog dismiss bug

yotam-wix:
## Description
Fixed Picker dialog bug where on dismiss it 'remembers' the selected values.

## Changelog
Picker - Fixed dialog onDismiss bug.

## Additional info
Ticket 4680


packages/react-native-ui-lib/src/components/picker/index.tsx
```diff
@@ -293,7 +294,7 @@ const Picker = React.forwardRef((props: PickerProps, ref) => {
         <ExpandableOverlay
           ref={pickerExpandable}
           useDialog={useDialog || useWheelPicker}
-          dialogProps={DEFAULT_DIALOG_PROPS}
+          dialogProps={{...DEFAULT_DIALOG_PROPS, onDismiss: handleCancelAction}}
```
M-i-k-e-l: You're creating a new object each render

packages/react-native-ui-lib/src/components/picker/helpers/usePickerSelection.tsx
```diff
@@ -35,13 +35,17 @@ const usePickerSelection = (props: UsePickerSelectionProps) => {
...
+  const handleCancelAction = useCallback(() => {
     setSearchValue('');
     setMultiDraftValue(multiFinalValue);
-    pickerExpandableRef.current?.closeExpandable?.();
     topBarProps?.onCancel?.();
   }, [multiFinalValue, topBarProps]);
 
+  const cancelSelect = useCallback(() => {
+    handleCancelAction();
+    pickerExpandableRef.current?.closeExpandable?.();
```
M-i-k-e-l, line 50: This fix changes the order of `topBarProps?.onCancel?.();` and `pickerExpandableRef.current?.closeExpandable?.();`.
Although I could not find a bug with it, it does not mean it does not break anything.
I am not saying that we **have to** find another solution but we should be aware of this and possibly try to find a solution that does not change the order.

packages/react-native-ui-lib/src/components/picker/helpers/usePickerSelection.tsx
```diff
@@ -35,13 +35,17 @@ const usePickerSelection = (props: UsePickerSelectionProps) => {
   },
   [multiDraftValue]);
 
-  const cancelSelect = useCallback(() => {
+  const handleCancelAction = useCallback(() => {
```
M-i-k-e-l: I don't really know what is the difference between `handleCancelAction` and `cancelSelect` (name wise), try to think of a better name

packages/react-native-ui-lib/src/components/picker/index.tsx
yotam-wix: @M-i-k-e-l Would a better solution be to create a memoized value using useMemo for this object of dialogProps that re-calculates for each change in handleCancelAction?

packages/react-native-ui-lib/src/components/picker/helpers/usePickerSelection.tsx
yotam-wix: will change it to resetSelectionState if that makes sense

packages/react-native-ui-lib/src/components/picker/helpers/usePickerSelection.tsx
yotam-wix: I've taken out the first mutual 2 lines out to another function so it can be called in the correct order now. only downside I believe is that we created one more function.

M-i-k-e-l: Nice

packages/react-native-ui-lib/src/components/picker/helpers/usePickerSelection.tsx
M-i-k-e-l: Sounds good

M-i-k-e-l (approved):

PR #3860, opened 2025-12-12
Fix/android typography font weight fix

ZivSa1:
## Description

Enable full typography font weight support (100-900) on Android API 28+, matching iOS behavior. Older Android versions continue using the simplified bold/regular approach. In addition, removed the font family name for android which was blocking the font weights. 

## Changelog
Typography: Added full font weight range support (THIN-BLACK) on Android 9.0+. Removed font family name on Android, which blocked the weights. 

## Additional info

NONE


src/style/typographyPresets.ts
```diff
@@ -12,92 +12,95 @@ export const WEIGHT_TYPES: {[key: string]: TextStyle['fontWeight']} = {
   BLACK: '900' as const
 };
 
+// Check if platform supports full range of font weights
+const supportsFullFontWeights = Constants.isIOS || (Constants.isAndroid && Number(Platform.Version) >= 28);
```
lidord-wix, line 16: `Platform.Version >= 28` is equivalent to Android 9.0+?

lidord-wix (approved):
Approved with a question

ethanshar (approved):

src/style/typographyPresets.ts
ZivSa1: I rechecked: yes, it is

PR #3862, opened 2025-12-14
Add support for rendering FloatingButton secondary button standalone

adids1221:
## Description
Adds support for rendering FloatingButton component with only the secondary button. When `secondaryButton` is provided without `button` prop, the component renders the secondary button with horizontal layout styling and outline stroke.

**Logic Changes:**
- Secondary-only mode automatically uses horizontal layout
- FullWidth support works correctly in horizontal layout with only primary button
- Breaking: `button` prop is now required to render primary button.

## Changelog
FloatingButton - added support for rendering only secondary button.

## Additional info
Support channel slack thread

### DIFF
--- a/packages/react-native-ui-lib/src/components/floatingButton/index.tsx
+++ b/packages/react-native-ui-lib/src/components/floatingButton/index.tsx
@@ -224,6 +224,7 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
 
     this.firstLoad && !visible ? (this.firstLoad = true) : (this.firstLoad = false);
 
+    // NOTE: On first load, don't show if it should not be visible
     if (this.firstLoad === true && !this.initialVisibility) {
       return false;
     }


M-i-k-e-l:
@adids1221 please merge master after the monorepo change 🙏

demo/src/screens/componentScreens/FloatingButtonScreen.tsx
```diff
@@ -38,6 +40,7 @@ export default class FloatingButtonScreen extends Component<{}, State> {
         {renderBooleanOption.call(this, 'Full Width Button', 'fullWidth')}
         {renderBooleanOption.call(this, 'Show Secondary Button', 'showSecondary')}
         {renderBooleanOption.call(this, 'Button Layout Vertical', 'showVertical')}
+        {renderBooleanOption.call(this, 'Render Only Secondary', 'renderOnlySecondary')}
```
M-i-k-e-l: We should have `Show Primary Button` and `Show Secondary Button` one below the other

packages/react-native-ui-lib/src/components/floatingButton/__tests__/index.spec.tsx
```diff
@@ -31,14 +31,14 @@ describe('FloatingButton', () => {
       const buttonDriver = ButtonDriver({renderTree, testID: `${TEST_ID}.button`});
       expect(await buttonDriver.exists()).not.toBeTruthy();
       
-      renderTree.rerender(<TestCase visible/>);
+      renderTree.rerender(<TestCase visible button={button}/>);
```
M-i-k-e-l, line 34: Why did you need to add this to the test? It's a problem if the default behavior has changed.

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -226,6 +250,9 @@ const styles = StyleSheet.create({
     marginTop: Spacings.s4,
     marginBottom: Spacings.s7,
     marginLeft: 20
+  },
+  fullWidthSecondary: {
```
M-i-k-e-l: Why is this needed?

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -157,18 +170,26 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
+      ]
+      : {marginBottom: bottomMargin || Spacings.s7};
+
     return (
       <Button
-        outline={isHorizontal}
-        flex={isHorizontal}
-        link={!isHorizontal}
+        outline={shouldUseHorizontalStyle}
+        flex={this.shouldSecondaryButtonFlex}
```
M-i-k-e-l: This is not only for secondary - please change the name

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -191,21 +212,24 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
+          pointerEvents="box-none"
+          animated
+          style={[styles.container, this.getAnimatedStyle()]}
+          testID={testID}
+        >
+          {this.renderOverlay()}
+          {shouldUseHorizontalLayout && this.renderSecondaryButton()}
+          {button && this.renderButton()}
+          {this.isSecondaryVertical && this.renderSecondaryButton()}
+        </View>
```
M-i-k-e-l, line 249: This is getting less readable IMO.
WDYT about splitting it to horizontal vs vertical?

packages/react-native-ui-lib/src/components/floatingButton/__tests__/index.spec.tsx
adids1221: I made a breaking change here, and I believe it's necessary. 
Let me explain:

**Previous behavior:** When rendering FloatingButton with only visible: true (without button or secondaryButton props), it rendered an empty button wrapper with no title/label, which was redundant and not useful.

**New behavior:** Users must now explicitly provide either button or secondaryButton props to render buttons. If neither is provided, the component returns null (no rendering).

Why this change:
- Removes redundant rendering of empty button wrappers
- Makes the API explicit: provide a button prop to render a button
- Improves clarity and prevents confusion from empty buttons

This breaking change is documented in the PR description under "Breaking Changes".
We can discuss if you'd like to adjust the approach.

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
adids1221: now you can see the usage in renderSecondaryButton

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
adids1221: done.

packages/react-native-ui-lib/src/components/floatingButton/__tests__/index.spec.tsx
M-i-k-e-l: @adids1221 
Let's talk about this - not sure it qualifies as a breaking change, but if it does we cannot merge it

M-i-k-e-l: Let's remove the breaking change part from the `Changelog` (it's also generally best to have the Changelog in a single line unless it's actually two features) and add a ⚠️ to the start of the line

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -157,55 +176,84 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
+    const {secondaryButton, bottomMargin, testID, fullWidth} = this.props;
+
+    if (secondaryButton) {
+      const bgColor = secondaryButton.backgroundColor || Colors.$backgroundDefault;
+      const shouldUseHorizontalStyle = this.isHorizontalLayout;
+
+      const buttonStyle = shouldUseHorizontalStyle
+        ? [
+          styles.shadow,
+          fullWidth && this.isSecondaryOnly ? styles.fullWidthSecondaryMargin : styles.secondaryMargin,
```
M-i-k-e-l: Isn't `secondaryMargin` and `fullWidthSecondaryMargin` the same?

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
M-i-k-e-l: Not sure I do, see [this](https://github.com/wix/react-native-ui-lib/pull/3862/changes#r2645226842)

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -111,34 +111,53 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
   }
 
   get isSecondaryVertical() {
-    const {secondaryButton, buttonLayout} = this.props;
-    return secondaryButton && buttonLayout === FloatingButtonLayouts.VERTICAL;
+    const {secondaryButton} = this.props;
+    return secondaryButton && !this.isHorizontalLayout;
+  }
+
+  get shouldSecondaryButtonFlex() {
```
M-i-k-e-l: This appears only once and the name is not too good, maybe we can remove the function and create a const below?
Not sure if there are other functions we can do that with.

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -157,55 +176,84 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
-      </View>
-    );
+    if (button || secondaryButton) {
+      const hasBothButtons = !!(button && secondaryButton);
+      const shouldCenter = !fullWidth || (this.isHorizontalLayout && hasBothButtons);
+
+      return (
+        <View
+          row={this.isHorizontalLayout}
+          center={!!shouldCenter}
```
M-i-k-e-l: Not sure you need the `!!`

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -157,55 +176,84 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
+        {this.renderSecondaryButton()}
+      </>
     );
   }
 
   render() {
-    const {withoutAnimation, visible, fullWidth, testID} = this.props;
-    // NOTE: keep this.firstLoad as true as long as the visibility changed to true
+    const {withoutAnimation, visible, fullWidth, testID, button, secondaryButton} = this.props;
+
```
M-i-k-e-l, line 224: Why did you remove the two comments?

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -111,34 +111,53 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
 
-    const bottom = this.isSecondaryVertical ? Spacings.s4 : bottomMargin || Spacings.s8;
-    const left = this.isSecondaryHorizontal || fullWidth ? Spacings.s4 : undefined;
-    const right = this.isSecondaryHorizontal ? 20 : fullWidth ? Spacings.s4 : undefined;
-    const shadowStyle = !button?.outline && !button?.link && styles.shadow;
-    const marginStyle = {marginTop: 16, marginBottom: bottom, marginLeft: left, marginRight: right};
+    if (button) {
+      const bottom = this.isSecondaryVertical ? Spacings.s4 : bottomMargin || Spacings.s8;
+      const left = this.isSecondaryHorizontal || fullWidth ? Spacings.s4 : undefined;
+      const right = this.isSecondaryHorizontal ? 20 : fullWidth ? Spacings.s4 : undefined;
```
M-i-k-e-l: I know this is not your code, WDYT about making it `marginBottom` etc?

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -111,34 +111,53 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
...
+      const shadowStyle = button && !button.outline && !button.link ? styles.shadow : undefined;
+      const marginStyle = {marginTop: 16, marginBottom: bottom, marginLeft: left, marginRight: right};
```
M-i-k-e-l: We can also change `20` and `16` to `Spacings.s5` and `Spacings.s4`, not a must

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -157,55 +176,84 @@ class FloatingButton extends PureComponent<FloatingButtonProps> {
   };
 
   renderSecondaryButton() {
-    const {secondaryButton, bottomMargin, testID, buttonLayout} = this.props;
+    const {secondaryButton, bottomMargin, testID, fullWidth} = this.props;
+
+    if (secondaryButton) {
+      const bgColor = secondaryButton.backgroundColor || Colors.$backgroundDefault;
+      const shouldUseHorizontalStyle = this.isHorizontalLayout;
```
M-i-k-e-l: Why not use `this.isHorizontalLayout`?

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
adids1221: you are right, fixed.

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
adids1221: not sure, revert this change

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
adids1221: There are multiple usage here of `shouldUseHorizontalStyle` I think it's better for readability

PR #3883, opened 2025-12-22
textField - a11y label improvements

yotam-wix:
## Description
textField - added better handling of textField a11y label. Added unification of all different parts of the component (Label, helperText, charCounter), ordered by a11y guidelines given.

## Changelog
textField - a11y label fix.

## Additional info
Ticket 4797

### DIFF
--- a/packages/react-native-ui-lib/src/components/picker/helpers/usePickerSelection.tsx
+++ b/packages/react-native-ui-lib/src/components/picker/helpers/usePickerSelection.tsx
@@ -35,12 +35,21 @@ const usePickerSelection = (props: UsePickerSelectionProps) => {
   },
   [multiDraftValue]);
 
-  const cancelSelect = useCallback(() => {
+  const resetSelectionState = useCallback(() => {
     setSearchValue('');
     setMultiDraftValue(multiFinalValue);
+  }, [multiFinalValue]);
+
+  const onDismiss = useCallback(() => {
+    resetSelectionState();
+    topBarProps?.onCancel?.();
+  }, [resetSelectionState, topBarProps]);
+
+  const cancelSelect = useCallback(() => {
+    resetSelectionState();
     pickerExpandableRef.current?.closeExpandable?.();
     topBarProps?.onCancel?.();
-  }, [multiFinalValue, topBarProps]);
+  }, [resetSelectionState, topBarProps]);
 
   const availableItems: PickerMultiValue = useMemo(() => {
     return items?.filter(item => !item.disabled).map(item => item.value) || [];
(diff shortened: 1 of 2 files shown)


.yarnrc.yml
```diff
@@ -12,4 +12,4 @@ npmRegistryServer: "https://registry.npmjs.org/"
 npmMinimalAgeGate: "14d"
 npmPreapprovedPackages: ["@wix/*"]
 
-yarnPath: .yarn/releases/yarn-4.12.0.cjs
+yarnPath: .yarn/releases/yarn-4.12.0.cjs
```
lidord-wix:
```suggestion
yarnPath: .yarn/releases/yarn-4.12.0.cjs

```

packages/react-native-ui-lib/src/components/textField/index.tsx
```diff
@@ -138,9 +138,35 @@ const TextField = (props: InternalTextFieldProps) => {
...
+      parts.push('required');
+    }
+
+    parts.push('textField');
+
+    if (helperText) {
+      parts.push(helperText);
+    } else if (placeholder) {
+      parts.push(placeholder);
+    }
```
lidord-wix, line 159: Have we decided to include the placeholder? it can also be some weird example..

packages/react-native-ui-lib/src/components/textField/index.tsx
```diff
@@ -138,9 +138,35 @@ const TextField = (props: InternalTextFieldProps) => {
...
+
+    if (showCharCounter && others.maxLength) {
+      parts.push(`you can enter up to ${others.maxLength} characters`);
+    }
+
+    return parts.join(', ');
```
lidord-wix, line 165: Seems like it will work, for English :) 
What about other languages?

yotam-wix: Should I use i18n? Do I need to change something with the reading order of the label in other languages?

packages/react-native-ui-lib/src/components/textField/index.tsx
yotam-wix: That's what Alexey said in the case helper text is empty :|

yotam-wix:
> Added some comments BTW, the decision of the reading order is made by the UX?

@lidord-wix 

that was discussed with Alexey and Naama and that's what Alexey said in the end

yotam-wix:
@lidord-wix 

I've created a PR that wraps this public with i18n, you can take a look: https://github.com/wix-private/wix-react-native-ui-lib/pull/5894

I'm assuming this current PR would need to be pushed first.

packages/react-native-ui-lib/src/components/textField/index.tsx
```diff
@@ -138,9 +138,37 @@ const TextField = (props: InternalTextFieldProps) => {
...
+  }, [label, context.isMandatory, helperText, placeholder, showCharCounter, others.maxLength]);
+
+  const accessibilityLabel = props.accessibilityLabel ?? defaultAccessibilityLabel;
```
lidord-wix: you can extract the prop like all other props instead of using `props.`

PR #3884, opened 2025-12-24
ProgressBar custom borderRadius

MIchaelZablianov:
## Description
ProgressBar - add support for custom borderRadius

## Changelog
ProgressBar - add support for custom borderRadius

## Additional info
Ticket 4930


packages/react-native-ui-lib/src/components/progressBar/progressBar.api.json
```diff
@@ -29,6 +29,12 @@
       "name": "customElement",
       "type": "JSX.Element",
       "description": "Custom element to render on top of the animated progress"
+    },
+    {
+      "name": "borderRadius",
+      "type": "number",
+      "description": "Custom border radius for the progress bar",
+      "default": "BorderRadiuses.br100"
```
M-i-k-e-l, line 37: This is not accurate, the default is more complex (depends if it's full width or not), I think we should remove it

demo/src/screens/componentScreens/ProgressBarScreen.tsx
```diff
@@ -82,6 +82,37 @@ export default class ProgressBarScreen extends Component {
...
+            Custom Border Radius 0
+          </Text>
+          <ProgressBar
+            progress={progresses[3]}
+            style={styles.progressBar}
+            borderRadius={BorderRadiuses.br0}
+            progressColor={Colors.blue50}
+          />
+
+         <Text $textDefault text70 style={styles.text}>
```
M-i-k-e-l, line 96: I get a lint error here (about the indentation).
If you don't see it we need to fix it.
Let's also make sure your prettify works.

M-i-k-e-l (commented):
A note about the PR's description: ticket 4930 goes to `Additional info` (I've changed ti already)

Regarding the code change: this seems like a very bug change relative to what I would have expected from the task.
1. Is it needed or is it more readable in your opinion.
2. How did you test it did not change the behaviour?

PR #3903, opened 2026-01-01
Infra/v9 tests

M-i-k-e-l:
## Description
Fix button and update @testing-library/react-native version
Remove unsupported modifiers methods, add docs and fix tests
The old TextField was the only usage of the removed methods and it was only used by MaskedInput which is deprecated
Opened 4946 regarding `MaskedInput`

## Changelog
None

## Additional info
Ticket 4928


lidord-wix:
Approved
1. regarding the masked input - in oneApp all usages are with the new, right?
2. regarding the modifiers - same?

PR #3905, opened 2026-01-05
screenFooter component

yotam-wix:
## Description
Added ScreenFooter, a new component for sticky bottom items (3 max) with support for:
- Horizontal & vertical layouts with flexible alignment/distribution.
- Solid, fading, and transparent backgrounds.
- Keyboard handling (sticky or hoisted) and visibility animations.

Added ScreenFooterScreen demo to showcase all configurations.
Added hook 'useScrollToHide' to allow functionality of 'hideOnScroll'.

## Changelog
screenFooter - added new component. 
screenFooterScreen - added demo screen for new component.
useScrollToHide - added new hook for controlling visibility during scroll. 

## Additional info
Ticket 4330

### DIFF
--- a/lib/android/build.gradle
+++ b/lib/android/build.gradle
@@ -1,10 +1,10 @@
 apply plugin: 'com.android.library'
 
 project.ext {
-    buildToolsVersion = rootProject.ext.has("buildToolsVersion") ? rootProject.ext.buildToolsVersion : '35.0.0'
+    buildToolsVersion = rootProject.ext.has("buildToolsVersion") ? rootProject.ext.buildToolsVersion : '36.0.0'
     minSdkVersion = rootProject.ext.has("minSdkVersion") ? rootProject.ext.minSdkVersion : 24
-    compileSdkVersion = rootProject.ext.has("compileSdkVersion") ? rootProject.ext.compileSdkVersion : 35
-    targetSdkVersion = rootProject.ext.has("targetSdkVersion") ? rootProject.ext.targetSdkVersion : 34
+    compileSdkVersion = rootProject.ext.has("compileSdkVersion") ? rootProject.ext.compileSdkVersion : 36
+    targetSdkVersion = rootProject.ext.has("targetSdkVersion") ? rootProject.ext.targetSdkVersion : 36
     supportLibVersion = rootProject.ext.has("supportLibVersion") ? rootProject.ext.supportLibVersion  : '28.0.0'
 }
 
--- a/packages/react-native-ui-lib/android/build.gradle
+++ b/packages/react-native-ui-lib/android/build.gradle
@@ -2,10 +2,10 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "35.0.0"
+        buildToolsVersion = "36.0.0"
         minSdkVersion = 26
-        compileSdkVersion = 35
-        targetSdkVersion = 34
+        compileSdkVersion = 36
+        targetSdkVersion = 36
         ndkVersion = "27.1.12297006"
         kotlinVersion = "2.0.21"
     }


packages/react-native-ui-lib/src/assets/internal/images/bottomGradient.png
lidord-wix, line 1: I think we need the png in 5 sizes to support devices with different densities

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+    SH10 = 'sh10',
+    SH20 = 'sh20',
+    SH30 = 'sh30'
+}
+
+export interface ScreenFooterProps extends PropsWithChildren<{}> {
+    /**
+     * The background style of the footer
+     */
+    backgroundType?: ScreenFooterBackgrounds | `${ScreenFooterBackgrounds}`;
```
lidord-wix: Nice! the type is well defined 😄 
Did you understand why we're doing it?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
+import React, {useState, useMemo} from 'react';
+import {StyleSheet, ScrollView, Image} from 'react-native';
```
lidord-wix: Please use UILib's Image

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+  ScreenFooter,
+  ScreenFooterLayouts,
+  ScreenFooterBackgrounds,
+  KeyboardBehavior,
+  FooterAlignment,
+  HorizontalItemsDistribution,
+  ItemsFit,
+  Switch,
+  TextField,
+  Slider,
```
lidord-wix: The Slider doesn't work so good, can you try the slider from Incubator?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+        {/* Item Width Slider (only when Fixed is selected) */}
+        {itemsFit === ItemsFit.FIXED && (
+          <View marginB-s4>
+            <Text text70M $textDefault marginB-s2>
+              Item Width: {itemWidth}px
+            </Text>
+            <Slider
+              value={itemWidth}
+              minimumValue={50}
+              maximumValue={500}
```
lidord-wix: the range is too big IMO, let's decrease the max value

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+  const contentContainerStyle = useMemo(() => {
+    const style: any[] = [
+      styles.contentContainer,
+      layout === ScreenFooterLayouts.HORIZONTAL ? styles.horizontalContainer : styles.verticalContainer,
+      {alignItems, justifyContent}
+    ];
+
+    if (useSafeArea && Constants.isIphoneX) {
+      style.push({paddingBottom: safeAreaBottom});
+    }
```
lidord-wix: can we use the useSafeArea prop of view instead of creating the style by ourselves?

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+  const {bottom: safeAreaBottom} = Constants.getSafeAreaInsets();
+
+  const contentContainerStyle = useMemo(() => {
+    const style: any[] = [
+      styles.contentContainer,
+      layout === ScreenFooterLayouts.HORIZONTAL ? styles.horizontalContainer : styles.verticalContainer,
+      {alignItems, justifyContent}
+    ];
+
+    if (useSafeArea && Constants.isIphoneX) {
```
lidord-wix: why do we check for iPhoneX?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+
+  const getButtonSize = (size: ItemSize) => {
+    switch (size) {
+      case ItemSize.SMALL: return Button.sizes.small;
+      case ItemSize.MEDIUM: return Button.sizes.medium;
+      case ItemSize.LARGE: return Button.sizes.large;
+    }
+  };
+
+  const getTextPreset = (size: ItemSize): {main: string; sub: string} => {
```
lidord-wix: let's use the whole word instead of abbreviations (subtext instead of sub)

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+    const textPreset = getTextPreset(extraContentSize);
+    const imageSize = getImageSize(extraContentSize);
+    
+    // Extra Text (Total price)
+    if (showExtraText) {
+      items.push(
+        <View key="extra-text" centerV marginR-s4={isHorizontal} marginB-s4={!isHorizontal} style={{flexShrink: 1}}>
+          <Text {...{[textPreset.main]: true}} $textDefault numberOfLines={1}>
+            Total: <Text {...{[textPreset.main]: true}} $textDefault style={{fontWeight: 'bold'}}>257$</Text>
+          </Text>
```
lidord-wix: $textDefault is the default, no need to pass it :)

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+
+  const renderFooterItems = useMemo(() => {
+    const items = [];
+    const textPreset = getTextPreset(extraContentSize);
+    const imageSize = getImageSize(extraContentSize);
+    
+    // Extra Text (Total price)
+    if (showExtraText) {
+      items.push(
+        <View key="extra-text" centerV marginR-s4={isHorizontal} marginB-s4={!isHorizontal} style={{flexShrink: 1}}>
```
lidord-wix: You can pass the `flexS` modifier instead of the style

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+        </View>
+      );
+    }
+
+    if (itemsCount >= 1) {
+      items.push(
+        <Button
+          key="btn1"
+          size={getButtonSize(buttonSize)}
+        //   style={itemsFit === ItemsFit.STRETCH && !isHorizontal ? {alignSelf: 'stretch'} : undefined}
```
lidord-wix: Can we delete the commented out code?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+          {...getButtonProps(button3Type)}
+        />
+      );
+    }
+    
+    return items;
+  }, [itemsCount, itemsFit, isHorizontal, button1Type, button2Type, button3Type, buttonSize, showExtraText, showImage, extraContentSize, useLongButtonText]);
+
+  return (
+    <View flex bg-$backgroundDefault>
```
lidord-wix: Do we need this View?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+            <Text text70M $textDefault>
+              Show Image
+            </Text>
+            <Switch value={showImage} onValueChange={setShowImage}/>
+          </View>
+        )}
+
+        {/* Extra Content Size */}
+        {showExtraText && (
+          <View marginB-s4>
```
lidord-wix, line 279: In general, please avoid adding Views only for margins

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+            <Text text70M $textDefault marginB-s2>
+              Distribution
+            </Text>
+            <View row>
+              <SegmentedControl
+                segments={DISTRIBUTION_OPTIONS}
+                initialIndex={DISTRIBUTION_OPTIONS.findIndex(opt => opt.value === distribution)}
+                onChangeIndex={index => setDistribution(DISTRIBUTION_OPTIONS[index].value)}
+              />
+               <View flex />
```
lidord-wix: Why you added this View?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+          />
+        </View>
+
+        {/* Distribution (for Horizontal layout) */}
+        {isHorizontal && (
+          <View marginB-s4>
+            <Text text70M $textDefault marginB-s2>
+              Distribution
+            </Text>
+            <View row>
```
lidord-wix: Looks like something we don't really need :)

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+          </Text>
+          <Text text70 $textDefault marginT-s4 style={styles.placeholder}>
+            Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu 
+            fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in 
+            culpa qui officia deserunt mollit anim id est laborum.
+          </Text>
+        </View>
+      </ScrollView>
+
+      {/* ScreenFooter Component */}
```
lidord-wix: I think this comment is not so necessary

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+};
+
+const styles = StyleSheet.create({
+  scrollContent: {
+    padding: 20,
+    paddingBottom: 150
+  },
+  placeholder: {
+    lineHeight: 22
+  }
```
lidord-wix: Why changing the line height?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+      >
+        <Text text60 $textDefault marginB-s4>
+          ScreenFooter Configuration
+        </Text>
+
+        <TextField
+          placeholder="Focus me to test keyboard behavior"
+          label="Test Input"
+          floatingPlaceholder
+          containerStyle={{marginBottom: 20}}
```
lidord-wix: Please use modifier for the margins

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
+import React, {PropsWithChildren, useCallback, useEffect, useMemo, useState} from 'react';
+import {DimensionValue, Image, LayoutChangeEvent, StyleSheet} from 'react-native';
```
lidord-wix: Please use Image from ui-lib

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+  const [height, setHeight] = useState(0);
+  const visibilityTranslateY = useSharedValue(0);
+
+  // Update visibility translation when visible or height changes
+  useEffect(() => {
+    visibilityTranslateY.value = withTiming(visible ? 0 : height, {duration: animationDuration});
+  }, [visible, height, animationDuration]);
+
+  // Combine keyboard hoisting + visibility into a single animated style
+  const animatedStyle = useAnimatedStyle(() => {
```
lidord-wix: when hoisted, the footer position ins't animated (on iOS simulator) 

https://github.com/user-attachments/assets/948c3260-72fb-440e-87e5-6c56bdf83f05

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+  const renderBackground = useCallback(() => {
+    if (isSolid) {
+      const shadowStyle = Shadows[shadow]?.top;
+      const dividerStyle = hideDivider ? undefined : {borderTopWidth: 1, borderColor: Colors.$outlineDefault};
+      return (
+        <View
+          absF
+          bg-$backgroundElevated
+          style={[styles.background, shadowStyle, dividerStyle]}
```
lidord-wix: Please take this style to `useMemo`

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    return null;
+  }, [isSolid, isFading, shadow, hideDivider]);
+
+  const renderChild = useCallback((child: React.ReactNode, index: number) => {
+    if (itemsFit === ItemsFit.FIXED && itemWidth) {
+      const fixedStyle = isHorizontal
+        ? {width: itemWidth, flexShrink: 1, overflow: 'hidden' as const}
+        : {width: itemWidth, maxWidth: '100%' as const};
```
lidord-wix: Please avoid using "as ..." for resolving typing issues, it kinda hack

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+    gap: Spacings.s5
+  },
+  verticalContainer: {
+    flexDirection: 'column',
+    gap: Spacings.s3
+  },
+  background: {
+    width: '100%',
+    height: '100%'
+  }
```
lidord-wix, line 269: is it needed for non-Image components?

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -0,0 +1,298 @@
...
+  } = props;
+
+  const keyboard = useAnimatedKeyboard();
+  const [height, setHeight] = useState(0);
+  const visibilityTranslateY = useSharedValue(0);
+
+  // Update visibility translation when visible or height changes
+  useEffect(() => {
+    visibilityTranslateY.value = withTiming(visible ? 0 : height, {duration: animationDuration});
+  }, [visible, height, animationDuration]);
```
lidord-wix: Please add visibilityTranslateY to the deps array

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,534 @@
...
+  {label: 'Center', value: FooterAlignment.CENTER},
+  {label: 'End', value: FooterAlignment.END}
+];
+
+const DISTRIBUTION_OPTIONS = [
+  {label: 'Stack', value: HorizontalItemsDistribution.STACK},
+  {label: 'Spread', value: HorizontalItemsDistribution.SPREAD}
+];
+
+const DISTRIBUTION_OPTIONS_SPACED = [
```
lidord-wix: This one is not in use

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix, line 1: Please consider taking all the types to a different file

lidord-wix (commented):
Looks good!
I added some comments, it may look a lot but it's because the PR is a bit big.. 
Some more general comments:
1. Some of the comments I wrote general and not specific only to the place I added them, so please re-review the whole PR when fixing.
2. Please run `yarn tsc` to verify the typescript is ok
3. Please add api.json file for the docs.
4. In general theres no need to pass default values, you get them by default :)
5. Please add render tests, you can see references in other components.
6. When adding text + image + button, on a stretch case in the example screen, the result is not so good, see the image:Image

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: I've seen it done in other places in ui-lib, I guess it's for letting devs write the literal string or to use an enum which supports changes in the enums value?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: it seems much more laggy tbh, do we still prefer it?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: 250 is ok or too high/low?

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: I've tried but it made the footer's layout look much worse, plus cursor suggests that on 'landscape mode' it adds unnecessary horizontal padding. I could try to make it work though if that's the best practice.

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: because we're looking to push the footer up on newer iphones, where there's a need for it because of the home indicator area.

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: should I replace it with an empty fragment?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: what is the alternative, applying the margin as a style to the segmentedControl?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: redundant, removed

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: the row direction prevents the segmentedControl from stretching to full screen width, which looks less nice imo

let me know if there's a better way of doing this

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: removed

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: not sure why, in my simulator it works well

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: what's the alternative for it? couldn't find one but seemed to be used throughout the repo.
perhaps using 'ViewStyle'?

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: no, I added the absF later to the views, so I now removed this styling from the 2 views that were using it.

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: done, let me know if what i've done was ok.

packages/react-native-ui-lib/src/assets/internal/images/bottomGradient.png
yotam-wix: where can I get the other sizes from?

packages/react-native-ui-lib/src/assets/internal/images/buttonFloatingOverlay.png
lidord-wix, line 1: Let's rename, maybe bottomGradient or something?

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix: exactly, it also supports auto-complete when using it, which is really nice as a user

packages/react-native-ui-lib/src/assets/internal/images/bottomGradient.png
lidord-wix, line 1: I see we still have this one

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix: I thinks it's better to retry
in general we don't support landscape in Wix mobile apps (yet), but since it's also for the community, let's take it into consideration :)

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix: but only iPhoneX needs this padding?

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,505 @@
...
+          </Text>
+        </View>
+      );
+    }
+
+    // Image (Basket icon)
+    if (showExtraText && showImage) {
+      items.push(
+        <View key="extra-image" centerV marginR-s4={isHorizontal} marginB-s4={!isHorizontal}>
+          <Image source={basketIcon} style={{width: imageSize, height: imageSize, tintColor: Colors.$iconDefault}} />
```
lidord-wix: You can use the Icon component

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
lidord-wix: Yes, as discussed, you can move this marginB to the segmentedControl or even make it a marginT of the component below it :)

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
lidord-wix: Hmm I see, so you can use the containerStyle to pass the same style: `containerStyle={{flexDirection: 'row'}}` (with styleSheet) 
But It's not a must since it's just the example screen

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -0,0 +1,505 @@
...
+      case ItemSize.SMALL:
+        return 24;
+      case ItemSize.MEDIUM:
+        return 32;
+      case ItemSize.LARGE:
+        return 40;
+    }
+  };
+
+  const renderFooterItems = useMemo(() => {
```
lidord-wix: No need to memo if it's not passed as a prop value :)

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix: Yes, try to use ViewStyle
most of the time it just shows that we don't use the types properly

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
lidord-wix: Works better for me with Slider from incubator 🙈

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
lidord-wix: looks good to me

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix: works much better with the keyboard accessory view

lidord-wix (commented):
Looks really good overall, good job 🙂 
I replied and added some small comments, please also see bullet 6 in my last review comment (still looks the same)

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: I've added it as a marginT to the component below it since it looks cleaner with a modifier

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
yotam-wix: moved to using containerStyle, removed the newly redundant view

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: I've replaced the 'as' with ViewStyle in the lines you've shown here. let me know if it could be used in other places in this file where the word 'as' came up in the return statements.

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: image

here's how it looks like with the useSafeArea applied straight to the view, it really messes up the looks

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix: Yes, please clean all the ".. as .."

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
lidord-wix: let's recheck that, I don't see any issue when running locally, maybe we need to pass the `useSafeArea` somewhere else.. 
also, we plan to deprecate the `Constants.getSafeAreaInsets()` soon

packages/react-native-ui-lib/src/assets/internal/images/buttonFloatingOverlay.png
lidord-wix: couldn't see any changes so I guess you didn't push last changes

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
lidord-wix: same, I guess you need to push some changes

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
lidord-wix: same, wasn't changes

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -187,7 +181,9 @@ const ScreenFooter = (props: ScreenFooterProps) => {
     }
 
     if (isHorizontal && React.isValidElement(child)) {
-      const flexStyle = itemsFit === ItemsFit.STRETCH ? {flex: 1, alignItems: 'center'} : {flexShrink: 1};
+      const flexStyle = itemsFit === ItemsFit.STRETCH 
+        ? {flex: 1, minWidth: 0, alignItems: 'center'} 
+        : {flexShrink: 1};
```
yotam-wix: adding a minWidth prop here to the stretched layout improved the correct division of the container between the items, seems like the button having a certain minWidth caused it to catch more unwanted space

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -258,6 +255,11 @@ const styles = StyleSheet.create({
     flexDirection: 'column',
     gap: Spacings.s3
   },
+  stretchItemWrapper: {
+    flexDirection: 'row',
+    flex: 1,
+    justifyContent: 'center'
+  },
```
lidord-wix: you can use modifiers instead this style, also I think you don't need the "row"

yotam-wix: it's actually necessary - the stretch happens due to the default `alignItems: 'stretch'`, and to not ruin this stretch I needed to kind of cheat it by combining row+justifyContent: center, which caused effect of `alignItems: center` without overriding the `alignItems: stretch`.

let me know if I'm missing something though.

lidord-wix: Oh, ok 👍🏽

lidord-wix: but still I think you can move to modifiers :)

yotam-wix:
@lidord-wix 
I think I managed to fix the issue we had by using `react-native-safe-area-context` as suggested.
I also tried to sort out the issue of the padding remaining when hoisting is activated and a keyboard is open by using `UseKeyboardHeight()`. let me know if it looks ok.

also, my build fails for some reason due to it not finding the `react-native-safe-area-context` module?

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -1,12 +1,14 @@
 import React, {useCallback, useEffect, useMemo, useState} from 'react';
 import {LayoutChangeEvent, StyleSheet, ViewStyle} from 'react-native';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
```
lidord-wix: please see the usage of useSafeAreaInsets in the tabController component, when using a native dependency in one of our components, we're doing it optionally for the users, so if they don't use this component they don't have to install it

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -119,7 +124,10 @@ const ScreenFooter = (props: ScreenFooterProps) => {
       {alignItems, justifyContent}
     ];
 
-
+    if (useSafeArea && !isKeyboardVisible) {
```
lidord-wix: I think we don't need the `useSafeArea` prop anymore, the native control the heights

PR #3536, opened 2025-02-26
implement dismiss state management for closeExpandable control

adids1221:
## Description
Picker when passing `onDismiss` from `customPickerProps` the `onDismiss` getting called twice.
Snippet:
```
 <Picker
        placeholder="Favorite Language"
        value={language}
        onChange={item => setLanguage(item)}
        items={options}
        customPickerProps={{
          modalProps: {
            onDismiss: () => console.log('dismissed')
          }
        }}
      />
```
This issue occurs both on Modal and Dialog.

Implemented a state machine pattern to track the component's dismiss state.

## Changelog
ExpandableOverlay - Implemented state machine pattern to prevent duplicate onDismiss calls

## Additional Info
MADS-4656

PR #3916, opened 2026-01-20
fix: prevent ExpandableOverlay onDismiss callback from being called twice

adids1221:
## Description

Fixes an issue where the `onDismiss` callback passed via `modalProps` or `dialogProps` was being called twice when the overlay was dismissed. 

The issue occurred because `closeExpandable` was being triggered from multiple sources:
1. From `onBackgroundPress` / `onRequestClose` when user dismissed the modal
2. From React Native Modal's internal `onDismiss` callback

Added a new `dismissOverlay` function that only closes the UI without calling the user's callback. Modal's `onDismiss` now uses `dismissOverlay`, while `onBackgroundPress` and `onRequestClose` continue to use `closeExpandable`.

**Verified that components using ExpandableOverlay are not affected:**
- Picker - all tests pass
- DateTimePicker - all tests pass

## Changelog

**ExpandableOverlay** - Fixed `onDismiss` callback being called twice when modal is dismissed.

## Additional info
MADS-4656
[Old PR](https://github.com/wix/react-native-ui-lib/pull/3536) (Closed because of conflicts, all the comments there are fixed here)

### DIFF
--- a/docs/getting-started/setup.md
+++ b/docs/getting-started/setup.md
@@ -42,6 +42,7 @@ UILib supports additional features through optional dependencies. These packages
 | @react-native-community/blur | Card component's blur effects |
 | @react-native-community/datetimepicker | Date/time picker components |
 | @react-native-community/netinfo | Network connectivity features |
+| react-native-safe-area-context | support useSafeAreaInsets inside and usages of Modal |
 | react-native-haptic-feedback | Haptic feedback functionality |
 | react-native-svg | SVG-based components |
 | react-native-shimmer-placeholder | Shimmer loading effects |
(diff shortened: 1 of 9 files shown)


adids1221:
Hi @M-i-k-e-l,
Verified in private repo using snapshot `8.2.2-snapshot.7711`
Tested components that use ExpandableOverlay:
- Picker - onDismiss called once ✅
- ActionSheet - onDismiss called once ✅
- TimePicker - value reset on dismiss ✅
- DatePicker - value reset on dismiss ✅
- DurationPicker - value reset on dismiss ✅

Tested scenarios:
- Background tap dismiss
- Selection dismiss

No regressions found.

PR #3926, opened 2026-02-02
react-native-safe-area-context package

adids1221:
## Description
Adds `react-native-safe-area-context` as an optional dependency and integrates `SafeAreaProvider` into the Modal component.
When the package is installed, Modal and Dialog automatically wrap their content with `SafeAreaProvider`, allowing `useSafeAreaInsets()` to work inside modals without additional setup.

## Changelog
Modal - Added SafeAreaProvider support for safe area insets inside modals.

## Additional info
N/A


M-i-k-e-l (approved):

Relax594:
@M-i-k-e-l how to disable this? it leaves a gap on the DateTimePicker iOS default style for no reason.

PR #3931, opened 2026-02-08
Infra/monorepo release script

M-i-k-e-l:
## Description
Since we now have a `workspace:*` in our code base, npm fails even when it's trying to publish a package which does not have it (trying to publish `uilib-native` fails even though it does not have `workspace:*` in it).
Solution: replace `workspace:*` with a temporary version and once `uilib-native` is released insert that version into `react-native-ui-lib`'s `package.json`.
Made this process extendable so we can later add other packages (`unicorn-demo-app` etc).

## Changelog
Make release script extendable

## Additional info
None

### DIFF
--- a/packages/uilib-native/scripts/.npmrc
+++ b/packages/uilib-native/scripts/.npmrc
@@ -1,2 +1,0 @@
-email=${NPM_EMAIL}
-//registry.npmjs.org/:_authToken=${NPM_TOKEN}


packages/react-native-ui-lib/scripts/release/release.js
```diff
@@ -90,7 +90,7 @@ function tryPublishAndTag(version) {
 function tagAndPublish(newVersion) {
   console.log(`trying to publish ${newVersion}...`);
   exec.execSync(`npm --no-git-tag-version version ${newVersion}`);
-  exec.execSync(`npm publish --tag ${VERSION_TAG} --workspace react-native-ui-lib`);
+  exec.execSync(`npm publish --tag ${VERSION_TAG}`);
```
adids1221: Shouldn't this revert be in the first PR ?
The commit of this change is there.
[commit link](https://github.com/wix/react-native-ui-lib/pull/3923/changes/b4c9a7f6fdad3b8182978de29c8610ba4e706237)

M-i-k-e-l: It was, forgot to push it - fixed

PR #3935, opened 2026-02-12
fix/added an update to targetLayout for each change in targ…

yotam-wix:
## Description
Hint - fix targetFrame prop not updating after mount
When targetFrame is passed as a prop, useHintLayout stored it only as the useState initial value, so any changes after mount were ignored. Added a useEffect to sync internal state when targetFrame changes.

## Changelog
Hint - targetFrame prop now correctly repositions when updated after mount

## Additional info
None

### DIFF
--- a/packages/react-native-ui-lib/src/components/hint/hooks/useHintLayout.ts
+++ b/packages/react-native-ui-lib/src/components/hint/hooks/useHintLayout.ts
@@ -1,4 +1,4 @@
-import {type ElementRef, useState, useCallback, useRef} from 'react';
+import {type ElementRef, useState, useCallback, useRef, useEffect} from 'react';
 import type {LayoutChangeEvent, LayoutRectangle, View as RNView} from 'react-native';
 import _ from 'lodash';
 import {HintProps} from '../types';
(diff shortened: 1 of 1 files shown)


packages/react-native-ui-lib/src/components/hint/hooks/useHintLayout.ts
```diff
@@ -11,6 +11,14 @@ export default function useHintLayout({onBackgroundPress, targetFrame}: UseHintL
   const [hintMessageWidth, setHintMessageWidth] = useState<number | undefined>();
   const targetRef = useRef<ElementRef<typeof RNView> | null>(null);
 
+  useEffect(() => {
+    if (targetFrame && !_.isEqual(targetFrame, targetLayoutState)) {
+      setTargetLayout(targetFrame);
+      setTargetLayoutInWindow(targetFrame);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
yotam-wix, line 19: not sure regarding this line, cursor was insisting it's necessary due to lint issues that may come up without it but I didn't see an issue when removing this line, so leaving it up to your opinion..

PR #3937, opened 2026-02-19
Use SafeAreaContext optional dependency for View useSafeArea

adids1221:
## Description
useSafeArea now renders SafeAreaView from react-native-safe-area-context when available, falling back to the previous behavior when the package is not installed. No breaking change.

## Changelog
View – useSafeArea now supports SafeAreaView from react-native-safe-area-context when available.

## Additional info
N/A

### DIFF
--- a/packages/react-native-ui-lib/src/components/view/index.tsx
+++ b/packages/react-native-ui-lib/src/components/view/index.tsx
@@ -1,14 +1,6 @@
 import {useModifiers, useThemeProps} from 'hooks';
 import React, {useEffect, useMemo, useState} from 'react';
-import {
-  View as RNView,
-  SafeAreaView as RNSafeAreaView,
-  Animated,
-  ViewProps as RNViewProps,
-  type StyleProp,
-  type ViewStyle,
-  type DimensionValue
-} from 'react-native';
+import {View as RNView, SafeAreaView, Animated, ViewProps as RNViewProps, type StyleProp, type ViewStyle, type DimensionValue} from 'react-native';
 import type {AnimateProps as RNReanimatedProps} from 'react-native-reanimated';
 import {Constants, ContainerModifiers} from '../../commons/new';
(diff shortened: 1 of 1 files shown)


packages/react-native-ui-lib/src/components/view/index.tsx
```diff
@@ -1,9 +1,18 @@
 import {useModifiers, useThemeProps} from 'hooks';
 import React, {useEffect, useMemo, useState} from 'react';
-import {View as RNView, SafeAreaView, Animated, ViewProps as RNViewProps, type StyleProp, type ViewStyle, type DimensionValue} from 'react-native';
+import {
+  View as RNView,
+  SafeAreaView as RNSafeAreaView,
```
M-i-k-e-l: Is this change needed?

packages/react-native-ui-lib/src/components/view/index.tsx
```diff
@@ -106,7 +115,12 @@ function View(props: ViewProps, ref: any) {
   }, []);
 
   const ViewContainer = useMemo(() => {
-    const container = useSafeArea && Constants.isIOS ? SafeAreaView : RNView;
+    let container: React.ComponentType<any> = RNView;
+    if (useSafeArea && SafeAreaContextPackage?.SafeAreaView) {
+      container = SafeAreaContextPackage.SafeAreaView;
+    } else if (useSafeArea && Constants.isIOS) {
+      container = RNSafeAreaView;
+    }
```
M-i-k-e-l, line 115: Personally I'd nest it:
```suggestion
    if (useSafeArea) {
      container = SafeAreaContextPackage?.SafeAreaView ?? RNSafeAreaView;
    }
```

adids1221: In this case we are not taking `isIOS` to consider.
We should leave it as is for now.

M-i-k-e-l (approved):

PR #3941, opened 2026-02-24
Handle missing react-native-safe-area-context in ScreenFooter screen

adids1221:
## Description
Fixes crash in ScreenFooterScreen demo when react-native-safe-area-context is not installed.
Replaced the direct import with a try/catch require and added a fallback screen with install instructions.
Fixed a circular import in ScreenFooter (`import {Image} from 'react-native-ui-lib'` → `import Image from '../image'`) that broke the barrel export when the optional dependency was missing.

## Changelog
ScreenFooterScreen - Gracefully handle missing react-native-safe-area-context dependency.

## Additional info
N/A


adids1221 (commented):
Reverted. The reason for using RN primitives was that uilib's View uses useSafeArea which also depends on react-native-safe-area-context, so the fallback screen would crash too. But happy to go with your solution!

PR #3942, opened 2026-02-26
Enhance color palette generation with saturation curve options

ZivSa1:
## Description
- Updated GeneratePaletteOptions to include saturationCurve, saturationThreshold, and saturationFloor for improved saturation adjustments.
- Refactored adjustSaturation function to utilize the new saturation curve logic.
- Modified color tint tests to match the changes + added tests for new logic.

## Changelog
Colors: ColorPalette new Lightness and Saturation logic

## Additional info
Ticket 4308

### DIFF
--- a/packages/react-native-ui-lib/src/style/colors.ts
+++ b/packages/react-native-ui-lib/src/style/colors.ts
@@ -27,10 +27,12 @@ export type GeneratePaletteOptions = {
   adjustSaturation?: boolean;
   /** Custom percentage-based saturation curve indexed by distance from the base color.
    * Overrides the default curve when provided. Each value represents the fraction of the base
-   * color's saturation to apply at that distance (e.g. [1.0, 0.89, 0.77, ...]) */
+   * color's saturation to apply at that distance (e.g. [1.0, 0.89, 0.77, ...]).
+   * The 'adjustSaturation' option must be true */
   saturationCurve?: number[];
   /** Array of additive saturation adjustments to apply per-index on the palette (from darkest to lightest).
-   * When provided, uses legacy per-index saturation logic instead of the default curve */
+   * When provided, uses legacy per-index saturation logic instead of the default curve.
+   * The 'adjustSaturation' option must be true */
   saturationLevels?: number[];
   /** Whether to add two extra dark colors usually used for dark mode (generating a palette of 10 instead of 8 colors) */
   addDarkestTints?: boolean; // TODO: rename 'fullPalette'


packages/react-native-ui-lib/src/style/colors.ts
```diff
@@ -19,11 +19,15 @@ export type GetColorByHexOptions = {validColors?: string[]};
 export type GeneratePaletteOptions = {
   /** Whether to adjust the lightness of very light colors (generating darker palette) */
   adjustLightness?: boolean;
-  /** Whether to adjust the saturation of colors with high lightness and saturation (unifying saturation level throughout palette) */
+  /** Whether to apply the saturation curve to unify saturation levels throughout the palette */
   adjustSaturation?: boolean;
-  /** Array of saturation adjustments to apply on the color's tints array (from darkest to lightest).
-   * The 'adjustSaturation' option must be true */
-  saturationLevels?: number[];
```
lidord-wix, line 26: what if someone is passing `saturationLevels`? 
it seems like a breaking change..

packages/react-native-ui-lib/src/style/colors.ts
```diff
@@ -268,7 +272,7 @@ export class Colors {
     const end = options?.addDarkestTints && colorLightness > 10 ? undefined : size;
     const sliced = tints.slice(start, end);
 
-    const adjusted = options?.adjustSaturation && adjustSaturation(sliced, color, options?.saturationLevels);
+    const adjusted = options?.adjustSaturation && adjustSaturationWithCurve(sliced, color, options);
```
lidord-wix: The function expect to get `CurveOptions` but gets `GeneratePaletteOptions`

lidord-wix: When `adjustSaturation` is `false` (or not provided), `adjustSaturationWithCurve` is never called. 
However, `defaultPaletteOptions` sets `adjustSaturation: true`, so calling `generateColorPalette` without `options` always applies the curve. This is a behavior change from before where the old fallback `adjustSaturation` function only modified colors when `lightness > 80 && saturation > 60`. Now the curve is applied to all colors with `baseSaturation > 50`.

This means previously unaffected palettes (e.g., colors with `lightness < 80 but saturation > 50`) will now get
  modified. This widens the blast radius of the change.

packages/react-native-ui-lib/src/style/colors.ts
```diff
@@ -354,50 +360,35 @@ function colorStringValue(color: string | object) {
...
-  hsl.color[1] = saturation;
-  return hsl.hex();
+  return colors.map((hex, i) => {
+    if (i === baseIndex) {
+      return hex;
+    }
+    const hsl = Color(hex).hsl();
+    const distance = Math.abs(i - baseIndex);
+    const percentage = curve[Math.min(distance, curve.length - 1)];
+    const newSaturation = Math.max(floor, Math.ceil(baseSaturation * percentage));
```
lidord-wix: why using `Math.ceil` here?

packages/react-native-ui-lib/src/style/__tests__/colors.spec.js
lidord-wix, line 1: please also cover the new logic where the `addDarkestTints: true`

packages/react-native-ui-lib/src/style/colors.ts
ZivSa1: changed to `round` :pray:

packages/react-native-ui-lib/src/style/colors.ts
ZivSa1: I asked Alexey and Yulia, we do want to change the logic entirely. 
If I'm not mistaken, at Wix this logic is passed only via the private, so it should break our modules, but it is breaking change for public users. what do we want to do? still allow the public users use the old logic, or make them adjust to the new logic?

packages/react-native-ui-lib/src/style/colors.ts
ZivSa1: Yes, we want the saturation curve to be applied as default, and only if the user decides to - they can still turn it off by sending false in the options. The curve also wont be applied if the base saturation is below 50 - it's part of the logic.

packages/react-native-ui-lib/src/style/colors.ts
```diff
@@ -12,17 +12,25 @@ import Scheme, {Schemes, SchemeType} from './scheme';
...
 export type GetColorTintOptions = {avoidReverseOnDark?: boolean};
 export type GetColorByHexOptions = {validColors?: string[]};
 export type GeneratePaletteOptions = {
   /** Whether to adjust the lightness of very light colors (generating darker palette) */
   adjustLightness?: boolean;
-  /** Whether to adjust the saturation of colors with high lightness and saturation (unifying saturation level throughout palette) */
+  /** Whether to apply the saturation curve to unify saturation levels throughout the palette */
   adjustSaturation?: boolean;
-  /** Array of saturation adjustments to apply on the color's tints array (from darkest to lightest).
-   * The 'adjustSaturation' option must be true */
```
lidord-wix, line 25: I'd keep this line and also add it to the `saturationCurve` prop

PR #3948, opened 2026-03-12
Migrate FloatingButton to use new ScreenFooter component internally

yotam-wix:
## Description
- Rewrote FloatingButton component to a function component that renders ScreenFooter internally, while keeping the same external API and visual behavior. 
- All existing props work as before, mapped to their ScreenFooter equivalents. 
- Added hoisted prop (default: true) that gives FloatingButton built-in keyboard-awareness using screenFooter's logic.
- Updated FloatingButtonScreen to use SafeAreaProvider

## Changelog
floatingButton - changed to use new ScreenFooter component internally. deprecated original component.

## Additional info
Ticket 5000

### DIFF
--- a/packages/react-native-ui-lib/src/components/floatingButton/__tests__/index.spec.tsx
+++ b/packages/react-native-ui-lib/src/components/floatingButton/__tests__/index.spec.tsx
@@ -14,11 +14,11 @@ const secondaryButton = {
 };
 
 const StickyTestCase = props => {
-  return <FloatingButton hoisted={false} {...props} testID={TEST_ID} />;
+  return <FloatingButton hoisted={false} {...props} testID={TEST_ID}/>;
 };
 
 const HoistedTestCase = props => {
-  return <FloatingButton {...props} testID={TEST_ID} />;
+  return <FloatingButton {...props} testID={TEST_ID}/>;
 };
 
 export const FloatingButtonDriver = (props: ComponentProps) => {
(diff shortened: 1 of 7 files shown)


packages/react-native-ui-lib/src/components/floatingButton/index.tsx
```diff
@@ -56,220 +67,101 @@ export interface FloatingButtonProps {
...
-    }
-  }
-}
+  const children = isHorizontal
+    ? [renderSecondaryButton(), renderPrimaryButton()]
+    : [renderPrimaryButton(), renderSecondaryButton()];
+
+  return (
+    <ScreenFooter
+      visible={visible}
```
lidord-wix, line 157: visible is undefined by default in the floating button and true in the screen footer, means that in case the user didn't pass anything and expected it to be invisible, now it will be shown

packages/react-native-ui-lib/src/components/floatingButton/__tests__/index.spec.tsx
```diff
@@ -15,22 +14,29 @@ const secondaryButton = {
 };
 
 const TestCase = (props) => {
-  return <FloatingButton {...props} testID={TEST_ID}/>;
+  return <FloatingButton hoisted={false} {...props} testID={TEST_ID}/>;
```
lidord-wix, line 17: All the tests are for the `hoisted={false}` case, let's add some for the default one

packages/react-native-ui-lib/src/components/floatingButton/__tests__/index.spec.tsx
```diff
@@ -66,91 +72,43 @@ describe('FloatingButton', () => {
     });
   });
 
-  describe('bottomMargin', () => {
```
lidord-wix, line 69: let's bring back all the `bottomMargin` tests to make sure the users aren't break

packages/react-native-ui-lib/src/components/floatingButton/index.tsx
lidord-wix, line 1: please format all files before merging the PR (cmd + shift + P -> Format Document) make sure you're doing it with Prettier ESlint (cmd + shift + P -> Format Document With)

lidord-wix (commented):
I added some comments, overall looks good :) 
Please make the changelog (in the PR description) more clear 🙏🏽 
Also - I wasn't able to run locally, had some issues with the yarn install.. will try again tomorrow

lidord-wix (dismissed):
Approved :) 
don't forget to format the files before merging :)

lidord-wix (approved):

PR #3951, opened 2026-03-15
[Slider] Fix Android hitSlop by replacing touchArea overlay with responder system

adids1221:
## Description
Fixes an Android bug where touches near the Slider thumb fail to start a drag.

**Slider** – The `touchArea` overlay (`absoluteFillObject`) intercepted touches in the thumb's `hitSlop` zone before they reached the PanResponder.
Replaced it with the container's Gesture Responder System (`onStartShouldSetResponder + onResponderRelease`) — the thumb (deepest child) wins for direct touches, the container handles track taps. 
`hitSlop` now works consistently on both platforms.

**Incubator.Slider** – Added `.hitSlop()` to the RNGH `Gesture.Pan()` on the thumb, so the gesture handler natively recognizes touches in the expanded hit area on Android. 
Also added `hitSlop` to the Pan gesture mock in jest setup.

## Changelog

**Slider, Incubator.Slider** - Fixed Android bug where touches near the thumb failed to start a drag.

## Additional info
- MADS-4866
- React Native [Gesture Responder System docs](https://reactnative.dev/docs/gesture-responder-system)

### DIFF
--- a/packages/react-native-ui-lib/src/incubator/slider/Thumb.tsx
+++ b/packages/react-native-ui-lib/src/incubator/slider/Thumb.tsx
@@ -62,6 +62,7 @@ const Thumb = (props: ThumbProps) => {
   const lastOffset = useSharedValue(0);
 
   const gesture = Gesture.Pan()
+    .hitSlop(hitSlop ?? DEFAULT_THUMB_HIT_SLOP)
     .onBegin(() => {
       onSeekStart?.();
       isPressed.value = true;
--- a/packages/react-native-ui-lib/jestSetup/jest-setup.js
+++ b/packages/react-native-ui-lib/jestSetup/jest-setup.js
@@ -77,6 +77,7 @@ jest.mock('react-native-gesture-handler',
       PanMock.onFinalize = getDefaultMockedHandler('onFinalize');
       PanMock.activateAfterLongPress = getDefaultMockedHandler('activateAfterLongPress');
       PanMock.enabled = getDefaultMockedHandler('enabled');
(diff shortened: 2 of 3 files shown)


packages/react-native-ui-lib/src/incubator/slider/Thumb.tsx
```diff
@@ -62,6 +62,7 @@ const Thumb = (props: ThumbProps) => {
   const lastOffset = useSharedValue(0);
 
   const gesture = Gesture.Pan()
+    .hitSlop(hitSlop ?? DEFAULT_THUMB_HIT_SLOP)
```
M-i-k-e-l, line 65: You already have the default in the destructor
```suggestion
    .hitSlop(hitSlop)
```

I think this (modified) change along with changing the internal implementation from `Slider` to `Incubator.Slider` is enough (I think we've discussed testing the incubator version).
I've tested it on my device and it works, with that said I would not HF it and make sure it's fully tested (and also - please test on iOS)

M-i-k-e-l: If you do agree then also edit the description and changelog.
Either way please remove "React Native [Gesture Responder System docs](https://reactnative.dev/docs/gesture-responder-system)"

M-i-k-e-l: In the private screen - it seems that the video is no longer available, you can use a link from [here](https://download.blender.org/peach/bigbuckbunny_movies/) IMO

PR #3954, opened 2026-03-18
ScreenFooter - fix overflow in Fit mode, fix stretching issues

yotam-wix:
## Description
- Fixed layout bugs in horizontal STRETCH, FIXED, and FIT modes caused by the previous row + justifyContent: center approach for centering items. That approach centered items but prevented buttons from stretching,

- Replaced with column-direction wrappers (where default alignItems: 'stretch' handles button stretching) and cloneElement with the center modifier to inject centering into children's internal layout. This is necessary because alignItems: 'stretch' and alignItems: 'center' can't both be applied to the same container - the wrapper handles stretch, and the clone handles centering.

- Restored flexShrink wrapping for horizontal FIT mode, which was lost during the earlier refactor.

## Changelog
ScreenFooter - overflow and stretching bug fixes.

## Additional info
N/A

### DIFF
--- a/packages/react-native-ui-lib/src/components/screenFooter/index.tsx
+++ b/packages/react-native-ui-lib/src/components/screenFooter/index.tsx
@@ -199,7 +199,7 @@ const ScreenFooter = (props: ScreenFooterProps) => {
     (child: React.ReactNode, index: number) => {
       if (itemsFit === ItemsFit.FIXED && itemWidth) {
         const fixedStyle: ViewStyle = isHorizontal
-          ? {width: itemWidth, flexShrink: 1, overflow: 'hidden', flexDirection: 'row', justifyContent: 'center'}
+          ? {width: itemWidth, flexShrink: 1, overflow: 'hidden'}
           : {width: itemWidth, maxWidth: '100%'};
         return (
           <View key={index} style={fixedStyle}>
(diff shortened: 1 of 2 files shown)


packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -179,22 +179,32 @@ const ScreenFooter = (props: ScreenFooterProps) => {
...
     if (itemsFit === ItemsFit.FIXED && itemWidth) {
       const fixedStyle: ViewStyle = isHorizontal
-        ? {width: itemWidth, flexShrink: 1, overflow: 'hidden', flexDirection: 'row', justifyContent: 'center'}
+        ? {width: itemWidth, flexShrink: 1, overflow: 'hidden'}
         : {width: itemWidth, maxWidth: '100%'};
       return (
         <View key={index} style={fixedStyle}>
-          {child}
+          {isHorizontal && React.isValidElement(child)
+            ? React.cloneElement(child as React.ReactElement<any>, {center: true})
```
lidord-wix: We previously talked about it that we don't want to clone the user's element and add it some props
It's reasonable that their component won't even have the center prop..

lidord-wix (commented):
Added a comment

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
yotam-wix: yeah I know. But we do need to handle the combination of 'alignItems: stretch' and centerization, which I couldn't find a proper solution for without cloning the children and injecting the centering to them somehow. 

All previous solutions were messing up one of the two, and I can't find an alternative to cloneElement so far :/

Also, the user can always override the 'center' prop for one of his items manually which I felt made sense for a 'generic' component like the footer.

yotam-wix: Generally this bug is only related to components that do not already handle centering their content - such as Image for example. So technically we can move the responsibility to the users - by letting them know they need to center manually the items they're passing in that case.

yotam-wix:
@lidord-wix I re did this PR with a fix that assumes the user passes a centered item, now it's much simpler. let me know if this one's ok in your mind

lidord-wix (approved):

PR #3959, opened 2026-03-29
Add accessibility focus feature to TabController component

gayablau:
## Description
- Introduced `accessibilityFocusOnSelected` state to manage focus for screen-reader users.
- Added toggle button for enabling/disabling accessibility focus on selected tab.
- Updated TabController and TabBarItem components to support accessibility focus behavior.
- Enhanced documentation for `TabControllerProps` to include new accessibility prop.

Note: more reproducible on Samsung

toggle off:

https://github.com/user-attachments/assets/b094b77e-4125-46c6-bfa6-7c15a6101866

toggle on:

https://github.com/user-attachments/assets/6c79e9c0-566d-4a8f-b963-009432e7b3a4

## Changelog
TabController component — new accessibilityFocusOnSelected prop. When true, accessibility focus moves automatically to the selected tab item on index change, ensuring screen-reader users hear the active tab announced without extra navigation. Defaults to false.

## Additional info
MAE-1413


packages/react-native-ui-lib/src/components/tabController/TabBarContext.ts
```diff
@@ -15,6 +15,7 @@ interface TabControllerContext {
   targetPage: Reanimated.SharedValue<number>;
   /* carouselOffset: Reanimated.SharedValue<number>; */
   setCurrentIndex: (index: number) => void;
+  accessibilityFocusOnSelected?: boolean;
```
M-i-k-e-l, line 18: I would not add a special prop for this, it should probably work about of the box

packages/react-native-ui-lib/src/components/tabController/TabBarItem.tsx
```diff
@@ -164,6 +164,20 @@ export default function TabBarItem({
     }
   });
 
+  const isMounted = useRef(false);
```
M-i-k-e-l, line 167: Can we remove this?
If not, please LMK what it fixes.

packages/react-native-ui-lib/src/components/tabController/TabBarItem.tsx
```diff
@@ -164,6 +164,20 @@ export default function TabBarItem({
...
+  useEffect(() => {
+    if (!isMounted.current) {
+      isMounted.current = true;
+      return;
+    }
+    if (accessibilityFocusOnSelected && isSelected) {
+      const node = findNodeHandle(itemRef.current as unknown as React.Component | null);
```
M-i-k-e-l, line 174: 1. I see this is used as `itemRef?.current` in the rest of the file.
2. Can we remove some\all of the types?

PR #3988, opened 2026-04-29
FloatingButton - improve initial visibility handling with useRef

ZivSa1:
## Description
Restore the old FloatingButton first-load behavior: if the button starts with visible=false, don’t render it until it first becomes visible. This prevents initially hidden floating buttons from appearing at the bottom before their first reveal. It is low-risk, only touches FloatingButton, but the first entrance animation is not restored because ScreenFooter mounts already visible.

## Changelog
DRAFT

## Additional info
DRAFT


M-i-k-e-l (commented):
Please make sure to edit the changelog, info and description - bug scenario and how this is fixing it (IMO ScreenFooter etc is irrelevant)
Did you manage to find a playground scenario this fixes?

PR #3989, opened 2026-04-30
[ScreenFooter] - Fix taps absorbed in footer area on Android (physical devices)

MIchaelZablianov:
## Description

On Android physical devices, taps in the empty padding area of `ScreenFooter` (and therefore `FloatingButton`, which is now backed by `ScreenFooter`) are absorbed instead of falling through to content behind. This breaks `FlatList` items that scroll under a `FloatingButton` — their `onPress` never fires.

Root cause: `ScreenFooter` wraps everything in a Reanimated `Animated.View` and tries to enable fall-through via `pointerEvents="box-none"`. That **prop on a Reanimated `Animated.View` is unreliable on Android** (see referenced upstream issues — `pointerEvents` is silently ignored, taps in the band are captured).

Reproduces only on Android physical devices (not iOS, not Android emulators).

When the caller doesn't need an entrance animation (`animationDuration === 0` — which is what `FloatingButton`'s `withoutAnimation` prop sets), no `transform` is needed at all. We add a static-visible short-circuit to both branches (HOISTED and STICKY) that bypasses Reanimated entirely and renders a plain RN `View` tree. Plain `View`s honour `pointerEvents` correctly on Android.

For HOISTED, `pointerEvents="box-none"` is also forwarded to `KeyboardAccessoryView`, which propagates it via `{...others}` to the inner `KeyboardTrackingView` (a plain RN `View` on Android). The prop isn't part of `KeyboardAccessoryView`'s typed surface, so the spread + cast `{...({pointerEvents: 'box-none'} as object)}` is used.

The non-static (animated) branches are intentionally left unchanged — those still rely on `pointerEvents` on `Animated.View` and are still subject to the same upstream bug. Solving that requires a separate, more invasive change (e.g. shrinking the `Animated.View` to wrap only the buttons, or replacing it).

Repro snippet:

```tsx
<View flex>
  <FlatList ... />
  <FloatingButton
    button={{label: 'Add', onPress: () => {}}}
    visible
    withoutAnimation
    bottomMargin={32}
  />
</View>
```

On an Android physical device: scroll `FlatList` items into the FAB band, taps on those items don't fire. With this fix, they do.

Upstream references (why `pointerEvents` is broken on Reanimated `Animated.View` on Android):
- https://github.com/software-mansion/react-native-reanimated/issues/8497
- https://github.com/software-mansion/react-native-gesture-handler/issues/1270

## Changelog

**ScreenFooter** - Fix taps in the footer's transparent area being absorbed instead of falling through to content behind on Android (physical devices) when `animationDuration` is `0`. Affects `FloatingButton` with `withoutAnimation`.

## Additional info

N/A

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/react-native-ui-lib/src/components/screenFooter/index.tsx
+++ b/packages/react-native-ui-lib/src/components/screenFooter/index.tsx
@@ -226,6 +226,8 @@ const ScreenFooter = (props: ScreenFooterProps) => {
 
   const childrenArray = React.Children.toArray(children).slice(0, 3).map(renderChild);
 
+  const isStaticVisible = animationDuration === 0 && visible;
+
   const renderFooterContent = useCallback(() => {
     return (
       <>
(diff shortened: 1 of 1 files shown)


packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -226,6 +226,8 @@ const ScreenFooter = (props: ScreenFooterProps) => {
 
   const childrenArray = React.Children.toArray(children).slice(0, 3).map(renderChild);
 
+  const isStaticVisible = animationDuration === 0 && visible;
```
M-i-k-e-l, line 229: This line has a good potential for a better solution (a little less footprint and a little more readable IMO).

<details>
  <summary>Click to see a hint</summary>
  How is `animationDuration === 0` called in `FloatingButton`?
</details>

M-i-k-e-l:
Also, did you go over the description? :)

MIchaelZablianov:
work on it continued in https://github.com/wix/react-native-ui-lib/pull/3990

PR #3994, opened 2026-05-05
ScreenFooter - add animationType

M-i-k-e-l:
## Description
ScreenFooter - add animationType
Review each commit separately

Note:
There was another solution created by Ziv: #3987 however I feel like it is more bug prone; and even though this solution requires the user to change the animation from `slide` to `fade` (approved by UX) it is overall safer.

## Changelog
ScreenFooter - add animationType ['slide' (default) | 'fade' | 'none']

## Additional info
Ticket 5037

### DIFF
--- a/demo/src/screens/componentScreens/ScreenFooterScreen.tsx
+++ b/demo/src/screens/componentScreens/ScreenFooterScreen.tsx
@@ -10,6 +10,7 @@ import {
   ScreenFooterLayouts,
   ScreenFooterBackgrounds,
   KeyboardBehavior,
+  ScreenFooterAnimationTypeProp,
   FooterAlignment,
   HorizontalItemsDistribution,
   ItemsFit,
@@ -88,6 +89,12 @@ const KEYBOARD_BEHAVIOR_OPTIONS = [
   {label: 'Hoisted', value: KeyboardBehavior.HOISTED}
 ];
 
+const ANIMATION_TYPE_OPTIONS = [
+  {label: 'Slide', value: 'slide'},
+  {label: 'Fade', value: 'fade'},
+  {label: 'None', value: 'none'}
+];
+
 const KEYBOARD_BEHAVIOR_OPTIONS_SPACED = [
   {label: 'Sticky', value: KeyboardBehavior.STICKY},
   {label: 'Hoisted', value: KeyboardBehavior.HOISTED},
@@ -115,6 +122,7 @@ const ScreenFooterContent = () => {
   const [layout, setLayout] = useState<ScreenFooterLayouts>(ScreenFooterLayouts.HORIZONTAL);
   const [background, setBackground] = useState<ScreenFooterBackgrounds>(ScreenFooterBackgrounds.SOLID);
   const [keyboardBehavior, setKeyboardBehavior] = useState<KeyboardBehavior>(KeyboardBehavior.STICKY);
+  const [animationType, setAnimationType] = useState<ScreenFooterAnimationTypeProp>('slide');
   const [alignment, setAlignment] = useState<FooterAlignment>(FooterAlignment.CENTER);
   const [horizontalAlignment, setHorizontalAlignment] = useState<FooterAlignment>(FooterAlignment.CENTER);
   const [distribution, setDistribution] = useState<HorizontalItemsDistribution>(HorizontalItemsDistribution.STACK);
@@ -390,6 +398,21 @@ const ScreenFooterContent = () => {
           </View>
         </View>
 
+        {/* Animation type */}
+        <View marginB-s4>
+          <Text text70M marginB-s2>
+            Animation Type
+          </Text>
+          <View row>
+            <SegmentedControl
+              segments={ANIMATION_TYPE_OPTIONS}
+              initialIndex={ANIMATION_TYPE_OPTIONS.findIndex(opt => opt.value === animationType)}
+              onChangeIndex={index => setAnimationType(ANIMATION_TYPE_OPTIONS[index].value)}
+            />
+            <View flex />
+          </View>
+        </View>
+
         {/* Alignment (Cross Axis) */}
         <View marginB-s4>
           <Text text70M marginB-s2>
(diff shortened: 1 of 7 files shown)


demo/src/screens/componentScreens/ScreenFooterScreen.tsx
```diff
@@ -390,6 +398,21 @@ const ScreenFooterContent = () => {
...
+            Animation Type
+          </Text>
+          <View row>
+            <SegmentedControl
+              segments={ANIMATION_TYPE_OPTIONS}
+              initialIndex={ANIMATION_TYPE_OPTIONS.findIndex(opt => opt.value === animationType)}
+              onChangeIndex={index => setAnimationType(ANIMATION_TYPE_OPTIONS[index].value)}
+            />
+            <View flex />
+          </View>
```
lidord-wix, line 413: this is minor, but you can remove these two `Views` and pass `containerStyle={styles.rowContainer}` to the `SegmentedControl` instead

packages/react-native-ui-lib/src/components/screenFooter/useAnimatedFooterStyle.ts
```diff
@@ -0,0 +1,75 @@
...
+) => {
+  const {
+    animationType: animationTypeProp = 'slide',
+    animationDuration: animationDurationProp = 200,
+    keyboardBehavior,
+    visible,
+    isAndroidEdgeToEdge = !!androidVersion && androidVersion >= 35 ? true : undefined
+  } = props;
+
+  const animationType = animationDurationProp === 0 ? 'none' : animationTypeProp;
```
lidord-wix, line 21: animationType (a plain JS value) is used inside useAnimatedStyle, but Reanimated worklet closures capture the value at creation time. Since animationType is derived from props and can change, the worklet will reference a stale value.
If the user switches from 'slide' to 'fade' at runtime, the worklet may still branch on the old value. Consider using a useSharedValue for animationType and updating it in a useEffect, so the worklet always reads the current value. Same concern applies to keyboardBehavior.

IMO it's not a common case but worth fixing it

packages/react-native-ui-lib/src/components/screenFooter/useAnimatedFooterStyle.ts
```diff
@@ -0,0 +1,75 @@
...
+  const animationDuration = animationType === 'none' ? 0 : animationDurationProp;
+
+  const keyboard = useAnimatedKeyboard({
+    isNavigationBarTranslucentAndroid: isAndroidEdgeToEdge,
+    isStatusBarTranslucentAndroid: isAndroidEdgeToEdge
+  });
+
+  const [height, setHeight] = useState(0);
+  const animatedValue = useSharedValue(animationType === 'fade' && visible ? 1 : 0);
```
lidord-wix, line 30: When animationType='slide' and visible=false, the initial animatedValue is 0 — but for slide, hidden state means animatedValue = height. Since height starts at 0, the footer will briefly flash visible until the first layout fires and useEffect updates the value. This is likely a subtle flash-of-content bug on mount.
You can see it in the FloatingButtonScreen, change `visible` initial value to `false` and add `animationType='slide'`

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
```diff
@@ -260,26 +231,27 @@ const ScreenFooter = (props: ScreenFooterProps) => {
...
+  }, [keyboardBehavior, renderFooterContent]);
 
   return (
-    <Animated.View testID={testID} onLayout={onLayout} style={[styles.container, stickyAnimatedStyle]}>
-      {renderFooterContent()}
+    <Animated.View
+      testID={testID}
+      style={containerStyle}
+      onLayout={keyboardBehavior === 'hoisted' ? undefined : onLayout}
+      pointerEvents={keyboardBehavior === 'hoisted' ? (visible ? 'box-none' : 'none') : 'auto'}
```
lidord-wix: Cursor said that, I couldn't reproduce: 
When the footer fades out (opacity 0) or uses 'none' animation, it is still positioned on screen and can intercept touches. For the 'slide' type, the footer moves offscreen so it's not an issue. But with 'fade' or 'none', the footer should have pointerEvents='none' when visible=false, regardless of keyboard behavior. Currently, sticky + fade/none would leave an invisible but touchable footer.

packages/react-native-ui-lib/src/components/screenFooter/screenFooter.api.json
```diff
@@ -58,10 +58,16 @@
       "description": "If true, the footer is visible. If false, it slides down",
```
lidord-wix: If false, the footer is hidden using the configured animation type

lidord-wix (commented):
Looks good, I added some comments, most of them are small or not so important but worth fixing :)

demo/src/screens/componentScreens/ScreenFooterScreen.tsx
M-i-k-e-l: I copy-pasted the other parts, technically it should be using `ExampleScreenPresenter` all over this

packages/react-native-ui-lib/src/components/screenFooter/useAnimatedFooterStyle.ts
M-i-k-e-l: What is the possible bug?
The `animationType` is changed in the screen and it works as expected.

packages/react-native-ui-lib/src/components/screenFooter/useAnimatedFooterStyle.ts
M-i-k-e-l: To fix this we'll probably need a bigger change like I've mentioned in the PR description.
#3987 by Ziv might fix it but it's riskier (bigger regression chance) IMO

packages/react-native-ui-lib/src/components/screenFooter/index.tsx
M-i-k-e-l: Fixed - it's Android only

packages/react-native-ui-lib/src/components/screenFooter/useAnimatedFooterStyle.ts
M-i-k-e-l: It reproduces in the screen with `Hide on Scroll`

lidord-wix (approved):

PR #4023, opened 2026-05-27
fix(Chip): expose selected prop to screen readers via accessibilityState

adids1221:
## Description
The `Chip` component had no way to communicate selection state to screen readers.
Passing `accessibilityState` via spread worked as a workaround but was undocumented and fragile.

Added `selected` as a first-class optional prop — when provided, it maps directly to `accessibilityState={{selected}}` on the container.
When omitted, behavior is identical to today.

## Changelog
Chip - Added `selected` prop that exposes selection state to screen readers via `accessibilityState`.

## Additional info
N/A


adids1221:
Closing in favor of #4024 which contains only the Chip accessibility fix, without the unrelated skills changes.

PR #4033, opened 2026-07-19
fix: replace url-parse with native URL API

alexandergolbergwix:
## Description

Removes the `url-parse` dependency from `react-native-ui-lib` and uses the native `URL` / `URLSearchParams` APIs in `AvatarHelper` instead.

- `isGravatarUrl` and `patchGravatarUrl` now use the global `URL` constructor (no import shadowing)
- `patchGravatarUrl` sets `d=404` and removes `default` via `searchParams`
- Drops `url-parse` and `@types/url-parse` from package manifests and lockfile

This fully resolves the OSV supply-chain issue without needing to bump `url-parse` (see #4031).

## Changelog

**AvatarHelper** — Removed `url-parse` dependency; Gravatar URL helpers now use native `URL` APIs.

## Additional info

All existing `AvatarHelper` tests pass (`yarn workspace react-native-ui-lib test AvatarHelper.spec.js`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "@types/react": "19.0.0",
     "@types/react-test-renderer": "^19.0.0",
     "@types/tinycolor2": "^1.4.2",
-    "@types/url-parse": "^1.4.3",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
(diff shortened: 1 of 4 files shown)


adids1221 (approved):
🚀

